PR #2, opened 2024-03-03
Move to bin-lib workspace

pinkforest:
This is initial PR to do all the initial renames

There are some work to be done to separate argument processing as currently it is coupled in the library code

I generally use a trait and then implement trait at argument side in binary

**Note: This PR needs to be merged - not squashed - so the renames are tracked in github history correctly**

Since this cannot be squashed I would prefer to keep this PR minimal so I can send other PRs to do further decoupling


davidlattimore:
Thanks!

I thought a little overnight about the idea of using a trait for arguments. It seems like it would add considerable extra boilerplate for each argument. Where currently an argument needs:

1) A field to hold the value
2) Some code to parse the flag and store into the field.

With a trait, we'd still need both of those, but in addition, we'd also need:

1) A method on the trait to access the value.
2) An implementation of that method that returns the value of the field.

So it roughly doubles the number of places that need new code when a new flag is added.

The code that uses the flag values would need to either:

1) Be generic over the trait implementation
2) Use dynamic dispatch with a potential performance penalty for frequently accessed config values.
3) Call the trait methods once and store the returned values into a different configuration struct, but that then adds even more per-flag boilerplate.

One other related consideration is that some users of wild_lib could potentially want to use the library by just passing the flags as strings, or potentially pass some flags as strings, but also supply input files via some other API. I'm thinking specifically if you wanted to integrate the linker into a compiler. Compilers often have flags like rustc's `-C link-arg=`, which require that the compiler just pass a flag as a string onto the linker without necessarily understanding what that flag will do.

Fortunately all this mostly doesn't affect this PR, but is stuff to keep in mind for later.

Issue #4, opened 2024-04-23
Use of `object` crate


philipc:
I was having a look at the code because I'm interested in seeing how people make use of the `object` crate.

You are making extensive use of the unified API for reading ELF files. Is this intentionally so that you can handle other file formats in future, or just because it was easiest? A concern I have with this approach is that, to support the unified API, the relocation parsing in `object` is doing a bunch of work that you don't need, and this may be an area that is performance sensitive.

Also, you have defined your own ELF types for use in writing (e.g. `FileHeader`, `ProgramHeader`). Is there a reason that the types defined in `object::elf` weren't suitable?

davidlattimore:
Thanks for having a look!

I'll have a look at using the lower level APIs. I have on a few occasions found the APIs that I was using a little too high level. In particular, I was wanting at one point to store slices of relocations, but because the APIs were all iterator-based, I ended up needing to do something else instead. But you're right, if it I just read the relocation section bytes, then call object::slice_from_bytes, I can just get the raw relocations directly.

I defined my own structs for ELF types mostly because I initially didn't realise that I could write via the types that the object crate provided. Once I did realise, I figured object's types were doing some things that I figured I didn't need - supporting writing to unaligned bytes and supporting big endian. But I suspect I won't pay any performance penalty for those features, so I should probably just switch over.

davidlattimore:
I've switched several structs over to those provided by the `object` crate. One downside is that because there are some structs that `object` doesn't define, such as the format of the eh_frame header, I now need to work with two different `Pod` interfaces - `object`'s  and `bytemuck`'s. But it's not too bad.

philipc:
> I'll have a look at using the lower level APIs.

I could work on this if you want. The lower level APIs haven't seen a lot of use, so there's probably some rough edges in `object` that need fixing.

> I now need to work with two different Pod interfaces - object's and bytemuck's

You could use `object`'s `Pod` for everything. Or we could switch `object` to use `bytemuck`. (The reason I haven't done that is because `bytemuck` expects you to have the exact size, but when reading it is common to have trailing bytes, so I would still have to repeat all the size calculations that `bytemuck` does. This is similar to why https://crates.io/crates/bytes-cast exists.)

davidlattimore:
> I could work on this if you want. The lower level APIs haven't seen a lot of use, so there's probably some rough edges in `object` that need fixing.

[...]

PR #5, opened 2024-04-25
Use Rela instead of object::Relocation

philipc:
This is a draft for #4. It needs some small helpers in `object` first.

In doing this I've been merrily deleting all the support for `RelocationTarget::Section` because [`object` never returns that for ELF](https://github.com/gimli-rs/object/blob/9650d6429667938a01b41a19e6de32f8f36b1aab/src/read/elf/relocation.rs#L474-L478), instead returning a `RelocationTarget::Symbol` that may refer to a section symbol. However, maybe I shouldn't be deleting this, and instead fixing it to handle section symbols, even though it previously didn't? I don't understand enough of the code to know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -313,7 +313,7 @@ dependencies = [
  "clap",
  "iced-x86",
  "linker-layout",
- "object",
+ "object 0.34.0",
  "rayon",
  "symbolic-demangle",
 ]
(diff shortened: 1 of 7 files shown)


davidlattimore:
Great! I had noticed that `RelocationTarget::Section` didn't seem to be coming up. I'd assumed it was somewhat obscure and wasn't used often. I guess I should have investigated further. Deleting unnecessary code is fine. It's also fine to leave it for later if you'd prefer.

philipc:
This is ready. I've updated object to use a git version. I've left one function that was handling section symbols as dead code; it can probably be deleted. There definitely are relocations referencing section symbols in practice, and the regular symbol handling appears to work fine for section symbols too.

davidlattimore:
Thanks for doing this!

davidlattimore:
I deleted the dead code. It definitely seems that it's not needed. Symbols with no names and only a section are handled via a different path that's more or less the same as any other symbol. That code was for handling relocations that reference sections not via a symbol.... which erm, seems to not be a thing.

PR #6, opened 2024-05-09
Support for object crate not returning null symbol/section

philipc:
The current git version of `object` no longer returns the null symbol or section at index 0.

For sections, check for this and add an extra section slot if needed.

For symbols, there are multiple places where we iterate over the input symbols, and adding an extra symbol every time is a bit ugly. Instead, add `SymbolIdRange` to recording the starting index and add methods to convert between id, vec index, and ELF index. These methods also include debug assertions to validate the values are in the correct range.

### DIFF
--- a/wild_lib/src/elf_writer.rs
+++ b/wild_lib/src/elf_writer.rs
@@ -679,9 +679,7 @@ impl<'data> ObjectLayout<'data> {
         for rel in &self.plt_relocations {
             plt_got_writer.write_ifunc_relocation(rel, &mut relocation_writer)?;
         }
-        for (symbol_id, resolution) in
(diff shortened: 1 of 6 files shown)


philipc:
I was trying to get back to finishing off #5, but I discovered that https://github.com/gimli-rs/object/pull/679 has broken wild. Not sure how you feel about the changes in this PR. Other options would be to switch wild to using `object`'s lower level API for sections and symbols too, or consider reverting the `object` change.

davidlattimore:
Thanks for the fix! I think the `SymbolIdRange` abstraction worked out nicely.

BTW, if it makes it easier at any point, I'm happy to temporarily switch to using `object` from github rather than crates.io.

PR #7, opened 2024-05-10
Fix error when running integration tests

philipc:
rustc doesn't appear to support using /dev/null as the output. This error occurs even when running rustc manually.

error: couldn't create a temp dir: Permission denied (os error 13) at path "/dev/rmetayXAwWn"


davidlattimore:
I was wondering why your rustc and mine were behaving differently, but I just figured it out. My terminal runs the shell in a sandbox and it turns out that my sandbox configuration makes /dev a writable tmpfs. Thanks for the fix!

PR #8, opened 2024-05-13
Use low level API from object crate

### LINKED ISSUE #4: Use of `object` crate

philipc:
I think this is worth doing because the code is closely tied to the ELF format, and using the lower level API removes possibilities for ambiguity due to the abstractions in the unified API in `object`. Other than the need to define our own `File` and sometimes pass an extra parameter, it doesn't add any significant complexity.

Closes #4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -364,7 +364,7 @@ dependencies = [
 [[package]]
 name = "object"
 version = "0.35.0"
-source = "git+https://github.com/gimli-rs/object?rev=017624aa70a1a52e36dbb61e693ba3c7f3bf010e#017624aa70a1a52e36dbb61e693ba3c7f3bf010e"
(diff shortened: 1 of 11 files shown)


davidlattimore:
Thanks!

PR #16, opened 2024-07-04
add bevy dynamic linking project test

tbillington:
Conceptualising what more involved testing might look like, probably don't merge this as is :) ?

Runs with `cargo test run_bevy_dynamic`

- Tests using wild through cargo (arguably the most common scenario)
- Tests with cargo dependencies
- This particular test is with a dependency that relies upon dynamic linking
- Tests the actual program can run, this example currently compiles using dynamic linking, but errors when running

I'm not sure if bigger tests like this should live in the repo or not. This example is small, but if we wanted to add tests for something like ripgrep it seems clumsy to bring that in.

Bigger tests _could_ live in a separate repo that is dedicated to integration testing wild with larger projects. Then it wouldn't feel as hacky to have lots of git submodules. It would also make it more straight forward to develop integration testing tooling for running lots of cargo compiles.

### DIFF
--- a/wild/tests/build_projects.rs
+++ b/wild/tests/build_projects.rs
@@ -1,0 +1,55 @@
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+    process::Command,
+};
+
+const PROJECTS_ROOT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/tests/projects");
+const WILD_TEST_DIR: &str = "wild_project_tests";
+const RUSTFLAGS_LINK_WITH_WILD: &str = concat!(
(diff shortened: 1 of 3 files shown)


davidlattimore:
I think you're probably right that a separate repository would make sense - especially with git submodules.

Another thought is that I'd be somewhat tempted to leave the compilation directories in-place, then just delete all executables and .so files from the target directory before rerunning. That'd probably save a good bit of time by avoiding the need to rerun rustc.

At a larger scale, I could imagine eventually doing something like a crater run and just bulk downloading lots of crates from crates.io, then running their tests with wild set at the linker.

marxin:
Can we please close this PR as it's probably not going to land in a near future?

PR #23, opened 2024-07-31
openSUSE: switch to rolling distribution

marxin:
Please update to the more modern distribution (compared to Leap with GCC 7).

### DIFF
--- a/docker/opensuse.Dockerfile
+++ b/docker/opensuse.Dockerfile
@@ -3,13 +3,11 @@
 # docker build -t wild-dev-opensuse . -f docker/opensuse.Dockerfile
 # docker run -it wild-dev-opensuse
 
-FROM opensuse/leap:latest AS chef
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks for the PR. I'm happy to give tumbleweed a try for a bit. One concern I have it with and it's the reason I switched to leap is that I'd like to have a base image that's not changing all the time. If I could pin a specific tumbleweed image for a few months, that'd give that, but unless I'm missing something, I don't see tags for older tumbleweed images. Anyway, I'll give tumbleweed a try for a few days at least, but if the end result is that I end up needing to download a new image every day that I want to run it, I'll likely need to switch back.

marxin:
Yeah, it's a rolling distribution and so one can't have separate tags for it (there's only one: `tumbleweed/latest`). However, it seems you found a feasible solution by sticking to a specific hash.

PR #28, opened 2024-08-13
Support single and double dashes for multi-letter option names

marxin:
This implements what other linkers do:

man ld:
```
For options whose names are multiple letters, either one dash or two can precede the option name; for example, -trace-symbol and --trace-symbol are equivalent.  Note---there is one exception to this rule.  Multiple letter options that start with a lower case ’o’ can only be preceded by two dashes.  This is to reduce confusion with the -o option.  So for
       example -omagic sets the output file name to magic whereas --omagic sets the NMAGIC flag on the output.
```

### DIFF
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -109,26 +109,25 @@ pub(crate) const FILES_PER_GROUP_ENV: &str = "WILD_FILES_PER_GROUP";
 // feature.
 const IGNORED_FLAGS: &[&str] = &[
     // TODO: Support build-ids
-    "--build-id",
+    "build-id",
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks! I had noticed a bit of inconsistency with single and double dashes, but hadn't looked into what the rules were. Great to have this sorted out.

I'm unsure what the future is for the argument parsing code. I'm sure eventually it'll need some substantial refactoring in order to support `--help`.

marxin:
> I'm unsure what the future is for the argument parsing code. I'm sure eventually it'll need some substantial refactoring in order to support `--help`.

Sure! The question is if _Clap_ can support more complex option parsing like this (`modifier_stack` handling and options that support one and two dashes)?

PR #30, opened 2024-08-15
Extend supported cmdline arguments from file

marxin:
The goal of the pull request is to support rules specified by BFD:

```
       @file
           Options  in  file  are  separated  by  whitespace.   A whitespace character may be included in an option by surrounding the entire option in either single or double quotes.  Any character (including a backslash) may be included by prefixing the character to be included with a backslash.  The file may itself contain additional @file options; any such
           options will be processed recursively.
```

My changes don't utilize `Cow`; they use a normal string. Hopefully, it's something hard to spot in the performance stats. What do you think?


davidlattimore:
Looks good. Thanks!

Issue #36, opened 2024-08-18
Make it easier to add built-in sections


davidlattimore:
When adding new built-in sections, or sections that we need to reference from linker code, there's currently quite a few places that need to be changed. The tests help a bit, since they generally tell you if you miss one of the places that needs changing. It's still not ideal though.

There are a few things that make fixing this complicated.

* Sections that come from input files, we bucket by alignment, whereas for sections that we generate internally, we don't.
* Some internally generated sections have two parts that we track separately, but then merge into a single section for output. For example, `.symtab` is split into `symtab_locals` and `symtab_globals`, since all the locals needs to come before all the globals. A similar split occurs with `.rela.dyn` which has all relative relocations before all others.

Since this is a pretty large change, it'd be good to try to do it in a few stages, although it's entirely possible that once the thread is pulled, it'll all unravel and end up as one massive change.

I think probably the key will be to introduce `SectionPartId` or even just `PartId`. A section can then be composed of multiple `Part`s. Each `Part` would have a single alignment.

Change `OutputSectionPartMap` to do all access via `PartId` rather than via direct references to named fields. `AlignmentMap` would go away and be flattened into `OutputSectionPartMap`.

We might, at least temporarily, need a way to map between `PartId` and `OutputSectionId`. This is probably easier if we put all the 1-1 parts together, then the weird 2-1 parts (mentioned above) then all the 16-1 parts, which are categorised by alignment.

A careful eye needs to be kept on performance when doing this to make sure we don't regress. I think that should be entirely doable, so long as things that are currently together in a single heap allocation remain in a single heap allocation, we continue to use constant offsets into Vecs etc. I'm hopeful that we might even see a slight performance improvement due to the code being a bit simpler with data being treated in a more homogenous way.

davidlattimore:
I think this is done, at least for now. There's still a few steps to adding a new section, but they should all be relatively straightforward. I think further reduction would require some kind of code generation. e.g. macro, proc-macro or a build script. I'm not sure it's worthwhile doing that - especially since those often have costs around how easy it is to understand what's going on in the code.

PR #52, opened 2024-08-19
Define constants based on object's constants

marxin:
Theoretically, the constants can be used directly, however, they are of u32 type and would introduce a lot of `as u64` casting between types in the codebase. What do you think, @davidlattimore?

Do you want to merge it now, or wait for another `object` release that will introduce `SHF_GNU_RETAIN`?


davidlattimore:
Happy to merge this as-is for now if you like. Referencing all the usage sites that reference these constants might hit merge conflicts with my work on #36, since I've got lots of edits in stuff related to output sections (also lots in layout.rs).

marxin:
Good, so let's merge it now.

PR #53, opened 2024-08-19
Use rstest for integration_tests.

marxin:
The main benefit is that one can run all the integration tests in parallel without stopping at the first error. And one can also run a single integration test if needed.

Note `rstest` also provides a neat filesystem discovery option: https://docs.rs/rstest/0.22.0/rstest/attr.rstest.html#files-path-as-input-arguments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,6 +15,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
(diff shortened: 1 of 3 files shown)


marxin:
With the applied change, the tests can finish in Debug mode in `1.6s`, while it took `2.6s` before.

davidlattimore:
Nice!

PR #57, opened 2024-08-20
Fix race when parallel threads build the same object.

marxin:
The cold start of tests takes now 3.5s, while it took about 5.5s before we started using the parallel execution of tests (`integration_test`).


marxin:
After 5b2302112ed5f04089a5f55098d349dd0bbef68e, this goes down to `2.1s` which is really good. To be honest, I was quite close to more fine-grained Mutexes, but I forgot to use `Arc` :)

PR #58, opened 2024-08-20
Add support for clang compiler in `integration_tests`

marxin:


davidlattimore:
I just noticed that prior to this change, trivial-main.c didn't have any non abstract config. Ideally that should probably have been an error. 

I was also wondering, does trivial-main add much beyond libc-integration.c or is it intended to be a subset?

marxin:
> I just noticed that prior to this change, trivial-main.c didn't have any non abstract config. Ideally that should probably have been an error.

Sure, done in #59.

> I was also wondering, does trivial-main add much beyond libc-integration.c or is it intended to be a subset?

It's a much simpler input and it actually exposed the `.gnu.hash` problem.

PR #60, opened 2024-08-20
Use the default gcc, ld.lld when running tests

marxin:
I think all distros support a symlink of `ld.lld`. The `gcc-13` is probably a leftover from times when you tests `opensuse:leap` Docker container, right?


marxin:
The PR changes Ubuntu to the latest release so that we can test version 18 (15 is tested now) of the LLD linker.

PR #62, opened 2024-08-20
Add -z now to default Linker flags + add -z lazy

marxin:


davidlattimore:
BTW, I'm open to at some point changing the default to not be `-z now` to match the other linkers, however I'm not sure what outstanding bugs there are in the lazy binding code, so those would need to be fixed first.

PR #65, opened 2024-08-20
Use declarative macro for `rel_type_to_string`

marxin:
I think the function would be useful also in `wild_lib` for error messages like:

```
1: Failed to apply relocation of type 1 to section `.fini` (39 local=2) in file #512 (2/0) (/usr/lib64/gcc/x86_64-suse-linux/13/../../../../lib64/crti.o) (ADDRESS | CAN_BYPASS_GOT) at offset 0x55
```

Do you have a crate that is shared between `linker-diff` and `wild_lib`?

### DIFF
--- a/linker-diff/src/asm_diff.rs
+++ b/linker-diff/src/asm_diff.rs
@@ -744,51 +744,80 @@ impl<'elf, 'data> RelocationDisplay<'elf, 'data> {
     }
 }
 
+macro_rules! const_name_by_value {
+    ($needle: expr, $( $const:ident ),*) => {
+        match $needle {
+            $(object::elf::$const => Some(stringify!($const)),)*
+            _ => None
+        }
+    };
+}
+
 fn rel_type_to_string(r_type: u32) -> Cow<'static, str> {
-    Cow::Borrowed(match r_type {
-        0 => "R_X86_64_NONE",
-        1 => "R_X86_64_64",
-        2 => "R_X86_64_PC32",
-        3 => "R_X86_64_GOT32",
-        4 => "R_X86_64_PLT32",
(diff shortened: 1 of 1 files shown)


davidlattimore:
> Do you have a crate that is shared between `linker-diff` and `wild_lib`?

Thus far I've mostly relied on `object` being a common crate, but you're right, there probably is stuff that would be easier to put into our own common crate. Actually there are already two common crates, `linker-layout` and `linker-trace`, but they're both for specific things, not general purpose.

I guess looking up strings for object's constants could either go into `object`, or a crate that just contains functions to map object's constants to strings - `object-names` or `object-strings` or something like that.

Having a general purpose `linker-utils` crate would also be OK.

marxin:
@philipc Btw. isn't the change I made in this PR something you will potentially welcome in the `object` crate?

philipc:
Maybe. There's a bunch of them defined in the [`readobj` example](https://github.com/gimli-rs/object/blob/3872c290728dabef0ad69ea04025cc9866ea8542/crates/examples/src/readobj/elf.rs#L976), but they are a different format from what you've done here. What would be nice is if there was a way for object to provide the list of x86_64 relocation type constants, but leave it up to the user to define their own macro that uses that list. I'm not sure if rust provides a way to do that easily.

Also consider that moving this to `object` makes it harder to add values that `object` is missing to the list, so you might become more dependent on waiting for an `object` release for missing constants.

marxin:
> I'm not sure if rust provides a way to do that easily.

Me neither.

I would be happy with using the `&[Flag<u8>]` as defined in the `readobj` example. Maybe you can provide them as a feature flag?

PR #75, opened 2024-08-26
Use more elf::shf constants

marxin:


marxin:
I'm curious if https://github.com/davidlattimore/wild/blob/fe307369b9b815ea5226fe1c2eac7b2e0f9beedd/linker-diff/src/header_diff.rs#L93-L123 can also be moved to `wild_lib::elf::shf` module to a helper function?

If so, should I then add a dependency of `wild_lib` for `linker-diff`? Or should I move it to `linker-util`?

davidlattimore:
> If so, should I then add a dependency of `wild_lib` for `linker-diff`? Or should I move it to `linker-util`?

Either is fine with me. At one point `wild_lib` had an optional dependency on `linker-diff`, which would have prevented the reverse dependency, but I removed that in favour of just executing linker-diff, since actually depending on it inflated the binary size too much.

Actually, thinking about it some more, for it to be used from `linker-diff`, it'd need to be `pub`. If you're looking through the API docs for the linker, then a function like that might be surprising, since it's probably not an API that a user of the linker would even need. So maybe `linker-util` makes more sense.

marxin:
> So maybe `linker-util` makes more sense.

Good, I've just made #78 that addresses that.

Issue #37, opened 2024-08-18
Add support for DWARF debug info


davidlattimore:
I'm not 100% sure what's involved here. I gather that eh_frame info, which we already support, is somewhat related to, or a more limited form of dwarf debug info.

davidlattimore:
> I would like to try implementing this feature ;)

Great! Thanks!

> 3\. GC of the sections should not be inhibited by a debug-info section and the corresponding relocation of a debug-info section should use a tombstone value (`0` or `1` value depending on the section name). Can you help me how to integrate that into the GC algorithm?

Do you know if we need to split the sections then GC the parts of the sections that are for functions that get GCed? I had to do that with the `.eh_frame` support and it certainly added some complexity.

If you don't need to split the sections up and you also don't want relocations in those sections to prevent other things from being GCed, then you can possibly skip reading the relocations for those sections during the layout phase. e.g `Section::create` could skip the bit where it iterates over relocations. That also assumes that the relocations don't need any allocations - i.e. that none of the relocations need to be turned into runtime relocations.

> 4\. Should we list all the supported debug-info section names (~20 sections for DWARF 4 and 5) (`OutputSectionId::regular`)? Or do we want to use a different mechanism?

If you need to refer to the sections from code - e.g. if you need special logic for a particular section, then yes. Making the sections "regular" sections means that they'll be split by alignment. I guess if a particular section needs to only ever have one particular alignment, then we could use a "generated" section - although it'd make the name "generated" slightly less appropriate.

If you don't need special logic for debug sections, or if the only special logic is that you want to not process the relocations during the layout (GC) phase, then you could add a field on `SectionDetails` - like `is_debug_info`.

marxin:
> Do you know if we need to split the sections then GC the parts of the sections that are for functions that get GCed? I had to do that with the `.eh_frame` support and it certainly added some complexity.

No, for debug info sections, we'll only need to make the relocation resolution (with exception of `.debug_str` and `.debug_line_str` which can be candidate for string merging).

> That also assumes that the relocations don't need any allocations - i.e. that none of the relocations need to be turned into runtime relocations.

Yep, that's how I understand it as debug info is never read by a dynamic linker. It's the consumer like `gdb` or `valgrind` which loads the debug info from a (potentially different) file.

> If you don't need special logic for debug sections, or if the only special logic is that you want to not process the relocations during the layout (GC) phase, then you could add a field on `SectionDetails` - like `is_debug_info`.

[...]

PR #77, opened 2024-08-26
Support compressed input sections

marxin:
This is a foundation for supporting of the ELF compressed sections. The ergonomics is not ideal as one needs to go from `SectionHeader::sh_size` to `obj.object.section_size(section)` to get the correct section size, similarly for the alignment.
Do we have any other option?

For the actual uncompressed data, there is an existing [`object`'s `decompression` function](https://docs.rs/object/0.36.3/object/read/struct.CompressedData.html#method.decompress), but it comes from the unused `object::read` API. Do you know if we can use the function? If not, then we will need to do the decompression ourselves.

Partially implements #37 (part 1).

### DIFF
--- a/wild_lib/src/elf.rs
+++ b/wild_lib/src/elf.rs
@@ -5,6 +5,7 @@ use anyhow::Context;
 use bytemuck::Pod;
 use bytemuck::Zeroable;
 use linker_utils::elf::rel_type_to_string;
+use object::read::elf::CompressionHeader;
 use object::read::elf::FileHeader as _;
 use object::read::elf::ProgramHeader as _;
 use object::read::elf::RelocationSections;
@@ -122,6 +123,20 @@ impl<'data> File<'data> {
 
     pub(crate) fn section_data(&self, section: &SectionHeader) -> Result<&'data [u8]> {
         Ok(section.data(LittleEndian, self.data)?)
+    }
+
+    pub(crate) fn section_size(&self, section: &SectionHeader) -> Result<u64> {
+        Ok(section.compression(LittleEndian, self.data)?.map_or_else(
+            || section.sh_size.get(LittleEndian),
+            |compression| compression.0.ch_size(LittleEndian),
+        ))
+    }
+
+    pub(crate) fn section_alignment(&self, section: &SectionHeader) -> Result<u64> {
+        Ok(section.compression(LittleEndian, self.data)?.map_or_else(
+            || section.sh_addralign(LittleEndian),
+            |compression| compression.0.ch_addralign(LittleEndian),
+        ))
     }
 
     pub(crate) fn relocations(&self, index: object::SectionIndex) -> Result<&'data [Rela]> {
--- a/wild_lib/src/elf.rs
+++ b/wild_lib/src/elf.rs
@@ -125,7 +125,7 @@ impl<'data> File<'data> {
     pub(crate) fn section_data(&self, section: &SectionHeader) -> Result<&'data [u8]> {
         let data = section.data(LittleEndian, self.data)?;
 
-        if let Some((compression, offset, _)) = section.compression(LittleEndian, self.data)? {
+        if let Some((compression, _, _)) = section.compression(LittleEndian, self.data)? {
             let format = match compression.ch_type.get(LittleEndian) {
                 object::elf::ELFCOMPRESS_ZLIB => object::CompressionFormat::Zlib,
                 object::elf::ELFCOMPRESS_ZSTD => object::CompressionFormat::Zstandard,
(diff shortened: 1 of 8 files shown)


marxin:
Apparently, one can manually construct `object::read::CompressedData` and then call `decompress` ;)

davidlattimore:
> Do we have any other option?

What you've got doesn't look unreasonable to me. From what I can see, the call to `compression` should be relatively cheap. If it's not cheap, then I guess the alternative would be to store the size on the `UnloadedSection` so that we only look it up once.

wild_lib/src/elf.rs
```diff
@@ -121,7 +123,40 @@ impl<'data> File<'data> {
...
+            };
+
+            let decompressed = CompressedData {
+                format,
+                data: &data[(offset as usize)..],
+                uncompressed_size: compression.ch_size.get(LittleEndian),
+            }
+            .decompress()?;
+            // TODO: find a proper lifetime for the uncompressed data!!!
+            Ok(decompressed.into_owned().leak())
```
marxin, line 143: @davidlattimore: The following code should support getting data for a compressed section. The main question is how to cache the results and how to properly deallocate the uncompressed vector. Any suggestions?

davidlattimore: Is it only debug sections that use compression?

Do we need access to the section data during layout, or only when we're writing the output?

For executable sections, we currently look at the instruction bytes during layout to work out if we can apply certain optimisations. If we can apply those optimisations, then that affects whether we need to allocate space in the GOT, PLT etc. I'd hope that we don't need to do anything like that for debug info.

If we don't, then it might be possible to decompress directly into the output. That would be ideal for efficiency, since there's less copying and less heap allocation.

davidlattimore: Although I just looked at the code that calls `Relaxation::new` in `layout.rs` and it does it unconditionally. I suspect that's (a) potentially wrong and (b) potentially inefficient. So that bit of code can probably be changed to only call it if the section has the execute bit set. That would eliminate one call to `section_data`.

I see that I'm also storing a slice of section data on `Section`. From a quick look that's mostly not used until we write. Possibly something there can be refactored to delay that call to `section_data`.

marxin: > Is it only debug sections that use compression?

Yes. Some time ago, `maskray` suggested an LLD option for compression of any section, but he abandoned the effort: https://maskray.me/blog/2023-07-07-compressed-arbitrary-sections.

> Do we need access to the section data during layout, or only when we're writing the output?

Yes. As shown [here](https://github.com/davidlattimore/wild/issues/37#issuecomment-2310198009), debug info sections utilize relocations quite a lot and thus the section should be properly loaded and the relocations resolved (AFAIK debug info should not use any dynamic resolutions).

davidlattimore: > Yes. As shown [here](https://github.com/davidlattimore/wild/issues/37#issuecomment-2310198009), debug info sections utilize relocations quite a lot and thus the section should be properly loaded and the relocations resolved (AFAIK debug info should not use any dynamic resolutions).

For relocations in executable code, we process the twice, once during layout to (a) do a graph traversal to work out what is referenced and (b) work out if we need to allocate any space for a runtime relocation - e.g. for a GOT entry or similar. Then during file writing, we process the relocations again to actually apply them. It sounds to me like perhaps only the second lot of processing is needed for debug sections. i.e. the relocations need to be applied, but we don't need to have done anything with them during graph traversal (`layout.rs`).

marxin: Yes, I understand it the same. That sounds promising!

marxin: Regarding the current approach: `decompressed.into_owned().leak()`: is it something you can live for now or do you have any recommendations on what to do instead?

davidlattimore: I guess it only runs when we encounter compressed sections, which we currently don't support at all, so it's not regressing anything that we currently support. In that regard, I'd be OK with leaking in the short-term - assuming we want to make the refactorings required to not need it afterwards

marxin: All right then.

marxin:
@davidlattimore Can we merge this PR as it is now? The actual decompression is not used yet as it's only used for debug info sections, but I tested the functionality and the compression works fine.

davidlattimore:
I did a few additional commits that should reduce how much we need to leak. I think the only bit left is figuring out how to avoid leaking the decompressed data of string-merge sections. That one's trickier. It would be possible if I added an extra iteration through the sections, but I'd prefer not to do that. I'm going to see if I can avoid that extra iteration by using an arena.

PR #79, opened 2024-08-27
Introduce `SectionFlags` and provide Debug impl

marxin:
Sample output after the change:

```
[wild_lib/src/layout.rs:3685:9] &defs = SectionDetails {
    name: .comment,
    ty: 1,
    section_flags: MS,
    element_size: 1,
    retain: true,
    packed: false,
}
```


davidlattimore:
Nice!

PR #83, opened 2024-08-28
Do not skip first argument if it comes from a file

marxin:
I noticed that with:

```
$ cat /tmp/response.txt 
-m elf_x86_64 -pie -o /tmp/a.out /tmp/trivial.o -dynamic-linker /lib64/ld-linux-x86-64.so.2
$ ./ld @/tmp/response.txt
Error: Failed to open input file `elf_x86_64`

Caused by:
    No such file or directory (os error 2)
```


davidlattimore:
Ah, good catch!

PR #84, opened 2024-08-28
Enable debug info for dev mode

marxin:
It took me a while to realize that VS Code cannot put a breakpoint when debugging because I was missing the debug info.


davidlattimore:
Yeah, I turn it off on all my projects because it slows down building and I pretty much never use it. But I just checked and turning it back on only adds about a second, so I'm sure I can live with that for the time being if I know you're using it.

Issue #73, opened 2024-08-26
Do not assigned addresses to non-alloc sections (and Program Headers)


marxin:
As seen, Wild assigns addresses to all sections, while other linkers only for sections that have `ALLOC` section flag set:

```
$ gcc ~/Programming/main.c
$ eu-readelf -a a.out
...
Section Headers:
[Nr] Name                 Type         Addr             Off      Size     ES Flags Lk Inf Al
[ 0]                      NULL         0000000000000000 00000000 00000000  0        0   0  0
[ 1] .interp              PROGBITS     0000000000400318 00000318 0000001c  0 A      0   0  1
[ 2] .note.gnu.property   NOTE         0000000000400338 00000338 00000040  0 A      0   0  8
[ 3] .note.gnu.build-id   NOTE         0000000000400378 00000378 00000024  0 A      0   0  4
[ 4] .note.ABI-tag        NOTE         000000000040039c 0000039c 00000020  0 A      0   0  4
[ 5] .hash                HASH         00000000004003c0 000003c0 00000028  4 A      7   0  8
[ 6] .gnu.hash            GNU_HASH     00000000004003e8 000003e8 0000001c  0 A      7   0  8
[ 7] .dynsym              DYNSYM       0000000000400408 00000408 00000078 24 A      8   1  8
[ 8] .dynstr              STRTAB       0000000000400480 00000480 0000006d  0 A      0   0  1
[ 9] .gnu.version         GNU_versym   00000000004004ee 000004ee 0000000a  2 A      7   0  2
[10] .gnu.version_r       GNU_verneed  00000000004004f8 000004f8 00000020  0 A      8   1  8
[11] .rela.dyn            RELA         0000000000400518 00000518 00000060 24 A      7   0  8
[12] .init                PROGBITS     0000000000401000 00001000 0000001b  0 AX     0   0  4
[13] .text                PROGBITS     0000000000401020 00001020 00000101  0 AX     0   0 16
[14] .fini                PROGBITS     0000000000401124 00001124 0000000d  0 AX     0   0  4
[15] .rodata              PROGBITS     0000000000402000 00002000 00000004  4 AM     0   0  4
[16] .eh_frame_hdr        PROGBITS     0000000000402004 00002004 00000024  0 A      0   0  4
[17] .eh_frame            PROGBITS     0000000000402028 00002028 00000064  0 A      0   0  8
[...]

PR #87, opened 2024-08-29
Do not assigned addresses to non-alloc sections (and Program Headers)

marxin:
Fixes: #73

### DIFF
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -68,6 +68,7 @@ use object::read::elf::VerdefIterator;
 use object::LittleEndian;
 use object::SectionIndex;
 use smallvec::SmallVec;
+use std::collections::HashMap;
 use std::ffi::CString;
 use std::fmt::Display;
 use std::mem::size_of;
@@ -1364,14 +1365,18 @@ fn compute_segment_layout(
                 .expect("SegmentEnd without matching SegmentStart");
             complete.push(record);
         }
-        OrderEvent::Section(section_id, _section_details) => {
+        OrderEvent::Section(section_id, section_details) => {
             let part = section_layouts.get(section_id);
-            for rec in active_records.values_mut() {
-                rec.file_start = rec.file_start.min(part.file_offset);
-                rec.mem_start = rec.mem_start.min(part.mem_offset);
-                rec.file_end = rec.file_end.max(part.file_offset + part.file_size);
-                rec.mem_end = rec.mem_end.max(part.mem_offset + part.mem_size);
-                rec.alignment = rec.alignment.max(part.alignment);
+            // Segments only cover sections that are allocated and have a non-zero address.
+            if section_details.section_flags.contains(shf::ALLOC) {
+                assert!(part.mem_offset != 0);
+                for rec in active_records.values_mut() {
+                    rec.file_start = rec.file_start.min(part.file_offset);
+                    rec.mem_start = rec.mem_start.min(part.mem_offset);
+                    rec.file_end = rec.file_end.max(part.file_offset + part.file_size);
+                    rec.mem_end = rec.mem_end.max(part.mem_offset + part.mem_size);
+                    rec.alignment = rec.alignment.max(part.alignment);
+                }
             }
         }
     });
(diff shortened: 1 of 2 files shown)


marxin:
@davidlattimore For `integration_test::program_name_12___internal_syms_c__` I face the following problem:

```
readelf -SW /home/marxin/Programming/wild/wild/tests/build/internal-syms.c-default.wild
There are 10 section headers, starting at offset 0x120:

Section Headers:
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .phdr             NULL            0000000000400040 000040 000000 00   A  0   0  0
  [ 2] .shdr             NULL            0000000000400120 000120 000000 00   A  0   0  0
  [ 3] .eh_frame         PROGBITS        00000000004003a0 0003a0 000070 00   A  0   0  8
  [ 4] .text             PROGBITS        0000000000401410 000410 000063 00  AX  0   0  1
  [ 5] .rela.plt         RELA            0000000000402478 000478 000000 18  AI  8   0  1
  [ 6] .comment          PROGBITS        0000000000000000 000478 000035 01  MS  0   0  1
  [ 7] .shstrtab         STRTAB          0000000000000000 0004ad 00004a 00      0   0  1
  [ 8] .symtab           SYMTAB          0000000000000000 0004f8 0000a8 18      9   3  8
  [ 9] .strtab           STRTAB          0000000000000000 0005a0 00004e 00      0   0  1
```

As seen `rela.plt` is empty and thus we produce a segment with **zero size**:
`LOAD           0x000478 0x0000000000402478 0x0000000000402478 0x000000 0x000000 RW  0x1000`

That's something the dynamic linker is unhappy about. The section itself is preserved because it contains 2 symbols that are referenced:

```
     4: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __rela_iplt_start
     5: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __rela_iplt_end

```

and so the following condition is met:
https://github.com/davidlattimore/wild/blob/de472c8f2e25559497e452f608f8ec7a1f1a4a34/wild_lib/src/layout.rs#L2411

Do you have idea how to address that?

marxin:
And the second failure comes from the linker-diff of `integration_test::program_name_25___libc_integration_c__` that might be again related to an empty section that Wild keeps:
`[ 6] .preinit_array       PREINIT_ARRAY 0000000000000000 00030c54 00000000  0        0   0  1`

```
  wild 0x00448a2e 48 83 3d 7a 8f 07 00 00 cmpq $0,0x790D2  // Cmp_rm64_imm8(0x4c19b0) GOT(__preinit_array_start) GOT(__preinit_array_end)
  ld   0x0040426e 48 83 3d 0a dd 0b 00 00 cmpq $0,0xBDE62  // Cmp_rm64_imm8(0x4c1f80) POINTER-TO(byte in (RW))
  ORIG            48 83 3d 00 00 00 00 00 cmpq $0,0x158  // R_X86_64_GOTPCREL -> `__pthread_initialize_minimal` -5
```

And suggestion what can be wrong?

davidlattimore:
For `program_name_12___internal_syms_c__`, it looks like the test was wrong. It actually was dereferencing the pointers that the comment said it shouldn't - it just happened to work before. I've pushed a fix for the test.

davidlattimore:
For the linker-diff failure, I think it's a problem in linker-diff rather than a problem in the linker. There's a bunch of work needed to improve the asm diffing. I think part of the problem in this case is that the instructions in question have have two operands, a 4 byte one and a 1 byte one. This results in the addend being 5, but linker-diff is probably assuming that it'll be 4. At least that's my guess. Previously probably both GNU ld and Wild's outputs were effectively failing to match anything and two non-matches are considered OK. But now Wild, by chance is happening to match to a symbol (or a couple of symbols) and that's causing a failure.

One possible fix would be to say that if the reference linker fails to match (`POINTER-TO(byte in (RW))`), then we should accept anything from Wild.

A bigger change would be to improve the matching to take better advantage of information from the original relocation.

It's also possible if we shift some sections around that we'll get lucky and the test will pass - although linker-diff will likely still need these fixes eventually.

It looks like you're changing a few things with this PR. There are sections moving about, additional debug sections being output and sections that previously had addresses assigned not being assigned. One option might be to split this PR into a few commits that make smaller changes. It might be that the change that is actually triggering this isn't one that you need now.

davidlattimore:
I think you could probably add it in `UnifiedInstruction::new`, just after the call to `extract_undefined_behaviour`. I'm off to bed. If you can't get it working and want me to look during my day tomorrow, let me know.

marxin:
Please do it for me tomorrow, I'm finishing for today.

PR #92, opened 2024-08-30
Fix bug in `copy_section_data`

marxin:


davidlattimore:
I've got to stop breaking your stuff. Sorry, thought I remembered making sure I preserved that fix when I merged, but apparently not. The good news is that I think I'm done changing that code for now.

marxin:
Yeah, that happens :) Hopefully, we'll have a test coverage soon as debug info support will be added.

PR #94, opened 2024-08-30
Strengthen validation for sections&segments in compute_segment_layout

marxin:
It would be rather nice to propagate an error (with `ensure` macro), but the callback used in `sections_and_segments_do` returns the unit type :/


davidlattimore:
> but the callback used in `sections_and_segments_do` returns the unit type :/

We could replace `sections_and_segments_do` with a `Vec<OrderEvent>` on `OutputSections`. Then code that currently calls `sections_and_segments_do` could iterate over the events however they liked, including returning errors.

marxin:
Sure, done in #95.

PR #105, opened 2024-09-03
Come up with `SymbolInfo` type and use it

marxin:
I'm still working on the Debug info support and I think I'll need filtering in `process_relocation` where a debug info section should be able to "load" only another debug info section. Otherwise, I'm going to end up with all symbols (and sections) linked by debug info. As mentioned, the garbage collected sections linked by debug info should be relocated by a tombstone value.

So the question is: Are you happy with the introduction of `symbol_info`, which I can use for the aforementioned filtering?

I'm also thinking about the possibility to use tracing in a classical way:

```rust
info!(
  object = object.input.file.filename.to_str().unwrap(),
  source_section = source_section_name,
  symbol = format!("{}", symbol_info),
  "send_symbol_request"
);
```

```
2024-09-03T11:56:56.053953Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.debug_abbrev`"
2024-09-03T11:56:56.053964Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.debug_line`"
2024-09-03T11:56:56.053969Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.text`"
2024-09-03T11:56:56.053973Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.debug_str`"
2024-09-03T11:56:56.053977Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.debug_line_str`"
2024-09-03T11:56:56.053983Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="section `.note.ABI-tag`"
2024-09-03T11:56:56.053987Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crt1.o" source_section=".debug_info" symbol="symbol `_IO_stdin_used`"
2024-09-03T11:56:56.053992Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crti.o" source_section=".debug_info" symbol="section `.debug_abbrev`"
2024-09-03T11:56:56.053996Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crti.o" source_section=".debug_info" symbol="section `.debug_line`"
2024-09-03T11:56:56.053999Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crti.o" source_section=".debug_info" symbol="section `.debug_rnglists`"
2024-09-03T11:56:56.054003Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crti.o" source_section=".debug_info" symbol="section `.debug_str`"
2024-09-03T11:56:56.054007Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crti.o" source_section=".init" symbol="symbol `__gmon_start__`"
2024-09-03T11:56:56.054011Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/crtbegin.o" source_section=".init_array" symbol="section `.text`"
2024-09-03T11:56:56.054017Z  INFO wild_lib::layout: send_symbol_request object="a-main.o" source_section=".debug_info" symbol="section `.debug_abbrev`"
2024-09-03T11:56:56.054020Z  INFO wild_lib::layout: send_symbol_request object="a-main.o" source_section=".debug_info" symbol="section `.debug_str`"
2024-09-03T11:56:56.054024Z  INFO wild_lib::layout: send_symbol_request object="a-main.o" source_section=".debug_info" symbol="section `.debug_line_str`"
2024-09-03T11:56:56.054027Z  INFO wild_lib::layout: send_symbol_request object="a-main.o" source_section=".debug_info" symbol="section `.text`"
2024-09-03T11:56:56.054031Z  INFO wild_lib::layout: send_symbol_request object="a-main.o" source_section=".debug_info" symbol="section `.debug_line`"
2024-09-03T11:56:56.054041Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crtn.o" source_section=".debug_info" symbol="section `.debug_abbrev`"
2024-09-03T11:56:56.054045Z  INFO wild_lib::layout: send_symbol_request object="/usr/lib64/gcc/x86_64-suse-linux/14/../../../../lib64/crtn.o" source_section=".debug_info" symbol="section `.debug_line`"
```

Can we add another subscribed that will be of type https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html? Or is it going to collide with the custom subscriber used in the `--time` option?

### DIFF
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -2296,6 +2296,15 @@ fn process_relocation(
         }
 
         if previous_flags.is_empty() {
+            let symbol_info = resources.symbol_db.symbol_info(symbol_id)?;
+            let source_section_name =
+                String::from_utf8_lossy(object.object.section_name(section).unwrap()).to_string();
+            println!(
+                "{} {} -> {}",
+                object.input.file.filename.to_str().unwrap(),
+                source_section_name,
+                symbol_info
+            );
             queue.send_symbol_request(symbol_id, resources);
         }
 
--- a/wild_lib/src/symbol_db.rs
+++ b/wild_lib/src/symbol_db.rs
@@ -25,6 +25,7 @@ use itertools::Itertools;
 use object::read::elf::Sym as _;
 use object::LittleEndian;
 use std::collections::hash_map;
+use std::fmt::Display;
 
 pub struct SymbolDb<'data> {
     pub(crate) args: &'data Args,
@@ -199,6 +200,24 @@ struct SymbolLoadOutputs<'data> {
     pending_symbols: Vec<PendingSymbol<'data>>,
 }
 
+pub(crate) enum SymbolInfo<'data> {
+    Symbol(SymbolName<'data>),
+    Section(SymbolName<'data>),
+    Unnamed(&'static str),
+}
+
+impl<'data> Display for SymbolInfo<'data> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SymbolInfo::Symbol(symbol_name) => write!(f, "symbol `{symbol_name}`")?,
+            SymbolInfo::Section(section_name) => write!(f, "section `{}`", section_name)?,
+            SymbolInfo::Unnamed(name) => write!(f, "{}", name)?,
+        }
+
+        Ok(())
+    }
+}
+
 impl<'data> SymbolDb<'data> {
     #[tracing::instrument(skip_all, name = "Build symbol DB")]
     pub fn build(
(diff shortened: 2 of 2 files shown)


marxin:
Related to #37.

davidlattimore:
I'm open to other uses of tracing. There's already two subscribers, although they can't be used at the same time. One potential pitfall is that extra tracing stuff can potentially slow down `--time`, making it less useful, since the total time no longer matches what you get if you run without `--time`. If that gets too bad though, we could always move away from tracing for reporting timing.

I'd have thought that if you want to not trigger loading of things that are referenced by debug sections, that you'd change `Section::create`  so that if the section is a debug section, it just ignores the relocations.

davidbarsky:
(Hi, I help maintain `tracing`; I've been watching y'all work on `wild` with a lot of excitement!)

> Can we add another subscribed that will be of type https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html? Or is it going to collide with the custom subscriber used in the `--time` option?

`EnvFilter` is... funky. I'd recommend reading env variables and using the more limited [Targets](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html?) instead.

> There's already two subscribers, although they can't be used at the same time.

I'm 95% sure that you are already aware of this, but on the very slim chance that you're not, you should be able to combine different _layers_ in a single subscriber by chaining `with` calls. The other thing to be aware of: `Option<Layer>` is also a layer, so might be able to unify `tracing` setup into a single call.

> One potential pitfall is that extra tracing stuff can potentially slow down --time, making it less useful, since the total time no longer matches what you get if you run without --time. If that gets too bad though, we could always move away from tracing for reporting timing.

Let me know if you do: I know we still have some low-hanging performance optimizations available to do first.

davidlattimore:
> I'm 95% sure that you are already aware of this, but on the very slim chance that you're not, you should be able to combine different _layers_ in a single subscriber by chaining `with` calls. The other thing to be aware of: `Option<Layer>` is also a layer, so might be able to unify `tracing` setup into a single call.

Thanks, yep. I'm happy with the way it's set up at the moment, since one use of tracing is for getting timing information and the other is for diagnosing problems. If I had the latter active at the same time, then it would make the timing information less accurate.

> Let me know if you do: I know we still have some low-hanging performance optimizations available to do first.

I had previously disabled anything below info-level for release builds, because the trace-level stuff was slowing down the timing. But I just went back and looked a bit closer and found that I hadn't implemented `max_level_hint` on my timing subscriber. Now that I've done that, I've reenabled trace-level for release. It does mean that I need to use INFO level for timing stuff and TRACE level or below for debugging, but that seems OK.

I saw about a 1% slowdown on release builds due to trace-level instrumentation no longer being compiled out, but I think that's acceptable for now.

marxin:
Then we can probably enable a default subscriber: #107.

PR #108, opened 2024-09-05
Simplify tracing emission of events and spans

marxin:


wild_lib/src/elf_writer.rs
```diff
@@ -1555,7 +1555,7 @@ fn apply_relocation(
 ) -> Result<RelocationModifier> {
     let section_address = section_info.section_address;
     let place = section_address + offset_in_section;
-    let _span = tracing::span!(tracing::Level::TRACE, "relocation", address = place).entered();
+    let _span = tracing::trace_span!("relocation", place).entered();
```
davidlattimore: I think removing `address = ` will break this, since the subscriber won't know that this is the address. See the code in `output_trace.rs`.

You can test that this is still working by making a change that will cause link-diff to fail one of the tests. e.g. comment out this code in `relaxation.rs`:

```rs
            object::elf::R_X86_64_PLT32 if can_bypass_got => {
                return create(RelaxationKind::NoOp, object::elf::R_X86_64_PC32);
            }
```

If it's working, you should see the `TRACE` line in the output. e.g.:

```
  wild 0x0040146d e8 de ff ff ff call 0xFFFFFFFFFFFFFFF0  // Call_rel32_64(0x401450) PLT(exit_syscall)
  ld   0x0040100d e8 03 00 00 00 call 0x0000000000000015  // Call_rel32_64(0x401015) exit_syscall
  lld  0x00201219 e8 03 00 00 00 call 0x0000000000000015  // Call_rel32_64(0x201221) exit_syscall
  ORIG            e8 00 00 00 00 call 0x0000000000000012  // R_X86_64_PLT32 -> `exit_syscall` -4
  TRACE           value_flags=ADDRESS | CAN_BYPASS_GOT resolution_flags=DIRECT | GOT | PLT
```

marxin: Whoops, restored.

PR #113, opened 2024-09-09
Support for DWARF debug info

marxin:
Implements #37.

### DIFF
--- a/wild_lib/src/elf_writer.rs
+++ b/wild_lib/src/elf_writer.rs
@@ -1222,7 +1222,6 @@ impl<'data> ObjectLayout<'data> {
         sec: &Section,
         buffers: &mut OutputSectionPartMap<&mut [u8]>,
     ) -> Result {
-        let _span = debug_span!("write_debug_section", section = %self.object.section_display_name(sec.index)).entered();
         let out = self.write_section_raw(layout, sec, buffers)?;
         self.apply_debug_relocations(out, sec, layout)
             .with_context(|| {
@@ -1379,7 +1378,6 @@ impl<'data> ObjectLayout<'data> {
         section: &Section,
         layout: &Layout,
     ) -> Result {
-        tracing::debug!("apply_debug_relocations");
         let section_address = self.section_resolutions[section.index.0]
             .as_ref()
             .unwrap()
@@ -1758,13 +1756,6 @@ fn apply_debug_relocation(
     {
         resolution
     } else if let Some(section_index) = object_layout.object.symbol_section(sym, symbol_index)? {
-        // TODO: remove
-        let section_name = String::from_utf8_lossy(
-            object_layout
-                .object
-                .section_name(object_layout.object.section(section_index)?)?,
-        );
-        tracing::debug!(%section_name);
         object_layout.section_resolutions[section_index.0].unwrap_or({
             // TODO: it's likely a string merge section
             Resolution {
@@ -1783,7 +1774,6 @@ fn apply_debug_relocation(
     let addend = rel.r_addend.get(e) as u64;
     let r_type = rel.r_type(e, false);
     let rel_info = RelocationKindInfo::from_raw(r_type)?;
-    tracing::trace!(%resolution.value_flags, %resolution.resolution_flags);
 
     let value = match rel_info.kind {
         RelocationKind::Absolute => resolution.value_with_addend(
--- a/wild_lib/src/elf_writer.rs
+++ b/wild_lib/src/elf_writer.rs
@@ -1131,7 +1131,7 @@ impl<'data, 'out> SymbolTableWriter<'data, 'out> {
     }
 }
 
-impl<'data> ObjectLayout<'data> {
+impl<'out> ObjectLayout<'out> {
     fn write_file(
         &self,
         buffers: &mut OutputSectionPartMap<&mut [u8]>,
@@ -1239,8 +1239,8 @@ impl<'data> ObjectLayout<'data> {
         &self,
         layout: &Layout,
         sec: &Section,
-        buffers: &'data mut OutputSectionPartMap<&mut [u8]>,
-    ) -> Result<&'data mut [u8]> {
+        buffers: &'out mut OutputSectionPartMap<&mut [u8]>,
+    ) -> Result<&'out mut [u8]> {
         if layout
             .output_sections
             .has_data_in_file(sec.output_section_id())
@@ -1258,7 +1258,7 @@ impl<'data> ObjectLayout<'data> {
             // Cut off any padding so that our output buffer is the size of our input buffer.
             let object_section = self.object.section(sec.index)?;
             let section_size = self.object.section_size(object_section)?;
-            let out: &'data mut [u8] = &mut out[..section_size as usize];
+            let out: &'out mut [u8] = &mut out[..section_size as usize];
             self.object.copy_section_data(object_section, out)?;
             Ok(out)
         } else {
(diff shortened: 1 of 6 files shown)


wild_lib/src/elf_writer.rs
```diff
@@ -1661,6 +1735,79 @@ fn apply_relocation(
...
+    } else if let Some(section_index) = object_layout.object.symbol_section(sym, symbol_index)? {
+        // TODO: remove
+        let section_name = String::from_utf8_lossy(
+            object_layout
+                .object
+                .section_name(object_layout.object.section(section_index)?)?,
+        );
+        tracing::debug!(%section_name);
+        object_layout.section_resolutions[section_index.0].unwrap_or({
+            // TODO: it's likely a string merge section
```
marxin: @davidlattimore  Where should I grab the resolution (other than manually constructing that) for the string merge debug info sections?

davidlattimore: Can you call Resolution::value_with_addend, or alternatively directly call get_merged_string_output_address? For at least some relocations, we're already computing offsets into string-merge sections at ELF writing time, and that happens via those two functions

davidlattimore: I guess probably `get_merged_string_output_address`, since you don't have a resolution - which makes sense, since non-debug relocations wouldn't have referenced debug strings, so they wouldn't have resolutions.

marxin: All right, I've just rewritten `apply_debug_relocation` function where I use the suggested function.

wild_lib/src/elf_writer.rs
```diff
@@ -1191,6 +1197,50 @@ impl<'data> ObjectLayout<'data> {
...
+                )
+            })?;
+        Ok(())
+    }
+
+    fn write_section_raw(
+        &self,
+        layout: &Layout,
+        sec: &Section,
+        buffers: &'data mut OutputSectionPartMap<&mut [u8]>,
```
davidlattimore: Elsewhere, I've used 'out for lifetimes of the output and 'data for inputs

wild_lib/src/resolution.rs
```diff
@@ -788,7 +794,13 @@ fn resolve_sections<'data>(
                                 section_flags,
                                 ty: SectionType::from_header(input_section),
                             });
-                            if section_flags.should_retain() {
+                            if custom_section_id.name.bytes().starts_with(b".debug_") {
+                                if !args.strip_debug {
```
davidlattimore: I think this would be slightly clearer if you took out the `!` and swapped the `if` and `else` branches

wild_lib/src/elf_writer.rs
```diff
@@ -1661,6 +1734,89 @@ fn apply_relocation(
...
+                &layout.merged_string_start_addresses,
+                false,
+            )?
+            .context("Cannot get merged string offset for a debug info section")?,
+            SectionSlot::Discard | SectionSlot::Unloaded(..) => {
+                let section_name = object_layout
+                    .object
+                    .section_name(object_layout.object.section(section_index)?)?;
+                // Return debug info tombstone value.
+                if section_name == b".debug_loc" || section_name == b".debug_ranges" {
```
davidlattimore: Fetching the section name and doing a comparison for every relocation that needs this could be a performance issue. Consider computing this outside the loop in `apply_debug_relocations`.

wild_lib/src/layout.rs
```diff
@@ -2980,14 +2979,19 @@ impl<'data> ObjectLayoutState<'data> {
...
                         self.object.section_display_name(section_id),
                     );
                 }
-                SectionSlot::Loaded(_) | SectionSlot::EhFrameData(..) => {}
-                SectionSlot::MergeStrings(_) => {
+                SectionSlot::Loaded(_)
+                | SectionSlot::EhFrameData(..)
+                | SectionSlot::MergeStrings(_)
+                | SectionSlot::LoadedDebugInfo(..) => {
                     // We currently always load everything in merge-string sections. i.e. we don't
```
davidlattimore, line 2995: Before, this comment was only in the `MergeStrings` arm of the match, but now it's in the arm for multiple rules. Perhaps move it before `| SectionSlot::MergeStrings(_)`?

wild_lib/src/elf_writer.rs
```diff
@@ -1661,6 +1734,89 @@ fn apply_relocation(
...
+                symbol_index,
+                object_layout,
+                &layout.merged_strings,
+                &layout.merged_string_start_addresses,
+            )?,
+            RelocationKind::DtpOff => resolution
+                .value()
+                .wrapping_sub(layout.tls_end_address())
+                .wrapping_add(addend),
+            _ => todo!(),
```
davidlattimore: It'll be quicker to figure out what's going on here if we instead do something like: `kind => bail!("Unsupported debug relocation kind {kind:?}")`

davidlattimore (commented):
Nice work! Just a few minor comments.

Also, do you want this PR squashed, or rebased? I'm happy to rebase multi-commit PRs, provided all the commits pass tests and have commit messages that make sense by themselves. Feel free to amend commits if you'd like me to rebase. Or if you're not sure the commits should be separate, then I'm also happy to just squash them.

marxin:
> Also, do you want this PR squashed, or rebased?

Please merge all the commits as the changes among individual commits are likely not useful in this pull request.

marxin:
> Please merge (squash) all the commits as the changes among individual commits are likely not useful in this pull request.

Apparently, you forget to do the squash merge ;) Never mind, hopefully, my commits haven't polluted the git history much.

davidlattimore:
> Apparently, you forget to do the squash merge ;) Never mind, hopefully, my commits haven't polluted the git history much.

Ooops. Sorry! I'm sure I remember selecting an option for how to merge. Apparently I need to be more careful - especially when it's late and I'm about to go to bed. I checked the itnermediate commits and there was at least one where there was a failing test. I possibly shouldn't have done this, but I squashed them just now after-the-fact.

In case it's relevant these are the updated commit hashes:

60ed2576c10c9aadb49a2ccb3102cb0e0643c9b5 -> ebdb566980a7ad5272f4c3eae0bf7d8073cef830
47b6ac86bb0251bddac2a2110fce1d93068360d3 -> 8eba39e589bee20e5982aecd3e975881c3d96199

wild_lib/src/elf_writer.rs
```diff
@@ -1661,6 +1753,82 @@ fn apply_relocation(
...
+                addend,
+                object_layout.object,
+                &object_layout.sections,
+                &layout.merged_strings,
+                &layout.merged_string_start_addresses,
+                false,
+            )?
+            .context("Cannot get merged string offset for a debug info section")?,
+            SectionSlot::Discard | SectionSlot::Unloaded(..) => {
+                section_tombstone_values[section_index.0]
```
davidlattimore, line 1815: Sorry I didn't notice this pre-submission, but is this right? It looks like you're checking if a section that was discarded or not loaded is one of the two special debug sections. But won't those debug sections be unconditionally loaded? I'd have thought that the tombstone value would value would depend on the section to which the relocation applied (source) rather than the section to which the relocation referred (target).

If that's the case, then you wouldn't need to create a `Vec` of tombstone values because all the tombstones for the section we're currently processing would have the same tombstone.

marxin: Yep, good catch. The tombstone must be related to the place where we apply relocations (source section).
Please merge: #115.

Issue #117, opened 2024-09-10
Wild performance with debug info support


marxin:
Once the recent support for DWARF has been added, we're falling behind mold linker when it comes to linking of something huge (like `clang`):

```
❯ bloaty ../../../../bin/clang-20
    FILE SIZE        VM SIZE    
 --------------  -------------- 
  74.5%  3.31Gi   0.0%       0    .debug_info
   9.4%   427Mi   0.0%       0    .debug_loclists
   5.1%   231Mi   0.0%       0    .debug_str
   4.9%   220Mi   0.0%       0    .debug_line
   2.0%  91.3Mi  61.6%  91.3Mi    .text
   1.7%  76.4Mi   0.0%       0    .debug_rnglists
   0.9%  41.5Mi  28.0%  41.5Mi    .rodata
   0.5%  23.6Mi   0.0%       0    .debug_abbrev
   0.5%  21.7Mi   0.0%       0    .strtab
   0.2%  9.01Mi   6.1%  9.01Mi    .eh_frame
   0.1%  4.89Mi   0.0%       0    .symtab
   0.1%  4.48Mi   3.0%  4.48Mi    .data
   0.1%  3.64Mi   0.0%       0    .debug_aranges
   0.0%  1.09Mi   0.7%  1.09Mi    .eh_frame_hdr
   0.0%       0   0.5%   704Ki    .bss
   0.0%   365Ki   0.0%       0    .debug_line_str
   0.0%  43.4Ki   0.0%  43.4Ki    .rela.dyn
   0.0%  14.8Ki   0.0%  14.4Ki    [17 Others]
   0.0%  8.21Ki   0.0%  8.21Ki    .dynstr
   0.0%  7.62Ki   0.0%  7.62Ki    .dynsym
   0.0%  5.77Ki   0.0%  5.77Ki    .got
 100.0%  4.44Gi 100.0%   148Mi    TOTAL

> wild ...
┌───    2.43 Open input files
├───    1.62 Split archives
├───    3.53 Parse input files
│ ┌───    2.76 Read symbols
│ ├───    7.66 Populate symbol map
├─┴─   11.47 Build symbol DB
│ ┌─── 1719.15 Resolve symbols
│ ├───    1.25 Assign section IDs
│ ├─── 2206.91 Merge strings
│ ├───    0.04 Canonicalise undefined symbols
│ ├───    6.99 Resolve alternative symbol definitions
├─┴─ 3934.42 Symbol resolution
[...]

PR #118, opened 2024-09-10
Implement parallel string merging using buckets

marxin:
Implements: #117

With the provided change, I get to: `583.72 Merge strings` from `2206.91 Merge strings`.


marxin:
Plus, I had to simplify `ExpectComment` directive as the comment order does not respect an order for a given object file.

Issue #24, opened 2024-08-05
`.rodata.*.str1.1` relocation is wrong


marxin:
I noticed that while building of the `mold` linker with `wild`:

The problematic code snippet lives in `main.cc`:

```assembly
	movl	$.LC96, %esi
	call	_ZStlsISt11char_traitsIcEERSt13basic_ostreamIcT_ES5_PKc.isra.0
...
	.section	.rodata._ZN4mold3elf8elf_mainINS0_6X86_64EEEiiPPc.str1.1,"aMS",@progbits,1
.LC92:
	.string	"-run"
.LC93:
	.string	"--run"
.LC94:
	.string	"--Bstatic"
.LC95:
	.string	"--Bdynamic"
.LC96:
	.string	"-m option is missing"
.LC97:
	.string	"all"
.LC98:
	.string	"chdir failed: "
```

Reproducer branch: https://github.com/marxin/wild/tree/reproduce-rodata-issue

```
$ docker build --progress=plain -t wild-dev-opensuse . -f docker/opensuse.Dockerfile
...
#30 [builder 18/19] RUN /mold/build/mold || true
#30 0.223 mold: fatal: -m option is missing
#30 DONE 0.2s

#31 [builder 19/19] RUN /mold/build-wild/mold || true
#31 0.276 mold: fatal: chdir failed: 
#31 DONE 0.4s
```

As seen, the `mold` binary built with binutils' ld correctly prints `-m option is missing`, but the binary built with `wild` prints: `chdir failed:` (which is a string label that belongs to a different label `.LC98`).

Can you take a look, please?

marxin:
All right, I've got a much smaller reproducer:

```shell
$ cat a.cc
#include <iostream>
using namespace std;

void bar();
void foo() {
  cout << "a" << ":" << "b" << ":" << "c" << endl;
}

int main() {
  foo();
  bar();

  return 0;
}

$ cat b.cc
#include <iostream>

using namespace std;

void bar() {
  cout << "a2" << ":" << "b2" << ":" << "c2" << endl;
}
$ g++ a.cc b.cc -O2 && ./a.out
a:b:c
a2:b2:c2
$ g++ a.cc b.cc -O2 -B ~/Programming/wild && ./a.out
a:b:c
a2bcb
```

davidlattimore:
Thanks for the awesome bug report!

[...]

PR #122, opened 2024-09-11
Do not walk strings in string-merge sections

marxin:
Are you aware of a test-case where there would be a pointer to the middle of a string? It seems to me weird walking back to a string beginning.

### DIFF
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -3838,14 +3838,6 @@ pub(crate) fn get_merged_string_output_address(
         );
     }
 
-    // The offset might point part way through a string. Look backward to find the start of the
-    // string so that we look up the whole string. We then reverse that offset in our final position
-    // computation.
-    let mut offset_into_string = 0;
-    while input_offset > 0 && data[input_offset as usize - 1] != 0 {
-        input_offset -= 1;
-        offset_into_string += 1;
-    }
     let string = StringToMerge::take_hashed(&mut &data[input_offset as usize..])?;
     let section_id = merge_slot.part_id.output_section_id();
     let strings_section = merged_strings.get(section_id);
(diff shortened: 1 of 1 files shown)


davidlattimore:
The commit where I introduced that logic was for issue #24. I just tried the minimal repro that you had there and it seemed fine with your change. So I'm happy to try removing that logic and see how we go. If it turns out that we do actually need it, we can always add it back in. At least if it turns out it is needed, we'll get a link error rather than a bad executable.

marxin:
I've just noticed Wild does not support the following test-case:
https://github.com/rui314/mold/blob/main/test/arch-x86_64-mergeable-strings.sh

It hasn't been supported before this PR. The question is if it's a valid test-case.

davidlattimore:
I assume you're saying that we did support that test, but with this PR, we don't?

Before I merged this PR, I did attempt to get a test that produced an object file with a reference to the middle of a string in a string-merge section. I wasn't able to using just C. I tried GCC and Clang with `-O3` in the hopes that the compiler might attempt something like that, but it didn't. If it requires hand-crafted assembly in order to get it to happen, then it doesn't seem super-important to support. Although I don't object to bringing back support for this if you'd like. If we're concerned about the performance of doing the start-of-string search, one option would be to only do the start-of-string search after we've attempted the hashmap lookup and failed.

marxin:
> I assume you're saying that we did support that test, but with this PR, we don't?

Yes :)

Sure, let's ask the test-case author: @rui314: Is the mentioned test-case `arch-x86_64-mergeable-strings.sh` hand-crafted or something you've seen emitted by a compiler? Thanks.

rui314:
I don't remember what compiler it was, but there was a compiler that generates code like that. You may not encounter this with gcc or clang, but when people started using your linker with Intel Compiler, the D compiler, OCaml, Fortran, Glasgow Haskell Compiler, etc., I'm sure you'll get this bug report.

marxin:
All right, so we might face it in the future. To be honest, I'm not sure if GHC can work with Wild compiler (AFAIK it does not support Mold either). For gfortran I would expect similar emitted code as from gcc, g++ as it's using the very same code generator. Similarly for Intel Compiler which is LLVM-bassed.

@davidlattimore I'm leaving the decision up to you if you want to revert the change. What's your approach based on the information provided by Rui?

PR #126, opened 2024-09-12
Switch to zstd Rust library (C binding of the official libzstd)

marxin:
The linking for Clang binary is now 30% faster!

Before:
```
  Time (mean ± σ):      6.132 s ±  0.092 s    [User: 109.349 s, System: 13.627 s]
  Range (min … max):    5.994 s …  6.292 s    10 runs
```

After:
```
  Time (mean ± σ):      4.736 s ±  0.144 s    [User: 74.308 s, System: 13.071 s]
  Range (min … max):    4.494 s …  4.971 s    10 runs
```


davidlattimore:
Nice!

PR #127, opened 2024-09-12
Use itertools

marxin:
In the future, we might want to use the following function: https://github.com/rust-itertools/itertools/issues/934.

### DIFF
--- a/linker-diff/src/header_diff.rs
+++ b/linker-diff/src/header_diff.rs
@@ -8,6 +8,7 @@ use crate::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context as _;
+use itertools::Itertools;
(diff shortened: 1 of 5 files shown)


davidlattimore:
I generally try to avoid putting partially inferred types (e.g. `Vec<_>`) on variables, since then rust-analyzer doesn't show the full type inline. e.g. in here we can see that `files` is a `Vec<FileLayout<'data>>`

image

But with a partial type, we can't:

image

Not a massive problem, but that's why I tend to use the turbofish perhaps more than some other people. I'm not strongly opposed to doing it with partially inferred types if you really want to though - I can still get the inferred type if I hover, it's just not always there like it is for a fully inferred type.

marxin:
Sure, I fully understand your point. I've removed all the `try_collect` calls from the PR.

PR #140, opened 2024-09-13
Clang by default in integration tests

marxin:
This is mostly an experimental branch right now. However, I would like to analyze the fallout reported by the linker-diff. Can you take a look @davidlattimore? Is there something interesting we should fix?

### DIFF
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -148,6 +148,7 @@ impl Config {
                 // GNU ld sometimes makes this writable sometimes not. Presumably this depends on
                 // whether there are relocations or some flags.
                 "section.eh_frame.flags",
+                "section.eh_frame.type",
             ]
             .into_iter()
             .map(|s| s.to_owned()),
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -273,7 +273,7 @@ impl Default for Config {
             section_equiv: Default::default(),
             is_abstract: false,
             deps: Default::default(),
-            compiler: "gcc".to_owned(),
+            compiler: "clang".to_owned(),
             should_diff: true,
             should_run: true,
         }
--- a/wild/tests/sources/comments.c
+++ b/wild/tests/sources/comments.c
@@ -19,6 +19,5 @@ void _start(void) {
     exit_syscall(42);
 }
 
-//#ExpectComment:GCC*
 //#ExpectComment:Foo
 //#ExpectComment:Bar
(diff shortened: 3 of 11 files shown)


davidlattimore:
Looking through the failures...

program_name_22___comments_c__: It looks like clang ignores `#ident`, whereas GCC puts the following string into the `.comment` section.

`program_name_04___global_definitions_c__` gives a few errors:

```
  wild 0x00401498 48 c7 c0 30 25 40 00 mov $0x402530,%rax  // Mov_rm64_imm32(0x402530) global_value
  ld   0x00401028 48 c7 c0 00 40 40 00 mov $0x404000,%rax  // Mov_rm64_imm32(0x404000) global_value
  lld  0x00201278 48 8d 05 91 20 00 00 lea 0x20A0,%rax  // Lea_r64_m(0x203310) global_value
  ORIG            48 8b 05 00 00 00 00 mov 0xF,%rax  // R_X86_64_REX_GOTPCRELX -> `global_value` -4
  TRACE           relaxation.kind=RexMovIndirectToAbsolute value_flags=ADDRESS | CAN_BYPASS_GOT resolution_flags=DIRECT
```

Here, Wild and LLD are loading the address of `global_value` as absolute values. LLD is loading the address of `global_value` as a relative value. The executable being produced is non-relocatable, so either is fine. I thought that I had code in linker-diff to allow this to pass if wild matches either `ld` or `lld`, so I would have thought this should have passed. Perhaps a bug in linker-diff.

```
section.got
  wild Section missing
  ld Section missing
  lld OK
```

Wild and LD both skip emitting the `.got` section. lld didn't omit the GOT because it put something in it. That something was the value of the absolute symbol `abs1`. lld and Wild both inlined the value of that symbol into the instruction where it was referenced.

The instructions with `abs1` inlined (ld and wild) looks like this:

```asm
 4014cd:       48 c7 c0 19 00 00 00    mov    $0x19,%rax
  4014d4:       89 c7                   mov    %eax,%edi
  4014d6:       e8 95 ff ff ff          call   401470 <black_box>
```

The instructions with it non-inlined (lld) looks like:

```asm
  2012ad:       48 8b 05 54 10 00 00    mov    0x1054(%rip),%rax        # 202308 <exit_syscall+0x1028>
  2012b4:       89 c7                   mov    %eax,%edi
  2012b6:       e8 95 ff ff ff          call   201250 <black_box>
```

```
section.relro_padding
  wild Section missing
  ld Section missing
  lld OK
```

I'm not sure why lld emitted `.relro_padding`, but no other RELRO related stuff. It's not a big problem though and should be safe to ignore.

The failures in `program_name_20___common_section_c__` look similar to the above failures.

`program_name_29___cpp_integration_cc__` exits with code 10 after printing the string `ELFELFELFELFELFELFELF`. This looks like a bug in string merging.

davidlattimore:
I filed #180 for the string-merge bug and #179 for making it easier to override the compiler.

PR #146, opened 2024-09-14
Support `.note.ABI-tag` section (and `NOTE` segment)

marxin:


marxin:
@davidlattimore Please review it once the conflict is resolved.

Issue #104, opened 2024-09-03
Occasional test flake due to "Text file busy"


davidlattimore:
I saw this yesterday once after I pushed some changes: https://github.com/davidlattimore/wild/actions/runs/10659030447/job/29540980339

then again just now on my laptop in the openSUSE docker image.

```
---- integration_test::program_name_29___rust_tls_rs__ stdout ----
Error: Failed to build program `rust-tls.rs` with linker `ld` config `global-dynamic`

Caused by:
    0: Failed to run command: OUT="/wild/wild/tests/build/rust-tls.rs-global-dynamic.ld" WILD_VALIDATE_OUTPUT="1" WILD_WRITE_LAYOUT="1" WILD_WRITE_TRACE="1" "/wild/wild/tests/build/rust-tls.global-dynamic-e9236c1fda854ff6/run-with" "/usr/bin/ld"
    1: Text file busy (os error 26)
```

I'm not sure how this can happen. It'd make sense if one thread tried to write that file concurrently with another thread executing it. However I tried stracing running the tests and it looks like the file only gets written once and executed once. Both of these should have been from the same thread so shouldn't have been concurrent.

marxin:
How often can you reproduce the issue on your laptop? I can't to be honest.

What about using `cargo t -j1`, does the issue demonstrate? Can you attach `strace` log, please?

davidlattimore:
I hadn't seen this happen again since I filed this issue, however I just tried running the following:

```sh
while strace --string-limit=1000 -f -ttt -o strace.log cargo test; do rm -rf wild/tests/build; done
```

It eventually failed with:

```
---- integration_test::program_name_30___rust_tls_rs__ stdout ----
Error: Failed to build program `rust-tls.rs` with linker `ld` config `initial-exec`

Caused by:
    0: Failed to run command: OUT="/home/david/work/wild/wild/tests/build/rust-tls.rs-initial-exec.ld" WILD_VALIDATE_OUTPUT="1" WILD_WRITE_LAYOUT="1" WILD_WRITE_TRACE="1" "/home/david/work/wild/wild/tests/build/rust-tls.initial-exec-8bcd8b52ca283440/run-with" "/usr/bin/ld"
    1: Text file busy (os error 26)
```

From what I've observed, the failure always seems to be in the `rust_tls_rs` test, but which config of that test fails varies. If I run just that test rather than all tests, it doesn't seem to fail - or maybe I just didn't let it run for long enough.

[...]

PR #150, opened 2024-09-15
Run script in bash sub-shell

marxin:
Fixes: #104


marxin:
I've run the following stress test (`while cargo test; do rm -rf wild/tests/build; done`) and I haven't reached the "Text file busy" issue.

Issue #158, opened 2024-09-22
Wrong caching for debug info in get_merged_string_output_address since  `8e608131778e`


marxin:
I noticed that while preparing a `gimli` verification check for `linker-diff`.
If I build a simple C program, I get the following with Wild:

```
$ eu-readelf -w a.out
...
DWARF section [23] '.debug_info' at offset 0x1180:
 [Offset]
 Compilation unit at offset 0:
 Version: 5, Abbreviation section offset: 0, Address size: 8, Offset size: 4
 Unit type: compile (1)
 [     c]  compile_unit         abbrev: 1
           stmt_list            (sec_offset) 0
           low_pc               (addr) 0x0000000000401d00 <_start>
           high_pc              (udata) 38 (0x0000000000401d26)
           name                 (strp) "../sysdeps/x86_64/start.S"
           comp_dir             (strp) "../sysdeps/x86_64/start.S"
           producer             (strp) "../sysdeps/x86_64/start.S"
           language             (data2) Mips_Assembler (32769)
 [    28]    subprogram           abbrev: 2
             name                 (strp) "../sysdeps/x86_64/start.S"
             external             (flag_present) yes
             type                 (ref_udata) [    37]
             low_pc               (addr) 0x0000000000401d00 <_start>
             high_pc              (udata) 38 (0x0000000000401d26)
 [    37]    unspecified_type     abbrev: 3
```

while the correct output should look like:
```
DWARF section [23] '.debug_info' at offset 0x1180:
 [Offset]
 Compilation unit at offset 0:
 Version: 5, Abbreviation section offset: 0, Address size: 8, Offset size: 4
 Unit type: compile (1)
 [     c]  compile_unit         abbrev: 1
           stmt_list            (sec_offset) 0
           low_pc               (addr) 0x0000000000401d00 <_start>
           high_pc              (udata) 38 (0x0000000000401d26)
           name                 (strp) "../sysdeps/x86_64/start.S"
           comp_dir             (strp) "/home/abuild/rpmbuild/BUILD/glibc-2.40/csu"
           producer             (strp) "GNU AS 2.42.0"
           language             (data2) Mips_Assembler (32769)
 [    28]    subprogram           abbrev: 2
             name                 (strp) "_start"
             external             (flag_present) yes
             type                 (ref_udata) [    37]
             low_pc               (addr) 0x0000000000401d00 <_start>
             high_pc              (udata) 38 (0x0000000000401d26)
[...]

PR #160, opened 2024-09-23
Include `addend` in calculation for `string_offset_cache`

marxin:
With the applied patch, `readelf -w` is identical for a simple C test-case w/ and w/o the caching.

Fixes: #158


davidlattimore:
Ah, that makes sense. Thanks!

PR #163, opened 2024-09-23
Remove string-merge caching

davidlattimore:
When it was added, it appeared to speed up string merging, however it was buggy and once the bug(s) were fixed, it turns out that it actually makes things slower.

Issue #117


DrewRidley:
Related to this:

https://reviews.llvm.org/D27146

I found this LLD patch that might be helpful. For debug builds, as long as you are willing to sacrifice on determinism, you can get most of the strings merged without requiring coordination between threads.

davidlattimore:
Thanks @DrewRidley. That's true, sacrificing non-determinism would indeed open up several options for faster string merging. A non-deterministic mode as an opt-in might be OK, but I'd definitely want to exhaust all other possibilities for improving deterministic string-merging first and I don't think we've done that.

PR #192, opened 2024-11-22
WIP Sub-process forking

andrewdavidmackenzie:
Initial commits as a prototype implementation using fork. 

This doesn't have the child communicating yet and is just to explore the methods around named pipes and forking.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = ["linker-diff", "linker-layout", "linker-trace", "linker-utils", "wild", "wild_lib"]
+default-members = ["wild"]
 
 resolver = "2"
 
--- a/wild/Cargo.toml
+++ b/wild/Cargo.toml
@@ -5,6 +5,9 @@ edition = "2021"
 
 [dependencies]
 wild_lib = { version = "0.2.0", path = "../wild_lib" }
+fork = "0.2.0"
+anyhow = "1.0.93"
+libc = "0.2"
 
 [dev-dependencies]
 anyhow = "1.0.93"
--- a/wild/src/main.rs
+++ b/wild/src/main.rs
@@ -1,4 +1,7 @@
+use std::env::args;
+
 fn main() -> wild_lib::error::Result {
-    let linker = wild_lib::Linker::from_env()?;
+    // Supply args to wildlib, skipping the program name
+    let linker = wild_lib::Linker::from_args(args().skip(1))?;
     linker.run()
 }
--- a/wild/src/main.rs
+++ b/wild/src/main.rs
@@ -1,7 +1,80 @@
+use anyhow::anyhow;
+use fork::{fork, Fork};
 use std::env::args;
+use std::ffi::CString;
+use std::fs::File;
+use std::io::Error;
+use std::io::{Read, Write};
 
 fn main() -> wild_lib::error::Result {
-    // Supply args to wildlib, skipping the program name
-    let linker = wild_lib::Linker::from_args(args().skip(1))?;
-    linker.run()
+    // skip the program name
+    let mut args: Vec<String> = args().skip(1).collect();
+
+    // See if there is a better way to do this, more canonical
+    let mut fork_subprocess = false;
+
+    // TODO decide if we want some short-form equivalent of "--fork"
+    args.retain(|a| {
+        if a == "--fork" {
+            fork_subprocess = true;
+            false
+        } else {
+            true
+        }
+    });
+
+    match fork_subprocess {
+        true => match make_named_pipe() {
+            Ok(path) => {
+                match fork() {
+                    Ok(Fork::Child) => {
+                        // Create a linker with remaining args and run it
+                        wild_lib::Linker::from_args(args.into_iter())?.run()?;
+
+                        // inform parent that we are done!
+                        let mut f = File::open(path)?;
+                        f.write_all("OK".as_bytes())?;
+                        Ok(())
+                    }
+                    Ok(Fork::Parent(_child)) => {
+                        // Wait for child to exit or pipe to be closed
+                        let mut f = File::open(path)?;
+                        let mut response = String::new();
+                        let child_exit_status = f.read_to_string(&mut response)?;
+                        println!("Child Exit Status: {child_exit_status}");
+                        Ok(())
+                    }
+                    Err(_) => {
+                        // Create a linker with remaining args and run it
+                        wild_lib::Linker::from_args(args.into_iter())?.run()
+                    }
+                }
+            }
+            Err(_) => {
+                // Create a linker with remaining args and run it
+                wild_lib::Linker::from_args(args.into_iter())?.run()
+            }
+        },
+        false => {
+            // Create a linker with remaining args and run it
+            wild_lib::Linker::from_args(args.into_iter())?.run()
+        }
+    }
+}
+
+/// Create a named pipe for communication between parent and child processes.
+/// If successful it will return Ok with the name of the file
+/// If errors it will return an error message with the errno set, if it can be read or -1 if not
+fn make_named_pipe() -> wild_lib::error::Result<String> {
+    let path = "named_pipe";
+    let filename = CString::new(path)?;
+    unsafe {
+        match libc::mkfifo(filename.as_ptr(), 0o644) {
+            0 => Ok(path.to_owned()),
+            _ => Err(anyhow!(
+                "Error creating named pipe. Errno = {:?}",
+                Error::last_os_error().raw_os_error().unwrap_or(-1)
+            )),
+        }
+    }
 }
--- a/wild_lib/src/lib.rs
+++ b/wild_lib/src/lib.rs
@@ -1,3 +1,4 @@
+use crate::args::parse;
 use args::Args;
 use tracing_subscriber::fmt;
 use tracing_subscriber::layer::SubscriberExt;
@@ -56,9 +57,9 @@ pub struct Linker {
 }
 
 impl Linker {
-    pub fn from_env() -> crate::error::Result<Self> {
+    pub fn from_args<S: AsRef<str>, I: Iterator<Item=S>>(args: I) -> error::Result<Self> {
         Ok(Linker {
-            action: crate::args::from_env()?,
+            action: parse(args)?,
         })
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -345,6 +345,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fork"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05dc8b302e04a1c27f4fe694439ef0f29779ca4edc205b7b58f00db04e29656d"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "futures"
 version = "0.3.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1115,7 +1124,9 @@ name = "wild"
 version = "0.2.0"
 dependencies = [
  "anyhow",
+ "fork",
  "itertools",
+ "libc",
  "linker-diff",
  "object",
  "rstest",
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -134,11 +134,6 @@ const IGNORED_FLAGS: &[&str] = &[
     "no-relax",
 ];
 
-pub(crate) fn from_env() -> Result<Action> {
-    // Skip program name here
-    parse(std::env::args().skip(1))
-}
-
 // Parse the supplied input arguments, which should not include the program name.
 #[allow(clippy::if_same_then_else)]
 pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Result<Action> {
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -1,7 +1,7 @@
-//! A hand-written parser for our arguments.
+//! A handwritten parser for our arguments.
 //!
 //! We don't currently use a 3rd party library because
-//! order is important for some arguments and it's not clear how easy it would be to get that
+//! order is important for some arguments, and it's not clear how easy it would be to get that
 //! correct with something like clap.
 
 use crate::error::Result;
@@ -45,7 +45,7 @@ pub(crate) struct Args {
     /// If set, GC stats will be written to the specified filename.
     pub(crate) write_gc_stats: Option<PathBuf>,
 
-    /// If set and we're writing GC stats, then ignore any input files that contain any of the
+    /// If set, and we're writing GC stats, then ignore any input files that contain any of the
     /// specified substrings.
     pub(crate) gc_stats_ignore: Vec<String>,
 
@@ -655,7 +655,7 @@ mod tests {
     use std::path::PathBuf;
     use std::str::FromStr;
 
-    const INPUT1: &[&str] = &[
+    const INPUT1: Vec<&str> = vec![
         "-pie",
         "-z",
         "relro",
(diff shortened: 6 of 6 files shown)


andrewdavidmackenzie:
I'm running afoul of the Item import sepcified in rustfmt.toml.

I'll see how I can get my editor to take heed of that and respect the existing format.

davidlattimore:
> A lot of the arg parsing is in the lib, and could be argued that that should be in the binary, and a struct passed to the lib (moving parse etc) but I wouldn't make such a change in this PR or just because of the sub-process change.

Generally argument parsing should be done in a binary not in a library, unless users of the library are likely to need that argument parsing too, which is often the case with things like linkers and compilers. For example, if the Wild linker were to be built into a compiler like `rustc`, we'd need a way to pass arbitrary user-specified arguments to the linker. Effectively, the interface to the linker is via arguments. That doesn't mean we shouldn't also have a more programmatic interface to linking as well, but it does mean that we need to keep argument parsing as part of the library.

davidlattimore:
I tried running your change. It looks like it's silently falling back to not forking because `named_pipe` already exists. i.e. the first time you run it, it creates the named pipe and forks, but each time after that it doesn't. While that could be fixed, I'd suggest not using a named pipe, but instead using an unnamed [pipe](https://docs.rs/libc/0.2.164/libc/fn.pipe.html). This has several advantages. Firstly, you don't need to worry about a file on disk. Having a file on disk could get accidentally left behind, which is messy. It could also be a problem if running two invocations of the linker at the same time. Another advantage of using an unnamed pipe is that the subprocess doesn't need to open it, since it inherits it from the parent. So if the subprocess crashes for some reason before it would have opened the file, that's fine because the pipe will still get closed, thus signalling the parent that the child has gone away.

I'm fine with just using the `libc` crate directly if you like. It avoids the dependency on the `fork` crate. I like to avoid unnecessary unsafe, but this usage doesn't seem unnecessary.

I'd probably suggest making fork be the default and adding `--no-fork` as a flag to prevent it. This would match what mold does. I think whatever testing we do prior to submitting is sufficient to make sure it works. It's not like we have lots of users yet :)

andrewdavidmackenzie:
On that last comment, the PR is very WIP so...
- I tool a very conservative "fall back to not forking" approach for any error
- I was thinking to create it in a tmp directory
- The name should be unique to the invocation - either with a random name or the pid of the parent as the name
- It doesn't handle already existing (it could easily check for that and use it or delete and re-crate)
- It doesn't delete it on exit, which it should.
- Inheriting open FD - I think the child will inherit the open FD of the named pipe too? So, I could skip that. Easy to check.
- I'll look at unnamed pipe also and compare
- I agree that Fork doesn't add much to the equation, so I can look at using [lib::fork](https://docs.rs/libc/latest/libc/fn.fork.html) directly
- OK, I can switch the default and add no-fork. Providing I can control it to benchmark before and after the change.

wild/src/main.rs
```diff
@@ -1,4 +1,91 @@
...
+                            // Create a linker with remaining args and run it
+                            wild_lib::Linker::from_args(args.into_iter())?.run()
+                        }
+                        _ => {
+                            // Success in the child
+                            // Create a linker with remaining args and run it
+                            wild_lib::Linker::from_args(args.into_iter())?.run()?;
+
+                            // inform parent that we are done!
+                            let mut f = File::open(path)?;
```
andrewdavidmackenzie: To have child re-use an already open FD after fork, I think we'd need to make it a global, static variable...

andrewdavidmackenzie:
Remaining points:

Named Pipe Option
* I was thinking to create it in a tmp directory, with unique name (either with a random name or the pid of the parent as the name). I will not create in the project/CWD for sure.
* I usually use tempdir or similar, but will look to see if any option without adding a new dependency

Unnamed Pipe Option
* I'll look at unnamed pipe also and compare
* Inheriting open FD -I think we'd need to use a global static in rust, and have lib use the global declared in the binary!! Pretty ugly.

andrewdavidmackenzie:
Switching the default causes the tests to hang, so not a good sign - will investigate

wild/Cargo.toml
```diff
@@ -5,6 +5,9 @@ edition = "2021"
 
 [dependencies]
 wild_lib = { version = "0.2.0", path = "../wild_lib" }
+anyhow = "1.0.93"
+libc = "0.2"
+tempdir = "0.3.7"
```
andrewdavidmackenzie: Just an experiment - does wild already have some standard method for creating tempfiles or tempdirs that could be reused?

andrewdavidmackenzie:
Since I am working on benchmarking in parallel, I thought I'd drop this in here:

screenshot-2024-11-23-17-38-53

andrewdavidmackenzie:
I'm not seeing any improvements usin gthis method, using hyperfine or poop - in fact maybe a 1% or 2% slow-down.
I'll try a version with an unnamed pipe and see.

wild/src/main.rs
davidlattimore: I'm pretty sure it should be possible to just have a local variable. Both parent and child process inherit the same stack

andrewdavidmackenzie: I think you're right. If you do the call and have the variable before fork()

andrewdavidmackenzie:
With a first (hacky) unnamed pipe implementation, things have "turned around", from -1% to +1% .... not exactly a huge gain.... may be I need a better benchmark link to test this one (using the ripgrep one from the benchmarking PR work)

davidlattimore:
It looks like you're not signalling the parent process until after all the shutdown work has been done. The +/- 1% you're seeing is probably measurement variation. I think you'd need to pass something into `wild_lib::link` - possibly a closure - then run that closure after the call to `shutdown::free_output` and before the call to `shutdown::free_layout`. It needs to be after `free_output` because until the output has been properly closed, we can't say that we're done - but all the remaining freeing work after that can be done in the background.

andrewdavidmackenzie:
Sounds right. I forgot that you do shutdown work explicitly. I was going to
ask for the safe point when to "signa" parent.

Will test that on named pipe option, and keep another branch for untamed
pipe and compare.

Thanks

andrewdavidmackenzie:
Still not seeing much difference (unless I'm missing something):

screenshot-2024-11-25-14-46-44

screenshot-2024-11-25-14-47-50

andrewdavidmackenzie:
I tried counting forks in each option with strace, with no differences - so I suspect things are not working as expected...I'll debug more.

andrewdavidmackenzie:
I've checked that fork is happening and working by:
- force an error in all the cases in the code (but I didn't commit that)
- used strace to count forks ("clones" in fact) performed when running "run-with wild"and there is one more than when running with "--no-fork" option.

Initial tests using ripgrep crate show between a 1% (1.5ms) and 5% (7ms)  improvement.

Maybe we need a more worse-case crate for shutdown time to see the difference?

andrewdavidmackenzie:
screenshot-2024-11-25-23-34-49

at last! Goddamn it!

andrewdavidmackenzie:
Comparing mold and wild

screenshot-2024-11-25-23-36-13

wild/src/main.rs
```diff
@@ -1,4 +1,104 @@
...
+                        }
+                        _ => {
+                            // Fork success in the parent - wait for the child to "signal" us it's done
+                            let exit_status = wait_for_child_done(&fds)?;
+                            process::exit(exit_status);
+                        }
+                    }
+                }
+                Err(_e) => {
+                    // TODO do we want to log the error, or output a warning?
```
andrewdavidmackenzie: Thoughts on handling errors like this one?

wild/src/main.rs
```diff
@@ -1,4 +1,104 @@
...
+use std::io::Error;
+use std::process;
+
 fn main() -> wild_lib::error::Result {
-    let linker = wild_lib::Linker::from_env()?;
-    linker.run()
+    // skip the program name
+    let mut args: Vec<String> = args().skip(1).collect();
+
+    // TODO See if there is a better way to do this, more canonical
```
andrewdavidmackenzie: Don't like this much. Will have a look to see if can come up with something more rusty

wild_lib/src/lib.rs
```diff
@@ -106,6 +110,9 @@ fn link<S: storage::StorageModel, A: arch::Arch>(args: &Args) -> crate::error::R
     let output_file = output.write::<S, A>(&layout)?;
     diff::maybe_diff()?;
 
+    if let Some(done_callback) = done_closure {
```
andrewdavidmackenzie: If this is not the right place to signal parent, might need some guidance to find it.

wild/src/main.rs
```diff
@@ -1,4 +1,104 @@
...
+/// or for its unexpected death by closure of the pipe before receiving anything back
+fn wait_for_child_done(fds: &[c_int]) -> wild_lib::error::Result<i32> {
+    unsafe {
+        // close our sending end of the pipe
+        libc::close(fds[1]);
+        // open the other end of the pipe for reading
+        let stream = libc::fdopen(fds[0], "r".as_ptr() as *const i8);
+
+        // Wait for child to send exit_status or pipe to be closed
+        let mut response: [u8; 4] = [0u8; 4];
```
andrewdavidmackenzie, line 91: Are you interested in the possibility of passing non-zero exit_status?
I don't know the linker code to know of other exit paths (couldn't see any!) that might return a non-zero/

We could remove code, by just waiting for pipe close, and closing from child...

davidlattimore: The pipe being closed could be because the subprocess is in the process of crashing, so I think we need to write something to the pipe, but it probably should be an exit status, since we should only do so when our exit status is success. Perhaps just write "OK" or something. The error-path should just be that the subprocess runs to completion and exits with a non-zero exit status. When the parent sees that the pipe was closed without any content, it should wait for the child to exit, which should happen almost immediately, then exit with the same exit code as the child.

wild_lib/src/lib.rs
davidlattimore: I think after the line `shutdown::free_output(output_file);` below would be safest

wild/src/main.rs
```diff
@@ -1,4 +1,104 @@
...
+        if a == "--no-fork" {
+            no_fork_subprocess = true;
+            false
+        } else {
+            true
+        }
+    });
+
+    match no_fork_subprocess {
+        false => {
```
davidlattimore: Matching on a boolean is probably more simply written as an if-else. Also, removing the negative from the variable name and inverting the logic might make it clearer. i.e. have a variable `should_fork` that defaults to `true`, but is set to `false` if `--no-fork` is specified.

wild/src/main.rs
davidlattimore: If something like `make_pipe` is failing, then we should probably report an error and shut down. I'd probably just do something like
```rust
make_pipe(&mut fds).context("make_pipe")?;
```

wild/src/main.rs
andrewdavidmackenzie: I usually avoid "early returns" (except using '?'), but here I think it's got some merit, makes it easy to ready and removes two levels of indenting.

If you prefer an if-else, then easy to change back.

wild/src/main.rs
andrewdavidmackenzie: Do you wish to add waiting for the child to exit?
I'm cautious about having the parent hang on the child (beyond the pipe closure).
That seems like an additional step in the case when "pipe was closed without getting status code back"?

andrewdavidmackenzie:
screenshot-2024-11-26-09-03-43

wild/src/main.rs
davidlattimore: So I guess it's odd at the moment that there's support for sending a status code from the parent to the child, but the child only ever sends a status code of 0. If the child fails for some reason, then it closes the pipe and the parent then exits with -1 regardless of the failure mode of the child process. That said, I'm happy with the state that the PR is in if you'd just like to merge it as-is.

andrewdavidmackenzie: I'm not sure how to best capture failures in the child, map them to non-zero status codes and send to the parent.

With use of result and ? there are a lot of of early exit points in linker....we would need to catch all the errors, map to an exit code and then send to the parent...making it pretty messy IMO...

Unless we could use some generic exit handler, OR handle the error case in main()
```
        0 => {
            // Fork success in child - Run linker in this process with remaining args
            let done_closure = move |exit_status: i32| inform_parent_done(&fds, exit_status);
            wild_lib::Linker::from_args(args.into_iter())?.run(Some(Box::new(done_closure)))
        }
```

instead of just returning the result from run(), we match on it and send exit status code to parent when not OK()?

davidlattimore: You can never really capture all the failure paths in that way, since things like out-of-memory, or a panic or maybe even a segfault could occur. That's why I was suggesting making the pipe just for sending success and for everything else, just check the exit status of the child. But we don't really define or distinguish based on exit codes, so it doesn't matter too much.

davidlattimore:
I'm not seeing a speedup when I benchmark with `poop`. I think that's because stdout / stderr aren't being closed. I'm guessing `poop` probably captures them and waits for them to close. If I add code to close them in the subprocess once it's done, then I do see a speedup with `poop`.

andrewdavidmackenzie:
OK, even if it only affects poop - sounds like it's worth it... Got preferences for where/how to do it?
(want to share a snippet of code?)

davidlattimore:
I added the following to `inform_parent_done`:

```rust
        libc::close(libc::STDOUT_FILENO);
        libc::close(libc::STDERR_FILENO);
```

andrewdavidmackenzie:
That does indeed cause poop to also report gains (e.g. 10ms on a 100ms link, or 11%), inline with the same gains reported by hyperfine

Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 members = ["linker-diff", "linker-layout", "linker-trace", "linker-utils", "wild", "wild_lib"]
+default-members = ["wild"]
```
davidlattimore: Might be good to do this in a separate PR, since it's not related. Also, I was trying to understand why some of the changes you made in `args.rs` were able to compile and it turns out that they don't, but CI didn't pick it up because those changes weren't being built / tested on account of this change. I've changed the CI now so that even with this change, the build failure will show up. So after https://github.com/davidlattimore/wild/pull/200 - you should now have a build failure to fix next time you run CI - or you can do `cargo test --workspace` locally.

andrewdavidmackenzie: Wow - I didn't see that one coming (obviously!)!!

I broken the golden rule when debugging the issue, undo all your changes vs. the version that works, even the ones you _think_ are unrelated :-(

Will fix this morning

andrewdavidmackenzie:
Now back to the .rs linker integration tests failing:
```
failures:
    integration_test::program_name_27___rust_integration_rs__
    integration_test::program_name_28___rust_integration_dynamic_rs__
    integration_test::program_name_30___rust_tls_rs__
```

marxin:
What a nice improvement! Thanks for the PR.

wild/src/main.rs
```diff
@@ -1,4 +1,118 @@
...
-    linker.run()
+    // skip the program name
+    let mut args: Vec<String> = args().skip(1).collect();
+
+    // The default is to fork a sub-process to do the actual linking
+    let mut fork_subprocess = true;
+
+    // But it can be de-activated using the "--no-fork" option. This option is acted upon here
+    // in main, but is removed and not passed onto the linker for parsing
+    args.retain(|a| {
```
marxin, line 28: Probably the most idiomatic approach for this will be (hopefully soon default feature): https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.extract_if.

andrewdavidmackenzie: Didn't know about that. Want to send a minor PR to make more idiomatic?

It's funny how you can look at some rust and suspect "there must be a better way to do this..." Usually happens when it looks like 'C'

marxin: > Usually happens when it looks like 'C'

Yeah, that's exactly my feeling :)
About `extract_if`, it's an unstable feature so that I would wait for stabilization.

PR #194, opened 2024-11-23
Add BENCHMARKING.md with description of how to setup and run benchmarks

### LINKED ISSUE #193: Document running benchmarks
Create a separate BENCHMARKING.md file linked from README.md where we describe how to do benchmarking.

When possible, add some scripts, or rust tools in a benchmarking subdirectory that will help automate some or all of the steps (over time, start with something small).

andrewdavidmackenzie:
Fixes #193 

I have moved a limited amout of text on benchmarking in README.md to the added BENCHMARKING.md file.

If you prefer to move more of it, either let me know or push a commit later?

Since adding markdown files and linking from README.md, I have also:
* Moved contributing info into a CONTRIBUTING.md file and linked from main
* Added a link-check step at the end of CI, so broken links between docs won't make it to main branch
 
I will now focus adding docs as per chat comments and maybe some minimal automation of some of that.

### DIFF
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -54,33 +54,32 @@ rustflags = [
 
 ### Run benchmark with hyperfine
 
-* To benchmark, you can then run something
-  like "`hyperfine --warmup 2 '/tmp/wild/ripgrep/6/run-with ld' '/tmp/wild/ripgrep/6/run-with mold'
-  '/tmp/wild/ripgrep/6/run-with wild'`"
-    * This will benchmark the linking stage between `ld`, `mold` and `wild`, discarding the first two runs of each to
-      reduce the effects of cache warmup
+Let's benchmark the linking stage between `ld`, `mold` and `wild`, discarding the first two runs of each to reduce the
+effects of cache warmup
+
+```shell
+hyperfine --warmup 2 '/tmp/wild/ripgrep/6/run-with ld' '/tmp/wild/ripgrep/6/run-with mold' '/tmp/wild/ripgrep/6/run-with wild'
+```
 
 That should produce output similar to this (with different values):
 
 ```text
 Benchmark 1: /tmp/wild/ripgrep/6/run-with ld
-  Time (mean ± σ):     931.0 ms ±   7.2 ms    [User: 680.8 ms, System: 249.1 ms]
-  Range (min … max):   923.1 ms … 944.6 ms    10 runs
+  Time (mean ± σ):     954.1 ms ±  13.6 ms    [User: 683.4 ms, System: 268.8 ms]
+  Range (min … max):   920.6 ms … 970.7 ms    10 runs
  
 Benchmark 2: /tmp/wild/ripgrep/6/run-with mold
-  Time (mean ± σ):     144.3 ms ±   6.1 ms    [User: 53.2 ms, System: 1.7 ms]
-  Range (min … max):   133.0 ms … 156.4 ms    18 runs
+  Time (mean ± σ):     146.1 ms ±   3.6 ms    [User: 52.0 ms, System: 2.4 ms]
+  Range (min … max):   139.1 ms … 154.7 ms    19 runs
  
-Benchmark 3: 
-  /tmp/wild/ripgrep/6/run-with wild
-  Time (mean ± σ):     101.9 ms ±   5.0 ms    [User: 389.8 ms, System: 135.4 ms]
-  Range (min … max):    95.3 ms … 118.4 ms    30 runs
+Benchmark 3: /tmp/wild/ripgrep/6/run-with wild
+  Time (mean ± σ):      87.7 ms ±   2.8 ms    [User: 2.4 ms, System: 2.0 ms]
+  Range (min … max):    81.5 ms …  92.5 ms    34 runs
  
 Summary
-  
   /tmp/wild/ripgrep/6/run-with wild ran
-    1.42 ± 0.09 times faster than /tmp/wild/ripgrep/6/run-with mold
-    9.14 ± 0.46 times faster than /tmp/wild/ripgrep/6/run-with ld
+    1.67 ± 0.07 times faster than /tmp/wild/ripgrep/6/run-with mold
+   10.88 ± 0.38 times faster than /tmp/wild/ripgrep/6/run-with ld
 ```
 
 ### Run benchmark with poop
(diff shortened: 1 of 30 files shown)


andrewdavidmackenzie:
@davidlattimore I will add some questions I have in a code review, so beside the text in question

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+There might be other flags that speed up the other linkers by letting them avoid some work that
+they're currently doing. If you know of such flags, please let me know.
+
+## How to benchmark
+
+For benchmarking the linker, it's preferable to run just the linker, not the whole build process.
+
+The way to do that is by capturing the linker invocation so that it can be rerun. Wild has a built-in way to do that.
+
+TODO: Is it best to benchmark a debug or a release build? or both for different reasons?
```
andrewdavidmackenzie: WDYT. What should we say here (if anything)?

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+
+Follow-these steps:
+
+* Chose the crate that you wish to use in your benchmark, clone it, `cd` into it's root directory and make sure it
+  builds with `cargo build` (for a rust project)
+    * Examples: [`ripgrep`](https://github.com/BurntSushi/ripgrep.git)
```
andrewdavidmackenzie, line 30: I couldn't find the rustc-driver project you mentioned (see below) and also you mentioned hacks to get it to work, so I went with a more standard rust/cargo example....

open to change though.

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+* Clean the build using `cargo clean`
+* Set `RUSTFLAGS` or modify `.cargo/config.toml` to have the build use wild
+    * `RUSTFLAGS`:
```
andrewdavidmackenzie: I documented the use of changes in .cargo/config.toml, like the warmup benchmark does - but if can try to add RUSTFLAGS also....WDYT?

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+    * `.cargo/config.toml`: add this (example for `ripgrep`)
+
+```toml
+  [target.x86_64-unknown-linux-gnu]
+linker = "/usr/bin/clang"
+
+rustflags = [
+    "-C", "relocation-model=static",
```
andrewdavidmackenzie: Not sure whether it's a good idea to force static in all benchmarks, or this one - or even if that has effect on a gnu (not musl) build....?

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+    "-C", "link-arg=--ld-path=wild"
+]
+```
+
+* Make sure that you have a version of wild in your `$PATH` so that it will be used
+* Run `WILD_SAVE_BASE=/tmp/wild/ripgrep cargo build` in the crate's root directory
+* You will get a few numbered subdirectories in `/tmp/wild/ripgrep`
+* The last numbered subdirectory will be the final link stage of the binary (TODO clarify that)
```
andrewdavidmackenzie: In your chat you said "probably". Can we clarify this? Is it always the highest number that we should use?

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+* You can then run `/tmp/wild/ripgrep/6/run-with wild` and that will rerun the link with wild
+* To benchmark, you can then run something
+  like `hyperfine --warmup 2 '/tmp/wild/ripgrep/6/run-with ld' '/tmp/wild/ripgrep/6/run-with mold' '
+  /tmp/wild/ripgrep/6/run-with wild'`
+    * This will benchmark the linking stage between `ld`, `mold` and `wild`, discarding the first two runs of each to
+      reduce the effects of cache warmup
+
+That should produce output similar to this (with different values):
+
+```text
```
andrewdavidmackenzie, line 66: Yeh! It worked!

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+* between different options passed to Wild - You can pass arbitrary additional arguments to run-with.
+  The first argument needs to be the name of the linker to use. All additional arguments are passed through to the
+  linker as-is
+
+### Caching
+
+The use of CPU caches affects linker performance, for that reason, when running hyperfine we use the `--warmup 2`
+option: so that the first two runs (before caches are used) get discarded.
+
+TODO: When users are running their build, it's not clear that the cache could be filled from a previous run and
```
andrewdavidmackenzie: WDYT?

Maybe something to move to an issue or a discussion?

BENCHMARKING.md
```diff
@@ -0,0 +1,120 @@
...
+taken advantage of, so it may be interesting to compare the first runs (without cache from a previous run)
+separately. However, to do so, we would need to ensure that caches are invalidated and re-run the first run a
+number of times to get reliable results.
+
+## What to benchmark
+
+### rustc - TODO document this
```
andrewdavidmackenzie, line 160: Possible to use this method?
Document the hack? (config.toml edit won't do it?)

BENCHMARKING.md
davidlattimore: A release build. For dev builds, all that really matters is how fast the tests run

BENCHMARKING.md
davidlattimore: Rustc-driver is a shared object that gets built when you build rustc. It is a bit of a pain to get it. Fine to leave it for now. Maybe I'll try to write up some instructions for it later

BENCHMARKING.md
davidlattimore: Depends what order things build in. If you're building a workspace with multiple binaries then they could be in different orders. To mitigate this you could pass flags to cargo to make sure just a single binary gets built. I'll still build proc macros and build scripts though. Another way to mitigate is to do one build without the environment variable set, then touch a source file before building again with the environment variable set

BENCHMARKING.md
andrewdavidmackenzie: I was thinking more along the lines of: people want to link (fast) debug builds with wild, and if it works well and is faster (less frequently) release builds too, no?

You mention that wild is behind mold when the debug info is very large...

So, there is a case for benchmarking release builds and debug builds, no?

BENCHMARKING.md
andrewdavidmackenzie: OK. I'll close thise conversation and leave it to you to to add other examples later, or maybe we create an issue about including some benchmark in CI, to catch regressions...?

BENCHMARKING.md
andrewdavidmackenzie: I looked quickly in the different directories, but nothing jumped out at me to say "this directory was used to build the binary I am interested in". Is there a way to tell which one to use?

BENCHMARKING.md
davidlattimore: Oh, right. Benchmarking debug builds are definitely useful then. Although I tend to turn off debug info in my debug builds. Given how much of a difference having debug info makes in terms of link time, it's probably good to benchmark with and without debug info

BENCHMARKING.md
davidlattimore: Dynamic linking is the default so probably more common, so benchmarking that is likely more important

BENCHMARKING.md
davidlattimore: The last line in the run-with file should say what was being linked

BENCHMARKING.md
davidlattimore: Running with a cold cache is an interesting thing to benchmark, but I think a warm cache is what we care about more. Even after a cold build, most of the files being linked (besides system libraries) will likely be in the cache, since the files were just written - during the build. The exception would be if there isn't enough RAM and they got evicted from cache. For a warm build, where we did a link, edited some code, recompiled and relinked, it's even more likely that most of the input files will be in the cache. A warm build is the use-case that we care about most - it's what we get when the user is iterating one their code.

BENCHMARKING.md
```diff
@@ -0,0 +1,173 @@
...
+Using this method, you can benchmark:
+
+* between Wild and one or more other linkers
+* between different options passed to Wild - You can pass arbitrary additional arguments to run-with.
+  The first argument needs to be the name of the linker to use. All additional arguments are passed through to the
+  linker as-is
+
+### Caching
+
+The use of CPU caches affects linker performance, for that reason, when running hyperfine we use the `--warmup 2`
```
davidlattimore: `--warmup 2` is about warming up Linux's filesystem cache, not the CPU cache. The CPU caches are much smaller and generally won't live between multiple runs of the linker. The filesystem cache can use as much RAM as you have free and avoids the need to reread stuff off disk.

davidlattimore (commented):
It'd be good to make this PR just be for doc changes and move any other changes out to separate PRs with their own description

BENCHMARKING.md
andrewdavidmackenzie: I will not force this in the example and mention to leave as user wishes

BENCHMARKING.md
andrewdavidmackenzie: Done RUSTFLAGS and tested it also

BENCHMARKING.md
andrewdavidmackenzie: fixed.

BENCHMARKING.md
andrewdavidmackenzie: hmmm this is the last line of all the files for ripgrep:
```
 tail -n 1 /tmp/wild/ripgrep/*/run-with 
==> /tmp/wild/ripgrep/0/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/serde_json-4bafc9c1f2c9627a/build_script_build-4bafc9c1f2c9627a
==> /tmp/wild/ripgrep/1/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/serde-e3e5c012db1f1a6a/build_script_build-e3e5c012db1f1a6a
==> /tmp/wild/ripgrep/2/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/crossbeam-utils-11ac647ecdc91c93/build_script_build-11ac647ecdc91c93
==> /tmp/wild/ripgrep/3/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/libc-4c352264c1c5a73e/build_script_build-4c352264c1c5a73e
==> /tmp/wild/ripgrep/4/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/anyhow-7a1288326b2ce95b/build_script_build-7a1288326b2ce95b
==> /tmp/wild/ripgrep/5/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/build/ripgrep-70e6da0716cbf1e6/build_script_build-70e6da0716cbf1e6
==> /tmp/wild/ripgrep/6/run-with <==
# Original output file: /home/andrew/workspace/ripgrep/target/debug/deps/rg-fb8fbfbb38ad493c
```

All seem to be build scripts, and one a deps.

Questions:
- Are the build script ones useful for us? Could they be surpressed and not written out?
- Is it safe to reccomend "use the deps one" or "use the one that is not a build script"?
- Any way that wild could change the names from just numbers, or creat ea link in th eparent directory with the name of the binary being linked - something to make it obvious (or "automatic") which file to use with "run-with"?

andrewdavidmackenzie:
The code changes are due to me creating this branch from main, with my initial changes for forking, which was a mistake (didn't expect to have multiple branches/PRs open so soon...).

They will definitely be removed one way or other before being ready to merge.

andrewdavidmackenzie:
Removed code changes via rebase on main.
Ready for review.

I have other scratch pad notes that I will continue working on to extend discussion about benchmarking - but I think these docs are correct as they stand with the current code, and so can be merged after review.

BENCHMARKING.md
```diff
@@ -0,0 +1,178 @@
...
+    "-C", "link-arg=--ld-path=wild"
+]
+```
+
+* Make sure that you have a version of wild in your `$PATH` so that it will be used (try `which wild` to check)
+* Run `WILD_SAVE_BASE=/tmp/wild/ripgrep cargo build` in the crate's root directory (include `RUSTFLAGS` as above
+  if you have chosen that method)
+* You will get a few numbered subdirectories in `/tmp/wild/ripgrep` for different binaries being linked in the build
+  process (such as build scripts and multiple binaries in a crate)
+* Usually the last numbered subdirectory will be the final link stage of the binary (TODO clarify that)
```
davidlattimore: The numbers will be allocated in whatever order the binaries are built. So if you have one main binary, that will get built after all the build scripts and proc macros. If you have multiple binaries, potentially any of them could be last.

BENCHMARKING.md
```diff
@@ -0,0 +1,178 @@
...
+cpu_cycles          1.24G ± 15.6M       1.19G … 1.27G 3 ( 6%)       ⚡- 44.4% ± 0.6%
+instructions        1.21G ± 443K        1.21G … 1.21G 0 ( 0%)       ⚡- 67.9% ± 0.0%
+cache_references    34.0M ± 505K        33.1M … 35.0M 0 ( 0%)       ⚡- 63.1% ± 0.5%
+cache_misses        14.2M ± 171K        13.9M … 14.5M 0 ( 0%)       ⚡- 62.9% ± 0.5%
+branch_misses       3.42M ± 11.0K       3.40M … 3.45M 0 ( 0%)       ⚡- 63.1% ± 0.1%
+```
+
+NOTE: `poop` uses the first command as the reference the others are compared against, so if focusing on wild, you might
+want to re-order the commands and invoke `poop` thus:
+
```
davidlattimore, line 134: It might also be worthwhile mentioning that the `peak_rss` for mold and wild will not be at all accurate due to them both forking and that if you're wanting to look at rss. Actually, for `hyperfine`, it's also worth mentioning that the `User` and `System` measurements for wild and mold will be meaningless for the same reason - although I guess the sample results you pasted above are without your fork PR.

CONTIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+
+If you'd like to help out, I'd love to hear from you. It's a good idea to reach out first to avoid
+duplication of effort. Also, it'll make it possible for me to provide hints that might make what
+you're trying to do easier.
+
+## Options for communicating
+
+* I like, where possible, to talk to people via video chat.
+* You can book a time in my [calendar](https://calendar.app.google/MBYQeATMNBvuK8AZ6). If time zones make this hard, let
+  me
```
davidlattimore: Line wrapping looks weird here. I generally use 100 column wrap. Although, I'm also happy to just clean up line wrapping next time I'm editing the document if you want to leave it as-is.

BENCHMARKING.md
andrewdavidmackenzie: Fixed

CONTIBUTING.md
andrewdavidmackenzie: Fixed (to 80 cols)

davidlattimore:
Thanks!

Issue #38, opened 2024-08-18
Add support for .note.gnu.property


davidlattimore:
See https://lists.llvm.org/pipermail/llvm-dev/2018-March/121951.html

Note however that currently adding new linker-handled sections is a bit of a pain. See #36

marxin:
It's funny the actual specification of `GNU_PROPERTY_X86_XXX` properties is not yet merged into the official document (https://gitlab.com/x86-psABIs/x86-64-ABI/-/merge_requests/13) while being de-facto used in the wild.

marxin:
Implemented in #196.

@davidlattimore: Please close it.

davidlattimore:
Thanks for implementing this @marxin!

Issue #17, opened 2024-07-11
Fix tests on openSUSE


marxin:
Noticed that for the current main branch commit (042615a):

```
test integration_test ... FAILED

failures:

---- integration_test stdout ----
wild: /home/marxin/Programming/wild/wild/tests/build/ifunc.c-default.wild
ld: /home/marxin/Programming/wild/wild/tests/build/ifunc.c-default.ld
asm._start
  ORIG            `/home/marxin/Programming/wild/wild/tests/build/ifunc.default-f778dc51c46a70e0.o`
                  push %rbp
                  mov %rsp,%rbp
                  sub $0x10,%rsp
                  call 0x00000000000000F8
                  mov %eax,-4(%rbp)
                  cmpl $0,-4(%rbp)
                  je 0x0000000000000020
                  mov -4(%rbp),%eax
                  mov %eax,%edi
                  call 0x000000000000015F
                  call 0  // 0xAAA=PLT(IFUNC(resolve_compute_value10))
                  cmp $0xA,%eax
                  je 0x0000000000000034
                  mov $1,%edi
                  call 0x000000000000015F
                  call 0xFFFFFFFFFFFFFFF0
                  cmp $0x20,%eax
                  je 0x0000000000000048
                  mov $2,%edi
                  call 0x000000000000015F
  
  wild 0x00401798 b8 09 18 40 00 mov $0x401809,%eax  // Mov_r32_imm32(0x401809) resolve_compute_value10 compute_value10
  ld   0x00401058 b8 08 10 40 00 mov $0x401008,%eax  // Mov_r32_imm32(0x401008) PLT(IFUNC(resolve_compute_value10)) PLT(IFUNC(compute_value10))
  ORIG            b8 00 00 00 00 mov $0,%eax  // R_X86_64_32 -> `compute_value10`
  TRACE           value_flags=IFUNC | CAN_BYPASS_GOT resolution_flags=DIRECT | GOT | PLT
  
                  call *%rax
                  cmp $0xA,%eax
                  je 0x000000000000005E
                  mov $3,%edi
                  call 0x000000000000015F
                  mov 0,%eax  // 0xAAA=resolve_count
                  cmp $2,%eax
                  je 0x0000000000000073
                  mov $4,%edi
                  call 0x000000000000015F
  
  wild 0x004017c3 b8 09 18 40 00 mov $0x401809,%eax  // Mov_r32_imm32(0x401809) resolve_compute_value10 compute_value10
[...]

PR #196, opened 2024-11-24
Support .note.gnu.property

marxin:
The PR implements the feature where I would like to discuss the following things:

1. The merging of the sections happens for all object files where the output section is present just once and should be synthesized. That's why I chose to implement it in `EpilogueLayout`, where the corresponding sections are identified based on the newly added `SectionSlot` value. Is it a good design choice?
2. The implementation supports (similarly to Mold) only properties with 4B property data. Do you have a hint on how to do a smarter serialization in elf_writer.rs? Theoretically, `GnuProperty` [^1] from _object_ create can be converted to not only read mode.
3. I face an issue in testing:
```
Error: Unexpected memory offsets:
Part #17 (section `.note.gnu.property` alignment: 8) expected: 0x400408 actual: 0x4003d8 bumped by: 0x0 requested size: 0x30
```

Can you help me with that?

4. The output (`.gnu.note.property`) should be binary equal to what other linkers do. That said, will we need any special handling in linker-diff`, or is the binary diff the default one?

Thanks for review in advance!

Implements: #38

[^1]: https://docs.rs/object/latest/object/read/elf/struct.GnuProperty.html

### DIFF
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -326,14 +326,7 @@ fn merge_gnu_property_notes(group_states: &mut [GroupState]) -> Result {
                 })
                 .or_insert_with(|| {
                     type_index += 1;
-                    (
-                        type_index,
-                        if matches!(property_class, PropertyClass::And) {
-                            u32::MAX
-                        } else {
-                            0
-                        },
-                    )
+                    (type_index, prop.data)
                 });
         }
     }
@@ -349,7 +342,8 @@ fn merge_gnu_property_notes(group_states: &mut [GroupState]) -> Result {
                     .any(|prop| prop.ptype == property_type)
             });
             if match property_class {
-                PropertyClass::Or | PropertyClass::And => property_value != 0,
+                PropertyClass::Or => property_value != 0,
+                PropertyClass::And => type_present_in_all && property_value != 0,
                 PropertyClass::AndOr => type_present_in_all,
             } {
                 Some(GnuProperty {
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -305,6 +305,8 @@ fn merge_gnu_property_notes(group_states: &mut [GroupState]) -> Result {
         })
         .collect_vec();
 
+    // Merge bits of each property type based on type: OR or AND operation. When a property type
+    // is newly added to the map, we start either with zero or all bits-set (PropertyClass::And).
     let mut type_index = 0usize;
     let mut property_map = HashMap::new();
     for file_props in &properties_per_file {
--- a/wild_lib/src/layout.rs
+++ b/wild_lib/src/layout.rs
@@ -4,6 +4,7 @@
 
 use self::elf::NoteHeader;
 use self::elf::GNU_NOTE_NAME;
+use self::elf::GNU_NOTE_PROPERTY_ENTRY_SIZE;
 use self::output_section_id::InfoInputs;
 use crate::alignment;
 use crate::alignment::Alignment;
@@ -304,63 +305,63 @@ fn merge_gnu_property_notes(group_states: &mut [GroupState]) -> Result {
         })
         .collect_vec();
 
-    // First of all, merge (OR) all property values that belong to a property type.
     let mut type_index = 0usize;
-    // TODO: simplify
     let mut property_map = HashMap::new();
     for file_props in &properties_per_file {
         for prop in *file_props {
+            let property_class = get_property_class(prop.ptype).ok_or(anyhow::anyhow!(format!(
+                "unclassified property type {}",
+                prop.ptype
+            )))?;
             property_map
-                .entry(prop.r#type)
-                .and_modify(|e: &mut (usize, u32)| e.1 |= prop.data)
+                .entry(prop.ptype)
+                .and_modify(|e: &mut (usize, u32)| {
+                    if matches!(property_class, PropertyClass::And) {
+                        e.1 &= prop.data;
+                    } else {
+                        e.1 |= prop.data;
+                    }
+                })
                 .or_insert_with(|| {
                     type_index += 1;
-                    (type_index, prop.data)
+                    (
+                        type_index,
+                        if matches!(property_class, PropertyClass::And) {
+                            u32::MAX
+                        } else {
+                            0
+                        },
+                    )
                 });
         }
     }
 
-    let mut output_properties = Vec::new();
-    for (property_type, (_, property_value)) in property_map.into_iter().sorted_by_key(|x| x.1 .0) {
-        let property_class = get_property_class(property_type).ok_or(anyhow::anyhow!(format!(
-            "unclassified property type {property_type}"
-        )))?;
-        let mut output_value = 0;
-
-        // Merge bits for a particular property type.
-        for bit in 0..u32::BITS {
-            let mask = 1 << bit;
-            if property_value & mask != 0
-                && match property_class {
-                    PropertyClass::And => properties_per_file.iter().all(|props_per_file| {
-                        props_per_file
-                            .iter()
-                            .any(|prop| prop.r#type == property_type && prop.data & mask != 0)
-                    }),
-                    PropertyClass::Or | PropertyClass::AndOr => true,
-                }
-            {
-                output_value |= mask;
-            }
-        }
-
-        let type_present_in_all = properties_per_file.iter().all(|props_per_file| {
-            props_per_file
-                .iter()
-                .any(|prop| prop.r#type == property_type)
-        });
-        if (output_value != 0 && matches!(property_class, PropertyClass::Or | PropertyClass::And))
-            || (matches!(property_class, PropertyClass::AndOr) && type_present_in_all)
-        {
-            output_properties.push(GnuProperty {
-                r#type: property_type,
-                data: output_value,
+    let output_properties = property_map
+        .into_iter()
+        .sorted_by_key(|x| x.1 .0)
+        .filter_map(|(property_type, (_, property_value))| {
+            let property_class = get_property_class(property_type).unwrap();
+            let type_present_in_all = properties_per_file.iter().all(|props_per_file| {
+                props_per_file
+                    .iter()
+                    .any(|prop| prop.ptype == property_type)
             });
-        }
-    }
+            if match property_class {
+                PropertyClass::Or | PropertyClass::And => property_value != 0,
+                PropertyClass::AndOr => type_present_in_all,
+            } {
+                Some(GnuProperty {
+                    ptype: property_type,
+                    data: property_value,
+                })
+            } else {
+                None
+            }
+        })
+        .collect_vec();
 
     let epilogue = get_epilogue_mut(group_states);
-    epilogue.gnu_property_notes = dbg!(output_properties);
+    epilogue.gnu_property_notes = output_properties;
     Ok(())
 }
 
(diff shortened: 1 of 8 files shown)


marxin:
> 2\. The implementation supports (similarly to Mold) only properties with 4B property data. Do you have a hint on how to do a smarter serialization in elf_writer.rs? Theoretically, `GnuProperty` [1](#user-content-fn-1-d5c52a67ac4b57beeee265adc3703ffe) from _object_ create can be converted to not only read mode.

FYI: @philipc

davidlattimore:
> Error: Unexpected memory offsets:
> Part #17 (section `.note.gnu.property` alignment: 8) expected: 0x400408 actual: 0x4003d8 bumped by: 0x0 requested size: 0x30

Allocation of addresses within the output file goes through two phases. In the first phase, objects make requests for how much space they want in a particular section. In the second phase (`finalise_layout`), objects take addresses in the sections in which they requested space. They do this by getting the next address for that section from the argument `memory_offsets`. Once they take their address, they need to increment the next address for that section so that any subsequent files that also put stuff into that section don't get overlapping addresses.

So one possible fix is to make sure that `finalise_layout` increments `memory_offsets` for the `.note.gnu.property` section by whatever size it is using.

Alternatively, assuming the only "file" that ever does stuff with `.note.gnu.property` is the epilogue, then you can just ignore that PartId, by adding it in `verification::clear_ignored`.

davidlattimore:
> The merging of the sections happens for all object files where the output section is present just once and should be synthesized. That's why I chose to implement it in `EpilogueLayout`, where the corresponding sections are identified based on the newly added `SectionSlot` value. Is it a good design choice?

I think that's reasonable. Wherever there are things that aren't obviously "owned" by one particular input object, I usually write them from either the prologue or the epilogue. In some other cases, I pick a "winning" input object to be responsible for writing a particular thing. For example if multiple input objects define something, just one of them will be selected to write that thing. Either approach is fine, so whichever you think would be cleanest.

> The implementation supports (similarly to Mold) only properties with 4B property data. Do you have a hint on how to do a smarter serialization in elf_writer.rs? Theoretically, `GnuProperty` [1](#user-content-fn-1-ea1bf1508f1be669e88b9bb87067b8bb) from _object_ create can be converted to not only read mode.

If `Object` doesn't (currently) provide a struct that we can use, then perhaps just define a struct in `elf.rs` then derive `Pod`. e.g. see `EhFrameHdr`.

wild_lib/src/layout.rs
```diff
@@ -227,21 +239,125 @@ fn finalise_all_sizes<'data, S: StorageModel>(
...
     epilogue.dynamic_symbol_definitions = dynamic_symbol_definitions;
     Ok(())
 }
 
+enum PropertyClass {
+    // A bit in the output pr_data is set if it is set in any relocatable input.
+    // If all bits in the the output pr_data field are zero, this property should be removed from output.
+    Or,
+    // A bit in the output pr_data field is set only if it is set in all relocatable input pr_data fields.
+    // If all bits in the the output pr_data field are zero, this property should be removed from output.
```
davidlattimore, line 273: Seems odd that this isn't the opposite of `Or`. i.e. I would have expected that the property would be removed if _any_ bit in the output pr_data field was zero. But what you've got here matches what's in the spec. I probably don't have a deep enough understanding of the intent behind these rules - although now that I look at your implementation, it looks like that doesn't match what's written here, but does match the behaviour that I would expect. I guess the question is, is the spec wrong or do the other linkers implement it as written here.

davidlattimore:
> The output (`.gnu.note.property`) should be binary equal to what other linkers do. That said, will we need any special handling in linker-diff`, or is the binary diff the default one?

Linker-diff currently won't compare section content by default. I'm not aware of any other sections where we expect all linkers to produce exactly equal content, so this would be the first.

wild_lib/src/layout.rs
```diff
@@ -227,21 +239,125 @@ fn finalise_all_sizes<'data, S: StorageModel>(
...
+        })
+        .collect_vec();
+
+    // Merge bits of each property type based on type: OR or AND operation. When a property type
+    // is newly added to the map, we start either with zero or all bits-set (PropertyClass::And).
+    let mut type_index = 0usize;
+    let mut property_map = HashMap::new();
+    for file_props in &properties_per_file {
+        for prop in *file_props {
+            let property_class = get_property_class(prop.ptype).ok_or(anyhow::anyhow!(format!(
```
davidlattimore: Unless I'm mistaken, this will always build this format string. Perhaps you want `ok_or_else`?

wild_lib/src/resolution.rs
```diff
@@ -769,7 +772,10 @@ fn resolve_sections_for_object<'data>(
                             SectionSlot::Unloaded(UnloadedSection::new(id))
                         }
                         TemporaryPartId::Custom(custom_section_id, _alignment) => {
-                            if custom_section_id.name.bytes().starts_with(b".debug_") {
+                            let section_name = custom_section_id.name.bytes();
+                            if section_name == b".note.gnu.property" {
```
davidlattimore: Making ".note.gnu.property" a custom section seems a bit odd, since you define a PartId for the section. It means that a custom section PartId for the section will be allocated, but never used.

davidlattimore (commented):
Nice! Thanks for doing this - it'll be good to have this gap filled

philipc:
> The implementation supports (similarly to Mold) only properties with 4B property data. Do you have a hint on how to do a smarter serialization in elf_writer.rs? Theoretically, GnuProperty from object create can be converted to not only read mode.

`GnuProperty` is not an appropriate struct for this. It contains a reference to a slice of data which must already be in the correct byte order. Really what you want is something that contains a u32 property and u32 value. It's pretty easy to write that yourself though in three lines of code, as [already done elsewhere in object](https://github.com/gimli-rs/object/blob/a87b65a2a9db3f2c1ddcb4cf7112020cbf103abf/src/write/elf/object.rs#L50-L53).

wild_lib/src/layout.rs
marxin: Well, the wording is a bit convoluted, but it's correct:

> `// A bit in the output pr_data field is set only if it is set in all relocatable input pr_data field`

This correctly maps to the fact it's AND property - so a bit is 1 if it's 1 in all inputs (we are merging bits in `pr_data`).

> `// If all bits in the the output pr_data field are zero, this property should be removed from output.`

This speaks about a property (can have property type e.g. `GNU_PROPERTY_X86_FEATURE_1_AND`), where if you merge all bits and there is set none, you can omit the entire property. Values of this particular property can be `GNU_PROPERTY_X86_FEATURE_1_IBT`, or `GNU_PROPERTY_X86_FEATURE_1_SHSTK`.

Is it clearer now?

wild_lib/src/resolution.rs
marxin: All right, let me use `TemporaryPartId::NoteGnuProperty`, hope it's the right approach?!

marxin:
> So one possible fix is to make sure that `finalise_layout` increments `memory_offsets` for the `.note.gnu.property` section by whatever size it is using.

Thanks for reminding me how the linker works :) So yes, I used the appropriate way and bumped `memory_offsets`.

marxin:
> Linker-diff currently won't compare section content by default.

Heh. I was too optimistic as we'll need something more complicated. I noticed LLD omits the section for a trivial test case where BFD produces:
```
Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000030	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 ISA needed: x86-64-baseline
	x86 feature used: x86
	x86 ISA used: 
```

And I also noticed the order of the properties depends on the order of the files, where stable output can be ensured by sorting the properties by their type. So a proper comparison (using [^1]) would be a better approach!

[^1]: https://docs.rs/object/latest/object/read/elf/struct.Note.html#method.gnu_properties

marxin:
> If `Object` doesn't (currently) provide a struct that we can use, then perhaps just define a struct in `elf.rs` then derive `Pod`. e.g. see `EhFrameHdr`.

Yeah, I've just done that and the code seems nicer (at least I hope).

wild_lib/src/layout.rs
davidlattimore: I think I understand. So I had (mis) interpreted this as saying if all inputs are 0 then the output should be omitted, but it sounds like what it's actually saying is if the output is 0, then the output should be omitted. Is that right?

wild_lib/src/resolution.rs
davidlattimore: Rather than introducing a new variant of `TemporaryPartId`, would it work to use `TemporaryPartId::BuiltIn(part_id::NOTE_GNU_PROPERTY:)`?

wild_lib/src/layout.rs
marxin: Yep, nice wording!

wild_lib/src/resolution.rs
davidlattimore: I'll wait in case you wanted to try out `BuiltIn`. If you decide you're happy with it as-is, let me know and I'll merge

marxin: Yes, I would like to adjust the code so that it uses `TemporaryPartId::BuiltIn`, I will go back to that over the weekend. Plus, I would like to integrate a linker-diff pass for this section - as mentioned the bit-wise comparison is not enough!

marxin: If I use `TemporaryPartId::BuiltIn(part_id::NOTE_GNU_PROPERTY:)` I get crash here:
```
thread '<unnamed>' panicked at wild_lib/src/output_section_id.rs:554:13:
part_id_with_alignment can only be called for regular sections
```

that's called here:
https://github.com/davidlattimore/wild/blob/c8bbc8df13ef5bb23d8bf6fd49cbb3922704d86a/wild_lib/src/part_id.rs#L184

Any thought on that?

davidlattimore: One option would be to construct the return value directly after you check the section name. e.g.

```rust
if section_name == b".note.gnu.property" {
    return Ok(Some(UnresolvedSection {
        part_id: TemporaryPartId::BuiltIn(part_id::NOTE_GNU_PROPERTY),
        is_string_merge: false
    }));
}
```

marxin: Sure, that works, thanks.

marxin:
> So a proper comparison (using [1](#user-content-fn-1-163abd58d6389d62484e09c55d82c14f)) would be a better approach!

I take this back :P There are situation where probably some baseline note values are omitted:

```
❯ echo "int main() { return 0; }" | gcc -x c - -fuse-ld=bfd && readelf -n a.out

Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000030	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 ISA needed: x86-64-baseline
	x86 feature used: x86
	x86 ISA used: 
❯ echo "int main() { return 0; }" | gcc -x c - -fuse-ld=lld && readelf -n a.out
(the section is omitted)
❯ echo "int main() { return 0; }" | gcc -x c - -fuse-ld=mold && readelf -n a.out

Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000020	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 ISA needed: x86-64-baseline
	x86 feature used: x86
❯ echo "int main() { return 0; }" | gcc -x c - -B ~/Programming/wild && readelf -n a.out

Displaying notes found in: .note.gnu.property
  Owner                Data size 	Description
  GNU                  0x00000030	NT_GNU_PROPERTY_TYPE_0
      Properties: x86 ISA needed: x86-64-baseline
	x86 feature used: x86
	x86 ISA used: 
```

Wild's behavior matches to what BFD does. Would it be fine to merge the PR now?

PR #198, opened 2024-11-24
Remove explicit use of `core::mem::` functions

marxin:


davidlattimore:
I never realised that `size_of` was in the prelude. Good to know. I'm slightly less sure about using `take` and `replace` unqualified, but I'll give it a try and see if I get used to it.

PR #205, opened 2024-11-30
Remove .cargo/config.toml and forget it

### LINKED ISSUE #206: Add a gnu triple and use generic clang in both
Looking at #117 ....

made me want to check what linker we're using to link wild?
in the root folder, I see a .cargo/config.html file with:

```
[target.x86_64-unknown-linux-musl]
linker = "/usr/bin/clang-15"
rustflags = [ "-C", "relocation-model=static", "-C", "link-arg=--ld-path=wild" ]
```

The only problem is, I don't have `clang-15` installed on my system, so it can't be using that!
(hence PRs to warm benchmark repo to change to generic clang)...

Editing it to put giberish in there doesn't affect the wild build, and I realized the problem is the musl target triple, which is not the default target when I build.
Duplicating that to a gnu triple, with clang-15 still, makes the build fail with clang-15 not found, as expected.

Modifying it to use clang, the build proceeds, but crashes later.

Removing the rustflags line allows it to build and link.

restoring only: ` "-C", "relocation-model=static"` and it fails with:
```
  = note: /usr/bin/ld: /home/andrew/workspace/wild/target/release/deps/serde_derive-5535363a2c15caa6.serde_derive.4f3ca28e3f32b76f-cgu.00.rcgu.o: relocation R_X86_64_32 against `.rodata..L__unnamed_1' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: failed to set dynamic section sizes: bad value
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

remove that and restoring only: `["-C", "link-arg=--ld-path=wild"` allows it to build and link (wiht wild presumably).

So, I propose that to ensure developers build and link wild, with wild (do we want that yet) we change config.toml to be
```
[target.x86_64-unknown-linux-musl]
linker = "/usr/bin/clang"
rustflags = ["-C", "relocation-model=static", "-C", "link-arg=--ld-path=wild"]

[target.x86_64-unknown-linux-gnu]
linker = "/usr/bin/clang"
rustflags = ["-C", "link-arg=--ld-path=wild"]
```

Thoughts?
I have sent [a PR](https://github.com/davidlattimore/wild/pull/205) to that effect for comment....

andrewdavidmackenzie:
Fixes #206 
Add a gnu triple in .cargo/config.toml.
Ensure both triples use generic clang the user will have installed on their system, not clang-15.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,0 @@
-[target.x86_64-unknown-linux-musl]
-linker = "/usr/bin/clang"
-rustflags = ["-C", "relocation-model=static", "-C", "link-arg=--ld-path=wild"]
-
-[target.x86_64-unknown-linux-gnu]
-linker = "/usr/bin/clang"
-rustflags = ["-C", "link-arg=--ld-path=wild"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,7 @@
 [target.x86_64-unknown-linux-musl]
-linker = "/usr/bin/clang-15"
-rustflags = [ "-C", "relocation-model=static", "-C", "link-arg=--ld-path=wild" ]
+linker = "/usr/bin/clang"
+rustflags = ["-C", "relocation-model=static", "-C", "link-arg=--ld-path=wild"]
 
+[target.x86_64-unknown-linux-gnu]
+linker = "/usr/bin/clang"
+rustflags = ["-C", "link-arg=--ld-path=wild"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,8 +32,8 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
       - run: sudo apt install bubblewrap lld
-      - run: cargo build --no-default-features
-      - run: cargo test
+      - run: cargo build --workspace --no-default-features
+      - run: cargo test --workspace
 
   clippy:
     name: Clippy
@@ -53,7 +53,7 @@ jobs:
           ~/.cargo/git/db/
           target/
         key: ${{ runner.os }}-clippy-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
-    - run: cargo clippy --target x86_64-unknown-linux-gnu
+    - run: cargo clippy --workspace --target x86_64-unknown-linux-gnu
 
   rustfmt:
     name: Check formatting
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,28 +22,6 @@ jobs:
         with:
           targets: x86_64-unknown-linux-gnu,x86_64-unknown-linux-musl
           components: rustc-codegen-cranelift-preview
-      #      - uses: actions/cache@v4
-      #        with:
-      #          path: |
-      #            ~/.cargo/bin/
-      #            ~/.cargo/registry/index/
-      #            ~/.cargo/registry/cache/
-      #            ~/.cargo/git/db/
-      #            target/
-      #          key: ${{ runner.os }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
-      - run: sudo apt install bubblewrap lld
-      - run: cargo build --no-default-features
-      - run: cargo test
-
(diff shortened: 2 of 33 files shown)


davidlattimore:
My concern with this is that people won't be able to build wild if they don't already have a copy of wild. I have the following in my `~/.cargo/config.toml`

```toml
[target.x86_64-unknown-linux-gnu]
linker = "/usr/bin/clang"
rustflags = ["-C", "link-arg=--ld-path=/home/david/apps/bin/ld.wild-stable"]
```

I then occasionally copy a release build of wild to that location. I don't want wild continuously updating as I make changes, because then if I'm working on something and introduce a bug that made the linker non-functional, I wouldn't be able to rebuild with the bug fixed until I disabled wild.

We should possibly delete `.config/cargo.toml` from the repo, since it's a bit confusing as-is.

andrewdavidmackenzie:
That makes sense, unless you start doing releases, and people can easily previously install it using cargo binstall, or download a binary or such like. If you consider it too early to do releases, that would be fine, but maybe also an argument for not linking with wild either?

I use cargo-dist to automatically do releases (when you push a tag), if you wanted to do that, feel free to create an issue and mention me or assign to me.

I guess you could do something fancy with a build script, detecting it and falling back if not installed...

Another argument would be to not link with wild, until wild produces equally fast or faster code than alternatives.

If you just want folks to link using their default host linker, then removing config.toml would achieve that also.

davidlattimore:
> Another argument would be to not link with wild, until wild produces equally fast or faster code than alternatives.

I benchmarked a version of Wild linked using Wild against a version linked with GNU ld and couldn't measure any difference in performance. With LTO disabled in both, this is more or less what I'd expect. Linkers do do various relaxations, that micro-optimise bits of code. These can make a difference to performance, but Wild already does quite a few of them.

andrewdavidmackenzie:
Cool.

So linking wild with wild by default would be fine, once bootstrapped.

Options that occur to me:
1) we leave this file out and link with the default linker. Developers can add it locally.

2) developer has to edit the config to link with ld the first time (maybe cargo install wild that time, they will run into issues after a "cargo clean") to bootstrap the process.

3) we provide pre-built binary that user can install to bootstrap the process. This has a side benefit that folks can use wild without having to build it from source.

I'd vote for the option that gets us to "eating your own dogfood" faster, together with enabling easy use of it, so 3) I think.

davidlattimore:
Developers of Wild can hopefully consider making Wild their default linker not just for Wild, but for all their Rust development. That's what I do. I'm unsure about the extra hoops that someone would need to jump through in order to bootstrap. So my inclination is to just delete `.cargo/config.toml` from this repo (option 1). I'm also interested as to what @marxin thinks.

marxin:
> I'm also interested as to what @marxin thinks.

Yep, I support the deletion of the file as well.

andrewdavidmackenzie:
Sounds good.

IIRC, the config.toml file is loaded following a hierarchy of places it
looks for it.

So, maybe developers can set wild as default in one in their home directory?

Maybe a comment in the docs on that topic could help guide others?

davidlattimore:
> IIRC, the config.toml file is loaded following a hierarchy of places it looks for it. So, maybe developers can set wild as default in one in their home directory?

Yep, that's what I do in order to set it so that Wild is used for all Rust projects.

> Maybe a comment in the docs on that topic could help guide others?

Definitely a good idea. I guess once you've got the release builds working, we'll be adding installation instructions, so maybe that'd be a good point to add such docs - although before then is also fine.

davidlattimore:
Ah, I see you already documented it in this PR. That's reasonable. At some point I think we'll want to put it in the main README as installation instructions, so we can get people trying out wild and providing feedback even if they're not otherwise contributors.

davidlattimore:
It looks like the file might not have been deleted - or it got restored. Also, there seem to be merge conflicts.

andrewdavidmackenzie:
You will need to merge this to delete (from git) and forget it.
It may still stay around on your local systems.
I don't see merge conflicts here....

andrewdavidmackenzie:
Screenshot 2024-12-02 at 9 05 06 PM

davidlattimore:
Ah, you're right about no conflicts. Github was telling me that the branch cannot be rebased due to conflicts, however it can be squashed and merged, which is what I'd be wanting to do anyway.

However, the file doesn't look to be deleted. If I go to the "Files" tab in github, it shows that there are edits to the file, not that it has been deleted.

davidlattimore:
image

andrewdavidmackenzie:
Don't know what happened there. Deleted now.

PR #208, opened 2024-11-30
Support .note.gnu.build-id

mostafa-khaled775:
Implements the --build-id flag to generate GNU build IDs in ELF files.

The build ID is enabled by default (matching GNU ld behavior) and can be disabled with --build-id=none. The implementation uses SHA256 to generate a 32-byte build ID stored in the .note.gnu.build-id section.

Note: The current implementation computes the hash serially over the entire output file which can be slow for large outputs. A future optimization could follow mold's parallel approach as described in their [design doc](https://github.com/rui314/mold/blob/5bb914553fc0d0dcfe2b87204e06ebd1a9eff9b7/docs/design.md?plain=1#L152C1-L163C1)

### DIFF
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -268,7 +268,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
                 "fast" | "md5"| "sha1" => BuildIdOption::Fast,
                 "uuid" => BuildIdOption::Uuid,
                 s if s.starts_with("0x") || s.starts_with("0X")=> {
-                    let hex_string = s.strip_prefix("0x").unwrap_or_else(|| s.strip_prefix("0X").unwrap());
+                    let hex_string = &s[2..];
                     let decoded_bytes = hex::decode(hex_string).with_context(|| format!("Invalid Hex Build Id `0x{hex_string}`"))?;
                     BuildIdOption::Hex(decoded_bytes)
                 }
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -42,6 +42,7 @@ pub(crate) struct Args {
     pub(crate) files_per_group: Option<u32>,
     pub(crate) gc_sections: bool,
     pub(crate) should_fork: bool,
+    pub(crate) build_id: bool,
 
     /// If set, GC stats will be written to the specified filename.
     pub(crate) write_gc_stats: Option<PathBuf>,
@@ -113,8 +114,6 @@ pub(crate) const FILES_PER_GROUP_ENV: &str = "WILD_FILES_PER_GROUP";
 // other linkers. On the other, we should perhaps somehow let the user know that we don't support a
 // feature.
 const IGNORED_FLAGS: &[&str] = &[
-    // TODO: Support build-ids
-    "build-id",
     // TODO: Think about if anything is needed here. We don't need groups in order resolve cycles,
     // so perhaps ignoring these is the right thing to do.
     "start-group",
@@ -175,6 +174,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
     let mut soname = None;
     let mut execstack = false;
     let mut should_fork = true;
+    let mut build_id = true;
     let max_files_per_group = std::env::var(FILES_PER_GROUP_ENV)
         .ok()
         .map(|s| s.parse())
@@ -252,7 +252,13 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
                 bail!("Unsupported hash-style `{style}`");
             }
             // Since we currently only support GNU hash, there's no state to update.
-        } else if long_arg_split_prefix("build-id=").is_some() {
+        } else if long_arg_eq("build-id") {
+            build_id = true;
+        } else if let Some(build_id_value) = long_arg_split_prefix("build-id=") {
+            match build_id_value {
+                "none" => build_id = false,
+                _ => build_id = true,
+            };
         } else if long_arg_eq("time") {
             time_phases = true;
         } else if let Some(rest) = long_arg_split_prefix("threads=") {
@@ -437,6 +443,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
         files_per_group: max_files_per_group,
         execstack,
         should_fork,
+        build_id,
     }))
 }
 
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -174,7 +174,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
     let mut soname = None;
     let mut execstack = false;
     let mut should_fork = true;
-    let mut build_id = true;
+    let mut build_id = false;
     let max_files_per_group = std::env::var(FILES_PER_GROUP_ENV)
         .ok()
         .map(|s| s.parse())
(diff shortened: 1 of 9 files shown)


wild_lib/src/output_section_id.rs
```diff
@@ -934,8 +943,11 @@ fn test_constant_ids() {
         (PLT_GOT, ".plt.got"),
         (NOTE_ABI_TAG, ".note.ABI-tag"),
         (NOTE_GNU_PROPERTY, ".note.gnu.property"),
+        (NOTE_GNU_BUILD_ID, ".note.gnu.build-id"),
     ];
     for (id, name) in check {
+        dbg!(std::str::from_utf8(id.built_in_details().name.bytes()).unwrap());
```
davidlattimore: I assume these aren't still needed here?

wild_lib/src/elf_writer.rs
```diff
@@ -282,9 +287,43 @@ impl SizedOutput {
...
+        name_out.copy_from_slice(GNU_NOTE_NAME);
+
+        rest.copy_from_slice(&computed_build_id.as_slice());
+
         Ok(())
     }
 
+    #[instrument(skip_all, name = "Compute build ID")]
+    fn compute_gnu_build_id_note(&self) -> GnuBuildId {
+        let output_buffer = match &self.out {
```
davidlattimore: I think this can probably be simplified a bit here. `OutputBuffer` implements `Deref`. So you can probably just do something like `let output_buffer = &*self.out;` or, if you prefer, `self.out.deref()`.

wild_lib/src/args.rs
```diff
@@ -175,6 +174,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
     let mut soname = None;
     let mut execstack = false;
     let mut should_fork = true;
+    let mut build_id = true;
```
davidlattimore: As far as I can tell, GNU ld and lld both default to no build-id.

davidlattimore (commented):
Nice work! Thanks for doing this!

wild_lib/src/elf_writer.rs
```diff
@@ -282,9 +287,43 @@ impl SizedOutput {
...
+            OutputBuffer::Mmap(mmap) => mmap.deref(),
+            OutputBuffer::InMemory(vec) => vec.deref(),
+        };
+
+        GnuBuildId::new(Sha256::digest(output_buffer).into())
```
marxin: Please use a faster hash algorithm (we can use BLAKE3 similar to Mold linker), and we can also make the hashing in parallel. Using the following code snippet, I reduced the time spent in hashing for building Clang from 1.5s to 100ms, which is about 2% of the total link time (w/ debug info). Pretty reasonable to me:

```rust
let mut hasher = blake3::Hasher::new();
        let chunks = output_buffer
            .chunks(bytesize::mib(16u64) as usize)
            .collect_vec();
        let hashes: Vec<_> = chunks.par_iter().map(|chunk| blake3::hash(chunk)).collect();
        for hash in hashes {
            hasher.update(hash.as_bytes());
        }

        GnuBuildId::new(hasher.finalize().into())
```

wild_lib/src/args.rs
```diff
@@ -252,7 +252,13 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
                 bail!("Unsupported hash-style `{style}`");
             }
             // Since we currently only support GNU hash, there's no state to update.
-        } else if long_arg_split_prefix("build-id=").is_some() {
+        } else if long_arg_eq("build-id") {
+            build_id = true;
+        } else if let Some(build_id_value) = long_arg_split_prefix("build-id=") {
```
marxin, line 265: Consider also supporting a user-defined set hash, similar to other linkers do:

```
$ ld.lld --help | less
  --build-id=[fast,md5,sha1,uuid,0x<hexstring>]
                          Generate build ID note
  --build-id              Alias for --build-id=sha1
```

marxin (changes requested):
Thanks for working on the feature!

Theoretically, we can start hashing the output content earlier, but this implementation is quite straightforward and simple. I value that.

wild_lib/src/args.rs
m04f: Just to confirm, are you requesting support for a --build-id=0x\<hexstring> style option where users can specify the Build ID explicitly?

marxin: Yes, please. For all the other option values (`fast`, `md5`, `sha1`) except `uuid`, I would use the BLAKE3 hashing algorithm.

wild_lib/src/elf_writer.rs
```diff
@@ -282,9 +285,51 @@ impl SizedOutput {
...
+
+        rest.copy_from_slice(&computed_build_id.as_slice());
+
+        Ok(())
+    }
+
+    #[instrument(skip_all, name = "Compute build ID")]
+    fn compute_gnu_build_id_note(&self) -> GnuBuildId {
+        let output_buffer = self.out.deref();
+        fn par_hash(buffer: &[u8]) -> blake3::Hash {
```
marxin: Please use my suggested `par_iter()` instead of this recursive tree combining the hashes, I think it's more idiomatic.

wild_lib/src/args.rs
```diff
@@ -252,7 +261,30 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
...
-        } else if long_arg_split_prefix("build-id=").is_some() {
+        } else if long_arg_eq("build-id") {
+            build_id = BuildIdOption::Fast;
+        } else if let Some(build_id_value) = long_arg_split_prefix("build-id=") {
+            build_id = match build_id_value {
+                "none" =>  BuildIdOption::None,
+                "fast" | "md5"| "sha1" => BuildIdOption::Fast,
+                "uuid" => BuildIdOption::Uuid,
+                s if s.starts_with("0x") => {
+                    let hex_string = s.strip_prefix("0x").unwrap();
```
marxin: Can we please use `hex` crate and `hex::decode` function?

wild_lib/src/elf_writer.rs
```diff
@@ -282,9 +286,65 @@ impl SizedOutput {
...
+        let name_out = crate::slice::slice_take_prefix_mut(&mut rest, GNU_NOTE_NAME.len());
+        name_out.copy_from_slice(GNU_NOTE_NAME);
+
+        rest.copy_from_slice(build_id);
+
         Ok(())
     }
 
+    #[instrument(skip_all, name = "Compute build ID")]
+    fn compute_hash(&self) -> blake3::Hash {
```
marxin, line 332: I really like the function changes you made in the function!

wild_lib/src/args.rs
```diff
@@ -252,7 +260,20 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(mut input: I) -> Resul
...
-        } else if long_arg_split_prefix("build-id=").is_some() {
+        } else if long_arg_eq("build-id") {
+            build_id = BuildIdOption::Fast;
+        } else if let Some(build_id_value) = long_arg_split_prefix("build-id=") {
+            build_id = match build_id_value {
+                "none" =>  BuildIdOption::None,
+                "fast" | "md5"| "sha1" => BuildIdOption::Fast,
+                "uuid" => BuildIdOption::Uuid,
+                s if s.starts_with("0x") || s.starts_with("0X")=> {
+                    let hex_string = s.strip_prefix("0x").unwrap_or_else(|| s.strip_prefix("0X").unwrap());
```
marxin: Just a small nit: I would rather use `s[2..]`.

marxin (approved):
LGTM, thank you.

davidlattimore:
Thanks for doing this @mostafa-khaled775! There's one clippy warning, but rather than doing another roundtrip and getting you to fix it, I'll just merge now then push a fix for the warning.

PR #209, opened 2024-11-30
Prepare releases, including pre-built binaries

### LINKED ISSUE #207: Set up binary releases
Looks like the two most popular options are [cargo-dist](https://crates.io/crates/cargo-dist) and [release-plz](https://crates.io/crates/release-plz). I'm not sure what the tradeoffs between them are or which one is likely to be better for our purposes, but it's probably a good idea to have a quick look at both. @andrewdavidmackenzie, would you like to take a look?

andrewdavidmackenzie:
Fixes #207 

Using cargo-dist to do the builds for selected targets and then create the release, upload sourc etarball as usual plus binary assets.

Adds a short RELEASING.md document to document the process.

Here is the first attempt at a release build in my own repo:

https://github.com/andrewdavidmackenzie/wild/actions/runs/12099771708

It's lacking something installed for musl so fails, which I'll try and fix.


andrewdavidmackenzie:
`warning: symbolic-demangle@12.12.1: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `musl-g++` installed?`

davidlattimore:
> `warning: symbolic-demangle@12.12.1: Compiler family detection failed due to error: ToolNotFound: Failed to find tool. Is `musl-g++` installed?`
> 
> on my machine I (seem to, never knew!!) have musl-gcc installed. I guess it needs that?

Is this for running tests? There are certainly some tests that check that static linking with musl works

davidlattimore:
Do you want to try this out on your fork to see if it works as expected? You might need to temporarily enable actions on your fork, since I think they're disabled by default.

andrewdavidmackenzie:
It's trying to do the release build in my forks actions.

I'll try and get it passing and a release created in my fork, before
removing draft/wip status

andrewdavidmackenzie:
`cargo build --target x86_64-unknown-linux-musl`

fails on my local machine with the same error.

`error occurred: Failed to find tool. Is `musl-g++` installed?`

I guess you installed it manually, but I'm confused why I don't see it in ci.yml

andrewdavidmackenzie:
I see `symbolic-demangle` is referenced in cackle.yml

andrewdavidmackenzie:
Without understanding all this, it seems that ci.yml installs bubblewrap

```
      - run: sudo apt install bubblewrap lld
```

which is referenced from cackle.yml as the sandbox...

andrewdavidmackenzie:
symbolic-demangle doesn't support building for musl. I asked them about that:

https://github.com/getsentry/symbolic/issues/880

andrewdavidmackenzie:
A bit of a rat-hole...

We could dodge the whole issue and reduce the binaries we would build as part of a release...

andrewdavidmackenzie:
Dropping musl is simple and would avoid this problem, even if releasing all binaries.

Releasing only linker (gnu and musl) would also avoid it.

Apart from for musl based systems like alpine, what other reason to release musl binary?

andrewdavidmackenzie:
I tried reducing build to just wild, but it has a dev-dependency on linker-diff, and I think that's triggering the linker-diff build on musl, which fails.

So then I removed musl also.

If we want to re-add musl, even if just for wild, we'll need to resolve the build on linker diff, or find a way to avoid it in release builds of the wild binary (tests aren't run, so strictly speaking it's not needed but AFAIK, cargo doesn't distinguish between build dependencies and test dependencies in the dev-dependencies list...)

andrewdavidmackenzie:
Successfull release build in GH Action in my repo: https://github.com/andrewdavidmackenzie/wild/actions/runs/12117386105

Here is the release it created: https://github.com/andrewdavidmackenzie/wild/releases

Once you have a look and we merge this, I might delete that release to avoid confusion.

davidlattimore:
Looks good to me, although there's merge conflicts

andrewdavidmackenzie:
Fixed

Issue #203, opened 2024-11-29
Names of sections: should be a const used


marxin:
What's your opinion about the following type of changes:

```diff
diff --git a/wild_lib/src/output_section_id.rs b/wild_lib/src/output_section_id.rs
index 19f6fcf..893b56a 100644
--- a/wild_lib/src/output_section_id.rs
+++ b/wild_lib/src/output_section_id.rs
@@ -106,6 +106,10 @@ pub(crate) const NOTE_ABI_TAG: OutputSectionId = OutputSectionId::regular(13);
 
 pub(crate) const NUM_BUILT_IN_REGULAR_SECTIONS: usize = 14;
 
+// List of sections names, so that we don't repeat the names in the linker.
+
+pub(crate) const NOTE_ABI_TAG_SECTION_NAME: &str = ".note.ABI-tag";
+
 pub(crate) struct OutputSections<'data> {
     /// The base address for our output binary.
     pub(crate) base_address: u64,
@@ -481,7 +485,7 @@ const SECTION_DEFINITIONS: [BuiltInSectionDetails; NUM_BUILT_IN_SECTIONS] = [
         ..DEFAULT_DEFS
     },
     BuiltInSectionDetails {
-        name: SectionName(b".note.ABI-tag"),
+        name: SectionName(NOTE_ABI_TAG_SECTION_NAME.as_bytes()),
         ty: sht::NOTE,
         section_flags: shf::ALLOC.with(shf::GNU_RETAIN),
         ..DEFAULT_DEFS
@@ -922,7 +926,7 @@ fn test_constant_ids() {
         (SECTION_HEADERS, ".shdr"),
         (GNU_HASH, ".gnu.hash"),
         (PLT_GOT, ".plt.got"),
-        (NOTE_ABI_TAG, ".note.ABI-tag"),
+        (NOTE_ABI_TAG, NOTE_ABI_TAG_SECTION_NAME),
     ];
     for (id, name) in check {
         assert_eq!(
diff --git a/wild_lib/src/part_id.rs b/wild_lib/src/part_id.rs
index a583457..1cbc8a9 100644
--- a/wild_lib/src/part_id.rs
+++ b/wild_lib/src/part_id.rs
@@ -1,3 +1,4 @@
+use self::output_section_id::NOTE_ABI_TAG_SECTION_NAME;
 use crate::alignment::Alignment;
 use crate::alignment::NUM_ALIGNMENTS;
 use crate::args::Args;
[...]

PR #210, opened 2024-12-01
Use constants for section names

marxin:
Resolves: #203


davidlattimore:
Sorry for the delay reviewing this. I think I knew it would conflict with the build-id PR, so thought I'd wait until that was in, but then forgot to come back to it. Anyway, this looks good, but there is indeed a conflict

PR #220, opened 2024-12-04
WIP - Macos tests run

andrewdavidmackenzie:
Just tinkering on mac (aarch64).


andrewdavidmackenzie:
I skipped using ld.lld on mac, as it looks a bit like it should be ld64 or something (more clarity required on that).

Then it starts failing tests due to unrecognized assembly instructions - which makes sense.

andrewdavidmackenzie:
Confirmed that macos-15 runs on arm64...
"Run uname -a
Darwin Mac-1733386436564.local 24.1.0 Darwin Kernel Version 24.1.0: Thu Oct [10] 21:04:31 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_VMAPPLE arm64"
Just FYI

marxin:
Can we please close this PR as it's only something triggering GitHub actions?

PR #226, opened 2024-12-07
Fix propagation of error code when forking

davidlattimore:


wild_lib/src/subprocess.rs
```diff
@@ -89,9 +89,9 @@ fn wait_for_child_done(fds: &[c_int], child_pid: pid_t) -> i32 {
             }
             _ => {
                 // Child closed pipe without sending a byte - get the process exit_status
-                let mut child_exit_status = -1i32;
-                libc::waitpid(child_pid, &mut child_exit_status as *mut c_int, 0);
-                child_exit_status
+                let mut status: libc::c_int = -1i32;
+                libc::waitpid(child_pid, &mut status, 0);
+                libc::WEXITSTATUS(status)
```
andrewdavidmackenzie, line 94: Does this set errorno as well, or what does it do?

davidlattimore: The status written by `waitpid` is a 32 bit value that has all sorts of stuff packed into it. What was happening before, was we were returning that whole 32 bit value then passing that to `exit`. However `exit` on Linux ignores all but the low 8 bits, which were zero, while the actual exit code is in the next 8 bits.

The definition of `WEXITSTAUS` is `(status >> 8) & 0xff`.

PR #227, opened 2024-12-07
Do string merging in batches

davidlattimore:
This allows us to reuse some heap allocations, rather than needing 32 heap allocations for each input string-merge section.

Before we did all splitting and hashing during section resolution, then deduplication later in the string merging phase.

Now, we do multiple batches where we alternate between splitting/hashing and deduplication. This allows us to reuse the bucket allocations for each batch.

This also makes timing output more meaningful, since all the string merging is now done in the string merging phase rather than some of it being done during section resolution.

### DIFF
--- a/wild_lib/src/output_section_map.rs
+++ b/wild_lib/src/output_section_map.rs
@@ -1,3 +1,4 @@
+use crate::error::Result;
 use crate::output_section_id::OutputSectionId;
 
 /// A map from output section IDs to something.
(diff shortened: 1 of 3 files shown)


davidlattimore:
I think this helps performance. It certainly reduces the number of heap allocations. I have been having trouble benchmarking because my laptop seems to be thermal throttling, which is making comparison very hard. Not sure what's up with my laptop. It only started doing it this morning.

PR #234, opened 2024-12-10
Use rayon feature of blake3

davidlattimore:


marxin (approved):
Neat, how have you noticed they provide the feature?

davidlattimore:
> Neat, how have you noticed they provide the feature?

I was profiling the linker and noticed a bunch of stuff happening near the end that didn't have debug info or even symbols. Took me a bit to realise that it was probably the build ID stuff. Then I went looking at the blake3 crate to see if it was C or Rust and what build options were available. I didn't get very far that with, but in the process, I looked at what features were available on the crate and noticed the rayon feature.

PR #239, opened 2024-12-14
Come up with `RelocationSize` for future use for AArch64

marxin:
This is a preparation change for AArch64 where various relocation values are stored partially:
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#id53

```
285	15	R_<CLS>_ LDST32_ABS_LO12_NC	S + A	Set the LD/ST immediate value to bits [11:2] of X. No overflow check
```

and so I introduce `RelocationSize::BitRange`. I wish the `write_to_buffer` implementation would be simpler, but it seems Rust is quite strict when it comes to overflow of the shift operations and thus I combine masking with SHR.

Plus, I would like to ask why do you add and subtract `rel_info.byte_size` for `RelocationKind::Relative` relocation value calculation? That would make it more complicated for `BitRange` type.

### DIFF
--- a/wild_lib/src/elf.rs
+++ b/wild_lib/src/elf.rs
@@ -454,6 +454,7 @@ pub(crate) enum RelocationKind {
 #[derive(Clone, Debug)]
 pub(crate) enum RelocationSize {
     ByteSize(usize),
+    #[allow(dead_code)]
     BitRange(Range<usize>),
 }
 
--- a/wild_lib/src/elf.rs
+++ b/wild_lib/src/elf.rs
@@ -2,6 +2,7 @@ use crate::error::Result;
 use crate::resolution::LoadedMetrics;
 use anyhow::anyhow;
 use anyhow::bail;
+use anyhow::ensure;
 use anyhow::Context;
 use bytemuck::Pod;
 use bytemuck::Zeroable;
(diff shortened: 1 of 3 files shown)


davidlattimore:
> Plus, I would like to ask why do you add and subtract `rel_info.byte_size` for `RelocationKind::Relative` relocation value calculation?

I guess adding, then later subtracting can give a different result if the relocation references a merged string. Adding the byte size of the relocation is likely assuming that the relocation is at the end of the instruction and thus the instruction pointer when the instruction is being evaluated will be pointing to the byte immediately after the relocation. Removing it, does appear to still pass the tests. So maybe we should remove it. If it turns out that it is needed, we can then add additional tests when we add it back again. Might be best to remove it in a separate commit though.

I wonder if the enum will optimise away on x86 where only a single branch is constructed. I guess we should watch the benchmarks as you add the ARM code to make sure x86 doesn't slow down. If it does, then we could consider making RelocationSize be an associated type on the Arch trait - that way the x86 code will only have to deal with bytes. Does ARM always deal with bits? If to does, maybe an associated type might be the way to go.

Removing `Copy` from a few types seems like a shame. Perhaps skip on using `Range` and just have `start` and `stop` fields?

marxin:
> Does ARM always deal with bits? If to does, maybe an associated type might be the way to go.

I would say the majority of the relocations would be of the `BitRange`, however, there are also standard relocations that will use the other type. Based on my preliminary measurements, I haven't noticed any measurable slowdown when linking Clang on x86_64 if I replaced all `ByteSize(X)` with `BitRange(0, 8 * X)`.

> Removing `Copy` from a few types seems like a shame. Perhaps skip on using `Range` and just have `start` and `stop` fields?

Yep, makes sense.

Anyway, I tend to close this PR as I have a more complex patch-set that can add the rudimentary support for AArch64.

Issue #240, opened 2024-12-14
Support `--no-add-needed`/`--no-copy-dt-needed-entries`


marxin:
For some strange reason, Fedora uses the legacy option `--no-add-needed` that was renamed to `--no-copy-dt-needed-entries`:

```
 --no-copy-dt-needed-entries
           This option affects the treatment of dynamic libraries referred to by DT_NEEDED tags inside ELF dynamic libraries mentioned on the command line.  Normally the linker won't add a DT_NEEDED tag to the output binary for each library mentioned in a DT_NEEDED tag in an input dynamic library.  With --copy-dt-needed-entries specified on  the  command  line
           however any dynamic libraries that follow it will have their DT_NEEDED entries added.  The default behaviour can be restored with --no-copy-dt-needed-entries.

           This option also has an effect on the resolution of symbols in dynamic libraries.  With --copy-dt-needed-entries dynamic libraries mentioned on the command line will be recursively searched, following their DT_NEEDED tags to other libraries, in order to resolve symbols required by the output binary.  With the default setting however the searching of
           dynamic libraries that follow it will stop with the dynamic library itself.  No DT_NEEDED links will be traversed to resolve symbols.
```

I am pretty sure it's the behavior Wild follows, but I would like to verify that from @davidlattimore :)

davidlattimore:
Yep, I don't recall implementing copying of `DT_NEEDED` entries, nor the associated symbol lookup behaviour, so I'd say wild currently behaves as if `--no-copy-dt-needed-entries` were supplied - should be fine to accept the `--no-` versions of those flags and do nothing.

PR #241, opened 2024-12-15
Add various flags that are mapped to the default behavior of the linker

marxin:
Fixes: #240

### DIFF
--- a/wild_lib/src/args.rs
+++ b/wild_lib/src/args.rs
@@ -135,11 +135,17 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "fatal-warnings",
     "color-diagnostics",
     "undefined-version",
-    "no-call-graph-profile-sort",
-    "relax",
 ];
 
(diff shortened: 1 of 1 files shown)


wild_lib/src/args.rs
```diff
@@ -139,7 +139,13 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "relax",
 ];
 
-const IGNORED_FLAGS: &[&str] = &["gdb-index", "no-relax", "disable-new-dtags"];
+const IGNORED_FLAGS: &[&str] = &[
```
davidlattimore: Adding the flags here will mean that we issue a warning when those flags are supplied. Was that your intention? Given that the flags should be no-ops, perhaps we shouldn't issue a warning? Some of the flags above are probably no-op flags too, like "relax" and possibly "no-call-graph-profile-sort". Perhaps the new ones here should be grouped with those.

marxin: Yeah, sorry, I forgot to test my changes :) It was not intentional.

Issue #43, opened 2024-08-18
Support aarch64


davidlattimore:
The main code in the linker that would be different will be the code in `relaxation.rs`. Any code that directly references x86-64 relocation types will also need to change, although outside of `relaxation.rs` there's not too much of that.

linker-diff will need some changes. A disassembler for aarch64 will need to be selected and used in `asm_diff.rs`.

This will likely be the first port of Wild. Some care will be needed to make sure that performance doesn't degrade. The simplest option would be to switch between different bits of code (e.g. different relaxation schemes) at each call site via either a dyn object, or more likely an enum. Using a dyn object would prevent inlining, so might affect performance. Using an enum wouldn't prevent inlining, so might be faster. Even with an enum, we'd still be branching. It's likely that an enum will be sufficient and that there will be no measurable slow-down. However if that's not the case, then 3rd option would be make the code generic over a parameter `P: Platform`. Switching between different implementations can then be done at some outer, less-hot point in the code.

I'd be very tempted to just jump straight to the generic-over-platform option, but don't want to preclude the other options.

One advantage of the code being generic over platform is that the relaxation code can use platform-specific types. e.g. maybe the enum `RelaxationKind` might need to be different depending on the platform. However maybe it'd be better to move the little bits of code that use a type like `RelaxationKind` into `relaxation.rs` and make `RelaxationKind` private, then the type can become an internal implementation detail of the platform and doesn't need to be exposed at all.

andrewdavidmackenzie:
This refers to supporting aarch64 targets for the code generated, not (necessarily) running on aarch64, right?

That (for running on mac apple silicon, Raspberry PI, or other ARM PCs that are emerging) would need a separate issue I assume.

davidlattimore:
> This refers to supporting aarch64 targets for the code generated, not (necessarily) running on aarch64, right?

Yep, it's about the code generated. I wouldn't expect any substantial work to be required to actually run the linker on the other targets. I'd certainly expect that it would already run fine on a Raspberry Pi (if you gave it X86_64 objects to link).

davidlattimore:
I should mention that I submitted a [change yesterday](https://github.com/davidlattimore/wild/pull/191) that added an `Arch` trait. This was done to make porting to ARM and other targets easier. I figured I'd add the trait because it's kind of an invasive change. I'm not currently working on actually porting to any other architectures, so you or anyone else should feel free to do so. Also, don't feel that the trait is set in stone. It can definitely be changed as needed.

[...]

PR #244, opened 2024-12-15
Rudimentary AArch64 architecture support

marxin:
This a very basic support for the AArch64 architecture where I could link a `trivial.C` test case (2 relocation types are needed) on the real hardware :confetti_ball: I included parsing of the `-m` option in args.rs where default architecture is selected based on target_arch. Later on, the option value is converted to the proper `Arch` implementation and I've added more template Arch arguments in elf_writer.rs.

Please review the changes before I'll carry on :)

Related to #43.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,5 @@
 {
     "rust-analyzer.linkedProjects": [
         "Cargo.toml",
-    ],
-    "files.associations": {
-        "*.dbclient-js": "javascript",
-        "*.tcc": "c"
-    }
+    ]
 }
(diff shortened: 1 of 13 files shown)


davidlattimore:
Nice! This looks like a great start. Did you want to merge this or wait until it's more complete?

BTW, I was able to measure a small slow down for x86-64. Somewhere around 1%, which is close to the limit of what I can measure, however it does seem reasonably consistent. That doesn't need to block this though, since we can always make RelocationSize an associated type later - assuming that's the cause of the slowdown that is.

marxin:
Thanks! I've extended the implementation where all non-TLS and non-dynamic relocation types should be supported now. There are plenty of failing tests, but I think it's a good starting point. Can we please merge it?

PR #250, opened 2024-12-24
Add branchless version of `parse_decimal_int`

CrazyboyQCD:
See if this improve the performance.


davidlattimore:
Thanks. I ran a micro benchmark of just the old vs new versions of the function and it runs in about 60% of the time that the old version takes. Unfortunately it doesn't make any measurable difference to the total time to link, since the function isn't called enough to make a difference. As the comment in the code mentions, parsing often shows up in profiles, but that's just because it's touching pages that haven't been mapped yet, so it's triggering page faults. In any case, the function is itself faster, so I'll merge this. Thanks :)

PR #252, opened 2024-12-30
AArch64: implement proper application of the insn-related relocations

marxin:
This adds yet another relocation application layer where the AArch64 instruction relocations have a handler that known how to encode a computed value into a given instruction.

With the change, the tests improved from:
`test result: FAILED. 5 passed; 26 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.25s`
to:
`test result: FAILED. 15 passed; 16 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.86s`

### DIFF
--- a/wild_lib/src/aarch64.rs
+++ b/wild_lib/src/aarch64.rs
@@ -1,5 +1,8 @@
+use crate::elf::extract_bits;
+use crate::elf::BitRange;
 use crate::elf::DynamicRelocationKind;
 use crate::elf::PageMask;
+use crate::elf::RelocationInsn;
 use crate::elf::RelocationKind;
 use crate::elf::RelocationKindInfo;
 use crate::elf::RelocationSize;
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):
A few questions about `RelocationInsn`...

Iis "insn" short for instruction? Perhaps spell it out in full.

Currently the `RelocationInsn` is part of the `RelocationSize`. I guess this is because it's only used in conjunction with BitMasking? It does seem slightly odd given that it's not really size-related. Maybe we just need a new name other than size, although I have no idea what might be suitable.

I assume it's ARM-specific? The variants of the enum look pretty ARM-specific anyway. It does look like at some point, that we'll probably want to have maybe an extra associated type on `Arch`, so that it can be fully ARM-specific. But it's fine to leave that until later.

Anyway, I'm happy for this to be merged as-is - perhaps with `insn` renamed. Unless insn is what it's referred to in the docs? I tried searching the PDF you linked and couldn't find it - although I had trouble loading the PDF due to it being 14k pages.

marxin:
> Is "insn" short for instruction? Perhaps spell it out in full.

Correct, it's a short variant for `Instruction` and I've just renamed it.

> I assume it's ARM-specific? The variants of the enum look pretty ARM-specific anyway. It does look like at some point, that we'll probably want to have maybe an extra associated type on `Arch`, so that it can be fully ARM-specific. But it's fine to leave that until later.

Exactly. Right now, it's all ARM64-specific, while the `RelocationSize` extraction mechanism is pretty generic and might be reused in the future for other architectures.

PR #261, opened 2024-12-30
Allow phase 2 of string merging to run in parallel on a single phase 1 output

davidlattimore:


wild_lib/src/string_merging.rs
```diff
@@ -919,3 +859,18 @@ impl std::fmt::Display for MergeString<'_> {
...
+
+/// Returns an empty `Vec<U>` that reuses the storage of the supplied `Vec<T>`. `T` and `U` must
+/// have the same size and alignment.
+fn reuse_vec<T, U>(mut v: Vec<T>) -> Vec<U> {
+    debug_assert_eq!(size_of::<T>(), size_of::<U>());
+    debug_assert_eq!(align_of::<T>(), align_of::<U>());
+    let old_storage = v.as_ptr();
+    v.clear();
+    // Convert the type of the vec. This relies on a specialised implementation of `collect`. Were
+    // it not for that, we'd get a new heap allocation, which would default the purpose.
```
lqd: I assume this is a tiny typo

```suggestion
    // it not for that, we'd get a new heap allocation, which would defeat the purpose.
```

davidlattimore: Thanks! Fixed

PR #268, opened 2025-01-02
linker-diff: Skip asm-diff based on object arch not host arch

davidlattimore:


marxin:
Was addition of `libwild/src/layout2.rs` intentional as part of this change?

davidlattimore:
> Was addition of `libwild/src/layout2.rs` intentional as part of this change?

Oops! Thanks for pointing that out. I've removed it.

PR #269, opened 2025-01-02
Add flag --update-in-place

davidlattimore:
When set, this causes us to overwrite the existing file instead of deleting it and putting a new file in its place. On one benchmark (rustc without debug) this gave about a 4% speedup. On another (clang with debug) the difference was small enough that it was hard to measure.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -45,6 +45,7 @@ pub(crate) struct Args {
     pub(crate) gc_sections: bool,
     pub(crate) should_fork: bool,
     pub(crate) build_id: BuildIdOption,
+    pub(crate) file_write_mode: FileWriteMode,
 
     /// If set, GC stats will be written to the specified filename.
     pub(crate) write_gc_stats: Option<PathBuf>,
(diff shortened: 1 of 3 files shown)


marxin:
I can measure a speed improvement of **~7%** for the Clang binary w/ debug info :confetti_ball:
What about making it the default option?

davidlattimore:
> I can measure a speed improvement of **~7%** for the Clang binary w/ debug info 🎊
> What about making it the default option?

It does have some caveats. If the file is currently executing, it'll fail with `ETXTBSY`. We could do what mold does and detect that error, then fall back to deleting the old file. This is complicated by the fact that they removed `ETXTBSY` in kernel 6.11. They then reverted the removal in a later version after Rui complained (see release notes for mold 2.35.0). I'm actually slightly bummed that it got reverted. I had kind of wanted `ETXTBSY` to go away for some time (mostly for reasons related to hot code reloading). It seems weird and inconsistent that we error for writing a binary that's being executed but not for a shared object that's being executed. Anyway, it's reverted now and I'd be surprised if `ETXTBSY` gets removed again at this point. That does however mean that `--update-in-place` is not safe on kernel 6.11 and not safe when writing shared objects. Mold apparently now detects the kernel version and doesn't overwrite in place if it detects 6.11. Presumably it also uses full deletion and recreation when the output is a shared object.

marxin:
Oh, what an interesting story! Thanks for the deep explanation and based on what you described, I lean towards keeping `--update-in-place` non-default option.

PR #270, opened 2025-01-03
AArch64: add write_plt_entry

marxin:
Right now, we share the entry size for x86_64 and AArch64; however, we could use a smaller size for AArch64. 

The problem is that the size is used in `const SECTION_DEFINITIONS`, a const array for which we don't currently make any special code based on the `Arch`. Do you have any idea how to tackle this in the future?

With the changes, I was able to emit a working PLT entry for a symbol from `libc` shared library on AArch64 platform.

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -6,11 +6,26 @@ use crate::elf::RelocationInstruction;
 use crate::elf::RelocationKind;
 use crate::elf::RelocationKindInfo;
 use crate::elf::RelocationSize;
+use crate::elf::DEFAULT_AARCH64_PAGE_IGNORED_MASK;
+use crate::elf::DEFAULT_AARCH64_PAGE_MASK;
+use crate::elf::DEFAULT_AARCH64_PAGE_SIZE;
+use crate::elf::PLT_ENTRY_SIZE;
 use anyhow::bail;
 use anyhow::Result;
 use linker_utils::elf::aarch64_rel_type_to_string;
 
 pub(crate) struct AArch64;
+
+const PLT_ENTRY_TEMPLATE: &[u8] = &[
+    0x10, 0x00, 0x00, 0x90, // adrp x16, page(&(.got.plt[n]))
+    0x11, 0x02, 0x40, 0xf9, // ldr x17, [x16, offset(&(.got.plt[n]))]
+    0x20, 0x02, 0x1f, 0xd6, // br x17
+    0x1f, 0x20, 0x03, 0xd5, // nop
+];
+
+const _ASSERTS: () = {
+    assert!(PLT_ENTRY_TEMPLATE.len() as u64 == PLT_ENTRY_SIZE);
+};
 
 impl crate::arch::Arch for AArch64 {
     type Relaxation = ();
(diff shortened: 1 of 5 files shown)


libwild/src/aarch64.rs
```diff
@@ -753,6 +768,33 @@ impl crate::arch::Arch for AArch64 {
...
+    fn write_plt_entry(
+        plt_entry: &mut [u8],
+        got_address: u64,
+        plt_address: u64,
+    ) -> crate::error::Result {
+        plt_entry.copy_from_slice(PLT_ENTRY_TEMPLATE);
+        let plt_page_address = plt_address & DEFAULT_AARCH64_PAGE_IGNORED_MASK;
+        let offset = got_address.wrapping_sub(plt_page_address) as i64;
+        anyhow::ensure!(
+            ((-1 << 32)..=1 << 32).contains(&offset),
```
davidlattimore: Is this correct that the end of the range is included? i.e we can represent 2^32? It also seems surprising that our range is +/- 4GiB. i.e. we have 33 bits (or 34 if we count being able to represent 2^32).

Looking below at the way the value is being packed, it looks like the limiting factor is the page offset in the ADR instruction. It looks like it takes 32 bits from the supplied value, which would mean that we could address 2^32 pages. The page size is 12 bits, so we can perhaps represent 2^44 offsets.

libwild/src/aarch64.rs
```diff
@@ -753,6 +768,33 @@ impl crate::arch::Arch for AArch64 {
...
+            "PLT is more than 4GiB away from GOT"
+        );
+        RelocationInstruction::Adr.write_to_value(
+            // The immediation value represents a distance in pages.
+            offset as u64 / DEFAULT_AARCH64_PAGE_SIZE,
```
davidlattimore: Division of a negative value cast to an unsigned value will give a very different result than division of the original negative value. This makes me think that perhaps this won't work if `offset` is negative. Using a bitmask instead of division might be more likely to produce the desired result. If we're not sure we want to handle negative offsets now, then you could change the ensure above to disallow negative offsets. That would mean that we wouldn't support placing the PLT after the GOT, but until such time as we support linker scripts, that's unlikely to be an issue.

libwild/src/elf.rs
```diff
@@ -355,18 +355,14 @@ pub(crate) const COMPRESSION_HEADER_SIZE: usize =
     size_of::<object::elf::CompressionHeader64<LittleEndian>>();
 
 pub(crate) const GOT_ENTRY_SIZE: u64 = 0x8;
-pub(crate) const PLT_ENTRY_SIZE: u64 = PLT_ENTRY_TEMPLATE.len() as u64;
+// TODO: Right now, both x86_64 and AArch64 have 16 byte long entires, but
```
davidlattimore, line 358: One way this could be handled is by adding an `element_size_fn` that could dynamically compute the element size. This would be similar to what's currently done with `info_fn`.

Another options would be a more drastic refactoring to make the whole `SECTION_DEFINITIONS` be computed dynamically. I'm not sure what the implications of that would be though.

davidlattimore (approved):
Nice!

libwild/src/aarch64.rs
marxin: Yes, the limiting factor is immediate in the ADRP instruction, which is 21 bits, and the page size is 12 bits. Together, it's 33 bits, which make it possible to address +/- 4GiB. It's also part of the instruction docs:
https://developer.arm.com/documentation/ddi0602/2024-12/Base-Instructions/ADRP--Form-PC-relative-address-to-4KB-page-

> s the program label whose 4KB page address is to be calculated. Its offset from the page address of this instruction, in the range +/-4GB, is encoded as "immhi:immlo" times 4096.

libwild/src/aarch64.rs
marxin: A very good point. I tend to adjust the code so that we will assume the GOT entry follows PLT entry.

libwild/src/elf.rs
marxin: Thanks for the ideas! I've briefly tried the `info_fn` path and it seems one needs to add the `Arch` trait into `BuiltInSectionDetails` anyway (due to the `info_fn` member).

libwild/src/aarch64.rs
davidlattimore: Gotcha. The source of my confusion for how many bits were available to represent the page was from this code:

```rust
                mask = ((extract_bits(extracted_value, 0, 2) as u32) << 29)
                    | ((extract_bits(extracted_value, 2, 32) as u32) << 5);
```

Since there, it looks like you're extracting 2 + 30 bits, I concluded 32 bits total. But looking at the docs, is that right? It seems like that 32 in the second line should perhaps be 21?

PR #278, opened 2025-01-05
Delete a couple of files that got accidentally added

davidlattimore:


davidlattimore:
For context, jj automatically adds (tracks) files that you create. I tried turning this off, however then I had the opposite problem - I'd forget to track files that I had added. These files were added when I reenabled automatic tracking of added files, but I wasn't on the commit where the files should have been added and I didn't think through the implications.

I think going forward I'm going to keep automatic tracking on, however check for added files when uploading the PR. I've amended my pre-upload-pr script to check for new files and print them (with colour, so that they stand out). Hopefully that will be enough for me to avoid either forgetting to add files or adding files that I shouldn't.

Issue #267, opened 2025-01-02
Support testing of non-host architecture via qemu


davidlattimore:
@marxin is in the process of porting to aarch64. Not all the tests pass on ARM yet, but some do.

It'd be great if we could run tests for arm even if we're developing on x86-64. This should be possible with qemu. For some helpful (debian / ubuntu based) instructions, see this [blog post](https://azeria-labs.com/arm-on-x86-qemu-user/).

@andrewdavidmackenzie - perhaps this is something you might be interested in?

I'm not sure exactly what it might look like. One possibility is that our tests could have an instruction in them to specify which architectures the test should run on. e.g. `//#Arch:x86_64, aarch64`. We'd then need to add additional flags to the compiler (and possibly linker) to specify the architecture. Then, when running the resulting binary, if the architecture isn't the same as the host, we'd run the binary using `qemu-aarch64` or equivalent.

andrewdavidmackenzie:
I won't be back active until the 7th, but then could try and help with
this...maybe adding the target arch/triple to the tests and using it....

That could be done in separate PR and all would have x86 initially...

I don't have Linux on my Apple Silicon machines (yet!) and don't have any
arm Linux box.... But thinking of that made me wonder if qemu would help
make that work, executing x86 on arm machine!!

davidlattimore:
I was thinking more the other way around - executing arm binaries on an x86-64 Linux machine, but it could plausibly work both ways.

Rather than adding the option to all tests, one possibility would be to make the option have a default value, which initially could be "x86_64, aarch64", then override that default for the tests that are x86_64-specific (e.g. because they're written in x86 assembly), or which just don't yet pass on aarch64.

marxin:
Adding the qemu support to integration tests might be non-trivial as one needs to properly install cross GCC/Clang, glibc and I am not sure about the qemu when it comes to the dynamically linked binary.

An initial step for me would be porting most of the integration tests away from the assembly to C/Rust code that can be run by both platforms. The second step, in my opinion, can be porting the remaining assembly bits to aarch64 as well. Simil;arly to what I did https://github.com/davidlattimore/wild/blob/80e2fc59db57424a379dc4fe18ef12913e29b930/wild/tests/sources/exit.c#L17.

davidlattimore:
[...]

PR #280, opened 2025-01-05
Introduce `Arch` directive in integration tests

marxin:
Add AArch64-specific ignores.

### DIFF
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -158,6 +158,20 @@ impl Config {
             .into_iter()
             .map(ToOwned::to_owned),
         );
+
+        #[cfg(target_arch = "aarch64")]
+        {
+            self.ignore.extend(
+                [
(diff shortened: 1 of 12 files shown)


linker-diff/src/lib.rs
```diff
@@ -158,6 +158,20 @@ impl Config {
             .into_iter()
             .map(ToOwned::to_owned),
         );
+
+        #[cfg(target_arch = "aarch64")]
```
davidlattimore, line 162: It'd be good to decide this at runtime rather than at compile time. e.g. https://github.com/davidlattimore/wild/pull/268. But I'm happy to leave this as is for now in order to get this merged.

davidlattimore (approved):

linker-diff/src/lib.rs
marxin: Sure, that's definitely something that needs to be covered in #267.

PR #281, opened 2025-01-05
Add -fno-stack-protector to `weak-vars.c`

mati865:
This fixes linking this test with ld.bfd on Arch Linux.

CC https://github.com/davidlattimore/wild/issues/230

### DIFF
--- a/wild/tests/sources/weak-vars.c
+++ b/wild/tests/sources/weak-vars.c
@@ -1,7 +1,7 @@
 //#AbstractConfig:default
 //#Object:weak-vars1.c
 //#Object:exit.c
-//#CompArgs:-ffreestanding -fno-builtin
+//#CompArgs:-ffreestanding -fno-builtin -fno-stack-protector
 
 #include "exit.h"
 


davidlattimore:
Thanks! Any idea why this fixes this test? There are other tests - e.g. `weak-fns.c` where we do the oppose - i.e. turn off the stack protector. In that case, it's necessary because with the stack protector enabled, the compiler was emitting a reference to a symbol that we hadn't defined - something from the compiler builtins, or libc that reported the stack overflow and aborted. So I'm somewhat surprised that turning on the stack protector would help.

davidlattimore:
One more question... if you run the test a few times with and without this change, is it consistent that it passes with the change and fails without?

mati865:
I don't know why manually enabling stack protector is necessary here. I'd have expected it to be automatically enabled, as Arch enables it quite aggressively:
https://wiki.archlinux.org/title/Arch_package_guidelines/Security#Stack_canary
https://gitlab.archlinux.org/archlinux/packaging/packages/clang/-/blob/main/enable-fstack-protector-strong-by-default.patch?ref_type=heads


The failure is consistent. I ran tests with this change two or three times with the same result, but I can run them in a loop later if you want.

davidlattimore:
I tried out your change on Arch and can confirm that it does indeed fix it. For me at least, so does `-fno-stack-protector`. Could you perhaps change to pass that instead? To me it makes more sense as the fix. Especially if the error you're getting is `undefined reference to `__stack_chk_fail'`.

mati865:
Indeed, disabling stack protector also helps.
I thought some of the linked objects were built with stack protection enabled (or maybe the symbols were just pulled by the headers?). Enabling stack protector for the final link was quite natural for me, since that used to be the recommendation years ago when many packages were failing with error like that.

mati865:
Ah, it makes sense now. Thank you for the explanation.

> That symbol is provided by either libc or the compiler's runtime (I suspect the latter, but I can't recall which)

Actually by both. GCC ships with `libssp` that provide generic stack smashing protection, but libc can provide more robust variants. Glibc and recent versions of mingw-w64 do that, IIRC musl do that as well, dunno about others.

PR #290, opened 2025-01-06
Solve easy todos

mati865:
- Turn TODO about archive groups into a comment
Mold and MinGW backend of LLD do the same thing.
ELF backend of LLD handles them but seems just fine without them.
- Use offset_of! instead of hardcoded value
It was stabilised in Rust 1.77
image


davidlattimore:
Thanks!

PR #298, opened 2025-01-09
Slightly improve CLI compatibility with other linkers

mati865:
- Add aliases for strip and group arguments
- Ignore `--sort-common` flag 
Arch Linux uses this flag when building packages:
https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/blob/1b80cb88cf499898d10bb04964107d684f95acaf/makepkg.conf#L48

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -154,6 +154,7 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "fatal-warnings",
     "color-diagnostics",
     "undefined-version",
+    "sort-common",
 ];
 
 const IGNORED_FLAGS: &[&str] = &["gdb-index", "disable-new-dtags"];
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -143,6 +143,8 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     // so perhaps ignoring these is the right thing to do.
     "start-group",
     "end-group",
+    "(",
+    ")",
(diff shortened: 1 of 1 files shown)


libwild/src/args.rs
```diff
@@ -143,6 +143,8 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     // so perhaps ignoring these is the right thing to do.
     "start-group",
     "end-group",
+    "(",
+    ")",
```
mati865, line 147: I'm not sure about these two. I debugged at least one package years ago that used this variant, but they are so rare I'm not sure if there is a point even in having them here.

davidlattimore:
Thanks for those. Adding `-(` and `-)` seems reasonable to me. They're in the ld man page and they don't cost us much. The `--sort-common` is interesting. I didn't know that was a flag, but as it turns out the behaviour it describes is what Wild already does - so that's nice :)

mati865:
> The `--sort-common` is interesting. I didn't know that was a flag, but as it turns out the behaviour it describes is what Wild already does - so that's nice :)

Oh, I couldn't figure it out from the code, but that's nice.  It shouldn't fall under the todo section, then :wink:.

davidlattimore:
> Oh, I couldn't figure it out from the code, but that's nice. It shouldn't fall under the todo section, then 😉.

There's no explicit sort step, which I guess makes it less obvious. Instead, it buckets each input section based on its alignment. It then emits buckets in order of descending alignment.

PR #301, opened 2025-01-11
Add typos config and fix typos

mati865:
Project repo: https://github.com/crate-ci/typos

Maybe it should be integrated with CI?
https://github.com/crate-ci/typos/blob/master/docs/github-action.md


davidlattimore:
Neat. I didn't know about that project. Adding it to CI seems like a good idea too

Issue #27, opened 2024-08-13
Various TLS tests failures (from Mold test-suite)


marxin:
It seems to me the Mold linker provides a nice way how to test various TSL-related cases:

```
$ /home/marxin/Programming/mold/build
$ cp ~/Programming/wild/target/release/wild mold 
$ ctest
...
$ grep 'Failed.*tls' /home/marxin/Programming/mold/build/Testing/Temporary/LastTest.log
Error: Failed to activate out/test/elf/x86_64/x86_64_tls-gd-mcmodel-large/b.o (768 (3/0))
Error: Failed to activate out/test/elf/x86_64/x86_64_tls-gd-to-ie/a.o (768 (3/0))
Error: Failed to load symbol `main` (78 local=10) in file #1024 (4/0) (out/test/elf/x86_64/x86_64_tls-ld-mcmodel-large/a.o) (ADDRESS | CAN_BYPASS_GOT) from out/test/elf/x86_64/x86_64_tls-ld-mcmodel-large/a.o (1024 (4/0))
Error: Failed to load symbol `get_foo` (70 local=2) in file #1024 (4/0) (out/test/elf/x86_64/x86_64_tls-module-base/a.o) (ADDRESS | CAN_BYPASS_GOT) from out/test/elf/x86_64/x86_64_tls-module-base/a.o (1024 (4/0))
Error: Failed to load symbol `main` (77 local=5) in file #1280 (5/0) (out/test/elf/x86_64/x86_64_tlsdesc/b.o) (ADDRESS | CAN_BYPASS_GOT) from out/test/elf/x86_64/x86_64_tlsdesc/b.o (1280 (5/0))
Error: Failed copying from out/test/elf/x86_64/tls-common/a.o (1024 (4/0)) to output file
Error: Failed to activate out/test/elf/x86_64/tlsdesc-dlopen/a.o (768 (3/0))
Error: Failed to load symbol `main` (72 local=4) in file #1024 (4/0) (out/test/elf/x86_64/tlsdesc-import/a.o) (ADDRESS | CAN_BYPASS_GOT) from out/test/elf/x86_64/tlsdesc-import/a.o (1024 (4/0))
[...]

PR #305, opened 2025-01-14
Initial support for TLS Descriptors (x86_64)

marxin:
I am in the process of implementation the feature for `x86_64` Linux target. Even though it's non-default for `x86_64` (one needs `-mtls-dialect=gnu2`), it's the default mechanism used for `global-dynamic` and `local-dynamic` on AArch64.

I've got a working prototype, however, I struggle with the static symbols when the TLS Descriptor is combined with `R_X86_64_DTPOFF32` relocation and I would like to ask for a second pair of eyes (@davidlattimore).

Having the following example:
```c
_Thread_local long g1 = 1;
static _Thread_local long l2 = 2;
static _Thread_local long l3 = 3;
static _Thread_local long l4 = 4;
_Thread_local long g5 = 5;
_Thread_local long g6 = 6;
_Thread_local long g7 = 7;

void foo() {
  __builtin_printf("g1: %p: %ld\n", &g1, g1);
  __builtin_printf("l2: %p: %ld\n", &l2, l2);
  __builtin_printf("l3: %p: %ld\n", &l3, l3);
  __builtin_printf("l4: %p: %ld\n", &l4, l4);
  __builtin_printf("g5: %p: %ld\n", &g5, g5);
  __builtin_printf("g6: %p: %ld\n", &g6, g6);
  __builtin_printf("g7: %p: %ld\n", &g7, g7);
}

#ifdef MAIN
int main() {
  foo();
  return 0;
}
#endif
```

The scenario where the code is built as a shared library is fine:
```
✦ ❯ gcc tls.c -mtls-dialect=gnu2 -fPIC -shared -o libx.so --save-temps -B ~/Programming/wild && echo "int foo(); int main() { foo(); return 0; }" | gcc -x c - -L. -lx -B ~/Programming/wild -mtls-dialect=gnu2 && ./a.out
WARNING: wild: --plugin /usr/lib64/gcc/x86_64-suse-linux/14/liblto_plugin.so is not yet supported
WARNING: wild: --plugin /usr/lib64/gcc/x86_64-suse-linux/14/liblto_plugin.so is not yet supported
g1: 0x7fb39699cb48: 1
l2: 0x7fb39699cb50: 2
l3: 0x7fb39699cb58: 3
l4: 0x7fb39699cb60: 4
g5: 0x7fb39699cb68: 5
g6: 0x7fb39699cb70: 6
g7: 0x7fb39699cb78: 7
✦ ❯ clang tls.c -mtls-dialect=gnu2 -fPIC -shared -o libx.so --save-temps -B ~/Programming/wild && echo "int foo(); int main() { foo(); return 0; }" | clang -x c - -L. -lx -B ~/Programming/wild -mtls-dialect=gnu2 && ./a.out
WARNING: wild: -z relro is not yet supported
WARNING: wild: -z relro is not yet supported
g1: 0x7fef43cd7b48: 1
l2: 0x7fef43cd7b68: 2
l3: 0x7fef43cd7b70: 3
l4: 0x7fef43cd7b78: 4
g5: 0x7fef43cd7b50: 5
g6: 0x7fef43cd7b58: 6
g7: 0x7fef43cd7b60: 7
```

However, it starts failing for Clang if being built as an executable:

```
$ gcc tls.c -mtls-dialect=gnu2 -B ~/Programming/wild -DMAIN -fPIC --save-temps && ./a.out
WARNING: wild: --plugin /usr/lib64/gcc/x86_64-suse-linux/14/liblto_plugin.so is not yet supported
g1: 0x7efe4a230708: 1
l2: 0x7efe4a230710: 2
l3: 0x7efe4a230718: 3
l4: 0x7efe4a230720: 4
g5: 0x7efe4a230728: 5
g6: 0x7efe4a230730: 6
g7: 0x7efe4a230738: 7
$ clang tls.c -mtls-dialect=gnu2 -B ~/Programming/wild -DMAIN -fPIC --save-temps && ./a.out
WARNING: wild: -z relro is not yet supported
g1: 0x7f81e666b708: 1
l2: 0x7f81e666b6f0: 0
l3: 0x7f81e666b6f8: 0
l4: 0x7f81e666b700: 0
g5: 0x7f81e666b710: 5
g6: 0x7f81e666b718: 6
g7: 0x7f81e666b720: 7
```

The main difference here is that Clang uses `R_X86_64_DTPOFF32` for the local TLS variables and the TLSDESCR relocations are based on `_TLS_MODULE_BASE_`:

```
  57:	48 8d 05 00 00 00 00 	lea    0x0(%rip),%rax        # 5e <foo+0x5e>
  5e:	ff 10                	call   *(%rax)
  60:	48 8d b4 08 00 00 00 	lea    0x0(%rax,%rcx,1),%rsi
...
000000000000005a  0000001000000022 R_X86_64_GOTPC32_TLSDESC 0000000000000000 _TLS_MODULE_BASE_ - 4
000000000000005e  0000001000000023 R_X86_64_TLSDESC_CALL  0000000000000000 _TLS_MODULE_BASE_ + 0
0000000000000064  0000000400000015 R_X86_64_DTPOFF32      0000000000000020 l2 + 0
```

I noticed the `R_X86_64_DTPOFF32` relocation resolution is different when shlib and exe are built:
https://github.com/davidlattimore/wild/blob/b9d22b891b2f59a365000409409462e58f685f81/libwild/src/elf_writer.rs#L1888-L1895. Making various tweeks, I was unable to make all 4 scenarios working :/

Maybe one needs to specially place the artificial symbol (`_TLS_MODULE_BASE_`) as briefly mention here:
https://www.fsfla.org/~lxoliva/writeups/TLS/RFC-TLSDESC-x86.txt

> Because of the way @DTPOFF has historically been relaxed in main
> executables, when using the GNU TLS extensions, _TLS_MODULE_BASE_ is
> handled differently on final executables than on dynamic libraries.

> 
> In general, _TLS_MODULE_BASE_ is defined at the lowest address in the
> TLS segment, i.e., so that adding its NTPOFF (computed in the sequence
> above) to the linker-computed DTPOFF of a variable yields the TP
> offset of the variable.
> 
> On main executables, the DTPOFF is relaxed into NTPOFF, so we must
> stop _TLS_MODULE_BASE_'s standard NTPOFF from affecting the direct LE
> addressing.  Therefore, for main executables, the call sequence must
> be relaxed in such a way that the base address computed into %eax is
> zero.  Whether this is accomplished by setting _TLS_MODULE_BASE_ to
> the end of the TLS section, by handling its relaxation especially, or
> by any other means, is unspecified.  _TLS_MODULE_BASE_ is a special
> internal symbol, not to be relied upon for any other purposes; it
> needs not even be defined as a symbol in the linker output.
>

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -773,6 +773,9 @@ impl<'data, 'layout, 'out> TableWriter<'data, 'layout, 'out> {
         if resolution_flags.contains(ResolutionFlags::GOT_TLS_OFFSET) {
             return Ok(());
         }
+        if resolution_flags.contains(ResolutionFlags::GOT_TLS_DESCRIPTOR) {
+            return self.process_got_tls_descriptor::<A>(res, got_address);
+        }
 
         let got_entry = self.take_next_got_entry()?;
 
@@ -869,6 +872,31 @@ impl<'data, 'layout, 'out> TableWriter<'data, 'layout, 'out> {
         Ok(())
     }
 
+    fn process_got_tls_descriptor<A: Arch>(
+        &mut self,
+        res: &Resolution,
+        got_address: u64,
+    ) -> Result {
+        self.take_next_got_entry()?;
+
+        let dynamic_symbol_index = res.dynamic_symbol_index.map_or(0, std::num::NonZero::get);
+        debug_assert_bail!(
+            *compute_allocations(res, self.output_kind).get(part_id::RELA_DYN_GENERAL) > 0,
+            "Tried to write TLS descriptor with no allocation. {}",
+            ResFlagsDisplay(res)
+        );
+        let addend = if self.output_kind.is_executable() {
+            res.raw_value.sub(self.tls.start) as i64
+        } else {
+            0
+        };
+        self.write_tls_descriptor_relocation::<A>(got_address, dynamic_symbol_index, addend)?;
+
+        // TLS descriptor occupies 2 entries
+        self.take_next_got_entry()?;
+        Ok(())
+    }
+
     fn write_plt_entry<A: Arch>(&mut self, got_address: u64, plt_address: u64) -> Result {
         let plt_entry = self.take_plt_got_entry()?;
         A::write_plt_entry(plt_entry, got_address, plt_address)
@@ -953,6 +981,20 @@ impl<'data, 'layout, 'out> TableWriter<'data, 'layout, 'out> {
             dynamic_symbol_index,
             A::get_dynamic_relocation_type(DynamicRelocationKind::DtpMod),
             0,
+        )
+    }
+
+    fn write_tls_descriptor_relocation<A: Arch>(
+        &mut self,
+        place: u64,
+        dynamic_symbol_index: u32,
+        addend: i64,
+    ) -> Result {
+        self.write_rela_dyn_general(
+            place,
+            dynamic_symbol_index,
+            A::get_dynamic_relocation_type(DynamicRelocationKind::TlsDesc),
+            addend,
         )
     }
 
@@ -1906,6 +1948,11 @@ fn apply_relocation<S: StorageModel, A: Arch>(
             .value()
             .wrapping_sub(layout.tls_start_address_aarch64())
             .wrapping_add(addend),
+        RelocationKind::TlsDesc => resolution
+            .tls_descriptor_got_address()?
+            .bitand(mask.got_entry)
+            .wrapping_add(addend)
+            .wrapping_sub(place.bitand(mask.place)),
         RelocationKind::None => 0,
     };
     rel_info
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -2542,21 +2542,16 @@ fn write_internal_symbols<S: StorageModel>(
             shndx = 1;
         }
 
-        let mut address = resolution.address()?;
-        let st_type = if symbol_name.bytes() == TLS_MODULE_BASE_SYMBOL_NAME.as_bytes() {
-            // TODO: handle properly the symbol address
-            if layout.args().output_kind != OutputKind::SharedObject {
-                address = layout.tls_end_address();
-            }
-            object::elf::STT_TLS
-        } else {
-            object::elf::STT_NOTYPE
-        };
-
+        let address = resolution.address()?;
         let entry = symbol_writer
             .define_symbol(false, shndx, address, 0, symbol_name.bytes())
             .with_context(|| format!("Failed to write {}", layout.symbol_debug(symbol_id)))?;
 
+        let st_type = if symbol_name.bytes() == TLS_MODULE_BASE_SYMBOL_NAME.as_bytes() {
+            object::elf::STT_TLS
+        } else {
+            object::elf::STT_NOTYPE
+        };
         entry.set_st_info(object::elf::STB_GLOBAL, st_type);
     }
     Ok(())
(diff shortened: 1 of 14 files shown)


davidlattimore:
I've read through the linked RFC, but I've yet to properly look through your change. I'll try to have a look through it tomorrow in the hope of coming up with some suggestion(s). In the meantime, all I can suggest is looking through the output of the same program when linked with GNU ld or lld and see if you can make sense of what those linkers have done with the relocations.

davidlattimore:
I tried switching `start_symbol_name... TLS_MODULE_BASE_SYMBOL_NAME` to `end_symbol_name... TLS_MODULE_BASE_SYMBOL_NAME` and that indeed appeared to result in the correct output from the last test case. Presumably that breaks the shared object output case. So I guess the selection of where to put the symbol needs to be more dynamic.

Another thing I noticed, is that GNU ld appears to emit the symbol with a value that looks like a TLS offset, whereas wild emits it as what looks like a virtual address. e.g.:

```sh
# readelf -W --all a.out.wild  | grep TLS_MODULE_BASE
    19: 0000000000003698     0 TLS     GLOBAL DEFAULT   26 _TLS_MODULE_BASE_
# readelf -W --all a.out.gnu  | grep TLS_MODULE_BASE
    22: 0000000000000038     0 TLS     LOCAL  DEFAULT   21 _TLS_MODULE_BASE_
```

Note the much larger value for the wild output. I'm not sure if this matters, since as I understand it, the symbol doesn't even need to be put in the symbol table.

davidlattimore:
I suspect long-term, we might want to refactor the way the section descriptors work, but this is fine for now. I don't have any great depths of knowledge into how TLS descriptors are supposed to work, but what you've done seems consistent with the RFC. So I'm happy for this to be merged. We can always adjust later if we find something isn't quite right.

marxin:
The PR makes 6 more mold tests green :)

marxin:
Once we merge this, #27 can be basically closed.

libwild/src/elf_writer.rs
```diff
@@ -869,6 +872,36 @@ impl<'data, 'layout, 'out> TableWriter<'data, 'layout, 'out> {
...
+        );
+        let addend = if res.dynamic_symbol_index.is_none() {
+            res.raw_value.sub(self.tls.start) as i64
+        } else {
+            0
+        };
+        self.write_tls_descriptor_relocation::<A>(got_address, dynamic_symbol_index, addend)?;
+
+        // TLS descriptor occupies 2 entries
+        self.take_next_got_entry()?;
```
davidlattimore: Do the two calls to `take_next_got_entry` need to be separated? Just wondering if it might make sense to put them together

libwild/src/layout.rs
```diff
@@ -3908,6 +3923,24 @@ impl Resolution {
...
     }
 
+    pub(crate) fn tls_descriptor_got_address(&self) -> Result<u64> {
+        debug_assert_bail!(
+            self.resolution_flags
+                .contains(ResolutionFlags::GOT_TLS_DESCRIPTOR),
+            "Called tls_descriptor_got_address without GOT_TLS_DESCRIPTOR being set"
+        );
+        // We might have both GOT_TLS_OFFSET and GOT_TLS_DESCRIPTOR at the same time
+        // for a single symbol. Then the TLS descriptor comes later and my reflect that in the GOT address.
```
davidlattimore: I think perhaps you meant some word other than "my"

davidlattimore (approved):
This is great! Thanks for doing this!

PR #310, opened 2025-01-19
Introduce openSUSE in CI as well

marxin:


marxin:
One of the motivations is that **Ubuntu** does not support (for unknown reason) Clang with TLS Descriptors: #305.

marxin:
An update on the TLSDESC support in Clang. It has been added in Clang 19 (for x86_64) and that's why it's unsupported on ubuntu:latest (`24.04` with Clang 18), while `ubuntu:rolling` (with Clang 19) supports the feature :)

PR #313, opened 2025-01-20
Refactor argument parsing to directly populate Args struct

davidlattimore:
This reduces the number of places that need changing when parsing a new argument from 4 to 3. Once `default_field_values` is stable, we can further reduce this to 2.

Note, there is a subtle behaviour change, which is that we no longer support running the linker on architectures that we don't target.


marxin (approved):
Neat refactoring!

Issue #20, opened 2024-07-12
Missing "undefined reference to" error reporting


marxin:
Please add error reports:

```
❯ echo "int foo(); int main() { return foo(); }" | gcc -x c++ - -B ~/Programming/wild/
✦ ❯ ./a.out 
./a.out: symbol lookup error: ./a.out: undefined symbol: _Z3foov
✦ ❯ echo "int foo(); int main() { return foo(); }" | gcc -x c++ -
/usr/lib64/gcc/x86_64-suse-linux/13/../../../../x86_64-suse-linux/bin/ld: /tmp/ccuujEtS.o: in function `main':
<stdin>:(.text+0x5): undefined reference to `foo()'
collect2: error: ld returned 1 exit status
```

davidlattimore:
I've had undefined symbols reports errors at various points in the past. I think this time when I fix it, it might be time to add a test :)

andrewdavidmackenzie:
Just reproducing this to make sure I understand it.

I deleted our config.toml file so it's using the default linker, or so I thought. Not sure how (yet) but it seemed to be using wild.

I don't get any compile time error, but the runtime error, the same as you.

I deleted the wild binaries, and now I get the build time error:
```
echo "int foo(); int main() { return foo(); }" | gcc -x c++ - -B ~/workspace/wild/ > ./a.out
/usr/bin/ld: /tmp/ccTCOIzO.o: in function `main':
<stdin>:(.text+0x9): undefined reference to `foo()'
collect2: error: ld returned 1 exit status
```

davidlattimore:
`.cargo/config.toml` only affects stuff built with Cargo, but this is C code. The `-B ~/workspace/wild/` tells gcc to look in that directory for tools like the linker. If you need to check which linker produced a binary, you can run `readelf -p .comment a.out`. If the binary was produced with any linker other than GNU ld, there will be a comment saying the name and version of the linker. GNU ld unfortunately doesn't add any such comment.

davidlattimore:
The bug is that wild no longer reports an error when there's a non-weak reference to an undefined symbol. It used to report errors for this, but some at some point, I must have accidentally lost the code to check for it.

References to undefined symbols are OK if the symbol reference is weak. However if it's not weak, then that should be an error.

[...]

PR #314, opened 2025-01-21
Implement errors about undefined symbols

mati865:
Fixes: #20

There are two regressions right now:
```
Error: Undefined symbols found!
  Undefined symbol does_not_exist, referenced by /home/mateusz/projects/wild/wild/tests/build/archive_activation.default-f0cd5ddfa601dc84.o

Error: Undefined symbols found!
  Undefined symbol __start_foo, referenced by /home/mateusz/projects/wild/wild/tests/build/custom_section.archive-95f9e368cdd19b2a.o
  Undefined symbol __stop_foo, referenced by /home/mateusz/projects/wild/wild/tests/build/custom_section.archive-95f9e368cdd19b2a.o
  Undefined symbol __start_bar, referenced by /home/mateusz/projects/wild/wild/tests/build/custom_section.archive-95f9e368cdd19b2a.o
  Undefined symbol __stop_bar, referenced by /home/mateusz/projects/wild/wild/tests/build/custom_section.archive-95f9e368cdd19b2a.o

test integration_test::program_name_19___archive_activation_c__ ... FAILED
test integration_test::program_name_15___custom_section_c__ ... FAILED
```

I don't know about the first one yet, but the second one looks like this function should take into account start/stop symbols from `canonicalise_undefined_symbols`. I'll look into it when I get more time to understand what I'm doing 😅

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3,6 +3,7 @@
 //! where in the output file then allocates addresses for each symbol.
 
 use self::elf::NoteHeader;
+use self::elf::Symbol;
 use self::elf::GNU_NOTE_NAME;
 use self::elf::GNU_NOTE_PROPERTY_ENTRY_SIZE;
 use self::output_section_id::InfoInputs;
@@ -2452,6 +2453,20 @@ fn process_relocation<S: StorageModel, A: Arch>(
 
         if previous_flags.is_empty() {
             queue.send_symbol_request(symbol_id, resources);
+            if is_symbol_undefined(
+                object.object.symbol(local_sym_index)?,
+                object.file_id,
+                symbol_db.file_id_for_symbol(symbol_id),
+                symbol_value_flags,
+                args,
+            ) {
+                let symbol_name = symbol_db.symbol_name(symbol_id)?;
+                resources.report_error(anyhow::anyhow!(
+                    "Undefined symbol {}, referenced by {}",
+                    String::from_utf8_lossy(symbol_name.bytes()),
+                    object.input
+                ));
+            }
         }
 
         if resolution_kind.contains(ResolutionFlags::COPY_RELOCATION)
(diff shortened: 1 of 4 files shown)


davidlattimore:
Thanks for working on this! I think both of the problems you're hitting will likely be fixed by changing where you're checking for undefined symbols. I'd suggest looking at putting the logic somewhere near the call to `send_symbol_request` in `process_relocation`. Putting the logic there will mean that you'll only be reporting errors for undefined symbols if the undefined symbol is referenced from a live object and in the case of `--gc-sections` (default in Wild), from reachable code.

libwild/src/layout.rs
```diff
@@ -2874,6 +2889,22 @@ fn create_start_end_symbol_resolution<S: StorageModel>(
...
+    sym_file_id: FileId,
+    sym_def_file_id: FileId,
+    symbol_value_flags: ValueFlags,
+    args: &Args,
+) -> bool {
+    if (args.output_kind() == OutputKind::SharedObject && !args.no_undefined) || symbol.is_weak() {
+        return false;
+    }
+
+    sym_file_id == sym_def_file_id
```
mati865, line 2903: The need for skipping symbols defined in objects other than the one that references it, was found by test `program_name_04___global_definitions_c__`, not `program_name_35___undefined_symbols_c__`.
Ideally, `program_name_35___undefined_symbols_c__` would verify that too, but I couldn't manage to reproduce it there without duplicating most of the code from the other test.

mati865:
Rebased to the latest main, reworked the check, and (hopefully) improved the test a bit.

Previously I had missed `report_error`, I think that's pretty neat, except for printing only one error, which is far from ideal for printing undefined references.
I'll address that issue in a followup.

davidlattimore:
Nice! Thanks for doing this!

PR #320, opened 2025-01-23
Fix handling of .ctors section and add test for it

davidlattimore:
We already handled .ctors.*, but not plain .ctors.


davidlattimore:
This was in response to https://github.com/rust-lang/rust/issues/128286

Issue #321, opened 2025-01-23
Undefined symbol __tls_get_addr, referenced by /usr/lib64/gcc/x86_64-suse-linux/14/libstdc++.a @ eh_globals.o


marxin:
Started after #314:

```
✦ ❯ cat t.cc
int main() {
  try {
    throw 0;
  } catch (int x) {
    return x;
  }
  return 1;
}
✦ ❯ g++ t.cc -B ~/Programming/wild -fPIC -static
WARNING: wild: --plugin /usr/lib64/gcc/x86_64-suse-linux/14/liblto_plugin.so is not yet supported
Error: Undefined symbol __tls_get_addr, referenced by /usr/lib64/gcc/x86_64-suse-linux/14/libstdc++.a @ eh_globals.o
collect2: error: ld returned 255 exit status
```

@mati865: Can you please take a look?

mati865:
@marxin sure, I can do that later today. Feel free to revert the change for now.

davidlattimore:
Some additional info that might be useful...

The problem is specific to static  linking. If you remove the `-static` argument, it links OK. Doing that, you can save the linker invocation:

```
WILD_SAVE_DIR=t-non-static g++ t.cc -B ~/Programming/wild -fPIC
```

You can then get information about the symbol with the `--sym-info` flag:

```sh
cd t-non-static
./run-with wild --sym-info __tls_get_addr

....
  Global Func: symbol_id=12709 -> 12709 DYNAMIC | FUNCTION 
    #18 in File #11776 (46/0) /lib64/ld-linux-x86-64.so.2
```

So the function is provided by `ld-linux-x86-64.so.2`, which is only available when dynamically linking.

The reason that the function isn't actually needed when statically linking is because all the references to it get optimised away by the linker when the linker processes the relocation prior to the reference to that function.

For example, in the following `objdump -drw` output, there's a `TLSGD` relocation following by a `PLT32` relocation that calls `__tls_get_addr`.

```
   7:   66 48 8d 3d 00 00 00 00         data16 lea 0x0(%rip),%rdi        # f <__dfp_set_round+0xf>      b: R_X86_64_TLSGD
        __bid_IDEC_glbround-0x4
   f:   66 66 48 e8 00 00 00 00         data16 data16 rex.W call 17 <__dfp_set_round+0x17>      13: R_X86_64_PLT32      __tls_get_addr-0x4
```

[...]

PR #322, opened 2025-01-23
account for relaxations when processing relocations

mati865:
Fixes https://github.com/davidlattimore/wild/issues/321

Also created an alternative version, but it didn't work out the way I had hoped and doesn't seem any better: https://github.com/mati865/wild/commit/366175cf6e08b7756996ccb47eb26cc9b1ce1ec7
Diff between these two versions is available at: https://github.com/mati865/wild/compare/8ea229ab67057dae5042a8f75d1ef5eeade6e0e9..4c0d05e2ff1befc253f02c5c85c646f792dba457

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -1,3 +1,4 @@
+use crate::arch::RelocationModifier;
 use crate::elf::extract_bits;
 use crate::elf::BitRange;
 use crate::elf::DynamicRelocationKind;
@@ -13,7 +14,6 @@ use anyhow::bail;
 use anyhow::Result;
 use linker_utils::elf::aarch64_rel_type_to_string;
 use linker_utils::elf::RelocationKind;
-use linker_utils::relaxation::RelocationModifier;
 
 pub(crate) struct AArch64;
 
@@ -875,14 +875,7 @@ impl crate::arch::Relaxation for () {
     }
 
     #[allow(unused_variables)]
-    fn apply(
-        &self,
-        section_bytes: &mut [u8],
-        offset_in_section: &mut u64,
-        addend: &mut u64,
-        next_modifier: &mut RelocationModifier,
-    ) {
-    }
+    fn apply(&self, section_bytes: &mut [u8], offset_in_section: &mut u64, addend: &mut u64) {}
 
     fn rel_info(&self) -> crate::elf::RelocationKindInfo {
         RelocationKindInfo {
@@ -894,6 +887,10 @@ impl crate::arch::Relaxation for () {
 
     fn debug_kind(&self) -> impl std::fmt::Debug {
         todo!()
+    }
+
+    fn next_modifier(&self) -> RelocationModifier {
+        RelocationModifier::Normal
     }
 }
 
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -7,6 +7,7 @@ use self::elf::TLS_MODULE_BASE_SYMBOL_NAME;
 use crate::alignment;
 use crate::arch::Arch;
 use crate::arch::Relaxation as _;
+use crate::arch::RelocationModifier;
 use crate::args::Args;
 use crate::args::BuildIdOption;
 use crate::args::FileWriteMode;
@@ -73,7 +74,6 @@ use linker_utils::elf::shf;
 use linker_utils::elf::sht;
 use linker_utils::elf::RelocationKind;
 use linker_utils::elf::SectionFlags;
-use linker_utils::relaxation::RelocationModifier;
 use memmap2::MmapOptions;
 use object::elf::NT_GNU_BUILD_ID;
 use object::elf::NT_GNU_PROPERTY_TYPE_0;
@@ -1829,7 +1829,8 @@ fn apply_relocation<S: StorageModel, A: Arch>(
     ) {
         tracing::trace!(kind = ?relaxation.debug_kind(), %value_flags, %resolution_flags);
         rel_info = relaxation.rel_info();
-        relaxation.apply(out, &mut offset_in_section, &mut addend, &mut next_modifier);
+        relaxation.apply(out, &mut offset_in_section, &mut addend);
+        next_modifier = relaxation.next_modifier();
     } else {
         tracing::trace!(%value_flags, %resolution_flags);
         rel_info = A::relocation_from_raw(r_type)?;
--- a/libwild/src/arch.rs
+++ b/libwild/src/arch.rs
@@ -7,7 +7,6 @@ use crate::error::Result;
 use crate::resolution::ValueFlags;
 use anyhow::bail;
 use linker_utils::elf::SectionFlags;
-use linker_utils::relaxation::RelocationModifier;
 use std::borrow::Cow;
 use std::str::FromStr;
 
@@ -61,15 +60,17 @@ pub(crate) trait Relaxation {
     where
         Self: std::marker::Sized;
 
-    fn apply(
-        &self,
-        section_bytes: &mut [u8],
-        offset_in_section: &mut u64,
-        addend: &mut u64,
-        next_modifier: &mut RelocationModifier,
-    );
+    fn apply(&self, section_bytes: &mut [u8], offset_in_section: &mut u64, addend: &mut u64);
 
     fn rel_info(&self) -> crate::elf::RelocationKindInfo;
 
     fn debug_kind(&self) -> impl std::fmt::Debug;
+
+    fn next_modifier(&self) -> RelocationModifier;
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RelocationModifier {
+    Normal,
+    SkipNextRelocation,
 }
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -11,6 +11,7 @@ use crate::alignment;
 use crate::alignment::Alignment;
 use crate::arch::Arch;
 use crate::arch::Relaxation as _;
+use crate::arch::RelocationModifier;
 use crate::args::Args;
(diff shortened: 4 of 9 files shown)


wild/a.out
mati865, line 1: Hmm, this file keeps returning when running the tests. I'll remove it from the commit later.

libwild/src/aarch64.rs
```diff
@@ -895,6 +888,10 @@ impl crate::arch::Relaxation for () {
     fn debug_kind(&self) -> impl std::fmt::Debug {
         todo!()
     }
+
+    fn next_modifier(&self) -> RelocationModifier {
+        todo!()
```
davidlattimore: It should be fine to just return `RelocationModifier::Normal` here, since we don't yet implement any relaxations for aarch64

wild/a.out
mati865, line 1: Hmm, this file keeps returning after running tests

wild/tests/sources/undefined_symbols.c
```diff
@@ -20,4 +20,5 @@ __attribute__((weak)) int undefined_weak();
 void _start(void) {
     undefined_weak();
     undefined_strong();
+
```
mati865: Gonna remove the newline later

libwild/src/elf_writer.rs
```diff
@@ -1803,6 +1803,8 @@ fn apply_relocation<S: StorageModel, A: Arch>(
         .symbol(e, false)
         .context("Unsupported absolute relocation")?;
     let local_symbol_id = object_layout.symbol_id_range.input_to_id(symbol_index);
+    // FIXME(Open point): We could bail here instead of duplicating next_modifier logic,
+    // but I worry that would negatively affect debugability when resolution fails due to a bug.
```
mati865: I'm torn here, I'm not happy with duplicating the logic of `next_modifier` but reducing traceability could be even worse.

wild/tests/sources/undefined_symbols.cc
```diff
@@ -0,0 +1,16 @@
+
+//#AbstractConfig:default
+//#DiffEnabled:false
+//#RunEnabled:false
+
+//#Config:executable:default
+//#LinkArgs:--cc=g++ -fPIC -static
```
davidlattimore: Tests that link against libc are moderately slow, so we do need to be a little bit conservative with adding new ones. An alternative to adding this would be to adjust one of the existing tests to trigger the bug. One way to do this would be to add the flag `-fPIC` to `libc-integration.c` in the config `clang-static-pie`.

wild/a.out
davidlattimore: a.out is the default output filename if no output filename is specified. Perhaps see if you can figure out which test is writing it? I just tried running the tests myself and I didn't get the file, so perhaps it's related to the new test, or perhaps some difference in your environment.

libwild/src/elf_writer.rs
davidlattimore: I'm not 100% sure that I understand. The caller of `apply_relocation` (`apply_relocations`) will skip calling `apply_relocation` if the previous call returned `SkipNextRelocation`, so we won't get here for relocations that need to be skipped.

Or perhaps what you're concerned about is that both layout and writing now make use of `RelocationModifier` to skip relocations? If that's the case, then there's plenty of other things that layout and writing do. For example, they both compute the relaxation. Layout, address allocation and writing are three phases that are pretty tightly coupled and need to be kept in sync. A small amount of duplication of work in these phases is to be expected - often it's faster to redo a bit of work than to just do it once, which would require that we store and retrieve the result of doing it the first time.

wild/a.out
davidlattimore: I assume you're also using jj. I've had issues with accidentally adding files. I tried turning off auto-track, but decided I didn't like that, so turned it back on. I now have a pre-push check that shows me any files that I've added in the code that I'm about to push, so I can look at that and confirm that I really wanted to add them. That seems to be working reasonably well. The following is the script that does that check in case you're interested:

```sh
#!/bin/bash
set -e

RED="\u001b[31m"
GREEN="\u001b[32m"
RESET="\u001b[0m"

if jj diff --from main --to @ --summary | grep ^A >/dev/null; then
    echo -e "${RED}The following files were added:${GREEN}"
    jj diff --from main --to @ --summary | grep ^A
    echo -e "${RESET}"
fi
```

davidlattimore: Actually, thinking about where this file is likely coming from, my guess would be from running the repro command-line on issue #321. It doesn't specify an output file, so would be creating `a.out`.

libwild/src/elf_writer.rs
mati865: Sorry if I wasn't clear enough.

> Or perhaps what you're concerned about is that both layout and writing now make use of RelocationModifier to skip relocations?

Yeah, that was my concern.
Although not from the performance point, because I trust the compiler will do a good job optimizing it. Only from a code maintainability perspective, I'm not a fan of having to couple distinct parts by duplicating the logic there. But if you say it's fine, then my concern is resolved.

wild/a.out
mati865: Actually it's `program_name_14___tlsdesc_c__`:
```
❯ ls wild/ && cargo t program_name_14___tlsdesc_c__ 2&> /dev/null && ls wild/
Cargo.toml  src/  tests/
Cargo.toml  a.out*  src/  tests/
```

mati865: Regarding jj, I also both like and dislike auto-track. I typically inspect the diff before pushing, but binary files aren't very visible in the diff, and I tend to miss them.

Thanks for the script, seems to work well in this case.

davidlattimore: I couldn't reproduce in my usual dev environment (Ubuntu-based), but was able to reproduce in the openSUSE docker image. I'm now looking into where it's coming from.

mati865: No worries, it's nothing serious.

davidlattimore: Should be fixed now: https://github.com/davidlattimore/wild/pull/323

mati865: Thanks, I'll continue later today (European time). Feel free to take over if you want to fix it sooner.

wild/tests/sources/undefined_symbols.cc
mati865: It works but seems kinda fragile. I doubt I'd be able to come up with something better, though.

davidlattimore:
This change looks reasonable to me. Was there more you wanted to do with it? If not, I'm happy to merge it (once the `wip:` prefix is removed)

mati865:
I want to move the check this fix works to `undefined_symbols.c` and I think it sort of worked out for me.
Also, reworded commit and PR's title. Hopefully it makes more sense now.

davidlattimore:
> I want to move the check this fix works to `undefined_symbols.c` and I think it sort of worked out for me. Also, reworded commit and PR's title. Hopefully it makes more sense now.

I'm a bit unsure about mixing doesn't-compile and does-compile tests in the one test file. That said, if you're sure that want to go that path, we can stick with that for now. If you do stick with it, then I think a few changes might be needed. The `DiffEnabled` and `RunEnabled` are set in the abstract config, but presumably only need apply to the doesn't-compile parts of the test. Possibly we should change `integration_test.rs` so that `ExpectError` sets both `DiffEnabled` and `RunEnabled` to false.

Also, the test currently fails on my system. I'm running on an ubuntu-22.04 based system, which isn't covered by CI currently. Not sure if that's the reason for the difference. I was able to make it pass with the following change:

```patch
diff --git a/wild/tests/sources/undefined_symbols.c b/wild/tests/sources/undefined_symbols.c
index b49fca4185..f11c556083 100644
--- a/wild/tests/sources/undefined_symbols.c
+++ b/wild/tests/sources/undefined_symbols.c
@@ -4,10 +4,10 @@
 //#RunEnabled:false
 
 //#Config:shared-lib:default
-//#LinkArgs:--shared
+//#LinkArgs:--cc=gcc --shared
 
 //#Config:no-undefined:default
-//#LinkArgs:--shared --no-undefined -pie
+//#LinkArgs:--cc=gcc --shared -Wl,--no-undefined -pie
 //#ExpectError:undefined_strong
 
 //#Config:executable:default
@@ -15,7 +15,7 @@
 
 //#Config:tls:default
 //#CompArgs:-fPIC -DSHOULD_PASS
-//#LinkArgs:-pie
+//#LinkArgs:--cc=gcc -pie
 
 int undefined_strong();
 __attribute__((weak)) int undefined_weak();
```

As it was before, it was directly invoking the linker, which is generally something we only do for low-level tests that don't use libc (and don't have a `main` function). If using libc, it's generally necessary to use the C compiler to invoke the linker, since the C compiler adds various arguments that are necessary in order to use the C runtime.

davidlattimore:
I'm also happy to just merge this as-is and fix the breakage that occurs on my machine after. Let me know if you'd like me to do this.

mati865:
Okay, I see your point. I'm used to having positive and negative scenarios in a single file (albeit with gherkin/cucumber style) or folder. My biggest concert with having this test as a part of libc integration is the potential to accidentally make it no longer reproduce the original problem.
That said, I don't mind reverting to the previous version.

> Also, the test currently fails on my system. I'm running on an ubuntu-22.04 based system, which isn't covered by CI currently. Not sure if that's the reason for the difference

Interesting, kinds defeats my intention here.

> As it was before, it was directly invoking the linker, which is generally something we only do for low-level tests that don't use libc (and don't have a main function). If using libc, it's generally necessary to use the C compiler to invoke the linker, since the C compiler adds various arguments that are necessary in order to use the C runtime.

Oof, I though I had mostly avoided involving libc there and only its startup code was used (maybe I underestimated its "weight"). That's one point for going back to `libc-integration` test.

> I'm also happy to just merge this as-is and fix the breakage that occurs on my machine after. Let me know if you'd like me to do this.

Let's go back. Eventual improvements can be made later on.

PR #337, opened 2025-01-26
Warn about `--sysroot` option

marxin:


mati865:
AFAIK wild doesn't have anything akin to `--with-sysroot` so the warning doesn't seem relevant. When `--with-sysroot` gets implemented, ideally `--sysroot` should be imported at the same time. The warning would make sense only when `--with-sysroot` is implemented without implementing `--sysroot` at the same time.

OFC it's just a warning, so it's rather a nitpick than an actual problem.

davidlattimore:
If we silently ignored --sysroot, then someone passed `--sysroot /foo -L$SYSROOT/bar` there would be no indication that we weren't doing what they probably expected us to do, so I'm inclined to keep the warning at least for now.

mati865:
Sorry for bumping old PR, but I stand corrected.
On a fresh Arch Linux install the cross-compilation tests fail for me with: `Error: Failed to open input file `/lib/ld-linux-aarch64.so.1`.

Creating symlink with `ln -s /usr/aarch64-linux-gnu/lib/ld-linux-aarch64.so.1 /lib/ld-linux-aarch64.so.1` workarounds the problem, but indeed `--sysroot` is required to make it fully work.

PR #340, opened 2025-01-26
CI: Run complete tests on x86_64

marxin:


.github/workflows/ci.yml
```diff
@@ -37,9 +37,9 @@ jobs:
         if: ${{ runner.arch == 'X64' }}
       - run: echo "ARCH=aarch64" >> $GITHUB_ENV
         if: ${{ runner.arch == 'ARM64' }}
-      - run: apt-get update && apt-get -y install gcc g++ clang lld curl bubblewrap
+      - run: apt-get update && apt-get -y install gcc g++ clang lld mold binutils-gold curl bubblewrap
```
davidlattimore: Does this change anything? I didn't think we had any integration tests that enabled mold or gold - i.e., we've only got `//#EnableLinker:lld`. Did you see failures due to them being missing when you made the change below? I just tried running the integration tests with mold not available on the path and they passed, so I wouldn't expect they'd be needed.

marxin: Oh, correct, I didn't know one has to explicitly enable a linker with `//#EnableLinker`. So I take the change back.

PR #347, opened 2025-01-27
Rewrite asm_diff.rs

davidlattimore:
Details about how the new implementation works are in comments in the code. There's still lots more to be done, but the new approach should be a much more solid foundation on which to build.

The new implementation relies on having a .layout file.

The new implmentation doesn't use a disassembler for diffing, only for reporting problems. This should make porting to aarch64 much more practical, since we no longer need to extract information from the disassembled instructions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -284,6 +284,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3fd119d74b830634cea2a0f58bbd0d54540518a14397557951e79340abc28c0"
 
 [[package]]
+name = "colored"
+version = "3.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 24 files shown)


davidlattimore:
Some of the things that are not yet done, but which can be in subsequent PRs:

* Diffing of non executable sections and their relocations
* Optimising runtime. It's pretty slow running on clang. Fortunately there's some low-hanging fruit.
* Verifying references to merged strings.
* Unification of sections based on references to those sections from code.
* Reporting of the percentage of section bytes that were diffed, probably broken down by section type.
* Porting to aarch64. Most arch-specific stuff should now be behind a trait - although I probably missed something.
* Improving some diagnostics when things go wrong.

Here's a sample of what the reported diffs look like now (if I comment out the code in Wild that optimises away PLT32 relocations).

image

I still need to fix a thing or two before I can merge this, but hopefully will get those done tomorrow.

PR #356, opened 2025-01-30
wip: jmp relaxation

mati865:
An attempt to reduce the number of failures in https://github.com/davidlattimore/wild/issues/230

Currently fails with errors like:
```
rel.match_failed.R_X86_64_GOTPCRELX
  `/usr/lib/gcc/x86_64-pc-linux-gnu/14.2.1/../../../../lib/libc.a` @ `setlocale.o` .text new_composite_name
  ORIG 0x00219: [ 41 5f             ] pop %r15
  ORIG 0x0021b: [ 5d                ] pop %rbp
  ORIG 0x0021c: [ ff 25 00 00 00 00 ] jmpq *0x222
                        ^^^^^^^^^^^ R_X86_64_GOTPCRELX
                                    __memcpy_chk -4
  wild 0x44119: [ 41 5f             ] pop %r15
  wild 0x4411b: [ 5d                ] pop %rbp
  wild 0x4411c: [ e9 00 f8 07 00    ] jmp 0x00000000000C3921
                        ^^^^^^^^^^^ R_X86_64_32 MovIndirectToAbsolute Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 CallIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 MovIndirectToLea Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_GOTPCRELX NoOp Prior bytes didn't match

  wild TRACE: message=relaxation applied kind=JmpIndirectToRelative
              value_flags=ADDRESS | CAN_BYPASS_GOT resolution_flags=DIRECT
              rel_info.kind=Relative symbol_name=__memcpy_chk
  wild 0x44121: [ 90                ] nop
  ld   0x58099: [ 41 5f             ] pop %r15
  ld   0x5809b: [ 5d                ] pop %rbp
  ld   0x5809c: [ e9 5f e4 fe ff    ] jmp 0x0000000000046500
                        ^^^^^^^^^^^ R_X86_64_32 MovIndirectToAbsolute Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 CallIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 MovIndirectToLea Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_GOTPCRELX NoOp Prior bytes didn't match

  ld   0x580a1: [ 90                ] nop
  lld  0x731d9: [ 41 5f             ] pop %r15
  lld  0x731db: [ 5d                ] pop %rbp
  lld  0x731dc: [ e9 5f 6d 02 00    ] jmp 0x0000000000099F40
                        ^^^^^^^^^^^ R_X86_64_32 MovIndirectToAbsolute Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 CallIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 MovIndirectToLea Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative Relaxation output didn't match
                        ^^^^^^^^^^^ R_X86_64_GOTPCRELX NoOp Prior bytes didn't match

  lld  0x731e1: [ 90                ] nop


Error: Validation failed.
Binary `/home/mateusz/projects/wild/wild/tests/build/libc-integration.c-gcc-static-pie.wild`.
```
Passed the tests before linker-diff rewrite, but I have little trust in this assembly. Maybe some changes to linker-diff are needed?


davidlattimore:
Great work! Yes, you'll likely need a small change to linker-diff. Linker-diff now needs to know which relocations are possible for a given input relocation. The place to update is [possible_relaxations_do](https://github.com/davidlattimore/wild/blob/main/linker-diff/src/x86_64.rs#L62).

I also noticed that you're adjusting the addend by 1. You might find that you need to not do that after one of the other changes I made yesterday - https://github.com/davidlattimore/wild/pull/355

davidlattimore:
Oh, wait. I see that you already made the required change to linker-diff. I guess maybe it's just the addend.

linker-diff/src/x86_64.rs
```diff
@@ -36,6 +36,7 @@ impl Arch for X86_64 {
             Self::RelaxationKind::RexSubIndirectToAbsolute => RelaxationMask::new(3, &[0xff; 3]),
             Self::RelaxationKind::RexCmpIndirectToAbsolute => RelaxationMask::new(3, &[0xff; 3]),
             Self::RelaxationKind::CallIndirectToRelative => RelaxationMask::new(2, &[0xff; 2]),
+            Self::RelaxationKind::JmpIndirectToRelative => RelaxationMask::new(2, &[0xff; 6]),
```
davidlattimore: Actually, the adjustment to the addend wouldn't explain why you're getting `JmpIndirectToRelative Relaxation output didn't match`. Rather, I think this line is the issue. You're comparing 6  bytes to see if the relaxation matched, however that would cover the operand. I think perhaps you want to compare the first byte and the 6th byte, but not the bytes in between. i.e. `&[0xff, 0, 0, 0, 0, 0xff]`.

mati865: Oh, right. I have read the comparison code, yet somehow it didn't occur to me it wouldn't handle offsets difference.

With that fix applied, linker-diff also very neatly confirms the addend change is no longer required:
```
rel.R_X86_64_PC32.R_X86_64_PC32
  `/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/14.2.1/../../../../lib64/libc.a` @ `dl-tls.o` .text _dl_deallocate_tls
  ORIG 0x007f7: [ ff 25 00 00 00 00 ] jmpq *0x7FD
                        ^^^^^^^^^^^ R_X86_64_GOTPCRELX
                                    free -4
  wild 0x94bf7: [ e9 15 49 fc ff    ] jmp 0x0000000000059511
                     ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative
                                 free
  wild TRACE: message=relaxation applied kind=JmpIndirectToRelative
              value_flags=ADDRESS | CAN_BYPASS_GOT resolution_flags=DIRECT
              rel_info.kind=Relative symbol_name=free
  wild 0x94bfc: [ 90                ] nop
  ld   0x27357: [ e9 34 e2 fe ff    ] jmp 0x0000000000015590
                     ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative
                                 free -1
  ld   0x2735c: [ 90                ] nop
  lld  0xb6997: [ e9 b4 a9 fa ff    ] jmp 0x0000000000061350
                     ^^^^^^^^^^^ R_X86_64_PC32 JmpIndirectToRelative
                                 free -1
  lld  0xb699c: [ 90                ] nop
```

mati865:
Also, great work on the linker-diff.
Truly a marvelous achievement.

Issue #230, opened 2024-12-08
Various tests fail on Arch Linux


kpcyrd:
hello, just wanted to let you know the following integration tests are currently failing for me:

I had to truncate some of the assembly output due to github's size limit - [full output.txt](https://github.com/user-attachments/files/18052332/output.txt)

```
     Running tests/integration_tests.rs (target/debug/deps/integration_tests-bccecb31354265dc)

running 31 tests
/usr/bin/ld: /home/user/repos/github/wild/wild/tests/build/weak-vars.16-187e5e57cedc9161.o: in function `_start':
weak-vars.c:(.text+0x171): undefined reference to `__stack_chk_fail'
test integration_test::program_name_06___weak_vars_c__ ... FAILED
test integration_test::program_name_05___data_c__ ... ok
test integration_test::program_name_01___trivial_c__ ... ok
test integration_test::program_name_04___global_definitions_c__ ... ok
test integration_test::program_name_10___init_test_c__ ... ok
test integration_test::program_name_12___internal_syms_c__ ... ok
test integration_test::program_name_03___link_args_c__ ... ok
test integration_test::program_name_14___old_init_c__ ... ok
test integration_test::program_name_11___ifunc_c__ ... ok
test integration_test::program_name_16___stack_alignment_s__ ... ok
test integration_test::program_name_02___trivial_main_c__ ... ok
test integration_test::program_name_15___custom_section_c__ ... ok
test integration_test::program_name_17___got_ref_to_local_c__ ... ok
test integration_test::program_name_18___local_symbol_refs_s__ ... ok
test integration_test::program_name_19___archive_activation_c__ ... ok
test integration_test::program_name_22___comments_c__ ... ok
test integration_test::program_name_21___string_merging_c__ ... ok
test integration_test::program_name_20___common_section_c__ ... ok
test integration_test::program_name_23___eh_frame_c__ ... ok
test integration_test::program_name_24___trivial_asm_s__ ... ok
test integration_test::program_name_25___non_alloc_s__ ... ok
error: failed to find a `codegen-backends` folder in the sysroot candidates:
       * /home/user/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu
       * /home/user/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu

test integration_test::program_name_27___rust_integration_rs__ ... FAILED
test integration_test::program_name_13___tls_c__ ... ok
test integration_test::program_name_31___input_does_not_exist_c__ ... ok
[...]

PR #357, opened 2025-02-01
linker-diff: Accept optimisation CallIndirectToRelative for static bins

davidlattimore:
We eliminate GOTPCRELX for static binaries even when the symbol is undefined. This is valid, since the binary cannot be provided by a shared object at runtime since the binary is statically linked. Other linkers don't do this.

Issue #230

### DIFF
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -172,7 +172,11 @@ impl Config {
                 "section.note.package",
                 // TLSDESC relaxations aren't yet implemented.
                 "rel.match_failed.R_X86_64_GOTPC32_TLSDESC",
-                "rel.R_X86_64_TLSDESC_CALL.R_X86_64_NONE",
+                "rel.missing-opt.R_X86_64_TLSDESC_CALL.SkipTlsDescCall.*",
+                // Wild eliminates GOTPCRELX in statically linked executables even for undefined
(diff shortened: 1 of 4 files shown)


linker-diff/src/lib.rs
```diff
@@ -172,7 +172,11 @@ impl Config {
                 "section.note.package",
                 // TLSDESC relaxations aren't yet implemented.
                 "rel.match_failed.R_X86_64_GOTPC32_TLSDESC",
-                "rel.R_X86_64_TLSDESC_CALL.R_X86_64_NONE",
+                "rel.missing-opt.R_X86_64_TLSDESC_CALL.SkipTlsDescCall.*",
+                // Wild eliminates GOTPCRELX in statically linked executables even for undefined
+                // symbols, whereas other linkers don't. This is a valid optimisation that other
+                // linkers don't currently do.
+                "rel.extra-opt.R_X86_64_GOTPCRELX.CallIndirectToRelative.static-*",
```
mati865, line 179: Maybe it'd be worth extracting these and AArch64 specific exclusions to the respective modules. On the other hand, having them all in one place makes it easier to find what is excluded.

davidlattimore: That sounds like something that could be good to do at some stage if the number of architectures grows more and / or the number of architecture-specific exclusions gets larger. Thanks for the suggestion!

Issue #376, opened 2025-02-04
(256 (1/0)): Don't know what segment to put `.debug_gdb_scripts` in, but it's referenced


dw:
Have tried full rebuild and cargo install --git for wild, issue persists. Is there anything I can provide to help diagnose the issue?

```
Description:	Debian GNU/Linux 12 (bookworm)
rustc 1.86.0-nightly (2f348cb7c 2025-01-27)
cargo 1.86.0-nightly (cecde95c1 2025-01-24)
```

I believe the section was introduced by one of the following Cargo.toml dependency changes:

```diff
-flate2 = { version = "1.0.35", default-features = false, features = ["libz-sys", "zlib"] }
+flate2 = { version = "1.0.35", default-features = false, features = ["zlib-ng"] }
-reqwest = { version = "0.12.12", default-features = false, features = ["default-tls", "blocking"] }
+reqwest = { version = "0.12.12", default-features = false, features = ["blocking", "rustls-tls"] }
````

davidlattimore:
What are you trying to link with Wild? If you can provide some instructions for how to reproduce the problem, then I'm sure we can fix it.

dw:
Just posted the deps I suspect are causing it. Trying to isolate a test case for you now

dw:
https://github.com/dw/wild-breakage-testcase triggers the problem on the distro and compiler versions listed above. Have confirmed the issue is still present after removing rust-toolchain.toml and running with rustc 1.84.0

davidlattimore:
Thanks! Looks like the problem is the combination of `debug = true` and `strip = true`. Because wild is stripping debug info, it drops the section `.debug_gdb_scripts`, but then has troubles when there's a reference to a symbol in that section. It looks like GNU ld keeps that section.

I haven't yet figured out what might be the best rule for us to determine that we shouldn't discard the section. The section has the ALLOC flag set, so perhaps that should be our signal to not discard it. Or maybe rather than discarding sections that start with `.debug`, we should keep them, then let the GC phase get rid of them if they're not referenced.

davidlattimore:
This should be fixed now. Thanks again for reporting it!

BTW, it's worth noting that turning on debug info causes rustc to emit slightly less optimal code. In particular, variables will sometimes be put on the stack, when they could instead be in registers. So if you don't need debug info, which I'm assuming you don't if you're stripping it, then you could consider turning it off for a slightly more optimised release build.

PR #378, opened 2025-02-05
Allow sections starting with '.debug_' to be kept if they're alloc

davidlattimore:
Issue #376


marxin (approved):
Thanks for taking care of this! I can confirm I've seen the mentioned section before during my work on debug info, but I didn't try using the strip option ;)

PR #351, opened 2025-01-28
x86_64: relax TLSDESC for statically linked executables

marxin:
Fixed the following Mold test: test/tlsdesc-static.sh.

PR #384, opened 2025-02-07
AArch64: relax TLSDESC for statically linked executables

marxin:
This is equivalent change I made in #351 and there are a couple of questions I have:

- As opposed to `x86_64`, this relaxation modifies 4 consecutive instructions (and also relocations):

```
 20c:   90000000        adrp    x0, 0 <_ZN3std2rt19lang_start_internal17hc40c77548151fb1dE>     20c: R_AARCH64_TLSDESC_ADR_PAGE21       _ZN3std3sys3pal4unix14stack_overflow3imp5GUARD29_$u7b$$u7b$constant$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$3VAL17h88df0c7ea033e002E
 210:   f9400001        ldr     x1, [x0]        210: R_AARCH64_TLSDESC_LD64_LO12        _ZN3std3sys3pal4unix14stack_overflow3imp5GUARD29_$u7b$$u7b$constant$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$3VAL17h88df0c7ea033e002E
 214:   91000000        add     x0, x0, #0x0    214: R_AARCH64_TLSDESC_ADD_LO12 _ZN3std3sys3pal4unix14stack_overflow3imp5GUARD29_$u7b$$u7b$constant$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$3VAL17h88df0c7ea033e002E
 218:   d63f0020        blr     x1      218: R_AARCH64_TLSDESC_CALL     _ZN3std3sys3pal4unix14stack_overflow3imp5GUARD29_$u7b$$u7b$constant$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$3VAL17h88df0c7ea033e002E
```

Should I somehow rely on the fact these are used together, or should I ensure that e.g. when a relaxation is built for the first one?

- The `RelaxationKind` would be very instructions specific on AArch64, are you happy with the selected names?
- I hit very interesting debug info size mismatch that I'm investigating right now:

```
debug_info.size_mismatch
  Size mismatch for /rust/deps/compiler_builtins-0.1.143/src/lib.rs/@/compiler_builtins.a69a77954fe228ee-cgu.014 (/rust/deps/compiler_builtins-0.1.143) in ld, expected: 87, got: 77
```

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -6,8 +6,10 @@ use crate::elf::PLT_ENTRY_SIZE;
 use crate::resolution::ValueFlags;
 use anyhow::anyhow;
 use anyhow::Result;
+use linker_utils::aarch64::relocation_type_from_raw;
 use linker_utils::aarch64::RelaxationKind;
 use linker_utils::elf::aarch64_rel_type_to_string;
+use linker_utils::elf::shf;
 use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::RelocationInstruction;
 use linker_utils::elf::RelocationKind;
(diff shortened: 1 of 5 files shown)


marxin:
> * I hit very interesting debug info size mismatch that I'm investigating right now:
> 
> 
> ```
> debug_info.size_mismatch
>   Size mismatch for /rust/deps/compiler_builtins-0.1.143/src/lib.rs/@/compiler_builtins.a69a77954fe228ee-cgu.014 (/rust/deps/compiler_builtins-0.1.143) in ld, expected: 87, got: 77
> ```

Seems it's related to fact that there are multiple CUs in the DWARF format, every time with a different abbreviation.

davidlattimore:
> Should I somehow rely on the fact these are used together, or should I ensure that e.g. when a relaxation is built for the first one?

Do you think any compiler would ever not emit all 4? Does the spec say that all 4 are to be used together? If it does, then it seems reasonable to rely on that.

I guess if we decided that we wanted to check, we'd need to change the way we currently do relaxations a bit.

marxin:
> Do you think any compiler would ever not emit all 4?

I am not aware of any. So yes, we can adjust it in the future if it causes troubles.

PR #386, opened 2025-02-08
integration_tests: Don't use lld when cross compiling

davidlattimore:


davidlattimore:
@mati865 - I found out why I had lld disabled when cross compiling. It's probably not too hard to fix, although the only fix that comes to mind, creating a temporary directory with a symlink to lld, feels a little hacky. Anyway, I don't need lld for the moment, so I've just disabled it.

mati865:
Ah, I'd expect this bug to have been fixed after all those years.
In the future, LLD could be enabled with Clang as the cross-compiler, but it is not a big deal.

PR #390, opened 2025-02-08
Skip one test as GCC on Ubuntu does not support -fPIC + large model

marxin:


wild/tests/sources/cpp-integration.cc
```diff
@@ -21,6 +21,8 @@
 //#LinkArgs:--cc=g++ -Wl,-z,now
 //#EnableLinker:lld
 //#Cross:false
+// TODO: Ubuntu: cc1plus: sorry, unimplemented: code model 'large' with '-fPIC'
```
davidlattimore, line 24: That was the same error that caused me to set `Cross:false` above :)

PR #393, opened 2025-02-08
Dump final value of a relocation

marxin:


libwild/src/elf_writer.rs
```diff
@@ -1991,6 +1990,13 @@ fn apply_relocation<S: StorageModel, A: Arch>(
         RelocationKind::None | RelocationKind::TlsDescCall => 0,
     };
 
+    let value_hex = format!("{value:x}");
```
davidlattimore: Does the compiler manage to optimise away this heap allocation when tracing isn't active? One alternative might be to create a struct HexU64 that wraps the value and implements Display.

marxin: Sure, good idea!

PR #394, opened 2025-02-08
Add -mno-fix-cortex-a53-835769 default linker flag

marxin:


wild/tests/integration_tests.rs
```diff
@@ -332,10 +333,16 @@ impl ArgumentSet {
...
-            // Wild linker uses -znow by default!
-            args: vec!["-z".to_owned(), "now".to_owned()],
+        // Wild linker uses -znow by default!
+        #[allow(unused_mut)]
+        let mut args = vec!["-z".to_owned(), "now".to_owned()];
+        #[cfg(target_arch = "aarch64")]
+        {
+            // Provide a workaround for ld.lld: error: unknown argument '--fix-cortex-a53-835769'
+            // Bug link: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105941
+            args.push("-mno-fix-cortex-a53-835769".to_owned());
```
davidlattimore: Should we unconditionally pass this as an argument rather than just having it be a default argument? Otherwise tests that override `LinkArgs` won't get it right?

Also, if it's practical, it could be good to apply it based on the arch that we're linking for rather than the host arch - although we don't yet cross compile with lld, so it doesn't matter too much at the moment.

marxin: Yeah, we should find a better place. Apparently, it's a compiler option we should append:
```
marxin@arm64box:~/Programming> gcc m.cpp -mno-fix-cortex-a53-835769 
marxin@arm64box:~/Programming> clang m.cpp -mno-fix-cortex-a53-835769 
marxin@arm64box:~/Programming> 
```

What place would fit the best?

davidlattimore: I think probably somewhere in `LinkCommand::new`. You can check the target architecture something like this:

```rust
        let arch = cross_arch.unwrap_or_else(get_host_architecture);
        if arch == Architecture::AArch64 {

        }
```

Issue #330, opened 2025-01-25
Cannot link ClickHouse project


alexey-milovidov:
I tried to use it as a drop-in replacement for linking [ClickHouse](https://github.com/ClickHouse/ClickHouse/) and found it needs more complete support:

Linking with LLD:
```
[...]

PR #405, opened 2025-02-11
Write merged_strings in parallel

marxin:
Improves the test-case mentioned in #330 by ~10% on my machine.


davidlattimore:
That's awesome! I love how straightforward the change is. I think I would have forgotten about par_bridge and made it much more complex than necessary.

marxin:
Same for me, the Rust ergonomics really shines!

PR #414, opened 2025-02-15
AArch64: add relocation range check

marxin:


linker-utils/src/aarch64.rs
```diff
@@ -100,522 +101,728 @@ impl RelaxationKind {
...
+            RelocationKind::Absolute,
+            RelocationSize::ByteSize(4),
+            None,
+            AllowedRange::new(-(2i64.pow(31)), 2i64.pow(32)),
+        ),
+        object::elf::R_AARCH64_ABS16 => (
+            RelocationKind::Absolute,
+            RelocationSize::ByteSize(2),
+            None,
+            AllowedRange::new(-(2i64.pow(15)), 2i64.pow(16)),
```
davidlattimore, line 131: Why are many of these ranges so asymmetrical? e.g. this one is [-32768..65536)

davidlattimore (commented):
Good idea!

linker-utils/src/aarch64.rs
marxin: Dunno, I just followed the SPEC:
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst

Should we file an issue against the documentation? It might be the case, they wrongly placed `-(2^15)` instead of `-(2^16 - 1)`.

davidlattimore: Ah, I see. The spec says:

> These overflow ranges permit either signed or unsigned narrow values to be created from the intermediate result viewed as a 64-bit signed integer. If the place is intended to hold a narrow signed value and INTn_MAX < X <= UINTn_MAX, no overflow will be detected but the positive result will be interpreted as a negative value.

So we could check a tighter bound if we knew that the value was signed, but we don't.

PR #415, opened 2025-02-15
Support `openSUSE` with WILD_TEST_CROSS & support WILD_TEST_CROSS in CI

marxin:


mati865:
The same issue happens on Arch, it'd be nice to get to the core of the problem. I feel this is just papering over the problem.

Issue #413, opened 2025-02-15
disarm64 compilation needs > 4GiB of memory


marxin:
I tried building `wild` on the Hetzner AArch64 server with 4 GiB of memory, but the compilation hit the out-of-memory limit even in `dev` mode without debug information. Even on a powerful machine, the compilation takes some time :)

Theoretically, we can simplify that by using directly `objdump`:
```
$ xxd snippet 
00000000: 0800 0037                                ...7
$ objdump  -b binary -m aarch64 -D snippet
snippet:     file format binary

Disassembly of section .data:

0000000000000000 <.data>:
   0:	37000008 	tbnz	w8, #0, 0x0
```

https://github.com/davidlattimore/wild/blob/eadfd8eebe0a7f4dc6ac51a5c9021279912ae750/Cargo.toml#L13-L20

PR #416, opened 2025-02-15
Replace disarm64 with objdump due to performance reasons

marxin:
Fixes: #413

### DIFF
--- a/linker-diff/src/aarch64.rs
+++ b/linker-diff/src/aarch64.rs
@@ -56,6 +56,14 @@ fn decode_insn_with_objdump(insn: &[u8]) -> Result<String> {
         .to_string())
 }
 
+#[test]
+fn test_align_up() {
+    assert_eq!(
+        decode_insn_with_objdump(&[0xe3, 0x93, 0x44, 0xa9]).unwrap(),
+        "ldp\tx3, x4, [sp, #72]"
+    );
+}
+
 impl Arch for AArch64 {
     type RType = RType;
 
--- a/linker-diff/src/aarch64.rs
+++ b/linker-diff/src/aarch64.rs
@@ -30,7 +30,12 @@ fn decode_insn_with_objdump(insn: &[u8]) -> Result<String> {
     tmpfile.write_all(insn)?;
     tmpfile.flush()?;
 
-    let command = Command::new("objdump")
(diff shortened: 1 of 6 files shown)


mati865:
Do we need whole disarm64? I wanted to try if building it with only some of the features instead of full would be sufficient, but I decided to fix the problem with tests on my box first.

Here's how the features are used there: https://github.com/kromych/disarm64/blob/master/disarm64%2Fsrc%2Flib.rs

marxin:
> Do we need whole disarm64? I wanted to try if building it with only some of the features instead of full would be sufficient, but I decided to fix the problem with tests on my box first.

Probably not, we are interested in the all instructions one can find in the table of relocations:
https://github.com/davidlattimore/wild/blob/eadfd8eebe0a7f4dc6ac51a5c9021279912ae750/linker-utils/src/elf.rs#L660-L671

Feel free to experiment if we can come up with a reasonable subset of the instructions.

davidlattimore:
> Do we need whole disarm64? I wanted to try if building it with only some of the features instead of full would be sufficient, but I decided to fix the problem with tests on my box first.
> 
> Here's how the features are used there: https://github.com/kromych/disarm64/blob/master/disarm64%2Fsrc%2Flib.rs

I just tried disabling features of disarm64 and it seems that the decoder, which we're using requires the "full" feature.

mati865:
Thanks, it looks like this isn't going to work.

linker-diff/src/aarch64.rs
```diff
@@ -14,16 +16,70 @@ use linker_utils::elf::RelocationInstruction;
...
+use std::io::Write;
+use std::process::Command;
+use std::process::Stdio;
+use std::sync::OnceLock;
+use tempfile::NamedTempFile;
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub(crate) struct AArch64;
 
+fn decode_insn_with_objdump(insn: &[u8]) -> Result<String> {
```
davidlattimore, line 28: I think we'll want to pass `--adjust-vma=` to objdump as well so that we correctly decode the target address of relative jumps. I was going to just merge and add that afterwards, but since there's a merge conflict...

marxin: Sure, good idea, can you please do that once you merge this one?

PR #418, opened 2025-02-15
host_supports_clang_with_tls_desc: do not create echo process

marxin:


wild/tests/integration_tests.rs
```diff
@@ -220,21 +221,19 @@ fn host_supports_clang_with_tls_desc() -> bool {
...
-
-        clang.status.success()
+        let mut stdin = clang.stdin.take().expect("Failed to open stdin");
+        std::thread::spawn(move || {
+            stdin
+                .write_all("int main() { return 0; }".as_bytes())
+                .expect("Failed to write to stdin");
+        });
+
+        dbg!(clang.wait().expect("Wait failed").success())
```
mati865: Leftover dbg!

wild/tests/integration_tests.rs
```diff
@@ -220,21 +221,19 @@ fn host_supports_clang_with_tls_desc() -> bool {
...
             .args(["-mtls-dialect=gnu2", "-x", "c", "-", "-o/dev/null"])
-            .stdin(Stdio::from(echo_out))
-            .output()
+            .stdin(Stdio::piped())
+            .spawn()
             .expect("Failed to run clang");
-
-        clang.status.success()
+        let mut stdin = clang.stdin.take().expect("Failed to open stdin");
+        std::thread::spawn(move || {
```
mati865: Is spawning a thread necessary? AFAIK, it shouldn't be.

wild/tests/integration_tests.rs
marxin: Good point, thanks.

wild/tests/integration_tests.rs
marxin: I followed the example here: https://doc.rust-lang.org/std/process/struct.Stdio.html#method.piped.

mati865: Make sense, although in this case using up the entire 4K buffer shouldn't be a problem.

marxin: I tried not using the separate thread:
```rust
let mut clang = Command::new("clang")
            .args(["-mtls-dialect=gnu2", "-x", "c", "-", "-o/dev/null"])
            .stdin(Stdio::piped())
            .spawn()
            .expect("Failed to run clang");
        let mut stdin = clang.stdin.take().expect("Failed to open stdin");
        stdin
            .write_all("int main() { return 0; }".as_bytes())
            .expect("Write of a source file failed");
        clang.wait().expect("Wait failed").success()
```

and the test-case execution is stuck :)

davidlattimore: It would have gotten stuck because clang was waiting for the end of the input stream. If you do `drop(stdin)`, it should work.

marxin: Great, not it works :)

PR #419, opened 2025-02-15
Implement sysroot handling

mati865:
Fixes https://github.com/davidlattimore/wild/issues/406

### DIFF
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -14,6 +14,7 @@ use crate::symbol::SymbolName;
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
+use normalize_path::NormalizePath;
 use std::collections::HashSet;
 use std::path::Path;
 
@@ -23,6 +24,7 @@ pub(crate) fn linker_script_to_inputs(
     bytes: &[u8],
     path: &Path,
     modifiers: Modifiers,
+    sysroot: Option<&Path>,
 ) -> Result<Vec<Input>> {
     let text = std::str::from_utf8(bytes)?;
     let directory = path
@@ -33,9 +35,34 @@ pub(crate) fn linker_script_to_inputs(
         .into_iter()
         .map(|mut input| {
             input.search_first = Some(directory.to_owned());
+            if let (Some(sysroot), InputSpec::File(file)) = (sysroot, &mut input.spec) {
+                if let Some(new_file) = maybe_apply_sysroot(path, file, sysroot) {
+                    *file = new_file;
+                }
+            }
+
             input
         })
         .collect())
+}
+
+fn maybe_apply_sysroot(
+    linker_script_path: &Path,
+    input_path: &Path,
+    sysroot: &Path,
+) -> Option<Box<Path>> {
+    if linker_script_path.normalize().starts_with(sysroot) {
+        Some(Box::from(sysroot.join(input_path.strip_prefix("/").ok()?)))
+    } else {
+        maybe_forced_sysroot(input_path, sysroot)
+    }
+}
+
+pub(crate) fn maybe_forced_sysroot(path: &Path, sysroot: &Path) -> Option<Box<Path>> {
+    path.strip_prefix("=")
+        .or_else(|_| path.strip_prefix("$SYSROOT"))
+        .ok()
+        .map(|stripped| Box::from(sysroot.join(stripped)))
 }
 
 /// A version script. See https://sourceware.org/binutils/docs/ld/VERSION.html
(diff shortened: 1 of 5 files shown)


mati865:
Right now this handles only linker scripts part **based on my guess how it should work**. Unfortunately I couldn't find any information in the docs but with this change integration tests fail with errors like:
```
wild: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-aarch64.wild
ld: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-aarch64.ld
.dynamic.DT_BIND_NOW
  wild
  ld 0x0

.dynamic.DT_FLAGS
  wild 0x8
  ld
```
So, at least linking is successful.

I still need to fix `-L=` according to the docs and reconsider if this could be done better.

davidlattimore:
Thanks for working on this! I tried patching your PR and the tests all passed for me. It's strange that ld isn't setting DT_FLAGS=0x8 (BIND_NOW), given that we should be passing `-Wl,-z,now` to the compiler when linking.

Boxing a String probably isn't necessary and removing the Box will likely help to eliminate the clippy warning.

mati865:
> Boxing a String probably isn't necessary and removing the Box will likely help to eliminate the clippy warning.

Yeah, it's still WIP. I had started with Path, hence this Box. I haven't touched this since publishing the draft because of other things, but I'll be back to it later today.

mati865:
So, I searched more extensively, because that sysroot application seemed dangerously broad to me and found where it's described: https://sourceware.org/binutils/docs/ld/File-Commands.html
My guess wasn't that far off, as I missed the condition to use sysroot only when linker script is within the sysroot for absolute paths 😉 Also  added `=` and `$SYSROOT` handling in the scripts.

This looks mostly good now but still needs some love. I'd like to get your opinion on [`556ccc3` (#419)](https://github.com/davidlattimore/wild/pull/419/commits/556ccc3e6dfd2063d100dfca3239008a06ecc7d7) "Is it any better?" commit. I'm not happy with either way, WDYT? Maybe you have a better idea?

Tests and `-L=` on the CLI are still on my todo, but probably I'll get to it tomorrow.

davidlattimore:
> So, I searched more extensively, because that sysroot application seemed dangerously broad to me and found where it's described: https://sourceware.org/binutils/docs/ld/File-Commands.html
> My guess wasn't that far off, as I missed the condition to use sysroot only when linker script is within the sysroot for absolute paths 😉 Also added `=` and `$SYSROOT` handling in the scripts.

Ah, nice find, that makes sense.

> This looks mostly good now but still needs some love. I'd like to get your opinion on [`556ccc3` (#419)](https://github.com/davidlattimore/wild/pull/419/commits/556ccc3e6dfd2063d100dfca3239008a06ecc7d7) "Is it any better?" commit. I'm not happy with either way, WDYT? Maybe you have a better idea?

Either looks fine to me.

libwild/src/linker_script.rs
```diff
@@ -33,11 +34,36 @@ pub(crate) fn linker_script_to_inputs(
...
         })
         .collect())
 }
 
+fn maybe_apply_sysroot(
+    linker_script_path: &Path,
+    input_path: &Path,
+    sysroot: &Path,
+) -> Option<Box<Path>> {
+    if matches!(linker_script_path.canonicalize(), Ok(p) if p.starts_with(sysroot)) {
```
mati865: `linker_script_path` can look like this: `/usr/lib/gcc/aarch64-linux-gnu/14.2.0/../../../../aarch64-linux-gnu/lib/../lib64/libgcc_s.so` and because `Path::starts_with` checks the path literally, without `Path::canonicalize` it would give false negative if we compare against `/usr/lib/aarch64-linux-gnu`.

`canonicalize` is the easy way to avoid carrying about it but not necessarily the best: it'll call libc to check if file exists and normalize (so, allocate) on every iteration.

mati865: Okay, I'm probably trying harder than necessary. Without `--sysroot` this won't affect performance at, with `--sysroot` hopefully it will be acceptable.

davidlattimore: If we're worried, there's always https://docs.rs/normalize-path/0.2.1/normalize_path/ as a possibility. Or if we worry about too many dependencies, we could roll our own, since it presumably would be a pretty simple function to implement.

Doing normalisation or canonicalisation earlier so we only do it once might also help, although I guess we wouldn't want to incur that cost if we didn't have a sysroot set.

But you're right, it's probably fine as-is.

mati865: I had something similar to that, but without creating intermediate PathBuf. However it looked unnecessarily complex for a problem that maybe isn't really a problem. Although I can use `normalize_path` if you want, an allocation in the loop is still better than allocation with IO.

mati865: Went for it, even the readability has improved.

davidlattimore:
Thanks for taking this on and for figuring out some of the more tricky details involved!

Issue #412, opened 2025-02-15
Failed to build wild with itself on aarch64


davidlattimore:
I tried to build wild with itself on a Raspberry Pi 5 and got the following error:

```
  error occurred: Command "cc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-4" "-fno-omit-frame-pointer" "-I" "zstd/lib/" "-I" "zstd/lib/common" "-I" "zstd/lib/legacy" "-fvisibility=hidden" "-DZSTD_LIB_DEPRECATED=0" "-DXXH_PRIVATE_API=" "-DZSTDLIB_VISIBILITY=" "-DZDICTLIB_VISIBILITY=" "-DZSTDERRORLIB_VISIBILITY=" "-DZSTD_LEGACY_SUPPORT=1" "-o" "/home/david/wild/target/debug/build/zstd-sys-2d052d44f16a11f8/out/690f776bf7b37aeb-huf_decompress_amd64.o" "-c" "zstd/lib/decompress/huf_decompress_amd64.S" with args cc did not execute successfully (status code exit status: 1).
```

If I run the command that it says failed, it succeeds. Looking at an strace of the successful invocation of `cc` vs the one that happens during a cargo build:

```
64014 execve("/usr/bin/cc", ["cc", "-O0", "-ffunction-sections", "-fdata-sections", "-fPIC", "-gdwarf-4", "-fno-omit-frame-pointer", "-I", "zstd/lib/", "-I", "zstd/lib/common", "-I", "zstd/lib/legacy", "-fvisibility=hidden", "-DZSTD_LIB_DEPRECATED=0", "-DXXH_PRIVATE_API=", "-DZSTDLIB_VISIBILITY=", "-DZDICTLIB_VISIBILITY=", "-DZSTDERRORLIB_VISIBILITY=", "-DZSTD_LEGACY_SUPPORT=1", "-o", "/home/david/wild/target/debug/build/zstd-sys-2d052d44f16a11f8/out/690f776bf7b37aeb-huf_decompress_amd64.o", "-c", "zstd/lib/decompress/huf_decompress_amd64.S"], 0x7fffeb1fe618 /* 34 vars */) = 0
```

vs the unsuccessful build

```
[...]

PR #427, opened 2025-02-16
Fix incorrect resolution flag for RelocationKind::Got

davidlattimore:
Issue #412


marxin:
May I ask you David why does it cause troubles only on AArch64? I expect the same logic has been thoroughly tested on x86_64.

davidlattimore:
> May I ask you David why does it cause troubles only on AArch64? I expect the same logic has been thoroughly tested on x86_64.

It's not used by x86-64 and was added just for aarch64 (#244). x86-64 tends to always use GotRelative instead.

PR #433, opened 2025-02-17
Detect object architecture when parsing

### LINKED ISSUE #424: Improve error when attempting to link incompatible object files
It's easy, with the wrong flags to end up trying to link a mix of objects from different architectures.

lld gives errors like the following:

```
ld.lld: error: foo.o is incompatible with elf64-x86-64
```

We currently error like this:

```
  = note: thread '<unnamed>' panicked at libwild/src/aarch64.rs:106:76:
          called `Result::unwrap()` on an `Err` value: Unsupported relocation type Unknown aarch64 relocation type 0x2a
          clang: error: linker command failed with exit code 101 (use -v to see invocation)
```

When we open each object file, we should check what architecture it's for in the header and if it doesn't match the architecture that we're linking for, emit a helpful error message.

inflation:
Fix #424

### DIFF
--- a/libwild/src/arch.rs
+++ b/libwild/src/arch.rs
@@ -8,6 +8,8 @@ use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::RelocationKindInfo;
 use linker_utils::elf::SectionFlags;
 use linker_utils::relaxation::RelocationModifier;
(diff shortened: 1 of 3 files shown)


davidlattimore:
Thanks!

PR #439, opened 2025-02-17
Set sysroot in Rust's integration tests when cross-compiling

mati865:
See discussion in https://github.com/davidlattimore/wild/pull/429

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -230,6 +230,10 @@ fn is_host_opensuse() -> bool {
     os_info::get().os_type() == Type::openSUSE
 }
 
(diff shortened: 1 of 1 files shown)


marxin:
Thanks for this change. Now it seems all except one test work for me on openSUSE. The one problem is about:
`/home/marxin/Programming/wild/wild/tests/build/cpp-integration.cc-no-pie-aarch64.ld: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory`

It's a library that's not present in the `/usr/aarch64-suse-linux/sys-root/lib64`, but in `/usr/lib64/gcc/aarch64-suse-linux/14/`. Anyway, for now, I am going to copy the file.

PR #449, opened 2025-02-19
Replace tests synchronisation mutex with file lock

mati865:
Fixes https://github.com/davidlattimore/wild/issues/445


davidlattimore:
Thanks for fixing this!

PR #452, opened 2025-02-20
Add `--whole-archive`

### LINKED ISSUE #272: Add support for --whole-archive
See `man ld` for a description of what `--whole-archive` needs to do.

`--whole-archive` is a modifier that is pushed / popped by `--push-state` / `--pop-state`, so the value of the flag should be added to the `Modifiers` struct (in `args.rs`).

The method where the `whole_archive` attribute should be used is `ParsedInputObject::is_optional`.

It'd be good to have a test that verifies that `--whole-archive` is behaving as expected. There are probably a few ways such a test could be written. The key observation is that whatever is defined in the archive needs to not be directly referenced from the main test object, otherwise the archived object would be included even without `--whole-archive`. One option is to have an archive that puts something into a custom section, then have the `_start_` function check the custom section for that value. You can look at the test `custom_section.c` for how this is done.

riverbl:
Closes #272.

I was getting some funny behaviour where the added test would pass but the option wouldn't work when actually using the linker, but I'm no longer seeing this after rebasing onto latest `main`.

### DIFF
--- a/libwild/src/parsing.rs
+++ b/libwild/src/parsing.rs
@@ -129,11 +129,12 @@ impl<'data> ParsedInputObject<'data> {
         self.is_dynamic
     }
 
-    /// Returns whether this input should be skipped if there are no non-weak reference to symbols
-    /// it defines. This is true for archive entries and shared objects for which --as-needed is
-    /// true.
+    /// Returns whether this input should be skipped if there are no non-weak references to symbols
+    /// it defines. This is true for archive entries for which --whole-archive is false and shared
+    /// objects for which --as-needed is true.
     pub(crate) fn is_optional(&self) -> bool {
-        self.input.entry.is_some() || (self.is_dynamic() && self.modifiers.as_needed)
+        (self.input.entry.is_some() && !self.modifiers.whole_archive)
+            || (self.is_dynamic() && self.modifiers.as_needed)
     }
 
     fn filename(&self) -> &'data Path {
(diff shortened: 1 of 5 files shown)


libwild/src/parsing.rs
```diff
@@ -133,7 +133,8 @@ impl<'data> ParsedInputObject<'data> {
     /// it defines. This is true for archive entries and shared objects for which --as-needed is
     /// true.
     pub(crate) fn is_optional(&self) -> bool {
-        self.input.entry.is_some() || (self.is_dynamic() && self.modifiers.as_needed)
+        !self.modifiers.whole_archive
+            && (self.input.entry.is_some() || (self.is_dynamic() && self.modifiers.as_needed))
```
davidlattimore: It looks like the parentheses might be in the wrong spot. The --whole-archive flag shouldn't affect shared objects, but it looks like setting --whole-archive here will cause shared objects to become non-optional even if they're --as-needed.

davidlattimore (commented):
Thanks!

libwild/src/parsing.rs
```diff
@@ -133,7 +133,8 @@ impl<'data> ParsedInputObject<'data> {
     /// it defines. This is true for archive entries and shared objects for which --as-needed is
     /// true.
```
mati865: Let's also update this comment regarding archive entries

mati865: Thanks!

PR #463, opened 2025-02-22
Fix wrong resolution related to TLS

marxin:
The following PR fixes the following issue:

```
Error: Unexpected memory offsets:
Part #5 (section `.got` alignment: 8) expected: 0x403f28 actual: 0x403f08 bumped by: 0x38 requested size: 0x58
```

However, it seems we need to implement TLSDESC -> TLS offset relaxation, otherwise we face a run-time crash if TLSDESC is combined with `TPOFF64`.


marxin:
@davidlattimore May I ask you for help?

I face the following issues right now:

1. on `x86_64`, the following assert is hit for the special start/stop symbol we introduce:

```
Caused by:
    0: Failed to process `symbol `_TLS_MODULE_BASE_` (9 local=9) in file #0 (0/0) (<prelude>) (ADDRESS | CAN_BYPASS_GOT)`
    1: GotTlsOffset resolves to address not in TLS segment 0x4035d8
```


2. On `AArch64` I managed to implement the same relaxation, however the binary fails at run-time:

```
marxin@pi:~/Programming$ cat once.cc
#include <iostream>
#include <mutex>
#include <thread>
 
std::once_flag flag1, flag2;
 
void foo()
{
    std::call_once(flag1, [](){ std::cout << "Simple example: called once\n"; });
}
marxin@pi:~/Programming$ cat once2.cc
#include <iostream>
#include <mutex>
#include <thread>
 
std::once_flag flag11;
 
void bar()
{
    std::call_once(flag11, [](){ std::cout << "Simple example: called once2\n"; });
}
marxin@pi:~/Programming$ cat main.cc
void foo();
void bar();

int main() {
	foo();
	bar();

	return 0;
}
$ marxin@pi:~/Programming$ g++ once.cc -c -ftls-model=initial-exec --save-temps && g++ once2.cc -c -fpic && g++ main.cc once*.o -B ~/Programming/wild/fakes-debug/ && ./a.out
WARNING: wild: --plugin /usr/libexec/gcc/aarch64-linux-gnu/14/liblto_plugin.so is not yet supported
WARNING: wild: --fix-cortex-a53-843419 is not yet supported
WARNING: wild: -z relro is not yet supported
Segmentation fault (core dumped)
$ marxin@pi:~/Programming$ g++ once.cc -c -ftls-model=initial-exec --save-temps && g++ once2.cc -c -fpic && RUST_LOG=trace g++ main.cc once*.o -B ~/Programming/wild/fakes-debug/ | grep once_callable
2025-02-22T19:20:24.409008Z TRACE libwild::elf_writer: Write .dynsym name=_ZSt15__once_callable
2025-02-22T19:20:24.411169Z TRACE write_file{filename=once.o}:relocation{address=72876 address_hex=11cac}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOff value=69632 value_hex=11000 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.411242Z TRACE write_file{filename=once.o}:relocation{address=72880 address_hex=11cb0}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOffGot value=139792 value_hex=22210 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.411611Z TRACE write_file{filename=once.o}:relocation{address=73012 address_hex=11d34}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOff value=69632 value_hex=11000 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.411692Z TRACE write_file{filename=once.o}:relocation{address=73016 address_hex=11d38}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOffGot value=139792 value_hex=22210 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.412142Z TRACE write_file{filename=once.o}:relocation{address=73180 address_hex=11ddc}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOff value=69632 value_hex=11000 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.412217Z TRACE write_file{filename=once.o}:relocation{address=73184 address_hex=11de0}: libwild::elf_writer: relocation applied value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOffGot value=139792 value_hex=22210 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.417401Z TRACE write_file{filename=once2.o}:relocation{address=73832 address_hex=12068}: libwild::elf_writer: relaxation applied kind=ReplaceWithNop value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=None value=0 value_hex=0 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.417477Z TRACE write_file{filename=once2.o}:relocation{address=73836 address_hex=1206c}: libwild::elf_writer: relaxation applied kind=ReplaceWithNop value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=None value=0 value_hex=0 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.417554Z TRACE write_file{filename=once2.o}:relocation{address=73840 address_hex=12070}: libwild::elf_writer: relaxation applied kind=AdrpX0 value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOff value=65536 value_hex=10000 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.417638Z TRACE write_file{filename=once2.o}:relocation{address=73844 address_hex=12074}: libwild::elf_writer: relaxation applied kind=LdrX0 value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOffGot value=139792 value_hex=22210 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.418004Z TRACE write_file{filename=once2.o}:relocation{address=73976 address_hex=120f8}: libwild::elf_writer: relaxation applied kind=ReplaceWithNop value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=None value=0 value_hex=0 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.418079Z TRACE write_file{filename=once2.o}:relocation{address=73980 address_hex=120fc}: libwild::elf_writer: relaxation applied kind=ReplaceWithNop value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=None value=0 value_hex=0 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.418155Z TRACE write_file{filename=once2.o}:relocation{address=73984 address_hex=12100}: libwild::elf_writer: relaxation applied kind=AdrpX0 value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOff value=65536 value_hex=10000 symbol_name=_ZSt15__once_callable
2025-02-22T19:20:24.418230Z TRACE write_file{filename=once2.o}:relocation{address=73988 address_hex=12104}: libwild::elf_writer: relaxation applied kind=LdrX0 value_flags=DYNAMIC resolution_flags=DIRECT | GOT_TLS_OFFSET | COPY_RELOCATION rel_info.kind=GotTpOffGot value=139792 value_hex=22210 symbol_name=_ZSt15__once_callable
$ marxin@pi:~/Programming$ readelf -rW a.out | grep once_callable
0000000000022490  0000001300000400 R_AARCH64_COPY         0000000000022490 _ZSt15__once_callable@GLIBCXX_3.4.11 + 0
0000000000022210  0000001300000406 R_AARCH64_TLS_TPREL64  0000000000022490 _ZSt15__once_callable@GLIBCXX_3.4.11 + 0
```

I suspect it's related to fact, we emit `R_AARCH64_COPY` for the symbol, while e.g. Mold not.

3. Do we have a possibility to specify an extra `CompArg` for an `//#Object:` file?

davidlattimore:
> GotTlsOffset resolves to address not in TLS segment

I haven't looked into this yet, but will do so.

> I suspect it's related to fact, we emit R_AARCH64_COPY for the symbol

Copy relocations for thread-locals definitely seems suspect to me, especially if the other linkers aren't doing them. It's quite possible that the C runtime wouldn't support them.

> 3\. Do we have a possibility to specify an extra `CompArg` for an `//#Object:` file?

I don't think we do at present, but it seems reasonable to add such a mechanism.

libwild/src/layout.rs
```diff
@@ -4007,7 +4007,8 @@ fn create_resolution(
         } else {
             resolution.got_address = Some(allocate_got(1, memory_offsets));
         }
-    } else if res_kind.contains(ResolutionFlags::GOT_TLS_MODULE)
+    }
+    if res_kind.contains(ResolutionFlags::GOT_TLS_MODULE)
         | res_kind.contains(ResolutionFlags::GOT_TLS_DESCRIPTOR)
     {
         resolution.got_address = Some(allocate_got(2, memory_offsets));
```
davidlattimore, line 4017: So one potential issue here is that got_address is being assigned twice, so the second assignment will overwrite the first. I think that was the idea behind the logic above, where it checks if GOT_TLS_OFFSET is set, then if it is it either assigns 3 slots or 1 slot depending on whether GOT_TLS_MODULE is also set. That way `resolution.got_address` will always point to the first got address for that symbol and the other got addresses can be computed via known offsets from that base. The alternative would be to store several GOT addresses on the resolution. e.g. `got_tls_offset_address`, `got_tls_module_address` etc, however that would substantially increase the size of the `Resolution` struct which would likely hurt performance.

One possibility is to restructure this code to compute how many GOT slots are needed. e.g.

```rust
let mut num_got_slots = 0;
if res_kind.contains(ResolutionFlags::GOT_TLS_OFFSET) {
  num_got_slots += 1;
}
if res_kind.contains(ResolutionFlags::GOT_TLS_MODULE) {
  num_got_slots += 2;
}
...
if num_got_slots > 0 {
   resolution.got_address = Some(allocate_got(num_got_slots, memory_offsets);
}
```

It'll probably be necessary to look at the code (used during the write phase) that computes different offsets relative to the base. That code would need to follow similar logic. e.g. the TLS_OFFSET is always at `resolution.got_address + 0`, then TLS_MODULE is at offset 0 if there's not TLS_OFFSET or offset 1 if there is a TLS_OFFSET and so on for TLS_DESCRIPTOR.

marxin:
> Copy relocations for thread-locals definitely seems suspect to me, especially if the other linkers aren't doing them. It's quite possible that the C runtime wouldn't support them.
> 
> >

I found the root cause, it's a funny coincidence. It's related to the following Relaxation:
https://github.com/davidlattimore/wild/blob/a226f4beaad13ae010cc49d460650df449a8505a/libwild/src/aarch64.rs#L147

where the `R_AARCH64_NONE` relocation is treated as a direct reference here:
https://github.com/davidlattimore/wild/blob/a226f4beaad13ae010cc49d460650df449a8505a/libwild/src/layout.rs#L2599

and thus we end up with a copy relocation.

libwild/src/layout.rs
marxin: The suggested change makes perfect sense to me. Would you be okay with me making it part of another PR that includes integration tests for all 3 access types?

davidlattimore: Sounds good

PR #483, opened 2025-02-26
Support all 3 access TLS ABIs at the same time

marxin:
Right now, we can properly handle the following test case: https://github.com/davidlattimore/wild/issues/448#issuecomment-2676699324 and hopefully the code is more readable right now.

I've also added support for extra compilation arguments for `Objects/Shared/Archive`, but it seems I would need help with addressing the linker-diff.

Related to: #463.

### DIFF
--- a/wild/Cargo.toml
+++ b/wild/Cargo.toml
@@ -16,8 +16,6 @@ mimalloc = { version = "0.1", default-features = false, optional = true }
 
 dhat = { version = "0.3.3", optional = true }
 os_info = "3.10.0"
-strum = { version = "0.27.1", features = ["derive"] }
-strum_macros = "0.27.1"
 
 [dev-dependencies]
 anyhow = "1.0.95"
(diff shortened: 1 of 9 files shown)


wild/Cargo.toml
```diff
@@ -16,6 +16,8 @@ mimalloc = { version = "0.1", default-features = false, optional = true }
 
 dhat = { version = "0.3.3", optional = true }
 os_info = "3.10.0"
+strum = { version = "0.27.1", features = ["derive"] }
```
davidlattimore: If these are only used from tests, can they be moved to dev-dependencies?

davidlattimore (commented):
Looks good. I'll see if I can figure out how to make linker-diff happy today.

marxin:
> Looks good. I'll see if I can figure out how to make linker-diff happy today.

Will appreciate that. Looking at the `x86_64` issue, there seems to be a relaxation from `global-dynamic` -> `initial-exec` being done by `ld.bfd` as one of the objects is marked as `initial-exec` and thus the final shared object has the flag:
```
❯ readelf -d /home/marxin/Programming/wild/wild/tests/build/tls-variant-1.gcc-shared-host-7d27e91bf8aba2ba.ld.so | rg FLAGS
 0x000000000000001e (FLAGS)              BIND_NOW STATIC_TLS
```

That's weird as mentioned in the Maskray's blog post about the TLS.

marxin:
Thanks for the changes @davidlattimore. Now it seems to build except I had to ignore unsupported TLSGD -> local exec relocation on AArch64 target.

PR #488, opened 2025-02-27
Implement support for thin archives

### LINKED ISSUE #361: Support thin archives
Currently wild bails when encountering thin archives: https://github.com/davidlattimore/wild/blob/ee2e0fafc81abaa38b61ee17c3940973de7d36ad/libwild/src/file_kind.rs#L55-L56

Thin archives themselves do not contain any data sections. Instead, they reference other archives the linker should read (with mmap perhaps?).

Example project: https://github.com/cisco/openh264

GlowingScrewdriver:
Thin archives are like regular archives except that they store references to files on the filesystem instead of the file contents themselves.

May fix #361

### DIFF
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -229,17 +229,33 @@ impl<'data> ArchiveContent<'data> {
     pub(crate) fn data_range(&self) -> Range<usize> {
         self.data_offset..self.data_offset + self.entry_data.len()
     }
+
+    // Parse the identifier as a reference to extended filenames
+    pub(crate) fn parse_as_thin_reference(&self, filenames: ExtendedFilenames<'data>) -> Result<&'data str> {
+        // TODO: dedupe?
+        let content = self.ident.strip_prefix("/")
+            .with_context(|| format!("Not a thin entry: {}", &self.ident))?;
+        let addr: usize = content.parse()
+            .with_context(|| format!("Invalid offset: {}", content))?;
+
+        if addr >= filenames.data.len() {
+            bail!("Thin entry filename offset ({}) exceeds extended filenames length ({})", addr, filenames.data.len());
+        }
+        let rest = &filenames.data[addr..];
+        let end = memchr::memchr(b'\n', rest)
+            .with_context(|| "Malformed filename")?;
+        if rest[end-1] != b'/' {
+            bail!("Malformed filename");
+        }
+        let res = std::str::from_utf8(&rest[..end-1])
+            .with_context(|| "Invalid UTF-8 in filename")?;
+        Ok(&res)
+    }
 }
 
 impl<'data> Identifier<'data> {
     pub(crate) fn as_slice(&self) -> &'data [u8] {
         let end = memchr::memchr(b'/', self.data).unwrap_or(self.data.len());
-        &self.data[..end]
-    }
-
-    // Get the filename indicated by this identifier
-    pub(crate) fn as_filename(&self) -> &'data [u8] {
-        let end = memchr::memchr(b'\n', self.data).unwrap_or(self.data.len());
         &self.data[..end]
     }
 }
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -236,6 +236,12 @@ impl<'data> Identifier<'data> {
         let end = memchr::memchr(b'/', self.data).unwrap_or(self.data.len());
         &self.data[..end]
     }
+
+    // Get the filename indicated by this identifier
+    pub(crate) fn as_filename(&self) -> &'data [u8] {
+        let end = memchr::memchr(b'\n', self.data).unwrap_or(self.data.len());
+        &self.data[..end]
+    }
 }
 
 impl<'data> Iterator for ArchiveIterator<'data> {
@@ -309,6 +315,9 @@ mod tests {
                     match entry {
                         ArchiveEntry::Regular(content) => {
                             our_entries.push(content);
+                        }
+                        ArchiveEntry::FileReference(_) => {
+                            // TODO: Implement
                         }
                         ArchiveEntry::Ignored => {}
                         ArchiveEntry::Filenames(table) => filenames = Some(table),
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -13,6 +13,7 @@ pub(crate) enum ArchiveEntry<'data> {
     Ignored,
     Regular(ArchiveContent<'data>),
     Filenames(ExtendedFilenames<'data>),
+    Thin(&'data str), // Stores the identifier
 }
 
 #[derive(Clone, Copy)]
@@ -39,7 +40,7 @@ pub(crate) struct ArchiveContent<'data> {
     ident: &'data str,
 
     /// `None` for file references in thin archives
-    pub(crate) entry_data: Option<&'data [u8]>,
+    pub(crate) entry_data: &'data [u8],
 
     /// The offset in the archive at which the data is from.
     pub(crate) data_offset: usize,
@@ -127,19 +128,21 @@ impl<'data> ArchiveIterator<'data> {
             },
         };
 
-        let (entry_data, size) = match ident_kind {
-            IdentifierKind::InlineContent
-            | IdentifierKind::SymbolTable
-            | IdentifierKind::Filenames => {
+        let entry_size = match ident_kind {
+            IdentifierKind::FileReference => {
+                // The size field of a thin reference indicates size of
+                // the references file, not the entry itself
+                0
+            }
+            _ => {
                 if self.data.len() < size {
                     bail!(
                         "Entry size is {size}, but only {} bytes left",
                         self.data.len()
                     );
                 }
-                (Some(&self.data[..size]), size)
+                size
             }
-            IdentifierKind::FileReference => (None, 0),
         };
         let entry = match ident_kind {
             IdentifierKind::SymbolTable => {
@@ -148,17 +151,16 @@ impl<'data> ArchiveIterator<'data> {
                 ArchiveEntry::Ignored
             }
             IdentifierKind::Filenames => ArchiveEntry::Filenames(ExtendedFilenames {
-                data: entry_data.unwrap(),
+                data: &self.data[..entry_size],
             }),
-            IdentifierKind::InlineContent | IdentifierKind::FileReference => {
-                ArchiveEntry::Regular(ArchiveContent {
-                    ident,
-                    entry_data,
-                    data_offset: self.offset,
-                })
-            }
+            IdentifierKind::InlineContent => ArchiveEntry::Regular(ArchiveContent {
+                ident,
+                entry_data: &self.data[..entry_size],
+                data_offset: self.offset,
+            }),
+            IdentifierKind::FileReference => ArchiveEntry::Thin(ident),
         };
-        let size_with_padding = size.next_multiple_of(2).min(self.data.len());
+        let size_with_padding = entry_size.next_multiple_of(2).min(self.data.len());
         self.data = &self.data[size_with_padding..];
         self.offset += size_with_padding;
         Ok(Some(entry))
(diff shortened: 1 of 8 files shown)


GlowingScrewdriver:
At the moment, the test infrastructure is updated to include a test for thin archives, and the linker itself is capable of determining the filenames referenced by the thin archive.

_Note:_ I'm not sure how to fix the merge conflict right now. I've done
```rust
enum InputType {
    // ...
    Archive {thin: bool},
    // ...
}
```
with a couple of conditional switches on the `thin` field to avoid replicating the code that handles archives. But I can't figure out how to direct `strum` to customize `thin`.

davidlattimore:
For the merge conflict, I suspect the easiest way might be to get rid of the bool and have a separate enum variant instead. e.g.

```rust
enum InputType {
    // ...
    Archive,
    ThinArchive,
    // ...
}
```

davidlattimore:
Some thoughts on reading the data in case they're helpful... Reading the data for regular files is done in `InputData::register_input`. I think you may want to reuse some of that code, but might first need to extract part of that method out into a separate function - one that opens the file and returns an `Mmap`.

`archive_splitter.rs` has a struct `InputBytes` which stores the data in a `&[u8]`, but for thin archives you'll have just opened the input file, so you have an owned value, not a borrowed value. Probably the easiest change here would be to change the field `data` to hold an enum with two variants, one a borrowed byte slice and one an `Mmap`.

GlowingScrewdriver:
Alright. Having 2 enum variants in `InputBytes.data` seems like a neat way to do it — I understand that we'd have to pass around ownership of the `Mmap`, and not an  `&[u8]`. Another option might be to have a separate `Some(Mmap)` field in `InputBytes`, while retaining `data: &[u8]`, so that the consumers of `InputBytes` can remain unchanged.

Thanks for the suggestions!

GlowingScrewdriver:
Rust is brittle to work with initially, but keeping the compiler happy teaches one so much about program safety, and just good programming in general. It's good fun.

GlowingScrewdriver:
Okay, Wild can now read and use thin archives! I tested it manually on a thin archive with object files, and the resulting executable works properly. Also, the executables that Wild produces from a normal archive and a thin one are identical byte-for-byte.

The crate's tests aren't passing though. Linker-diff is having some problems. Currently, `section_map::IndexedLayout::new` [reads a thin archive file](https://github.com/davidlattimore/wild/blob/9be16debd4500c8dd3f558b5373c2c0d396d9ea5/linker-diff/src/section_map.rs#L103) and then tries to load it as an ELF file immediately after that.

davidlattimore:
Nice work, this is looking good!

For the problem with linker-diff not knowing how to open the thin-archive entries, here's what I would suggest:

If there's an archive entry associated with the file, check the header of the file to see if it's a thin archive. If it is, then use the identifier on the `ArchiveEntryInfo` to determine what file to get the contents of.

You'll also need to change `LayoutAndFiles::from_base_path` to read all of the object files that have archive entries for thin archives. One change that might make this slightly easier would be if you changed `ArchiveEntryInfo.range` to be `Option<Range<usize>>`. Then the having no range could indicate that the data for the entry isn't in the file.

GlowingScrewdriver:
I'm trying to understand the role of the layout file. What I understand so far is that it captures the archive entries generated during linking, including their backing files and the address ranges in the backing files. And that is what linker-diff uses to figure out what objects to load from which ranges in which files, rather than duplicating the logic in `libwild`.

If that is the case, then I should probably modify `libwild::layout::Layout::layout_data` to emit enough info for thin archive entries, right? ATM, the layout emitted by `wild` is only populated with the archive's path and some section info.

What I'm thinking is that the object file referenced by a thin archive can go in `linker_layout::InputFile.path`. The reason I ended up here is that I'm trying to figure out how not to duplicate the archive-splitting logic from `libwild`, since I noticed that `linker-diff` does not call into `libwild`.

GlowingScrewdriver:
Also, is there a way to pretty-print the layout file? For now I'm resorting to `println!`ing the layout file's contents before they are serialized (using `serde`?).

davidlattimore:
> I'm trying to understand the role of the layout file. What I understand so far is that it captures the archive entries generated during linking, including their backing files and the address ranges in the backing files.

Yes, although not just for archive entries. linker-diff needs to know where each section of each object file got placed in the output file.

> If that is the case, then I should probably modify `libwild::layout::Layout::layout_data` to emit enough info for thin archive entries, right? ATM, the layout emitted by `wild` is only populated with the archive's path and some section info.

I would have thought that the information in the layout file currently would be sufficient. The main extra bit of information that you need is whether the archive is a thin archive or not. You could store that as an extra bool - `is_thin_archive` in the layout, but I don't think it's really necessary, since you can just check the file to see if it starts with the appropriate prefix.

If the entry is from a thin archive, then use the entry identifier as a path, open that path and use the contents of that file instead of trying to use bytes from the archive file.

> What I'm thinking is that the object file referenced by a thin archive can go in `linker_layout::InputFile.path`. The reason I ended up here is that I'm trying to figure out how not to duplicate the archive-splitting logic from `libwild`, since I noticed that `linker-diff` does not call into `libwild`.

I wouldn't think that there would be any need to split or parse thin archives in linker-diff. You already have the identifiers - i.e. the filenames of the relevant entries - from the layout file. So the only thing linker-diff should need to do with the archive file is to check if it's a thin archive by looking at the header.

> Also, is there a way to pretty-print the layout file? For now I'm resorting to `println!`ing the layout file's contents before they are serialized (using `serde`?).

I don't think there is. You could if you wanted add a flag to linker-diff to print the layout information, although I'd imaging for non-trivial programs that it could get quite large.

GlowingScrewdriver:
Is there a particular reason that `archive::ArchiveContent::data_range` is a function? Is there a problem with just storing a `Range<usize>` directly?
It's not getting in the way, but I'm curious.

davidlattimore:
> Is there a particular reason that `archive::ArchiveContent::data_range` is a function? Is there a problem with just storing a `Range<usize>` directly? It's not getting in the way, but I'm curious.

It could be a range, but then what happens if the size of the range is different than the length of `entry_data`? By not duplicating that length information in two fields, such an inconsistency is made impossible.

GlowingScrewdriver:
Alright, I've got the `linker-diff` tests passing. `libwild` tests are also passing, except for `rust-integration` (which I understand is designed to `panic!`?).

Just summing up the TODO's I've left for myself in this PR:
- [ ] Resolve the redundancy between `fn identifier` and `fn parse_as_thin_reference` in `archive::ArchiveContent`. Both of these parse an archive entry identifier to produce a filename. Apart from emitting error messages, the latter will return an error if the identifier contains a filename itself, rather than a reference to the extended filenames field.
- [ ] Verify that file names in the extended filenames really end with `/\n`, so that `archive::Archive::Identifier::as_slice()` can scan for `\n` instead of `/`. The latter causes problems when the extended filenames entries are absolute paths.
- [ ] Include a check on thin archives in `archive::tests::test_ar_consistency` (this will also clean up all the ugly `.unwrap()`s I've left in `archive::tests::check_consistency`)

Please let me know what else needs tweaking.

GlowingScrewdriver:
I've also applied Clippy's and Rustfmt's suggestions, so the CI tests should go off smoothly unlike the [last run](https://github.com/davidlattimore/wild/actions/runs/13581221782).


> libwild tests are also passing, except for rust-integration

Never mind, I figured out the issue. That test is also passing now.

libwild/src/archive_splitter.rs
```diff
@@ -32,16 +44,28 @@ pub fn split_archives<'data>(input_data: &'data InputData) -> Result<Vec<InputBy
                         ArchiveEntry::Ignored => {}
                         ArchiveEntry::Filenames(t) => extended_filenames = Some(t),
                         ArchiveEntry::Regular(archive_entry) => {
+                            let (from, data) = if let Some(entry_data) = archive_entry.entry_data {
```
davidlattimore: Do you think it'd be worthwhile experimenting to see if having `ArchiveEntry::Thin` makes the code simpler?

libwild/src/archive.rs
```diff
@@ -184,15 +223,51 @@ impl<'data> ArchiveContent<'data> {
...
+            std::str::from_utf8(&rest[..end - 1]).with_context(|| "Invalid UTF-8 in filename")?;
+        Ok(res)
     }
 }
 
 impl<'data> Identifier<'data> {
     pub(crate) fn as_slice(&self) -> &'data [u8] {
-        let end = memchr::memchr(b'/', self.data).unwrap_or(self.data.len());
-        &self.data[..end]
+        // TODO: Verify
```
davidlattimore: I think it's fine to assume that the filenames end in '\n', so you can probably remove this TODO/comment. Maybe add a comment about why the `- 1` is there (to ignore the trailing '/')

libwild/src/archive.rs
```diff
@@ -282,15 +357,16 @@ mod tests {
                         our_entries.len()
                     );
                 }
+                // TODO: Run on thin archives too
```
davidlattimore: Given that we have an integration test that uses thin archives, it's fine with me if you wanted to just remove this TODO and rely on the integration test to verify that thin archives work.

libwild/src/archive.rs
```diff
@@ -184,15 +223,51 @@ impl<'data> ArchiveContent<'data> {
...
 
-    pub(crate) fn data_range(&self) -> Range<usize> {
-        self.data_offset..self.data_offset + self.entry_data.len()
+    pub(crate) fn data_range(&self) -> Option<Range<usize>> {
+        self.entry_data
+            .map(|entry_data| self.data_offset..self.data_offset + entry_data.len())
+    }
+
+    // Parse the identifier as a reference to extended filenames
+    pub(crate) fn parse_as_thin_reference(
```
davidlattimore: I think to solve the TODO below, you can probably just get rid of this method. The caller of this method can just do something like:

```rust
let fname = archive_entry.identifier(extended_filenames).as_slice();
let bytes = mmap_file(Path::new(OsStr::from_bytes(fname)), false)?;
```

libwild/src/input_data.rs
```diff
@@ -229,6 +205,37 @@ impl Input {
     }
 }
 
+pub(crate) fn mmap_file(path: &PathBuf, prepopulate_maps: bool) -> Result<Mmap> {
```
davidlattimore: Might be slightly nicer to take a `&Path` here rather than a `&PathBuf`. Requiring a PathBuf means that the caller needs to heap-allocate.

davidlattimore (commented):
Great work!

libwild/src/archive.rs
GlowingScrewdriver: Done

libwild/src/archive.rs
GlowingScrewdriver: Sure

libwild/src/archive.rs
GlowingScrewdriver: Ah, I see. I never realized there was a cheaper way to construct the path other than via `str::from`!

GlowingScrewdriver: Done

libwild/src/input_data.rs
GlowingScrewdriver: Done

libwild/src/archive_splitter.rs
GlowingScrewdriver: Just thinking aloud here: If we were to have an `ArchiveEntry::Thin` alongside `ArchiveEntry::Regular`, that would remove the need for an `enum DataKind` to wrap `Mmap` and `&[u8]`.

On the flip side, we'd be duplicating some struct instantiation code between the arms that {generate, consume} `ArchiveEntry::Regular` and `ArchiveEntry::Thin`. But, then again, we'd have less redundant data since we can introduce an `ArchiveContentReference` (holding just a filename) alongside `ArchiveContent`. And we don't need to make `ArchiveContent::data_range` return `Option`.

EDIT: There may not be a need for the new `ArchiveContentReference`. I can probably just use an `Identifier`

GlowingScrewdriver: Alright, I'll refactor it.

GlowingScrewdriver:
Does `identifier` need to be a method on `ArchiveContent`? It gets awkward when you want to evaluate an identifier stored in an `Identifier` instead of an `ArchiveContent`.

Or, alternatively, we could keep the core logic in a separate global function and call into that from the struct method?

davidlattimore:
> Does `identifier` need to be a method on `ArchiveContent`? It gets awkward when you want to evaluate an identifier stored in an `Identifier` or a `[u8]` instead of an `ArchiveContent`.
> 
> Or, alternatively, we could keep the core logic in a separate global function and call into that from the struct method?

Feel free to move functions around if that makes them easier to reuse for your use-case.

libwild/src/archive_splitter.rs
GlowingScrewdriver: Done. What do you think?

davidlattimore:
Thanks for your work on this!

GlowingScrewdriver:
Thank you for your guidance!

PR #492, opened 2025-02-27
Port to bytesize 2.0 crate

marxin:


marxin:
The default formatting is using IEC (binary): e.g. `4.2 MiB`.

PR #495, opened 2025-02-28
Support RELRO

### LINKED ISSUE #42: Support RELRO
RELRO (relocations, read-only) is a GNU extension that remaps parts of the file as read-only after relocations have been applied. This makes the resulting binary more secure, since those bits can't be changed after startup.

It's probably not essential for using Wild as your linker for day-to-day development, so long as you use a more mature linker for your production builds (which you definitely should). There are a couple of arguments for doing this even for a development-only linker:

* If your program hits undefined behaviour that causes it to try to overwrite one of these sections, it'll crash sooner if they've been remapped read-only.
* If you foolishly write code that depends on these sections being writable, you might not discover until you go and link your binary with a linker that supports making them read-only. This feels a little far-fetched, but still.

inflation:
Fix #42.

One remaining issue:
https://github.com/davidlattimore/wild/blob/main/wild/tests/sources/trivial_asm.s#L12
customizes the flags and entsize of the section, but I don't see any general mechanics for passing it in our codebase. Do we need the ability to pass arbitrary `.section` parameters now?

### DIFF
--- a/libwild/src/output_section_id.rs
+++ b/libwild/src/output_section_id.rs
@@ -111,8 +111,9 @@ pub(crate) const BSS: OutputSectionId = OutputSectionId::regular(10);
 pub(crate) const COMMENT: OutputSectionId = OutputSectionId::regular(11);
 pub(crate) const GCC_EXCEPT_TABLE: OutputSectionId = OutputSectionId::regular(12);
 pub(crate) const NOTE_ABI_TAG: OutputSectionId = OutputSectionId::regular(13);
+pub(crate) const DATA_REL_RO: OutputSectionId = OutputSectionId::regular(14);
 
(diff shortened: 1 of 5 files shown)


davidlattimore:
I think GNU ld has some rule where if all the input sections that feed into a particular output section have some flag set, then it'll set it on the output section. Similar for entry size. We do something like that for custom sections, but not for built-in sections. I don't think us not doing that has any real consequences other than linker-diff complaining about it, so it hasn't thus far been a priority to fix. It should be fine for now to just add some ignore rules, either for that specific test, or even in the default ignore rules in `linker-diff/lib.rs`.

The change looks good. Nice work!

You have a typo in the name of the program segment constant "RELEO" vs "RELRO".

Issue #41, opened 2024-08-18
Support for symbol versions when building shared objects


davidlattimore:
We already support linking against shared objects that define symbol versions and emitting those symbol versions into the resulting binary so that the correct symbol can be selected at runtime.

However we don't as yet support building shared objects that define their own symbol versions.

Or put another way, we support referencing symbol versions, but don't support defining them.

When building a shared object, symbol versions can be defined by a version script. We have limited support for version scripts already, but only the part that allows symbols to be downgraded to locals. See `VersionScript` in `linker_script.rs`. This parsing would need to be extended to support more of the version script syntax. See https://sourceware.org/binutils/docs/ld/VERSION.html

Then version definitions and symbol -> version ID mappings would need to be written to the output shared object.

It'd also be good to extend linker-diff to verify as much of this as possible. i.e. check that the symbols in our output file have the same versions as those output by the reference linker.

davidlattimore:
Awesome! There may be some overlap with #444 which I'm currently working on. I'll try to wrap that work up quickly.

mati865:
No worries, I still have too many other things to do.

mati865:
Just wanted to say how it's going. The code is a pile of smoking mess, but looks like the core functionality is there: https://gist.github.com/mati865/34151eadc57b85c05dbf812e05efd3c5

Funny how much LLD differs from LD. Also, why the heck would it align `.gnu.version_d` to 4 instead of 8?
```
  [ 3] .gnu.version_d    VERDEF           000000000000029c  0000029c
       0000000000000054  0000000000000000   A       6     3     4
```
Dunno if I even want to know at this point.

Linker diff is still basically untouched, I have only enabled checking whether the sections are present.
Once I clean it up a bit, I'll open a draft. So, probably today or tomorrow evening.

mati865:
MVP for this issue has landed in https://github.com/davidlattimore/wild/pull/517. I have created #571 and #572 for things that were not implemented fully.

PR #517, opened 2025-03-03
Support outputting versioned symbols in shared objects

mati865:
Cc #41


mati865:
Likely not worth looking at yet. I will prepare a description of current state, some questions and explanations of my decisions this evening CET/UTC time.

libwild/src/elf_writer.rs
```diff
@@ -2275,7 +2315,88 @@ impl PreludeLayout {
...
+    // TODO: Maybe we can do better here?
+    let mut strings = AHashMap::new();
+
+    for (i, verdef) in verdefs.iter().enumerate() {
+        let verdef_out = table_writer.version_writer.take_verdef()?;
+        // Base version may use (already allocated)
+        let (name, name_offset) = match (soname, soname_offset) {
+            (Some(soname), Some(offset)) if i == 0 => (soname, offset),
+            _ => {
+                let offset = *strings.entry(&verdef.name).or_insert_with(|| {
```
davidlattimore: It looks like you're deduplicating verdef names here at write time, which would presumably mean that either you're deduplicating twice - once in layout and once here - or that you might not be allocating the correct amount of space at layout time.

libwild/src/layout.rs
```diff
@@ -3178,12 +3232,63 @@ impl<'data> EpilogueLayoutState<'data> {
             memory_offsets.increment(part_id::NOTE_GNU_BUILD_ID, build_id_sec_size);
         }
 
+        // TODO: Is it possible avoid this allocation?
```
davidlattimore: It's only a single allocation right? As in it's on the Epilogue, of which there's only a single instance. In that case, I definitely wouldn't worry about it.

libwild/src/elf_writer.rs
mati865: It's for deduplicating versions and their parents. For this version script:
```
                VERS_1.1 {
                    global:
                        foo1;
                    local:
                        old*;
                };

                VERS_1.2 {
                    foo2;
                } VERS_1.1;
```
We get:
```
Version definition section '.gnu.version_d' contains 3 entries:
 Addr: 0x0000000000000848  Offset: 0x00000848  Link: 6 (.dynstr)
  000000: Rev: 1  Flags: BASE  Index: 1  Cnt: 1  Name: symbol-versions.c-verdef-host.wild
  0x001c: Rev: 1  Flags: none  Index: 2  Cnt: 1  Name: VER_1.0
  0x0038: Rev: 1  Flags: none  Index: 3  Cnt: 2  Name: VER_2.0
  0x0054: Parent 1: VER_1.0
```
So there are 3 versions here: base (--soname or filename), VER_1.0 and VER_2.0. Also, there is parent (readelf naming) or dependency (LD docs naming) that shares the name/allocation with the version.

libwild/src/layout.rs
mati865: Yeah, it's not terrible, but I was wondering if some refactors to `VersionScript` structure could make it redundant.

mati865:
First of, I'm not sure about `struct VersionScript`, I had tried few different options, four of them can be seen as the first four commits in the file history: https://github.com/davidlattimore/wild/commits/f2e80df3e20a3ee7d948a06ae541ee9bd008f600/libwild/src/linker_script.rs
I think current one has some benefits, like possibility to easily error on duplicate symbol in version script (any exact symbol can appear only in a one version).

This is mostly working, despite small issues like versioned symbols always being global: `5: 0000000000001a28    17 FUNC    GLOBAL DEFAULT   12 bar_local`.

Some commits like [`7f6e433` (#517)](https://github.com/davidlattimore/wild/pull/517/commits/7f6e433c120859c96d6eaa81b2e3f8d80b5ff3a1) [`2e9f1db` (#517)](https://github.com/davidlattimore/wild/pull/517/commits/2e9f1db6f0df15085949e31070fce9e603cfc933) could be reviewed and landed separately to make this review easier.

libwild/src/elf_writer.rs
mati865: Ah, I see now how it can be confusing. And the reply I made while in a hurry is not really helpful.

I meant to say this map is to avoid duplicating parent version string since that string was already created by the version definition.
Consider the previously mentioned version script. After layout transformation, we get:
```
{
  name: "filename",
  parent: None
}
{
  name: "VERS_1.1",
  parent: None
}
{
  name: "VERS_1.2",
  parent: Some("VERS_1.1")
}
```
All names after the first record (base version) will appear in the strings map exactly once. There is no duplication there, but the parent string is duplicated with regard to the version it refers to. This is where we can improve.

libwild/src/elf_writer.rs
```diff
@@ -2456,7 +2588,13 @@ fn write_dynamic_symbol_definitions(
                 // now we just set them all to the global version.
                 if let Some(versym) = table_writer.version_writer.versym.as_mut() {
                     if let Some(version_out) = crate::slice::take_first_mut(versym) {
-                        version_out.0.set(LittleEndian, object::elf::VER_NDX_GLOBAL);
+                        let version = layout
+                            .symbol_db
+                            .version_script
+                            // TODO: can we avoid rehashing?
```
mati865: This is probably the biggest offender in this PR, but I have no idea how to solve it.

libwild/src/elf_writer.rs
```diff
@@ -2456,7 +2588,13 @@ fn write_dynamic_symbol_definitions(
                 // now we just set them all to the global version.
```
davidlattimore: While I'm thinking of it, this comment can probably be updated.

libwild/src/elf_writer.rs
davidlattimore: I guess if nothing else, we could make sure that we don't hash the symbol name here if the version script is empty, that way the cost is only paid if there's actually a version script. It's possible that the compiler might be able to do this for us, although I wouldn't be surprised if it doesn't.

Another possibility is that we could move all the symbol hashes into a `Vec<u64>`. That would allow us to look up the hash for an arbitrary SymbolId.

libwild/src/elf_writer.rs
davidlattimore: I think I understand. So the hashmap is for lookup purposes rather than deduplication purposes. The names should already be unique, but we might reference the one parent name multiple times, and during layout we'll allocate the space for each parent just once, but we don't need the hashmap there because we're not looking at the child-parent relationships. Does that sound right?

davidlattimore (commented):
I think I still need to review this some more, but wanted to send the comments that I had so far

davidlattimore:
Oh, I forgot to say, splitting out those changes that you mentioned sounds good. If you've got those or any other commits that you think are ready and make sense by themselves, then the sooner they're submitted, the less chances for merge conflicts.

libwild/src/elf_writer.rs
mati865: Sounds right. We could do this mapping in layout and store versions as a vector of strings in the state, similarly to verneed. I'll see how it goes.

wild/tests/integration_tests.rs
```diff
@@ -636,6 +636,14 @@ fn parse_configs(src_filename: &Path) -> Result<Vec<Config>> {
                 "RequiresClangWithTlsDesc" => {
                     config.requires_clang_with_tlsdesc = arg.to_lowercase().parse()?;
                 }
+                // TODO: This is but a poor hack, think of a better way
```
davidlattimore: Another option would be to add the version script to the `LinkArgs` in the `.c` file. For that to work, we'd either need some kind of path substitution e.g. `//#LinkArgs:--version-script=$TEST_SRCS/symbol-version-script` or we could change into the sources directory when doing all the builds, in which case just `//#LinkArgs:--version-script=symbol-version-script` would work.

libwild/src/elf_writer.rs
```diff
@@ -2275,7 +2315,88 @@ impl PreludeLayout {
...
+        }
+    }
+
+    Ok(())
+}
+
+fn write_epilogue_dynamic_entries(
+    layout: &Layout,
+    table_writer: &mut TableWriter,
+    // TODO: Do it better?
```
davidlattimore: Two options occur to me. One, would be to put it as a field in a struct, so that we have a more distinct type than `Option<u32>`. e.g.

```rust
struct EpilogueOffsets {
   /// The offset of the shared object name in .dynsym.
   soname: Option<u32>,
}
```

The other option, which is probably a bit more work would be to figure out the offset during layout and store it somewhere (e.g. on `EpilogueLayout` or `Layout`).

mati865:
Thank you for the review. Unfortunately, I'm short on time this week and couldn't manage to get them all processed.
I've got another idea how to improve it and pushed it just now as "refactor". Hopefully this doesn't get too convoluted at this point.

> Oh, I forgot to say, splitting out those changes that you mentioned sounds good. 

Split out one of them. The other one would need a bit more changes because it turns out changes in this PR enable that refactor.

wild/tests/integration_tests.rs
mati865: I've tried with that initially, but there is a problem the linkers are not rerun if only the version script changes. The current approach also suffers from it, but it shouldn't be too hard to invalidate the results.
I don't know if this is a big deal, though.

libwild/src/elf_writer.rs
mati865: I was considering the first option, but hoped for a some kind of silver bullet. I'll probably go for it when I'm back.

> The other option, which is probably a bit more work would be to figure out the offset during layout and store it somewhere (e.g. on EpilogueLayout or Layout).

Wouldn't figuring out the offset without actually writing it to the table be fragile? If a new record is added or records are reordered, that would break.

wild/tests/integration_tests.rs
davidlattimore: Oh, that's a good point about not detecting file changes. In that case, what you've got is perhaps the best option.

libwild/src/elf_writer.rs
davidlattimore: > Wouldn't figuring out the offset without actually writing it to the table be fragile? If a new record is added or records are reordered, that would break.

That's true, although it's kind of similar to many other things in the linker. For example, we figure out where we're putting sections and symbols before we write them. One mitigation would be to check that the offset we wrote the thing at was actually the offset that picked during layout. Another is to have linker-diff check it. But I guess things like sections and symbols are different in that we have to figure out their addresses before we start writing them, because we process relocations as we write and the relocations could reference sections that we haven't written yet. The SO name isn't like that, so only learning the offset at write time is OK. Anyway, the first option is fine.

davidlattimore:
> Thank you for the review. Unfortunately, I'm short on time this week and couldn't manage to get them all processed. I've got another idea how to improve it and pushed it just now as "refactor". Hopefully this doesn't get too convoluted at this point.

There's no hurry on the PR, so whenever you've got time again.

> > Oh, I forgot to say, splitting out those changes that you mentioned sounds good.
> 
> Split out one of them. The other one would need a bit more changes because it turns out changes in this PR enable that refactor.

No problem. If it would be extra effort to split the other one out, then I'd say just keep it all in this PR.

libwild/src/elf_writer.rs
mati865: Seems like it no longer does anything if there are no versions defined since your change in: https://github.com/davidlattimore/wild/commit/de06a0850e9c62b80c8582c53dd77bd75436c0a3#diff-24ad70f02b754b0cbf057979b0af5dbcb2657573587351469a4382942029bb4eR753

> Another possibility is that we could move all the symbol hashes into a Vec<u64>. That would allow us to look up the hash for an arbitrary SymbolId.

I'm not sure what you mean. Would you like to store them during symbol_db/layout handling?

mati865:
Rebased because of the conflicts and added commits `Add verdefs to asm_diff` and `linker-diff: compare gnu_version_d section`.
This only compares `version_d` right now, but I'll also add comparison for each versioned symbol, so like `bar@VER_2.0`.

libwild/src/elf_writer.rs
davidlattimore: > > Another possibility is that we could move all the symbol hashes into a Vec. That would allow us to look up the hash for an arbitrary SymbolId.
> 
> I'm not sure what you mean. Would you like to store them during symbol_db/layout handling?

That's what I was thinking, although I suspect that any such change should be left until later, if done at all. I did some experiments the other day with using hashbrown's HashTable instead of a HashMap for storing the symbol name to symbol ID mapping. That meant that I could store the symbol hashes separately. It didn't seem like a win for performance, so I discarded that experiment.

Performance wise, I'd say that so long as this PR doesn't regress performance on stuff that we already support, that it's fine. If performance for emitting versioned symbols has room for improvement, then that's something that fine to leave until later.

mati865:
Apparently, for Clang, which doesn't use `.gnu.version_d` and only `.gnu.version_r` this is marginally, albeit reproducibly slower:
```
❯ powerprofilesctl launch -p performance poop './run-with /tmp/clang/wild-main --no-fork' './run-with /tmp/clang/wild-versions-pr --no-fork'
Benchmark 1 (30 runs): ./run-with /tmp/clang/wild-main --no-fork
  wall_time           170ms ± 3.28ms     161ms …  177ms          1 ( 3%)        0%
  peak_rss            829MB ± 7.08MB     815MB …  840MB          0 ( 0%)        0%
  cpu_cycles         6.54G  ±  119M     6.34G  … 6.80G           0 ( 0%)        0%
  instructions       5.22G  ± 19.1M     5.19G  … 5.26G           0 ( 0%)        0%
  cache_references   90.6M  ±  627K     89.5M  … 91.9M           0 ( 0%)        0%
  cache_misses       24.5M  ±  193K     24.1M  … 25.1M           1 ( 3%)        0%
  branch_misses      9.06M  ± 57.2K     8.99M  … 9.24M           3 (10%)        0%
Benchmark 2 (30 runs): ./run-with /tmp/clang/wild-versions-pr --no-fork
  wall_time           172ms ± 4.59ms     165ms …  187ms          1 ( 3%)          +  1.4% ±  1.2%
  peak_rss            829MB ± 9.10MB     816MB …  851MB          0 ( 0%)          +  0.0% ±  0.5%
  cpu_cycles         6.74G  ± 92.2M     6.53G  … 6.96G           0 ( 0%)        💩+  3.1% ±  0.8%
  instructions       5.24G  ± 16.6M     5.21G  … 5.27G           0 ( 0%)          +  0.3% ±  0.2%
  cache_references   90.9M  ±  587K     89.7M  … 91.8M           0 ( 0%)          +  0.3% ±  0.3%
  cache_misses       24.7M  ±  197K     24.4M  … 25.1M           0 ( 0%)          +  1.0% ±  0.4%
  branch_misses      9.10M  ± 35.4K     9.04M  … 9.20M           0 ( 0%)          +  0.5% ±  0.3%
```
Reducing the number of threads to 16 (without SMT) nullifies any difference.

davidlattimore:
That's a small enough difference that we can probably accept it. Any idea why it might have slowed things down? I tried benchmarking on my laptop, but I couldn't see any slowdown - then again I've only got 4 cores (8 hyperthreads). That sort of difference could even just be due to code moving around.

mati865:
Unfortunately, I'm not that skilled in performance tuning to have a clear answer here.
The only idea I have so far is the increase in structures sizes. Maybe with more threads, lower level CPU's caches start to be insufficient?
Perhaps diffing main and this pr with cachegrind would give some insight.

> I tried benchmarking on my laptop, but I couldn't see any slowdown - then again I've only got 4 cores (8 hyperthreads).

Good to know, My CPU is a bit unusual because each physical core has its own L1 and L2 caches, but L3 is split between 2 CCX (8 physical cores in this case). Your CPU is probably more representative.

> That sort of difference could even just be due to code moving around.

Yeah, that's also very likely.

davidlattimore:
What else is there that you'd like to do on this PR before we land it? From my perspective, if you say it's ready to go in, I'd be happy to merge it as-is.

BTW, a PR I'm working on has a small merge conflict with this one. It's pretty trivial to resolve though - we just both changed adjacent arguments to `SymbolDb::build`.

mati865:
I want to rebase on top of https://github.com/davidlattimore/wild/commit/15e397cf1cceff15cfad1d7d8ca334a39d24e0ff
It enables a slight improvement.

The only bug I'm aware of is https://github.com/davidlattimore/wild/pull/517/files#diff-45ed8463683a5f815de1cb422b26180bc3a144c767ff85cf77b121a411d19985R41
That symbol should be local but is global, even though https://github.com/davidlattimore/wild/blob/804799462243d39201fabd2f15b23ec1dddf1773/libwild/src/symbol_db.rs#L842 works properly. Inside `elf_writer.rs`, `sym.is_local()` returns false, I wasn't successful in finding the reason yet.

There is one more improvement to linker-diff that I already decided to postpone. I'll write about it in the issue once this goes in.

> BTW, a PR I'm working on has a small merge conflict with this one. It's pretty trivial to resolve though - we just both changed adjacent arguments to `SymbolDb::build`.

No worries, I can rebase this PR after it lands.

Regarding cachegrind diff, I performed it few hours ago and it gave interesting results.
With the default 16 codegen units for release builds, this PR affected code generation of iterators. I haven't looked deeply, but at the very least inlining isn't as good. This is going to change back and forth with the development, so no worries there.

When I rebuilt with a single codegen unit, `poop` showed smaller cycles regression and other metrics were within the margin of error. However, the cachegrind diff shows something I didn't really expect. Somehow `matches` function became faster when linking Clang, but `steal` from crossbeam was used more (I ran wild with all threads at all time) and that throws off cycles count.

Overall, I see no issues with the performance here when no version script is used, and I don't expect regression when a "simple" version script (like the one created by Rust) is used.

davidlattimore:
> That symbol should be local but is global...

I just had a quick look. It looks like `ValueFlags::DOWNGRADE_TO_LOCAL` prevents the symbol from being dynamically exported  - i.e. it makes it not get listed in `.dynsym`, and also allows the GOT to be bypassed for references to that symbol, but doesn't currently affect whether the symbol is made a local in `.symtab`. It looks like for the output `.symtab`, visibility is currently only determined by the visibility in the input file. The visibility in `.symtab` for an executable / shared object shouldn't affect runtime behaviour - at most it would affect debugger behaviour. That said, it's nice to match the other linkers as closely as we can. I'm pretty sure that the difference in visibility would be observable without your PR, so don't feel that you have to fix it before this can land.

Thank you for the very thorough performance investigation! Sounds like we're good on that front.

mati865:
Ah, that explains it.

davidlattimore (approved):
Thanks for all your work on this!

mati865:
Thank you for bearing with me during this lengthy process.

PR #512, opened 2025-03-02
Build symbol name hashmap in parallel

davidlattimore:

PR #519, opened 2025-03-03
Move alternative symbol resolution from resolution.rs to symbol_db.rs

davidlattimore:
This seems like a better place for this code and allows some of the fields on SymbolDb to be made private.


davidlattimore:
I made this change while I was doing #512, but then decided that I should separate it out to make the diffs clearer.

PR #520, opened 2025-03-03
integration tests: Make ExpectError imply no running and diffing

davidlattimore:


davidlattimore:
Doing this uncovered that one of the test in `undefined_symbols.c` wasn't erroring, but also wasn't being diffed. When diffing it, one of the diffs was interesting not note, which is that GNU ld (and lld it turns out) set the entry point even when building a shared object. I'm not sure what the point of that is, so I just ignored it for now.

wild/tests/sources/undefined_symbols.c
```diff
@@ -1,9 +1,10 @@
 //#AbstractConfig:default
```
mati865, line 1: Nit: this redundant now

Issue #458, opened 2025-02-21
Refactor args.rs to support --help


davidlattimore:
We don't currently support --help, but it'd be nice to.

We could just write a separate help document and then try to make sure that we keep it in sync with the actual argument parsing, however I think it would be inevitable that the two would diverge.

See also https://github.com/davidlattimore/wild/discussions/302 for discussion as to why we can't use something like clap.

EDIT: The suggestion here is superseded by the suggestion to use bpaf instead, so be sure to read the whole issue. Some of what's written here is still relevant though - e.g. don't copy help text. Also, still a good idea to do an initial PR that doesn't add help, then add help texts in subsequent PR(s).

My current thought on how we might approach this is to run the same function when printing help as we run when parsing, but just pass a different argument. Currently the argument passed to `args::parse` is an iterator over the arguments. We could replace this with something that implements our own trait, of which we'd have two implementations, one for parsing and one for printing help.

To provide a concrete example, let's take a snippet of our argument parsing code:

```rust
       } else if long_arg_eq("static") || long_arg_eq("Bstatic") {
            modifier_stack.last_mut().unwrap().allow_shared = false;
        } else if long_arg_eq("Bdynamic") {
            modifier_stack.last_mut().unwrap().allow_shared = true;
        } else if arg == "-o" {
            args.output = input
                .next()
                .map(|a| Arc::from(Path::new(a.as_ref())))
                .context("Missing argument to -o")?;
       }
```

We might rewrite this something like:

```rust
parser
   .declare()
   .long("static")
   .long("Bstatic")
   .help("Disallow linking of shared libraries")
   .execute(|| {
     modifier_stack.last_mut().unwrap().allow_shared = false;
  });

parser
   .declare()
   .long("Bdynamic")
   .help("Allow linking of shared libraries")
   .execute(|| {
      modifier_stack.last_mut().unwrap().allow_shared = true;
   });

parser
   .declare_with_param()
   .short("-o")
   .help("Set the output filename")
   .execute(|a| {
      args.output = Arc::from(Path::new(a.as_ref()));
    });
```

So the argument parser trait would likely have a couple of associated types that get returned by the `declare` and `declare_with_param` methods. That will mean that we'll likely have 3 traits in total with each having two implementations. It'd also be possible to do it all with one trait and no associated types, albeit with slightly more runtime checking.

Some arguments might need to return errors from their closure. For that, we could have a `try_execute` that takes a closure returning a `Result` and which itself returns a `Result`.

[...]

PR #527, opened 2025-03-05
Use bpaf for argument parsing

yerke:
Use bpaf crate for argument parsing. Currently this PR is for discussion purposes only.

Relates to https://github.com/davidlattimore/wild/issues/458

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -169,6 +169,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bpaf"
+version = "0.9.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd343f000c4472dc4557e093f314c5210375d3e9349ad866fee0ddb7f4ed10bf"
+
+[[package]]
 name = "bumpalo"
 version = "3.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -649,6 +655,7 @@ dependencies = [
  "atomic-take",
  "bitflags",
  "blake3",
+ "bpaf",
  "bumpalo-herd",
  "bytemuck",
  "bytesize",
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -11,6 +11,7 @@ edition.workspace = true
 ahash = { version = "0.8.11", default-features = false, features = ["std"] }
 anyhow = "1.0.97"
 bitflags = "2.9.0"
+bpaf = "0.9"
 bytemuck = { version = "1.22.0", features = ["derive"] }
 crossbeam-queue = "0.3.12"
 crossbeam-utils = "0.8.21"
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -579,32 +579,51 @@ fn output() -> impl Parser<Arc<PathBuf>> {
         .map(Arc::new)
 }
 
-fn long_arg_flag(name: &'static str) -> impl Parser<bool> {
-    let long = bpaf::long(name)
-        .switch();
-
-    let single_dash = bpaf::any::<String, _, _>("", move |s| {
+fn single_dash_flag(name: &'static str) -> impl Parser<bool> {
+    bpaf::any::<String, _, _>("", move |s| {
         if let Some(rest) = s.strip_prefix('-') {
             (rest == name).then_some(true)
         } else {
             None
         }
     })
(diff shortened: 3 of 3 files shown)


yerke:
@davidlattimore I spent some time reading bpaf docs and tutorials. I think I understand how @pacak intended for it to be used. But that vision is different from what you described in https://github.com/davidlattimore/wild/issues/458, where you propose to manually consume arguments.
Can you please explain how you want `execute()` part to be implemented from your examples? I am not sure I figured a way to do it with bpaf.
```
parser
   .declare()
   .long("static")
   .long("Bstatic")
   .help("Disallow linking of shared libraries")
   .execute(|| {
     modifier_stack.last_mut().unwrap().allow_shared = false;
  });
```

davidlattimore:
> I spent some time reading bpaf docs and tutorials. I think I understand how @pacak intended for it to be used. But that vision is different from what you described in #458, where you propose to manually consume arguments.

Yep, the initial common on the issue was one design that was not using any 3rd party argument parsing libraries. The later comments should supercede that design. If using bpaf turns out to not work so well for some reason, then we might want to fall back to the original design, but otherwise you can disregard it.

> Another question: is it a valid combination to parameters to provide both `--gc-sections` and `--no-gc-sections`? They contradict one another, but the current parser will consume the last one and be happy.

Yes, later arguments should override earlier ones. If unsure, you can try doing the same thing with GNU ld. e.g, the following command completes without error or warning:

```sh
gcc -Wl,-gc-sections -Wl,-no-gc-sections main.c
```

lu-zero:
Do we have a corpus to use to make tests?

davidlattimore:
> Do we have a corpus to use to make tests?

I find corpus based testing tends to result in lots of redundant testing. So they're good for finding problems, at which point you write a proper, minimal test for that bug, but I wouldn't want to be running on a large test corpus with every PR for example. But I guess given that you're commenting on this PR, you're talking about specifically for argument parsing? There's some existing tests, although we could probably benefit from a few more. But I'd say the argument for hand-writing them would be even stronger there.

lu-zero:
> I find corpus based testing tends to result in lots of redundant testing. So they're good for finding problems, at which point you write a proper, minimal test for that bug, but I wouldn't want to be running on a large test corpus with every PR for example.

Yes :)

> But I guess given that you're commenting on this PR, you're talking about specifically for argument parsing? There's some existing tests, although we could probably benefit from a few more. But I'd say the argument for hand-writing them would be even stronger there.

Recently I got a request to minimize/normalize link lines in cargo-c so I was looking exactly at those corner cases to explain why collecting the arguments in a set may lead to wrong results.

lapla-cogito:
FYI: I proposed a different approach. The choice of which approach to use remains open for consideration. #1029

lu-zero:
It is a bit reinventing the wheel, but given that this PR is on hold since a long time...

davidlattimore:
We can still potentially switch to bpaf in future

lapla-cogito:
While it may initially appear that we're reinventing the wheel, there are several subtle requirements that existing argument parser frameworks like clap or bpaf cannot fulfill—primarily to maintain compatibility with existing linkers such as GNU ld.
For example, we'd like to support parsing long options even when they use a single hyphen (which would require using `any` parser for almost all options in bpaf, or employing workarounds like `.alias()` in clap). We also need to handle options with side effects, such as those that manipulate a stack for storing flag states (like `--push-state` and `--pop-state`).

davidlattimore:
There are ways to meet those requirements with bpaf. See comments by pacak on #458. I guess the question then would be whether the code to do these with bpaf is more or less readable than the without bpaf.

lu-zero:
Given the helper shown in #458 I would expect bpaf to be much more straightforward. If @yerke is busy I can try to give it a go myself.

davidlattimore:
> Given the helper shown in #458 I would expect bpaf to be much more straightforward. If @yerke is busy I can try to give it a go myself.

By all means. That'd be good to see what it'd look like with bpaf. We wouldn't want to use an external crate for argument parsing just for the sake of it though, so we'd only want to merge it if it gave us some benefit like better error messages or substantially simpler code.

marxin:
Given the age of the PR and the fact it conflicts with the `main`, can we please close the PR? Similarly to @davidlattimore, l am open to the integration of the bpaf in the future. Thanks.

PR #548, opened 2025-03-09
Set O_CLOEXEC on our output file

davidlattimore:
This means that when libwild is used in a multithreaded environment that is spawning subprocesses at the same time as we're writing our output file, that the inherited file lock is held only until exec is called rather than for the entire execution time of the subprocess.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -343,6 +343,14 @@ impl SizedOutput {
         should_write_trace: bool,
     ) -> Result<SizedOutput> {
         let mut open_options = std::fs::OpenOptions::new();
+
+        // If another thread spawns a subprocess while we have this file open, we don't want the
+        // subprocess to inherit our file descriptor. This unfortunately doesn't prevent that, since
+        // unless and until the subprocess calls exec, it will inherit the file descriptor. However,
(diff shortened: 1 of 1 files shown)


libwild/src/elf_writer.rs
```diff
@@ -343,6 +343,14 @@ impl SizedOutput {
         should_write_trace: bool,
     ) -> Result<SizedOutput> {
         let mut open_options = std::fs::OpenOptions::new();
+
+        // If another thread spawns a subprocess while we have this file open, we don't want the
+        // subprocess to inherit our file descriptor. This unfortunately doesn't prevent that, since
+        // unless and until the subprocess calls exec, it will inherit the file descriptor. However,
+        // assuming it eventually calls exec, this at least means that it inherits the file
+        // descriptor for less time. i.e. this doesn't really fix anything, but makes problems less bad.
+        std::os::unix::fs::OpenOptionsExt::custom_flags(&mut open_options, libc::O_CLOEXEC);
```
bjorn3, line 352: CLOEXEC should already be set by default by libstd for all files that are opened. No need to set it explicitly.

davidlattimore: Thanks! I'll remove it.

PR #565, opened 2025-03-15
Verify lld can do: rel.extra-opt.R_AARCH64_TLSIE_ADR_GOTTPREL_PAGE21.…

marxin:
…MovzXnLsl16

Once I added `lld`, then the tests passed and so I assume it means `linker-diff` was able to find the same relaxation for `ldd`. Am I right?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,6 @@ jobs:
           - ubuntu-24.04
           - ubuntu-24.04-arm
         container:
-          - 'ubuntu:22.04'
           - 'ubuntu:24.04'
           - 'ubuntu:24.10'
           - 'ubuntu:25.04'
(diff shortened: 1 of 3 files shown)


marxin:
Funny, seems older Ubuntu 22.04 lld was not capable of the relaxation. How to exclude it cleverly?

davidlattimore:
I guess the easiest thing to do would be to just say that we need a newer version of LLD when running tests, at least on aarch64, then skip aarch64 ubuntu:22:04 in CI

davidlattimore:
> Once I added `lld`, then the tests passed and so I assume it means `linker-diff` was able to find the same relaxation for `ldd`. Am I right?

That sounds right. linker-diff will accept what Wild does so long as it matches at least one of the reference linkers

PR #568, opened 2025-03-16
CI: unify used names for Ubuntu containers

marxin:
The tags 'rolling' and 'latest' are somewhat moving targets (in case of `25.05` will even point to the same version), thus I would like to use explicit numbered release versions.

Link: https://hub.docker.com/_/ubuntu


marxin:
And let's include `25.04`, which seems to work fine!

PR #570, opened 2025-03-16
Thin archives: handle relative paths correctly

### LINKED ISSUE #361: Support thin archives

GlowingScrewdriver:
Paths referenced from a thin archive are relative to the parent directory of the archive (unless they are absolute). This PR corrects `Wild` to behave this way.

Thanks @mati865 for pointing out the error.

While fixing this, another issue was encountered, detailed [here](https://github.com/davidlattimore/wild/issues/361#issuecomment-2727250152). TD;DR: entries with identifiers like
```
/0             /
```
(note the trailing `/`) need to be handled.

May fix #361 (again)


GlowingScrewdriver:
At the time of writing this comment, this branch can link [openh264](https://github.com/cisco/openh264). The solution is, however, a bit of a [hack](https://github.com/davidlattimore/wild/pull/570/files#diff-6768b6b2a35926708985f0084738a5efef6c48b72ced3a2b349f51cae444cf4c).

libwild/src/archive.rs
```diff
@@ -218,7 +218,9 @@ fn evaluate_identifier<'data>(
 ) -> Identifier<'data> {
     if let Some(filenames) = extended_filenames {
         if let Some(rest) = ident.strip_prefix('/') {
-            if let Ok(offset) = rest.parse() {
+            // TODO: What is this for exactly?
```
davidlattimore: Probably no need for a TODO here. I'd suggest just documenting the behaviour. Something like `GNU ar puts a trailing '/' as the last byte of the identifier, but only if the full filename is exactly 15 bytes long - e.g. '/48            /' if the filename is 15 character long, or '/48             ' if it's any other length.`

GlowingScrewdriver: Okay. I'll just use this

GlowingScrewdriver: Done

GlowingScrewdriver:
I'm wondering how we can test this. What I can think of is to run `wild actual/input.a` in a file layout like this:
```
actual/
    input.o
    input.a
```

Is this a good idea? Where does such a test belong?

libwild/src/archive.rs
mati865: According to GlowingScrewdriver's investigation, a slash is added when a name is at least 15 characters long, not just exactly 15 characters long.

GlowingScrewdriver: From what I [documented on the issue](https://github.com/davidlattimore/wild/issues/361#issuecomment-2727342388), it happens with exactly 15 characters.

In the issue comment linked above, see the last code block (the one demonstrating the creation of `many_weird.a`).
Notice the last 3 entries, with filenames of length 14, 15 and 16.

mati865: Ah, sorry. I had misunderstood your comment.

PR #564, opened 2025-03-15
Support `rel.match_failed.R_AARCH64_TLSDESC_LD64_LO12`

marxin:
After my change, `linker-diff` can properly find the corresponding parts of `R_AARCH64_TLSLE_MOVW_TPREL_G1`, `R_AARCH64_TLSLE_MOVW_TPREL_G0_NC`, but seems the symbol address look up is off:

```
$ WILD_TEST_CROSS=aarch64 cargo t tls_variant
...
rel.R_AARCH64_NONE.R_AARCH64_TLSLE_MOVW_TPREL_G1
  `/home/marxin/Programming/wild/wild/tests/build/tls-variant-1.gcc-aarch64-aarch64-c672ffb113a4c0aa.o` .text foo
  ORIG 0x000008: [ 00 00 00 90 ] adrp	x0, 0x0
                   ^^^^^^^^^^^ R_AARCH64_TLSDESC_ADR_PAGE21 
  ORIG 0x00000c: [ 01 00 40 f9 ] ldr	x1, [x0]
                   ^^^^^ R_AARCH64_TLSDESC_LD64_LO12 
  ORIG 0x000010: [ 00 00 00 91 ] add	x0, x0, #0x0
                   ^^^^^ R_AARCH64_TLSDESC_ADD_LO12 
  ORIG 0x000014: [ 20 00 3f d6 ] blr	x1
                   ^ R_AARCH64_TLSDESC_CALL 
       global_tls1
  wild 0x41122c: [ 1f 20 03 d5 ] nop
                   ^ R_AARCH64_NONE ReplaceWithNop 
  wild 0x411230: [ 1f 20 03 d5 ] nop
                   ^ R_AARCH64_NONE ReplaceWithNop 
  wild 0x411234: [ 00 00 a0 d2 ] movz	x0, #0x0, lsl #16
                   ^^^^^ R_AARCH64_TLSLE_MOVW_TPREL_G1 MovzX0Lsl16 
  wild 0x411238: [ 00 02 80 f2 ] movk	x0, #0x10
                   ^^^^^ R_AARCH64_TLSLE_MOVW_TPREL_G0_NC MovkX0 
       0x7940010 (symbol is too far away)
  wild TRACE: relaxation applied relaxation=ReplaceWithNop, value_flags=ADDRESS | CAN_BYPASS_GOT,
  wild TRACE: resolution_flags=DIRECT | GOT_TLS_MODULE, rel_kind=None,
[...]

PR #586, opened 2025-03-21
Implement RelocationKind::Relative AdrpToAdr relaxation

marxin:
This is an attempt to implement one of the possible `AdrpToAdr` relaxations as mentioned here:
https://github.com/ARM-software/abi-aa/blob/main/aaelf64/aaelf64.rst#id56

However, it seems to me there are many cases when even `bfd` can't make the relaxation :/
@davidlattimore Do you read it the same as me?

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -109,6 +109,7 @@ impl crate::arch::Relaxation for Relaxation {
         output_kind: crate::args::OutputKind,
         section_flags: linker_utils::elf::SectionFlags,
         non_zero_address: bool,
+        relative_value: Option<i64>,
     ) -> Option<Self>
     where
         Self: std::marker::Sized,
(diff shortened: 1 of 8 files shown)


davidlattimore:
For some reason, the link to the spec that you gave doesn't seem to scroll to the right spot in the document (tested in both Chrome and Firefox). However, I did eventually find where I needed to be looking at.

It says:

> The following TLS relocations may be optimized if the symbol is not a pre-emptable definition and the TLS offset fits in 16 bits

Your change doesn't seem to take `can_bypass_got` (aka "not a pre-emptable definition") into account. Perhaps that could be the reason for the difference?

marxin:
> Your change doesn't seem to take `can_bypass_got` (aka "not a pre-emptable definition") into account. Perhaps that could be the reason for the difference?

It's a kind of relaxation where we end up with the equal address calculation (adrp + add replaced with adr).

Right now, I have ended up with a working solution (at run-time):

```patch
 0000000000011d78 <_Z6char_dv>:
-   11d78:	90ffff80 	adrp	x0, 1000 <__abi_tag+0x580>
-   11d7c:	910f2000 	add	x0, x0, #0x3c8
+   11d78:	10f7b280 	adr	x0, 13c8 <_ZNSt8__detail30__integer_to_chars_is_unsignedIyEE+0x24>
+   11d7c:	d503201f 	nop
    11d80:	d65f03c0 	ret
```

However, I struggled with the linker-diff, where I used the approach where the next relocation (`R_AARCH64_ADD_ABS_LO12_NC`) was supposed to be skipped (`RelocationModifier::SkipNextRelocation`).

@davidlattimore Can you help me with what needs to be adjusted in `possible_relaxations_do` in order to make linker-diff happy? Plus, I am not convinced that the additional complexity is worth the relaxation. What do you think?

davidlattimore:
First, although it wasn't what you asked, I was curious about the x86-64 test failures. It seems that's due to moving where `place` is computed in `elf_writer.rs`. Moving it can cause its value to be different because it depends on `offset_in_section`, which on x86-64 can be changed by the relaxation.

I was also thinking about the next relocation... when you check if the relaxation can be applied, you check if the next instruction is an add instruction, but not that it has a relocation for the same symbol. I've seen the compiler interleave some relocations for multiple different symbols. Fix that would add more complexity, since we'd need to pass in some way for the relaxation computation code to peek at the upcoming relocations. I'm OK with that, especially if it's something that we need or might need for other relaxations, but can understand if it's not something you want to tackle right now.

The linker-diff failure is, I think, because SkipNextRelocation currently has no effect within a relocation group. The bit of code that checks for SkipNextRelocation is in `compare_sections`. `compare_sections` calls `resolve_group_traced`, which calls `determine_relaxations`, which is what iterates through the relocations in the group. So, I think, it'd be possible to change `determine_relaxations` to check `self.next_modifier` and skip the relocation.

marxin:
> I've seen the compiler interleave some relocations for multiple different symbols.

Good catch!

Given the fact, this change needs some more refactoring, and changes to the linker-diff, I tend to close it for now and open an Issue where it's going to be mentioned and linked with #564. I think this relaxation is not fundamental and we can live without it :)

PR #591, opened 2025-03-23
better error reporting for duplicate strong symbols

### LINKED ISSUE #521: Report error when there are duplicate definitions of a strong symbol
Error reporting is something that we've neglected a bit. At the moment, we don't report an error when there are multiple definitions of a symbol.

There are a few nuances to multiple-definition rules:

* It's only a problem if there's more than one strong definition. Multiple weak definitions, optionally with one strong definition is fine.
* It's only a problem if the object containing the symbol is loaded. e.g. it's fine to have strong definitions in archive entries that duplicate other symbols and so long as those archive entries aren't loaded, it's OK.

Selecting from multiple symbol definitions is handled in `resolve_alternative_symbol_definitions`, so that would be the place to start looking.

If there are multiple symbols that have duplicate definitions, it might be good to report all of them, although probably not essential.

We should have a test for this. See `undefined_symbols.c` for a similar kind of test to use as a reference.

lapla-cogito:
close #521 

The current implementation lacks proper error reporting logic for cases where multiple duplicate strong symbols exist in the link target.

### DIFF
--- a/libwild/src/symbol_db.rs
+++ b/libwild/src/symbol_db.rs
@@ -28,6 +28,7 @@ use ahash::HashMap;
 use ahash::RandomState;
 use anyhow::Context;
 use anyhow::bail;
+use crossbeam_queue::SegQueue;
 use itertools::Itertools;
 use object::LittleEndian;
 use object::read::elf::Sym as _;
@@ -37,6 +38,7 @@ use rayon::iter::IntoParallelRefMutIterator;
 use rayon::iter::ParallelIterator;
 use std::collections::hash_map;
 use std::fmt::Display;
+use std::fmt::Write;
 use std::mem::replace;
 use std::mem::take;
 
@@ -596,25 +598,27 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
     resolved: &[ResolvedGroup],
 ) -> Result {
     let mut buckets = take(&mut symbol_db.buckets);
+    let error_queue = SegQueue::new();
 
     let replacements: Vec<Vec<(SymbolId, SymbolId)>> = buckets
         .par_iter_mut()
         .map(|bucket| {
-            let mut replacements = Vec::new();
-
-            let mut add_replacement = |a, b| {
-                replacements.push((a, b));
-            };
+            let mut local_replacements = Vec::new();
 
             for (first, alternatives) in replace(
                 &mut bucket.alternative_definitions,
                 HashMap::with_hasher(RandomState::new()),
             ) {
-                let selected = select_symbol(symbol_db, first, &alternatives, resolved);
-                add_replacement(first, selected);
-
-                for alt in alternatives {
-                    add_replacement(alt, selected);
+                match select_symbol(symbol_db, first, &alternatives, resolved) {
+                    Ok(selected) => {
+                        local_replacements.push((first, selected));
+                        for alt in alternatives {
+                            local_replacements.push((alt, selected));
+                        }
+                    }
+                    Err(err) => {
+                        error_queue.push(err);
+                    }
                 }
             }
 
@@ -622,17 +626,35 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
                 &mut bucket.alternative_versioned_definitions,
                 HashMap::with_hasher(RandomState::new()),
             ) {
-                let selected = select_symbol(symbol_db, first, &alternatives, resolved);
-                add_replacement(first, selected);
-
-                for alt in alternatives {
-                    add_replacement(alt, selected);
+                match select_symbol(symbol_db, first, &alternatives, resolved) {
+                    Ok(selected) => {
+                        local_replacements.push((first, selected));
+                        for alt in alternatives {
+                            local_replacements.push((alt, selected));
+                        }
+                    }
+                    Err(err) => {
+                        error_queue.push(err);
+                    }
                 }
             }
-
-            replacements
+            local_replacements
         })
         .collect();
+
+    let mut duplicate_errors: Vec<anyhow::Error> = error_queue.into_iter().collect();
+    duplicate_errors.sort_by_key(|e| e.to_string());
+
+    if !duplicate_errors.is_empty() {
+        let error_details = duplicate_errors
+            .iter()
+            .map(|e| e.to_string())
+            .collect::<Vec<_>>()
+            .join("\n");
+        return Err(anyhow::Error::msg(format!(
+            "Duplicate symbols detected: {error_details}"
+        )));
+    }
 
     for (a, b) in replacements.iter().flat_map(|r| r.iter()) {
         symbol_db.replace_definition(*a, *b);
@@ -650,23 +672,23 @@ fn select_symbol(
     symbol_id: SymbolId,
     alternatives: &[SymbolId],
     resolved: &[ResolvedGroup],
-) -> SymbolId {
-    let first_strength = symbol_db.symbol_strength(symbol_id, resolved);
-    if first_strength == SymbolStrength::Strong {
-        return symbol_id;
-    }
+) -> Result<SymbolId, anyhow::Error> {
     let mut max_common = None;
-    for &alt in alternatives.iter().rev() {
-        // Dynamic symbols, even strong ones, don't override non-dynamic weak symbols.
+    let mut strong_symbols = Vec::new();
+
+    let all_symbols = std::iter::once(symbol_id).chain(alternatives.iter().copied());
+
+    for alt in all_symbols {
         if symbol_db
             .symbol_value_flags(alt)
             .contains(ValueFlags::DYNAMIC)
         {
             continue;
         }
+
         let strength = symbol_db.symbol_strength(alt, resolved);
         match strength {
-            SymbolStrength::Strong => return alt,
+            SymbolStrength::Strong => strong_symbols.push(alt),
             SymbolStrength::Common(size) => {
                 if let Some((previous_size, _)) = max_common {
                     if size <= previous_size {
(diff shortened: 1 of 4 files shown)


libwild/src/symbol_db.rs
```diff
@@ -678,19 +702,34 @@ fn select_symbol(
             _ => {}
         }
     }
+
+    if strong_symbols.len() > 1 {
+        return Err(anyhow::Error::msg(format!(
+            "Multiple strong definitions found for the same symbol: {strong_symbols:?}"
```
lapla-cogito: The current implementation reports `SymbolId`, which is not appropriate for users. I'm wondering whether a method or function to retrieve the symbol name from `SymbolId` already exists.

wild/tests/sources/duplicate_strong_symbols.c
```diff
@@ -0,0 +1,14 @@
+//#AbstractConfig:default
+//#Config:executable:default
+//#SkipLinker:ld
+//#CompArgs:-fcommon
+//#Object:duplicate_strong_symbols0.c
+//#Object:duplicate_strong_symbols1.c
```
davidlattimore: You probably only need one extra file, since you can put one of the definitions in this file.

wild/tests/sources/duplicate_strong_symbols.c
```diff
@@ -0,0 +1,14 @@
...
+//#Object:exit.c
+//#LinkerDriver:none
+//#LinkArgs:--shared
```
davidlattimore: It seems fine to build either a shared object (like you are now) or an executable. If you're building a shared object, then you don't need a `_start` function.

wild/tests/sources/duplicate_strong_symbols0.c
```diff
@@ -0,0 +1,5 @@
+int test_func(void) {
+    return 0;
+}
+
+int main(){}
```
davidlattimore: There's no need for a `main` symbol. It won't be used by a shared object and it won't be used by an executable that's using `LinkerDriver:none`.

wild/tests/sources/duplicate_strong_symbols.c
```diff
@@ -0,0 +1,14 @@
+//#AbstractConfig:default
+//#Config:executable:default
+//#SkipLinker:ld
+//#CompArgs:-fcommon
+//#Object:duplicate_strong_symbols0.c
+//#Object:duplicate_strong_symbols1.c
+//#Object:exit.c
+//#LinkerDriver:none
```
davidlattimore: Note that `LinkerDriver:none` is the default, so you can probably remove this line

wild/tests/sources/duplicate_strong_symbols.c
```diff
@@ -0,0 +1,14 @@
+//#AbstractConfig:default
```
davidlattimore: If you've only got one config, then there's no need for an abstract config, so you could delete this line and remove `:default` from the next line. If you do keep the abstract config, it'd be good to put a blank line between the two configurations.

libwild/src/symbol_db.rs
```diff
@@ -610,31 +612,53 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
...
 
-            replacements
+            (replacements, errors)
         })
         .collect();
 
-    for (a, b) in replacements.iter().flat_map(|r| r.iter()) {
+    let (replacements, duplicate_errors): (Vec<_>, Vec<_>) = results.into_iter().unzip();
+
+    let replacements: Vec<(SymbolId, SymbolId)> = replacements.into_iter().flatten().collect();
```
davidlattimore: Note that we're doing a `collect` here where before we were just iterating through the replacements in-place. I don't know that this is a performance problem, but it potentially could be. Given that errors should be relatively rare, I'd suggest emitting errors separately rather than via the main return type. This can be done using something like `SegQueue<Error>` which can be pushed to from multiple threads. The errors will end up in the `SegQueue` in a non-deterministic order, so it'd be good to move them out into a Vec, then sort the Vec before displaying the errors.

libwild/src/symbol_db.rs
davidlattimore: Yep, there's methods on `SymbolDb` that can give you the name of a symbol given its SymbolId - e.g. `symbol_name_for_display`. The symbols will all have the same name, so you'd only need to display the name of one of the symbols. However, it'd be good to say which files each of the symbols came from.

wild/tests/sources/duplicate_strong_symbols.c
```diff
@@ -0,0 +1,14 @@
...
+//#Config:executable:default
+//#SkipLinker:ld
+//#CompArgs:-fcommon
```
davidlattimore: Why is this flag needed?

davidlattimore (commented):
Thanks for working on this!

wild/tests/sources/duplicate_strong_symbols.c
lapla-cogito: Thanks! It looks like the file structure from when I was struggling to write the tests was still there. Similarly, the issues pointed out in this test are leftovers from debugging while writing the test (which, of course, isn’t ideal).

wild/tests/sources/duplicate_strong_symbols.c
lapla-cogito: Modified this so that it builds as an executable file.

wild/tests/sources/duplicate_strong_symbols0.c
lapla-cogito: Removed.

wild/tests/sources/duplicate_strong_symbols.c
lapla-cogito: Removed.

wild/tests/sources/duplicate_strong_symbols.c
lapla-cogito: Simplified this to use a single config.

wild/tests/sources/duplicate_strong_symbols.c
lapla-cogito: Removed.

libwild/src/symbol_db.rs
lapla-cogito: Thanks! I was able to include the symbol names in the error messages. I also made sure to output the filenames where they appear. So now, for the test code included in this patch, I get the following stderr, for example:
```
-- stderr --
Error: Duplicate symbols detected: test_func, defined in /home/lapla/repos/wild/wild/tests/build/duplicate_strong_symbols2.dup-host-2847626a4a6e7b0b.o, which previously defined in /home/lapla/repos/wild/wild/tests/build/duplicate_strong_symbols.dup-host-157411f97983d987.o
```

davidlattimore: Great! A small grammatical change. I'd either change "which previously defined in" to just "previously defined in" (remove "which") or "which was previously defined in" (add "was")

libwild/src/symbol_db.rs
lapla-cogito: Makes sense. I’ve made the changes.

libwild/src/symbol_db.rs
lapla-cogito: Thanks! Fixed.

libwild/src/symbol_db.rs
```diff
@@ -596,45 +598,68 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
...
+                    }
                 }
             }
-
-            replacements
+            local_replacements
         })
+        .collect::<Vec<_>>()
+        .into_iter()
+        .flatten()
```
davidlattimore: Can you revert these few lines here? You'd also need to change the loop over `replacements` below back to how it was. That way we avoid collecting all replacements into one large flat Vec.

lapla-cogito: I see. However, I don't think any changes are needed for the latter half of the loop. This is because, in the original implementation, `replacements` was a `Vec<Vec<(SymbolId, SymbolId)>>`, whereas in the current implementation, it's a `Vec<(SymbolId, SymbolId)>`.

libwild/src/symbol_db.rs
```diff
@@ -596,45 +598,65 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
...
+            .iter()
+            .map(|e| e.to_string())
+            .collect::<Vec<_>>()
+            .join("\n");
+        return Err(anyhow::Error::msg(format!(
+            "Duplicate symbols detected: {error_details}"
+        )));
+    }
+
+    for (a, b) in &replacements {
```
davidlattimore: So with your most recent push, it's still collecting all the replacements into one large Vec - it's just doing it in a slightly different way. What I was meaning was change this line back to:

```rust
for (a, b) in replacements.iter().flat_map(|r| r.iter()) {
```

Then the `Vec` that gets collected above contains many smaller Vecs like it did before, which means less copying.

lapla-cogito: Ah, I misunderstood what you were trying to say, sorry. It’s probably fixed now.

davidlattimore: Thanks, yep, that's good now. I realised it seems strange that a flat `Vec` might be worse performance-wise than a `Vec<Vec<..>>`. The key difference is that with the flat vec, we're building all the smaller vecs, then iterating over them and copying the values into the flat vec, then iterating over that. Whereas with the nested Vecs, we build all the smaller vecs then just iterate over them - there's no merging it into a larger Vec. Hopefully that makes sense and thanks for making the change :)

mati865:
Unfortunately, there is a real world regression from this PR: https://github.com/davidlattimore/wild/issues/598

PR #597, opened 2025-03-24
make tests pass on Alpine Linux

### LINKED ISSUE #286: Make tests pass on Alpine Linux
I added a dockerfile to make it easier for those of us that don't run Alpine natively to have a look. In that image, I see the following tests fail:

* integration_test::program_name_02___trivial_main_c__
  * `Insufficient .gnu.version allocation` - For some reason we're attempting to write more data to .gnu.version than we allocated during the layout phase.
* integration_test::program_name_11___ifunc_c__
  * Ifuncs aren't supported on musl. I guess we need some way to skip this test for musl.
* integration_test::program_name_32___ifunc2_c__
  * Same as above
* integration_test::program_name_26___libc_integration_c__
  * Ifuncs again. Probably don't want to skip the whole test, but just the part of this test that's related to ifuncs.
* integration_test::program_name_27___rust_integration_rs__
  * Initially failed because I forgot to add `clang` to the installs in the dockerfile.
  * After that, this is failing because we're trying to run `bash`. Need to either switch to `sh` or maybe install bash in the image.

The following 3 tests also failed for me, but I didn't look at them, since I figured they'd be more of the same.

* integration_test::program_name_28___rust_integration_dynamic_rs__
* integration_test::program_name_29___cpp_integration_cc__
* integration_test::program_name_30___rust_tls_rs__

I probably won't get a chance to look more at any of this for a while, since I'm focusing my time on the linker-diff rewrite. If anyone would like to have a go at fixing any of these and needs more guidance, I'm definitely happy to help where I can.

_Originally posted by @davidlattimore in https://github.com/davidlattimore/wild/issues/230#issuecomment-2571804072_

lapla-cogito:
close #286 (I commented on this issue regarding the latest test failures.)

By default, `rustc` attempts to statically link for the `x86_64-unknown-linux-musl` target. As a result, the dynamically linked library being tested does not seem to work correctly on Alpine Linux. By using [the `crt_link` option](https://rust-lang.github.io/rfcs/1721-crt-static.html), we can explicitly enable dynamic linking by adding `-C target-feature=-crt-static`.

I confirmed that all tests pass on Alpine Linux in Docker.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -366,6 +366,10 @@ fn is_host_debian_based() -> bool {
     )
 }
 
+fn is_musl_used() -> bool {
+    os_info::get().os_type() == Type::Alpine
+}
+
 fn host_supports_clang_with_tls_desc() -> bool {
     static CLANG_SUPPORTS_TLS_DESC: OnceLock<bool> = OnceLock::new();
 
@@ -1105,6 +1109,10 @@ fn build_obj(
                 );
 
                 command.arg(format!("-Clink-arg=--target={target}"));
+            }
+
+            if is_musl_used() {
+                command.args(["-C", "target-feature=-crt-static"]);
             }
 
             if input_type == InputType::SharedObject {


davidlattimore:
Nicely figured out! Thanks!

PR #618, opened 2025-03-27
introduce Alpine in CI as well

lapla-cogito:
I’ve made [this patch](https://github.com/davidlattimore/wild/pull/597) pass the tests on Alpine as well, so I think it’s ready to be added to CI. Adding Alpine as a test target will help catch differences between `musl` and `glibc`, like the one found in this patch, at an early stage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
           - 'ubuntu:24.10'
           - 'ubuntu:25.04'
           - 'opensuse/tumbleweed:latest'
+          - 'rust:1.85-alpine'
         test-qemu:
           - false
         include:
@@ -36,6 +37,9 @@ jobs:
           # Ubuntu 22.04 contains an old lld linker which cannot do a relaxation on AArch64 (required by linker-diff)
           - runs-on: ubuntu-24.04
             container: 'ubuntu:22.04'
+        exclude: # TODO: https://github.com/actions/runner/issues/1637
+          - runs-on: ubuntu-24.04-arm
+            container: 'rust:1.85-alpine'
       fail-fast: false
 
     runs-on: ${{ matrix.runs-on }}
@@ -52,7 +56,8 @@ jobs:
         if: ${{ matrix.test-qemu }}
       - run: zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap
         if: ${{ contains(matrix.container, 'opensuse') }}
-      - run: useradd -m -g users user && su user && cd ~
+      - run: apk add build-base lld clang bash curl
+        if: ${{ contains(matrix.container, 'alpine') }}
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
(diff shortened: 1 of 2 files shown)


davidlattimore:
Good idea. Alpine is different enough, that that should definitely expand our coverage. Thanks! Good luck getting it passing and let me know if you need any help

lapla-cogito:
@davidlattimore All tests are now passing. For now, testing on Alpine for AArch64 is skipped because some actions like `actions/checkout` cannot be used there. We could either skip it or implement an alternative, but I don't think it's a significant issue, so I decided to skip it this time. Also, in the second commit, I added a missing option that was required to make the tests pass.

.github/workflows/ci.yml
```diff
@@ -70,8 +75,12 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-${{ matrix.container }}-${{ matrix.runs-on }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
+        if: ${{ !contains(matrix.container, 'alpine') }}
       - run: cargo build --profile ci --workspace --no-default-features
+      - run: cargo test --profile ci --workspace
+        if: ${{ contains(matrix.container, 'alpine') }}
       - run: WILD_TEST_CROSS=$WILD_TEST_CROSS cargo test --profile ci --workspace
+        if: ${{ !contains(matrix.container, 'alpine') }}
```
davidlattimore, line 83: Why is this needed? It looks like `test_qemu` is set to false by default and isn't set to true for the alpine container

.github/workflows/ci.yml
```diff
@@ -70,8 +75,12 @@ jobs:
             ~/.cargo/git/db/
             target/
           key: ${{ runner.os }}-${{ matrix.container }}-${{ matrix.runs-on }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
+        if: ${{ !contains(matrix.container, 'alpine') }}
```
davidlattimore, line 78: Why do we need to disable the cache? I expect not having the cache will make CI take a good bit longer

lapla-cogito: This is because `actions/cache` uses the `--posix` option when running the tar command internally. Since this option doesn't work on Alpine, the cache ends up not being saved.

.github/workflows/ci.yml
lapla-cogito: At first, I tried to run tests for all targets using `WILD_TEST_CROSS=$WILD_TEST_CROSS cargo test --profile ci --workspace`. However, on Alpine, the behavior when the environment variable is undefined seems to be different, and tests for AArch64 were also being run. So, as a workaround, I ended up doing it this way.

.github/workflows/ci.yml
lapla-cogito: I tried this: https://github.com/davidlattimore/wild/actions/runs/14101156134/job/39497483080

.github/workflows/ci.yml
lapla-cogito: I tried this: https://github.com/davidlattimore/wild/actions/runs/14100681203/job/39496159742

davidlattimore: Ah, yep. I can see from the code why it would behave that way. I might change it later

davidlattimore (approved):
Thanks for doing this

Issue #614, opened 2025-03-26
Some of Rust proc-macro tests cause compiler to crash


mati865:
Failing tests:
- tests/ui/proc-macro/meta-macro.rs
- tests/ui/proc-macro/meta-macro-hygiene.rs

They both crash when calling `free` when rustc is linked with wild:
```
--- stdout -------------------------------
Def site: /home/mateusz/Projects/rust/tests/ui/proc-macro/auxiliary/make-macro.rs:7:9: 7:56 (#3)
Input: TokenStream []
Respanned: TokenStream []
------------------------------------------
--- stderr -------------------------------
free(): invalid pointer
------------------------------------------
```
It's not a crash in the test binary, but a crash inside stage1 rustc executable, likely making the debugging harder.

davidlattimore:
I'm having trouble reproducing this. I tried on my laptop and both tests passed. I then tried in an arch docker image and they passed there too.

```
# PATH=/wild/fakes:$PATH ./x t --force-rerun tests/ui/proc-macro/meta-macro.rs
...
   Compiling compiletest v0.0.0 (/rust/src/tools/compiletest)
    Finished `release` profile [optimized] target(s) in 35.92s
Testing stage1 compiletest suite=ui mode=ui (x86_64-unknown-linux-gnu)

running 1 tests
.

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 18814 filtered out; finished in 199.16ms

Build completed successfully in 0:12:57

# readelf -p .comment ./build/x86_64-unknown-linux-gnu/stage1/bin/rustc | grep Linker
  [    4f]  Linker: Wild version 0.4.0
```

mati865:
Maybe it requires specific commit or depends on debuginfo, I'll retry later.

davidlattimore:
Actually, although the stage1 compiler seems to have been linked with Wild, it seems that the test wasn't. Turned out that I hadn't configured things properly and it was using lld. I can now reproduce. I'll take a look at this later today or tomorrow and see if I can figure out what's going on.

davidlattimore:
I explored way too many dead-ends, but I eventually figured it out. The problem was that TDATA was empty and the two TLS variables that the shared object defined were both in TBSS. We were computing the offset into the TLS segment by subtracting the start address of TDATA. However because TDATA wasn't present, its address didn't take into account padding required by TBSS, so the address subtracted was actually a few (7 bytes) lower than the actual start address of TBSS.

[...]

PR #624, opened 2025-03-28
Emit dynamic relocations for direct refs to interposable syms

davidlattimore:


davidlattimore:
I noticed that we weren't getting this right when investigating #614. I thought it could potentially be the cause of that issue, but it doesn't seem so.

PR #627, opened 2025-03-30
aarch64: Adjust placement of movn/movz value bits

davidlattimore:


davidlattimore:
Hi @marxin. I'm working on a change to linker-diff that has it doing more checking of TLS relocations and hit some issues with some match failures. This change I think matches what the ARM spec says and makes the test pass with my linker-diff change, but I wanted to double-check with you that it seems right

marxin:
Good catch! I am surprised it hasn't caused any run-time issues so far.

marxin (approved):

Issue #612, opened 2025-03-26
Some of Rust backtrace tests fail


mati865:
I don't know if this is necessary, but in `bootstrap.toml` I've got:
```toml
[rust]
lld = false
debug = true
debuginfo-level = 2
```

The tests:
- tests/ui/backtrace/dylib-dep.rs
- tests/ui/runtime/backtrace-debuginfo.rs

fail because they cannot find code location in the debuginfo:
```
--- stderr -------------------------------
-----------------------------------
looking for:
	/home/mateusz/Projects/rust/tests/ui/backtrace/dylib-dep.rs:70
	/home/mateusz/Projects/rust/tests/ui/backtrace/auxiliary/dylib-dep-helper.rs:18
	/home/mateusz/Projects/rust/tests/ui/backtrace/dylib-dep.rs:69
found:
Backtrace [
    { fn: "dylib_dep::verify" },
    { fn: "dylib_dep::main::{{closure}}" },
    { fn: "core::ops::function::FnOnce::call_once" },
    { fn: "foo", file: "/home/mateusz/Projects/rust/tests/ui/backtrace/auxiliary/dylib-dep-helper.rs", line: 18 },
    { fn: "dylib_dep::main" },
    { fn: "core::ops::function::FnOnce::call_once" },
    { fn: "std::sys::backtrace::__rust_begin_short_backtrace", file: "/rustc/FAKE_PREFIX/library/std/src/sys/backtrace.rs", line: 152 },
    { fn: "std::rt::lang_start::{{closure}}", file: "/rustc/FAKE_PREFIX/library/std/src/rt.rs", line: 199 },
    { fn: "std::rt::lang_start_internal" },
    { fn: "std::rt::lang_start", file: "/rustc/FAKE_PREFIX/library/std/src/rt.rs", line: 198 },
    { fn: "main" },
    { fn: "__libc_start_main" },
    { fn: "_start" },
]

thread 'main' panicked at /home/mateusz/Projects/rust/tests/ui/backtrace/dylib-dep.rs:53:9:
expected backtrace to contain /home/mateusz/Projects/rust/tests/ui/backtrace/dylib-dep.rs
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
             at /home/mateusz/Projects/rust/library/core/src/panicking.rs:75:14
   2: dylib_dep::verify
   3: dylib_dep::main::{{closure}}
   4: core::ops::function::FnOnce::call_once
[...]

PR #640, opened 2025-04-02
Fix typo in name of the debug info sections

marxin:
Fixes: #612.


marxin:
It shows the test-case mentioned in #612 wrongly sets the tombstone values of debug info sections.
Sorry for the stupid mistake.

Issue #44, opened 2024-08-18
Support for linker scripts


davidlattimore:
Wild currently supports the bare minimum of linker script syntax needed in order to handle loading shared objects that are actually small linker scripts that say to load some other shared object, possibly with additional linker flags.

What it doesn't support is a user-provided linker script that specifies the layout of the file we're creating.

It'd be interesting to see how much of the linker script syntax and functionality we can support without killing performance.

Issue #36 needs to be done first though.

jaskij:
As mentioned in #1110, I'm posting a semi-complicated linker script I use for microcontrollers, hopefully it'll be of help. This is a project that, should be able to support various microcontrollers by simply changing include paths for the linker and nothing else.

The first file, `sizes.ld`, is where I define the various memory banks of the microcontroller

```ld
SECTOR_SIZE = 128k;

FLASH_ORIGIN   = 0x8000000;
FLASH2_ORIGIN  = 0x8100000;
FLASH_SIZE     = 512K;

BOOTLOADER_ORIGIN  = FLASH_ORIGIN;
BOOTLOADER2_ORIGIN = FLASH2_ORIGIN;
BOOTLOADER_SIZE    = 128k;

FIRMWARE_ORIGIN    = FLASH_ORIGIN + BOOTLOADER_SIZE;
FIRMWARE2_ORIGIN   = FLASH2_ORIGIN + BOOTLOADER_SIZE;
FIRMWARE_SIZE      = 384k;

ASSERT(((BOOTLOADER_SIZE + FIRMWARE_SIZE) <= FLASH_SIZE), "bootloader and firmware size must fit in flash")
ASSERT(((FLASH_SIZE % SECTOR_SIZE) == 0), "WTF? flash size not divisble by sector size")
ASSERT(((BOOTLOADER_SIZE % SECTOR_SIZE) == 0), "WTF? bootloader size not divisble by sector size")
ASSERT(((FIRMWARE_SIZE % SECTOR_SIZE) == 0), "WTF? firmware size not divisble by sector size")

MEMORY
{
    DTCMRAM : ORIGIN = 0x20000000, LENGTH = 128k
    ITCMRAM : ORIGIN = 0x00000000, LENGTH = 64k

    D1_RAM  : ORIGIN = 0x24000000, LENGTH = 512k

    D2_RAM1 : ORIGIN = 0x30000000, LENGTH = 128k
    D2_RAM2 : ORIGIN = 0x30020000, LENGTH = 128k
[...]

PR #641, opened 2025-04-03
Reimplement implicit linker script parsing to use winnow

davidlattimore:
This should make it easier to extend what's supported to include more of the linker script syntax.

Issue #44

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,6 +681,7 @@ dependencies = [
  "tracing-subscriber",
  "typed-arena",
  "uuid",
+ "winnow",
  "zstd",
 ]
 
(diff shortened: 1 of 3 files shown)


mati865:
IDK if you know about this https://github.com/rosetta-rs/parse-rosetta-rs, so I'm posting it just in case.

davidlattimore:
> IDK if you know about this https://github.com/rosetta-rs/parse-rosetta-rs, so I'm posting it just in case.

Thanks, yep, I read that just recently. I haven't made much use of Rust parser frameworks, so I don't have any strong preferences or opinions. winnow seemed pretty easy to use though and the performance looks pretty good. I don't expect performance of parsing linker scripts to often be an issue with the exception of version scripts, which can potentially have a lot of symbols listed (e.g. I've seen rustc write a version script with over 300K symbols listed). I'm unsure at this stage what to do with the version script code. i.e. whether to switch it to winnow, or perhaps move it out to a separate `version_script.rs`. The GNU does says that version scripts are part of the linker script syntax and even says that you can pass a version script without the `--version-script=` flag, however when I tried that, it failed to parse it. Perhaps it only works with some version scripts.

mati865:
Unfortunately, I have very little experience with parsers too. Just thought that reference might be useful for finding the balance between the performance and the ease of use.

> I'm unsure at this stage what to do with the version script code. i.e. whether to switch it to winnow

I admit it was tempting to try how it goes with a generic parser. I'd expect a small performance hit there.

> or perhaps move it out to a separate `version_script.rs`. The GNU does says that version scripts are part of the linker script syntax

I wouldn't follow the docs too strictly. Maybe it worked out well for them, but with a different design it might be just an unnecessary complication.

PR #643, opened 2025-04-04
add test configuration support

### LINKED ISSUE #626: Add test configuration file
Was just talking with @mati865 and learned that Wild has been packaged for Arch Linux, but some tests needed to be [skipped](https://gitlab.archlinux.org/archlinux/packaging/packages/wild/-/blob/main/PKGBUILD?ref_type=heads#L40) due to using nightly rustc.

We discussed having a configuration file for the tests. I propose `test-config.toml` in the repository root. The config file wouldn't be checked in and would be git-ignored, so people could supply their own configuration file as they see fit. We would check in a `test-config.toml.example` and a `test-config-ci.toml`.

The two main things that currently would be good to control via such a file are:

* Whether to use nightly rustc. The cranelift backend test would need to be skipped if we weren't using nightly.
* Whether to cross compile tests for aarch64 and run with qemu. This is currently disabled by default and only enabled via setting the environment variable `WILD_TEST_CROSS=aarch64`. I think we perhaps should keep the environment variable. I guess environment variable takes precedence, but I'm not sure.

The goal here is to make it easier for people to just `cargo test` wild without needing a lot of esoteric things installed, but still allow core developers and CI to run those more unusual tests.

lapla-cogito:
close #626 

This patch adds two options for configuring test behavior:

- `rustc_channel`: Specifies which Rust compiler channel to use when running tests that build Rust code.
- `use_qemu`: Determines whether to run tests for architectures different from the host. This setting is overridden by the `$WILD_TEST_CROSS` environment variable.

These options are specified in TOML format and take effect by setting the path to the configuration file via an environment variable.

Additionally, an option was added to explicitly indicate that a test requires the nightly Rust compiler (`RequiresNightlyRustc`).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,22 @@ disable cross compilation.
 
 Cross compilation is set up in docker/debian.Dockerfile.
 
+## Configuration file for tests
+
+Currently, the behavior for the following test options can be configured using the TOML format:
+
+- `rustc_channel`: Specifies which Rust compiler channel to use when running tests that build Rust code. The default value is "default", which means no explicit toolchain is specified.
+
+- `use_qemu`: Determines whether to run tests for architectures different from the host. This setting is overridden by the `$WILD_TEST_CROSS` environment variable. The default value is `false`.
+
+A sample configuration file is provided as `test-config.toml.sample`.
+By default, Wild uses `test-config.toml` as the configuration file.
+If you have written your configuration in a different file, specify its location using the `WILD_TEST_CONFIG` environment variable as follows:
+
+```sh
+WILD_TEST_CONFIG=path_to_config cargo test
+```
+
 ## GitHub workflow
 
 TL;DR: We're pretty relaxed. Feel free to force push or not. Squash, rebase, merge, whatever you
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -415,6 +415,22 @@ struct Config {
     requires_glibc: bool,
     requires_clang_with_tlsdesc: bool,
     version_script: Option<PathBuf>,
+    rustc_channel: Option<RustcChannel>,
+}
+
+#[derive(serde::Deserialize)]
+struct TestConfig {
+    // These configs are used by the config file specified in `$WILD_TEST_CONFIG`
+    rustc_channel: RustcChannel,
+    use_qemu: bool,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, serde::Deserialize, Debug)]
+#[serde(rename_all = "snake_case")]
+enum RustcChannel {
+    Stable,
+    Beta,
+    Nightly,
 }
 
 #[derive(Clone, Copy, PartialEq, Eq)]
@@ -429,12 +445,13 @@ impl Default for DirectConfig {
 }
 
 impl Config {
-    fn should_skip(&self, arch: Architecture) -> bool {
+    fn should_skip(&self, arch: Architecture, test_config: &TestConfig) -> bool {
         !self.support_architectures.contains(&arch)
             || self.requires_glibc && !cfg!(target_env = "gnu")
             || (self.requires_clang_with_tlsdesc && !host_supports_clang_with_tls_desc())
             || (arch != get_host_architecture()
                 && (self.compiler == "clang" || !self.cross_enabled))
+            || (test_config.rustc_channel != RustcChannel::Nightly && self.requires_nightly_rustc)
     }
 
     fn is_linker_enabled(&self, linker: &Linker) -> bool {
@@ -580,6 +597,7 @@ impl Default for Config {
             requires_glibc: false,
             requires_clang_with_tlsdesc: false,
             version_script: None,
+            rustc_channel: None,
         }
     }
 }
@@ -742,6 +760,7 @@ fn parse_configs(src_filename: &Path) -> Result<Vec<Config>> {
             }
         }
     }
+
     let mut configs = config_by_name
         .into_values()
         .filter(|c| !c.is_abstract)
@@ -1072,10 +1091,16 @@ fn build_obj(
         }
         CompilerKind::Rust => {
             let wild = wild_path().to_str().context("Need UTF-8 path")?.to_owned();
+            let rustc_channel = match config.rustc_channel {
+                Some(RustcChannel::Stable) => Some("+stable"),
+                Some(RustcChannel::Beta) => Some("+beta"),
+                Some(RustcChannel::Nightly) => Some("+nightly"),
+                None => None,
+            };
 
             command
                 .env("WILD_SAVE_SKIP_LINKING", "1")
-                .arg("+nightly")
+                .args(rustc_channel)
                 .args(["-C", "linker=clang"])
                 .args(["-C", &format!("link-arg=--ld-path={wild}")]);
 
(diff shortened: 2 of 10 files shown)


wild/tests/integration_tests.rs
```diff
@@ -414,7 +417,25 @@ struct Config {
...
+    use_qemu: bool,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+enum RustcChannel {
+    Stable,
+    Beta,
+    Nightly,
 }
```
lapla-cogito: I separated the struct because I didn't want to derive `serde::Deserialize` for the other fields in `Config`.

wild/tests/integration_tests.rs
```diff
@@ -1072,10 +1100,15 @@ fn build_obj(
         }
         CompilerKind::Rust => {
             let wild = wild_path().to_str().context("Need UTF-8 path")?.to_owned();
+            let rustc_channel = match config.rustc_channel {
+                RustcChannel::Stable => "+stable",
```
davidlattimore: I'm wondering if we should add a 4th option - `Default` that doesn't add the argument at all. When distributions build and test wild, they likely won't have rustup installed, which I think would mean `+toolchain` wouldn't work.

wild/tests/integration_tests.rs
```diff
@@ -2188,17 +2221,52 @@ fn integration_test(
 
     // We only currently support cross compilation from x86_64 to aarch64, so we don't need to track
     // which targets are enabled, since there's only one.
-    let cross_enabled = std::env::var("WILD_TEST_CROSS").is_ok_and(|v| v == "aarch64");
+    let mut cross_enabled = std::env::var("WILD_TEST_CROSS").is_ok_and(|v| v == "aarch64");
+    let mut rustc_channel = RustcChannel::Nightly;
+    // parse configs specified in `$WILD_TEST_CONFIG`
+    if let Ok(config_path) = std::env::var("WILD_TEST_CONFIG") {
```
davidlattimore: Consider extracting this code out into a `read_test_config` function

CONTRIBUTING.md
```diff
@@ -59,6 +59,21 @@ disable cross compilation.
...
+
+- `rustc_channel`: Specifies which Rust compiler channel to use when running tests that build Rust code.
+
+- `use_qemu`: Determines whether to run tests for architectures different from the host. This setting is overridden by the `$WILD_TEST_CROSS` environment variable.
+
+A sample configuration file is provided as `test-config.toml.sample`.
+If you define these settings in your own `test-config.toml`, you can run tests using that configuration as follows:
+
+```sh
+WILD_TEST_CONFIG=test-config.toml cargo test
```
davidlattimore: I think it'd be good if we defaulted to reading `test-config.toml` even if the environment variable isn't set. The idea is that a developer can customise the config for their particular system based on what stuff they've got installed. If the environment variable isn't specified and the file doesn't exist, then just use the defaults and don't error. If the environment variable is set, then read the specified file and report an error if it doesn't exist (like you do now).

CONTRIBUTING.md
```diff
@@ -59,6 +59,21 @@ disable cross compilation.
 
 Cross compilation is set up in docker/debian.Dockerfile.
 
+## Configuration file for tests
+
+Currently, the behavior for the following test options can be configured using the TOML format:
+
+- `rustc_channel`: Specifies which Rust compiler channel to use when running tests that build Rust code.
+
+- `use_qemu`: Determines whether to run tests for architectures different from the host. This setting is overridden by the `$WILD_TEST_CROSS` environment variable.
```
davidlattimore: We may eventually want to be able to specify which cross-compilation toolchains are available. e.g. `cross_toolchains = ["aarch64], "riscv64"]`, but this is fine for now.

davidlattimore (commented):
Awesome. Thanks for working on this!

CONTRIBUTING.md
lapla-cogito: I had initially considered explicitly specifying the location of the configuration file, anticipating that more settings might be added in the future and that developers would need to use multiple configuration files simultaneously. However, I realized that defining a default file is a good idea, so I went ahead and did that. Thanks!

wild/tests/integration_tests.rs

wild/tests/integration_tests.rs
lapla-cogito: Makes sense. That’s probably how it is now. Thanks!

linker-diff/Cargo.toml
```diff
@@ -40,5 +40,8 @@ tempfile = "3.19.1"
 which = "7.0.2"
 ascii_table = "4.0.6"
 
+[dev-dependencies]
+toml = "0.8.20"
```
davidlattimore: Why does linker-diff need toml?

wild/tests/integration_tests.rs
```diff
@@ -1072,10 +1099,18 @@ fn build_obj(
...
-
+            let rustc_channel = match config.rustc_channel {
+                RustcChannel::Stable => "+stable",
+                RustcChannel::Beta => "+beta",
+                RustcChannel::Nightly => "+nightly",
+                RustcChannel::Default => "",
+            };
+
+            command.env("WILD_SAVE_SKIP_LINKING", "1");
+            if config.rustc_channel != RustcChannel::Default {
```
davidlattimore: I think this would be slightly cleaner if you made `rustc_channel` an `Option`, then set it to `None` for `RustcChannel::Default`. Then this if statement could be:

```rust
if let Some(channel) = rustc_channel {
   command.arg(channel);
}
```

Or, possibly (although I haven't tried it, just:

```rust
command.args(rustc_channel);
```

Since `Command::args` accepts anything that can be converted into an iterator, which includes `Option`.

wild/tests/integration_tests.rs
```diff
@@ -2186,21 +2221,73 @@ fn integration_test(
...
     for &arch in ALL_ARCHITECTURES {
-        if arch != host_arch && !cross_enabled {
+        if arch != host_arch && !test_config.use_qemu {
             continue;
         }
 
-        let config_it = configs.iter().filter(|config| !config.should_skip(arch));
+        let config_it = configs.iter().filter(|config| {
+            !(config.should_skip(arch)
+                || (test_config.rustc_channel != RustcChannel::Nightly
```
davidlattimore: It'd be nice if this logic could just be in `should_skip`. That could be done if you moved the code that sets `config.rustc_channel` (currently below) into `parse_configs`. i.e. move the call to `read_test_config` earlier and pass the TestConfig into `parse_configs`.

linker-diff/Cargo.toml
lapla-cogito: It was a mistake (I had mistakenly written a test in `linker-diff` initially).

wild/tests/integration_tests.rs
lapla-cogito: Since it worked fine in the latter part as well, I went with that approach.

wild/tests/integration_tests.rs

wild/tests/integration_tests.rs
```diff
@@ -429,12 +449,18 @@ impl Default for DirectConfig {
 }
 
 impl Config {
-    fn should_skip(&self, arch: Architecture) -> bool {
+    fn should_skip(
+        &self,
+        arch: Architecture,
+        test_config: &TestConfig,
+        requires_nightly_rustc: bool,
```
davidlattimore: I don't think it's necessary to pass this as an argument, since it's available on `self`.

linker-diff/Cargo.toml
davidlattimore: Makes sense. It looks like this change is still here though. Either that, or the github UI is just confusing

wild/tests/integration_tests.rs
lapla-cogito: Right. Fixed.

linker-diff/Cargo.toml
lapla-cogito: Ah, it seems I misunderstood the commit where I originally added the `toml` dependency during the rebase, so the diff for adding `toml` ended up remaining in the final.

PR #675, opened 2025-04-16
CI: Use `ubuntu-latest` in release

lapla-cogito:
It looks like Ubuntu 20.04 has been retired from GitHub-hosted runners, causing the release CI to fail.
e.g.: https://github.com/davidlattimore/wild/actions/runs/14485332930/job/40629776730

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,7 @@ on:
 jobs:
   # Run 'dist plan' (or host) to determine what tasks we need to do
   plan:
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-latest"
     outputs:
       val: ${{ steps.plan.outputs.manifest }}
       tag: ${{ !github.event.pull_request && github.ref_name || '' }}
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks! I was looking at switching away from dist because of this, but this is definitely an easier option for now :)

davidlattimore:
One thing I'm not sure about is whether this means that the binary that's built will be linked against the glibc in ubuntu-latest. For example, if I take wild built on ubuntu-latest and try to run it on ubuntu-22.04, the following happens:

```sh
❯ ./wild
./wild: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by ./wild)
```

If that's the case here, then we could perhaps do the build in a container, which would let us build on an older version. Binaries built with an older glibc work fine on systems with newer glibc, just not the other way around.

lapla-cogito:
I see. Then I'll try updating it to use the latest stable version of Debian. Fortunately, it seems that `cargo-dist` allows us to configure that as well. (Though we're not sure how long we'll continue using `cargo-dist`, I expect we'll stick with it at least until the next release.)

PR #676, opened 2025-04-16
Use sections headers instead of program headers

### LINKED ISSUE #611: Rust tests/run-make/raw-dylib-elf fails because binary tries to load a wrong shared object
For reproducing the problem, I recommend doing this change:
```diff
diff --git a/tests/run-make/raw-dylib-elf/rmake.rs b/tests/run-make/raw-dylib-elf/rmake.rs
index 59f901ac1e..4a28256c3c 100644
--- a/tests/run-make/raw-dylib-elf/rmake.rs
+++ b/tests/run-make/raw-dylib-elf/rmake.rs
@@ -17,6 +17,7 @@
         .crate_type("bin")
         .input("main.rs")
         .arg(&format!("-Wl,-rpath={}", cwd().display()))
+        .arg("-Csave-temps")
         .run();

     // Now, *after* building the binary, we build the library...
```
It prevents removal of `/tmp/rust*` that contain the raw-dylib.

---

Import library created by Rust looks good at the first sight:
```
❯ readelf -Wa rustc947gHX/raw-dylibs/libf0ila6u5kpn02icz3y0bvel03.so
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x0
  Start of program headers:          0 (bytes into file)
  Start of section headers:          112 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           0 (bytes)
  Number of program headers:         0
  Size of section headers:           64 (bytes)
  Number of section headers:         6
  Section header string table index: 1

Section Headers:
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .shstrtab         STRTAB          0000000000000000 000040 00002a 00      0   0  1
  [ 2] .text             PROGBITS        0000000000000000 000000 000000 00      0   0  1
  [ 3] .dynstr           STRTAB          0000000000000000 0001f0 00002a 00   A  0   0  1
  [ 4] .dynsym           DYNSYM          0000000000000000 000220 000030 18   A  3   1  8
  [ 5] .dynamic          DYNAMIC         0000000000000000 000250 000020 10  WA  3   0  8
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), l (large), p (processor specific)

There are no section groups in this file.

There are no program headers in this file.

There is no dynamic section in this file.

There are no relocations in this file.
No processor specific unwind information to decode

Symbol table '.dynsym' contains 2 entries:
     1: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT    2 this_is_a_library_function

No version information found in this file.

❯ readelf -W -p .dynstr rustc947gHX/raw-dylibs/libf0ila6u5kpn02icz3y0bvel03.so

String dump of section '.dynstr':
  [     1]  liblibrary.so
  [     f]  this_is_a_library_function
```
But wild adds wrong dependency to DT_NEEDED:
```
❯ WILD_REFERENCE_LINKER=ld ./run-with ~/Projects/wild/target/release/wild
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/14.2.1/liblto_plugin.so is not yet supported
ld: warning: type and size of dynamic symbol `this_is_a_library_function' are not defined
wild: ./bin
ref: ./bin.ref-linker
.dynamic.DT_NEEDED
  wild libgcc_s.so.1,libc.so.6,ld-linux-x86-64.so.2,libf0ila6u5kpn02icz3y0bvel03.so
  ref libgcc_s.so.1,libc.so.6,ld-linux-x86-64.so.2,liblibrary.so
```

`libf0ila6u5kpn02icz3y0bvel03.so` is the name of the temporary import library created by rustc. Which means somehow there was no `soname` at this point: https://github.com/davidlattimore/wild/blob/3aa88140ee4e89da1ee76a399e6ba57ef4a0f44e/libwild/src/layout.rs#L4442-L4444
Following this trail, I've noticed `program_headers` are empty at this point for this import library: https://github.com/davidlattimore/wild/blob/3aa88140ee4e89da1ee76a399e6ba57ef4a0f44e/libwild/src/elf.rs#L258
Naturally, that's why `soname` is `None`.

I haven't investigated deeper, but I should be able to find time for it if nobody else steps up.

mati865:
Fixes #611 

In Rust repo:
```
❯ PATH=~/Projects/wild:$PATH ./x.py clean

❯ PATH=~/Projects/wild:$PATH ./x.py t --skip tests/debuginfo/ --skip tests/run-make/include-all-symbols-linking
...
Build completed successfully in 0:17:50
```

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2,7 +2,6 @@ use crate::arch::Architecture;
 use crate::error::Result;
 use crate::resolution::LoadedMetrics;
 use anyhow::Context;
-use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::ensure;
 use bytemuck::Pod;
@@ -19,7 +18,6 @@ use object::LittleEndian;
 use object::read::elf::CompressionHeader;
 use object::read::elf::Dyn;
 use object::read::elf::FileHeader as _;
-use object::read::elf::ProgramHeader as _;
 use object::read::elf::RelocationSections;
 use object::read::elf::SectionHeader as _;
 use std::borrow::Cow;
@@ -60,7 +58,6 @@ pub(crate) struct File<'data> {
     pub(crate) sections: SectionTable<'data>,
     /// This may be symtab or dynsym depending on the file type.
     pub(crate) symbols: SymbolTable<'data>,
-    pub(crate) program_headers: &'data [ProgramHeader],
     pub(crate) versym: &'data [Versym],
 
     /// An iterator over the version definitions and the corresponding linked string table index.
@@ -118,19 +115,11 @@ impl<'data> File<'data> {
             }
         }
 
-        let program_headers = get_entries(
-            data,
-            header.e_phoff(endian) as usize,
-            header.e_phnum(endian) as usize,
-        )
-        .context("Failed to read program headers")?;
-
         Ok(Self {
             arch: architecture,
             data,
             sections,
             symbols,
-            program_headers,
             versym,
             verdef,
             verdefnum,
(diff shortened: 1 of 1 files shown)


mati865:
Ah, forgot to override the linker for tests.

Issue #694, opened 2025-04-21
Rustc debug info regression from #684


mati865:
In Rust repo:
```
❯ jj status
The working copy has no changes.
Working copy  (@) : yqrxqwww 2a044cfc (empty) (no description set)
Parent commit (@-): qumqxmmq b8005bff master | (empty) Auto merge of #140079 - ChrisDenton:rollup-2h5cg94, r=ChrisDenton

# Only `lld = false` is set
❯ rg -v '^#|^$' bootstrap.toml
35:change-id = 138986
40:[llvm]
171:[gcc]
181:[build]
458:[install]
485:[rust]
710:lld = false
809:[target.x86_64-unknown-linux-gnu]
962:[dist]

❯ PATH=~/Projects/wild/:$PATH ./x.py t --skip tests/debuginfo --no-fail-fast
...
failures:
    [ui] tests/ui/backtrace/dylib-dep.rs
    [ui] tests/ui/backtrace/line-tables-only.rs
    [ui] tests/ui/runtime/backtrace-debuginfo.rs
...
failures:
    [run-make] tests/run-make/embed-source-dwarf
    [run-make] tests/run-make/repr128-dwarf
    [run-make] tests/run-make/strip
```

I have focused on `tests/ui/backtrace/line-tables-only.rs` for no specific reason and found out this test builds shared object and an executable. The problem lies within shared object created by Wild, the binary itself works fine when linked by Wild but with shared object created by ld.bfd. The failure:
```
~/rustc_debug/1
❯ OUT=auxiliary/libline_tables_only_helper.so ../0/run-with ~/Projects/wild/target/debug/wild && ./bin
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/14.2.1/liblto_plugin.so is not yet supported
Backtrace [
    { fn: "line_tables_only_helper::capture_backtrace::{{closure}}" },
    { fn: "baz" },
    { fn: "bar" },
    { fn: "foo" },
    { fn: "line_tables_only_helper::capture_backtrace" },
    { fn: "line_tables_only::main" },
    { fn: "std::sys::backtrace::__rust_begin_short_backtrace" },
    { fn: "std::rt::lang_start::{{closure}}" },
    { fn: "std::rt::lang_start_internal" },
    { fn: "main" },
    { fn: "__libc_start_main" },
    { fn: "_start" },
]

[...]

PR #697, opened 2025-04-22
Alternative fix and test for #694

davidlattimore:
See individual commits for details. This also adds propagation of sh_entsize, which was necessary to make tests not fail now that we're not ignoring "section.debug*"


mati865:
Ah, I wasn't sure whether string sections are supposed to be loaded. I suspected they were supposed to use some other mechanism.

Is entsize observable using tools like readelf or debuginfo? I  compared them manually and didn't see any differences.

mati865 (approved):

Issue #678, opened 2025-04-17
RISC-V support


davidlattimore:
Martin has already started work on this :)

davidlattimore:
We're intending to do development and testing on VisionFive 2 boards.

Martin suggested using [these instructions](https://canonical-ubuntu-boards.readthedocs-hosted.com/en/latest/how-to/starfive-visionfive-2/) to set up the board with Ubuntu. I've followed these and they worked well.

Parts that I needed to do this:

* VisionFive 2 board (8GiB of RAM recommended)
* SSD (optional, but likely gives better speeds than an SD card)
* USB C power supply. I'm using a 42W supply, but I suspect a 30W would do fine. If you're not using an SSD, then a lower wattage can be used.
* Case. Optional, but recommended.
* Ethernet cable and a router to plug it into. It may alternatively be possible to use a USB wifi adapter, but I haven't tried.
* SD card. Needed for setup, but not needed after that provided you're using an SSD.
* USB to UART adapter. Needed for setup, but not needed after that. Needs to plug into the DuPont pins on the board. If your adapter has pins, then you'll likely need some female to female DuPont wires.

I'm not yet sure what will work well for development. vscode remote doesn't appear to work - it complains UnsupportedArch. I'd considering just editing the files on my laptop, then using lsyncd to keep the device in sync.

mati865:
I suppose embedded boards like the Sipeed MaixDuino (64-bit CPU) are out of consideration for the near future.

davidlattimore:
> I suppose embedded boards like the Sipeed MaixDuino (64-bit CPU) are out of consideration for the near future.

From what I can see, the CPU has 8MiB of SRAM. Unless the board itself has more memory (I couldn't see any mention of any) then I suspect that running a modern Linux on it might be a bit on the tricky side, especially if we want enough free RAM to be able to run sshd, gcc, rustc etc.

mati865:
I wouldn't dare to run Linux there. Only to cross-compile the binary and flash it directly (like any other embedded platform).

lu-zero:
In my experience boards with a spacemit cpu are way nicer since you can have a nvme and more ram, but I'd still cross-build and then execute ^^

marxin:
Just wanted to mention, I've also followed the official Ubuntu guide for StarFive2, and it's great! And I've just switched to the latest 25.04 release image available here: https://ubuntu.com/download/risc-v#starfive-visionfive-2.

marxin:
[...]

PR #704, opened 2025-04-26
Rudimentary RISC-V support

marxin:
Implements the very basic support for RISC-V architecture. Right now, I'm going to open a draft PR to get CI run results available with QEMU.

#678


marxin:
Finally, I have reached a point where I can successfully run all the wild integration tests, and I think it's time to review this PR. There are a bunch of missing pieces for the port, and I'm going to create a subsequent issue in the #678.

libwild/src/elf_writer.rs
```diff
@@ -1944,11 +1978,131 @@ struct SectionInfo {
...
+        .with_context(|| {
+            format!(
+                "Missing resolution for: {}",
+                layout.symbol_db.symbol_debug(local_symbol_id)
+            )
+        })?;
+    Ok((resolution, symbol_index, local_symbol_id))
+}
+
+const LOW6_MASK: u64 = 0b0011_1111;
```
davidlattimore: If this is only used in the following function, I'd be somewhat inclined to move it into the function

libwild/src/layout.rs
```diff
@@ -2865,11 +2872,19 @@ fn resolution_flags(rel_kind: RelocationKind) -> ResolutionFlags {
...
+        | RelocationKind::AbsoluteAddition
+        | RelocationKind::AbsoluteSubtraction
+        | RelocationKind::AbsoluteSubtractionWord6
         | RelocationKind::Relative
+        | RelocationKind::RelativeRISCVLow12
         | RelocationKind::DtpOff
         | RelocationKind::TpOff
         | RelocationKind::TpOffAArch64
-        | RelocationKind::SymRelGotBase => ResolutionFlags::DIRECT,
+        | RelocationKind::TpOffRISCV
```
davidlattimore: Other acronyms only have their first letter uppercase (e.g. "TP"), so it'd be more consistent to do the same for RISCV. Whether the V is lower or uppercase, I don't have any opinion on

test-config.toml.sample
```diff
@@ -1,3 +1,3 @@
 rustc_channel = FILLHERE # `"stable"`, `"beta"`, `"nightly"` or "default"
-use_qemu = FILLHERE # `true` or `false`
+qemu_arch = FILLHERE # `aarch64` or `riscv64`
```
davidlattimore, line 2: Do you think we should make this a list instead of a single value? That way we can test all three architectures in one run.

linker-utils/src/riscv64.rs
```diff
@@ -0,0 +1,403 @@
...
+        size,
+        mask,
+        range,
+        alignment,
+    })
+}
+
+impl RISCVInstruction {
+    // Encode computed relocation value and store it based on the encoding of an instruction.
+    // A handy pages where one can easily find instruction encoding:
```
davidlattimore: s/pages/page

Cargo.toml
```diff
@@ -44,12 +44,13 @@ iced-x86 = { version = "1.17.0", default-features = false, features = [
...
 ] }
 itertools = "0.14.0"
+leb128 = "0.2.5"
 libc = "0.2.152"
 memchr = "2.6.0"
 memmap2 = "0.9.0"
 mimalloc = { version = "0.1", default-features = false }
 normalize-path = "0.2.0"
-object = { version = "0.36.7", default-features = false, features = [
+object = { git = "https://github.com/gimli-rs/object", rev = "61474b7de8074f91b78ad24ff2424e54a22f8b32", default-features = false, features = [
```
davidlattimore, line 53: I guess it's a good thing that we just did a release of wild, since our next release will need to wait until after this is resolved :)

libwild/src/elf_writer.rs
```diff
@@ -1658,12 +1671,19 @@ impl<'data> ObjectLayout<'data> {
...
+        for (i, rel) in relocations.iter().enumerate() {
             if modifier == RelocationModifier::SkipNextRelocation {
                 modifier = RelocationModifier::Normal;
                 continue;
             }
             let offset_in_section = rel.r_offset.get(LittleEndian);
+            // The iterator is used for e.g. R_RISCV_PCREL_HI20 & R_RISCV_PCREL_LO12_I pair of relocations where the later
+            // one actually points to a label of the HI20 relocations and thus we need to find it. The relocation is typically
+            // right before the LO12_* relocation.
+            let relocations_to_search = relocations[..i]
```
davidlattimore: When I benchmark a build of this PR against main, I see about a 6% increase in instruction count. The effect on walltime is less (about 2.9-4.6%). I'm not sure what the cause is. One possibility is just from having more branches in some of our large match statements. If that's the case, then it's likely hard to do much about at present. The other possibility is some bit of code in a common path like this code here.

Do you see a small slow down too?

If it is this bit of code here - perhaps unlikely, but possible - then the two options that come to mind are (a) pass a closure that produces this iterator or (b) pass `relocations` and `i`.

davidlattimore (commented):
Looks good. Nice work! A few small comments. I'd say don't spend too long on any of them - if any look like they might take too long, then we can leave them until after merging.

libwild/src/layout.rs
marxin: Do you mean using `TpOffRiscV` ? Or any other variant? Anyways, something #774 should address so that we don't need the special target variants.

test-config.toml.sample
marxin: Yes, please! Could you make a refactoring of this once we merge it?

davidlattimore: Sure

Cargo.toml
marxin: Yeah! I've created #778 to track the thing. If we really need to do a release, we can still switch to a proper release and add a couple of newly added constants.

libwild/src/layout.rs
davidlattimore: Yeah, either `TpOffRiscV` or `TpOffRiscv` would be fine

libwild/src/elf_writer.rs
marxin: > Do you see a small slow down too?

No, I've just tried building `mold` both w/ and w/o debug info, and the speed might be 1% slower now, not more.

> If it is this bit of code here - perhaps unlikely, but possible - then the two options that come to mind are (a) pass a closure that produces this iterator or (b) pass `relocations` and `i`.

Yeah, I can experiment with that. It's true we currently build the iterator for every relocation, where in most cases, it's unused.

libwild/src/layout.rs
marxin: Done!

libwild/src/elf_writer.rs
marxin: @davidlattimore I reworked the logic around `relocations_to_search`, can you re-measure the numbers after the change?

davidlattimore: That helped a bit. The instruction count (which is very stable from run to run) now only increases by 5.4% as opposed to 6.7%.

Do you see any significant change on instruction counts? I'm using https://github.com/andrewrk/poop to profile.

marxin: Yesterday, I spent some time with the bisection and I can confirm the instruction count increase with this change, where I managed to identify 2 steps of the growth:

1. Addition of handling of `AbsoluteAddition`/`AbsoluteSubtraction`: 29.6G -> 30.5G insns
1. Addition of  `get_pair_subtraction_relocation_value` to `apply_debug_relocation`: 31.7G -> 32.1 insns

Which is kind of expected, given the logic was added to the hottest functions. Moreover, I noticed the switch from anyhow to the custom error time increased the instruction count by ~0.7G (noticed that because I started merging `main` branch at some point).

libwild/src/layout.rs
```diff
@@ -1597,8 +1597,13 @@ impl<'data> Layout<'data> {
     /// Returns the memory address of the start of the TLS segment used by the RISC-V.
     pub(crate) fn tls_start_address_riscv64(&self) -> u64 {
         let tdata = self.section_layouts.get(output_section_id::TDATA);
-        let tls_start = tdata.mem_offset;
-        tdata.alignment.align_down(tls_start)
+        let tbss = self.section_layouts.get(output_section_id::TBSS);
```
davidlattimore: Would it work to do the same as `tls_start_address` and use segment addresses rather than section addresses?

marxin: Yep, could be simplified!

davidlattimore:
> I'm happy to announce the current RISC-V branch can bootstrap `Wild` itself and all tests are green 🎊

Awesome!

> One interesting change I had to include is [6893073](https://github.com/davidlattimore/wild/commit/689307322f9d97016ea24cd324c9cb4ead8a7711), where we were missing emission of the TLSMOD (that was allocated) for an executable, where it references symbols from a dynamic library. Other archs probably don't suffer from it as they process a GD -> IE relaxation as the shared library presence is marked with `DT_NEEDED`?

Thanks for pointing out that change. That change looks right to me. I think you're right about initial-exec relaxations being the reason we hadn't noticed that it was wrong.

Issue #705, opened 2025-04-26
copy_relocations_c fails with WILD_TEST_CROSS=aarch64 on openSUSE TW


marxin:
The following fails for me on the `main` branch:

```
failures:

---- integration_test::program_name_33___copy_relocations_c__ stdout ----
wild: /home/marxin/Programming/wild/wild/tests/build/copy-relocations.c-default-aarch64.wild
ld: /home/marxin/Programming/wild/wild/tests/build/copy-relocations.c-default-aarch64.ld
dynsym.foo3.section
  wild .bss
  ld 
```

@davidlattimore Can you please take a look?

davidlattimore:
The openSUSE docker image doesn't have the required packages setup to cross compile for aarch64. Do you know what packages are needed?

PR #706, opened 2025-04-27
[suse docker] add missing setup steps

lapla-cogito:
related: #705 

Some setup steps are missing to run tests with `WILD_TEST_CROSS=aarch64` inside the OpenSUSE Docker container.


davidlattimore:
Thanks!

PR #713, opened 2025-04-27
Ignore -Bsymbolic* arguments until implemented

mati865:
Ignoring them hurts the performance of resulting libs, but doesn't justify failing with an error.
Print a warning instead.


mati865:
You can close the PR if you disagree.

PR #716, opened 2025-04-27
integration tests: Run test configs in the order they're listed

davidlattimore:
We were previously running them in a non-deterministic order based on iteration of a hashmap. This could get confusing when rerunning a failing test and expecting the same output file as last time to get updated.


mati865:
Could have been changed to BTreeMap, would be simpler.

davidlattimore:
> Could have been changed to BTreeMap, would be simpler.

I initially sorted the output, but then decided that insertion order made more sense than alphabetical order.

Issue #721, opened 2025-04-29
Diff program segments in linker-diff


davidlattimore:
I noticed that Wild sets the alignment on the `GNU_STACK` segment to 1, whereas GNU ld sets it to 0x10. We should fix this, but ideally we should fix it in conjunction with making linker-diff detect this difference.

Not everything about program segments can be diffed, since things like file offsets, sizes etc are expected to be different. However, the flags (write, execute etc) should be the same as should the alignment.

I'd suggest perhaps using `version_diff.rs` as a guide for how this might be written - mostly because it's pretty short. The idea is that you write a function that takes a `&crate::Binary` as an argument and returns `Result<FieldValues>`. The field values can be of the form `{segment_type}.{property}` e.g. `GNU_STACK.alignment`. For the `LOAD` segments, you can either ignore them or include their flags as part of their key rather than as a value. e.g. `LOAD.RX.alignment` for the alignment of the executable `LOAD` segment.

Once this is implemented, you should find tests failing due to the incorrect alignment of GNU_STACK. The failure should look something like this:

```
segment.GNU_STACK.alignment
  wild 0x1
  ld 0x10
```

I'd probably suggest adding `segment.GNU_STACK.alignment` to the default ignore rules in `linker-diff/src/lib.rs`, then in a subsequent PR, removing it and fixing the alignment.

lapla-cogito:
I’ll work on this.

PR #723, opened 2025-04-29
ignore `segment.GNU_STACK.alignment` temporary

lapla-cogito:
related: #721 

Once this PR is merged, I'll start working on rewriting linker-diff.


davidlattimore:
Thanks for volunteering to work on #721! I'm not sure it makes sense to add an ignore rule that linker-diff doesn't yet produce. Part of testing the new diff code is making sure that it does indeed detect the stack alignment problem. Also, it's possible that in the process of implementing segment diffing, you might find other things that need to be temporarily ignored. I'd probably suggest just adding them all in the same PR as the one that adds the diff support. That said, I don't feel strongly, so I'm prepared to just merge this if that's what you want.

If you comment on the issue, I'll assign it to you. I can't see a way to assign it unless you first comment on it.

Issue #729, opened 2025-04-29
Support version scripts without newlines


mati865:
Trying to build LLVM libs fails on parsing of version scripts, minimal reproducer here:
```diff
diff --git a/libwild/src/version_script.rs b/libwild/src/version_script.rs
index 32e1394852..06b1683f52 100644
--- a/libwild/src/version_script.rs
+++ b/libwild/src/version_script.rs
@@ -474,4 +474,13 @@
             ["foo2"],
         );
     }
+
+    #[test]
+    fn single_line_version_script() {
+        let data = VersionScriptData {
+            raw: r#"LLVM_21.0 { global: *; };"#,
+        };
+        let script = VersionScript::parse(data).unwrap();
+        assert!(script.globals.matches_all);
+    } 
 }
 
 
```
With message `Unsupported symbol pattern 'global: *; }'`.

davidlattimore:
I'll take a look. I'm probably going to investigate whether switching to winnow for parsing can be done without slowdown first.

PR #736, opened 2025-05-03
Parse version scripts as bytes and use winnow for parsing

davidlattimore:
Two separate commits. The change to bytes for parsing speeds up parsing by about 7%. The change to winnow seems to be performance neutral., This fixes parsing of version scripts on a single line. #729


mati865:
Winnow version looks much better and LLVM with the default config (with shared objects included) builds fine.

PR #740, opened 2025-05-04
linker-diff: Detect differences between program segments

### LINKED ISSUE #721: Diff program segments in linker-diff

lapla-cogito:
close #721 

Modify linker-diff to detect differences in program segments. Also, update libwild to set `GNU_STACK.alignment` to `0x10`, consistent with GNU ld.

### DIFF
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -43,6 +43,7 @@ mod gnu_hash;
 mod header_diff;
 mod init_order;
 pub(crate) mod section_map;
+mod segment;
 mod symbol_diff;
 mod symtab;
 mod trace;
@@ -570,6 +571,7 @@ impl Report {
         version_diff::report_diffs(self, objects);
         debug_info_diff::check_debug_info(self, objects);
         symbol_diff::report_diffs(self, objects);
+        segment::report_diffs(self, objects);
 
         match arch {
             ArchKind::X86_64 => {
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -235,6 +235,8 @@ impl Config {
                 // If we don't optimise a TLS access, then we'll have references to __tls_get_addr,
                 // when GNU ld doesn't.
                 "dynsym.__tls_get_addr.*",
+                // temporary
+                "segment.LOAD.*",
             ]
             .into_iter()
             .map(ToOwned::to_owned),
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -235,8 +235,8 @@ impl Config {
                 // If we don't optimise a TLS access, then we'll have references to __tls_get_addr,
                 // when GNU ld doesn't.
                 "dynsym.__tls_get_addr.*",
-                // temporary
-                "segment.LOAD.*",
+                // GNU ld emits two segments, whereas wild emits only a single segment.
+                "segment.LOAD.R.*",
             ]
             .into_iter()
             .map(ToOwned::to_owned),
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -237,6 +237,13 @@ impl Config {
                 "dynsym.__tls_get_addr.*",
                 // GNU ld emits two segments, whereas wild emits only a single segment.
                 "segment.LOAD.R.*",
+                // We haven't provided an implementation that is compatible with existing linkers.
+                "segment.PT_NOTE.*",
+                "segment.PT_INTERP.*",
+                "segment.PT_PHDR.*",
+                "segment.PT_GNU_RELRO.*",
+                "segment.PT_GNU_STACK.*",
+                "segment.PT_GNU_PROPERTY.*",
             ]
             .into_iter()
             .map(ToOwned::to_owned),
--- a/linker-diff/src/segment.rs
+++ b/linker-diff/src/segment.rs
@@ -1,0 +1,55 @@
+use crate::header_diff::Converter;
+use crate::header_diff::DiffMode;
+use crate::header_diff::FieldValues;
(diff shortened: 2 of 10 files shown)


lapla-cogito:
I'm marking this PR as a draft because I'm still unsure about how to handle `LOAD` segments. Currently, `LOAD` segments are excluded in `linker-diff/src/lib.rs`, but if I comment out that exclusion, a large number of differences like the following are reported:
```
segment.LOAD.R.alignment
  wild 0x1000
  ld 0x1000,0x1000
```
I don't believe these differences should be reported in the first place, so I'll look into whether they can be suppressed.

linker-diff/src/segment.rs
```diff
@@ -0,0 +1,55 @@
...
+    let e = LittleEndian;
+    let mut values = FieldValues::default();
+
+    for segment in object.elf_file.elf_program_headers() {
+        let p_type = segment.p_type(e);
+        let p_flags = segment.p_flags(e);
+        let p_align = segment.p_align(e);
+
+        match p_type {
+            PT_GNU_STACK => {
```
davidlattimore: It'd be good to do this for all segments except PT_LOAD. I'd suggest having a function that can convert the segment type to a string. e.g. see x86_64_rel_type_to_string for how this might be done.

linker-diff/src/lib.rs
```diff
@@ -234,6 +235,8 @@ impl Config {
                 // If we don't optimise a TLS access, then we'll have references to __tls_get_addr,
                 // when GNU ld doesn't.
                 "dynsym.__tls_get_addr.*",
+                // temporary
```
davidlattimore: I think it's reasonable to ignore these differences. If you wanted to ignore less, you could perhaps just ignore `segment.load.R.*`, since I think that's the main one where GNU ld emits two segments and we only emit one.

davidlattimore (commented):
Thanks for working on this!

linker-diff/src/lib.rs
lapla-cogito: Thanks! I’ve adjusted it to specifically ignore only `segment.LOAD.R.*`.

linker-diff/src/segment.rs
lapla-cogito: Thanks! I implemented `segment_type_to_string`. As a result, many differences are now being detected, so I'll take a closer look to decide which ones should be ignored and which ones should be addressed.

lapla-cogito: Looking at the [CI results](https://github.com/davidlattimore/wild/actions/runs/14826999703/job/41621556206), I feel that `segment.PT_GNU_PROPERTY.*` should be fixed, while the others (`segment.PT_NOTE.*`, `segment.PT_INTERP.*`, `segment.PT_PHDR.*`, `segment.PT_GNU_STACK.*`, and `segment.PT_GNU_RELRO.*`) can reasonably be ignored. Does that approach make sense?

linker-utils/src/elf.rs
```diff
@@ -209,6 +209,29 @@ pub fn aarch64_rel_type_to_string(r_type: u32) -> Cow<'static, str> {
...
+        PT_PHDR,
+        PT_TLS,
+        PT_GNU_EH_FRAME,
+        PT_GNU_STACK,
+        PT_GNU_RELRO,
+        PT_GNU_PROPERTY
+    ] {
+        Cow::Borrowed(name)
+    } else {
+        unreachable!()
```
davidlattimore: This could be reached if one of the linkers emitted a segment type that we don't know about. I'd suggest changing this to return some string that contains the numeric value. e.g.  `Cow::Owned(format!("UNKNOWN_P_TYPE_{r_type}"))`

linker-diff/src/segment.rs
davidlattimore: It's fine by me to ignore all of them, then we can fix whichever ones we think should be fixed in subsequent commits. It's also fine to fix one or two in this PR, provided the fixes are trivial and leave the rest as ignored. As to which ones should eventually be fixed and which ones aren't worth ever trying to make the same, it's hard to say without looking at why each one is different. That's probably easier to do once this change is merged.

linker-utils/src/elf.rs
lapla-cogito: Thanks! Fixed.

linker-diff/src/segment.rs
lapla-cogito: Alright. For now, I’ve made it so that all the items mentioned above are ignored. If anything seems in need of fixing later, I (or someone else) will address it separately.

PR #743, opened 2025-05-06
Update dependabot schedule and replace deprecated link checker

lapla-cogito:
Commits:
1. Set dependabot GitHub Actions update interval to weekly
Daily updates for `github-actions` via dependabot are likely unnecessary, so I changed the interval to weekly to align with `cargo`.

2. Replace `github-action-markdown-link-check` with `action-linkspector`
`github-action-markdown-link-check` is deprecated and was producing [false positives for crates.io](https://github.com/davidlattimore/wild/actions/runs/14595234863). I tested the new workflow by temporarily modifying the trigger conditions in a forked repository: https://github.com/lapla-cogito/wild/actions/runs/14855707088/job/41708454860

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: github-actions
     directory: /
     schedule:
-      interval: daily
+      interval: weekly
   - package-ecosystem: cargo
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thanks! Those sound like good changes

PR #744, opened 2025-05-06
Identify kind of archive members instead of copying from their parent

mati865:
It also means when we encounter an archive with LTO objects, we correctly reject them. Previously:
```
Error: Failed to parse object file `librustc_codegen_llvm-6a14a3ed70dffd38.rlib @ ef10e86dc40538c1-PassWrapper.o`

Caused by:
    Unsupported ELF header
```
Now:
```
Error: Failed to parse archive `librustc_codegen_llvm-6a14a3ed70dffd38.rlib @ ef10e86dc40538c1-PassWrapper.o`

Caused by:
    LLVM IR (LTO mode) is not supported yet
```

Initially observed in https://github.com/rust-lang/rust/pull/140624

### DIFF
--- a/libwild/src/archive_splitter.rs
+++ b/libwild/src/archive_splitter.rs
@@ -6,6 +6,8 @@ use crate::error::Result;
 use crate::file_kind::FileKind;
 use crate::input_data::InputData;
 use crate::input_data::InputRef;
+use anyhow::Context;
+use anyhow::bail;
 use rayon::iter::IntoParallelRefIterator;
 use rayon::iter::ParallelIterator;
 use std::fmt::Display;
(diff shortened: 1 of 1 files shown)


libwild/src/archive_splitter.rs
```diff
@@ -33,7 +34,17 @@ pub fn split_archives<'data>(input_data: &InputData<'data>) -> Result<Vec<InputB
...
+                                    .with_context(|| {
+                                        format!(
+                                            "Failed to parse archive `{} @ {}`",
+                                            f.filename.to_string_lossy(),
+                                            archive_entry
+                                                .identifier(extended_filenames)
+                                                .as_path()
+                                                .to_string_lossy()
+                                        )
+                                    })?,
```
mati865: Maybe it should be extracted into `kind` variable?

davidlattimore: I think it's fine like this. Extracting to a variable would also be fine.

davidlattimore (approved):
I was curious as to what uses `InputBytes::kind`. It seems like the only use is in parsing where we check if the file kind is a shared object. So I guess now if we encounter a shared object inside an archive, we'll treat it as a shared object, which would be weird, but it would have been weird treating it a regular object too.

So this change, at least at the moment, is really just about the improved error message. Given that, I wonder if it might make sense to check if the detected kind is `FileKind::Object` and error if not. But up to you. It's also fine to just submit this as-is.

mati865:
Rust's rlibs (which are archives) can contain unexpected things.
A bit offtopic, but on Windows, they can contain import libraries alongside normal objects, which turned out to be a problem for ld.bfd when some conditions are met.

I hope such a thing doesn't happen on ELF platforms, but I'll test it later on to be on the safe side.

mati865:
Appears to work fine with Rust.

PR #747, opened 2025-05-10
Use astral-sh fork of cargo-dist

mati865:
This reduces the baseline from Ubuntu 24.04 to Ubuntu 22.04.


mati865:
Some of the folks over the r/Rust suggest using https://github.com/crabnebula-dev/cargo-packager, but it doesn't have such a nice, ready to go workflow.

PR #753, opened 2025-05-11
Set DF_1_NODELETE when -z nodelete is passed

mati865:


libwild/src/args.rs
```diff
@@ -58,6 +58,7 @@ pub struct Args {
     pub(crate) file_write_mode: Option<FileWriteMode>,
     pub(crate) no_undefined: bool,
     pub(crate) needs_origin_handling: bool,
+    pub(crate) needs_nodelete_handling: bool,
```
mati865, line 61: I'm not sure about the naming, just followed `needs_origin_handling`.

PR #764, opened 2025-05-17
remove useless use of anyhow from slice.rs

m-hugo:
fixes #763


m-hugo:
this pull request is best reviewed commit-by-commit to minimize diff noise

marxin:
Good catch, thanks for the fix!

PR #768, opened 2025-05-22
Use debug info to show where error came from

davidlattimore:


wild/tests/sources/undefined_symbols.c
```diff
@@ -12,8 +12,10 @@
 //#ExpectError:  undefined_strong
 
 //#Config:executable:default
+//#CompArgs:-g
 //#ExpectError:Undefined symbols:
 //#ExpectError:  undefined_strong
+//#ExpectError:undefined_symbols.c
```
mati865: Note this will only match the last occurrence: https://github.com/davidlattimore/wild/blob/11e92314026a1af44d5ae1f9e90948e428e337f0/wild/tests/integration_tests.rs#L772-L773

davidlattimore: Thanks for noticing that! I've made a separate PR to fix that. https://github.com/davidlattimore/wild/pull/769 - I ended up needing to remove one of the existing "ExpectError" commands, because GNU ld doesn't emit the string "Undefined symbols".

mati865: IIRC ld.bfd uses either "undefined symbols" or "undefined references", maybe it uses the second form here. Anyway, skipping it is not a big deal.

PR #770, opened 2025-05-24
Support `-Bsymbolic-functions`

### LINKED ISSUE #494: Support `-Bsymbolic-functions`
This obscure flag is used by LLVM when building it with `-DLLVM_LINK_LLVM_DYLIB=ON`. 
Might depend on https://github.com/davidlattimore/wild/issues/41

lapla-cogito:
close #494

### DIFF
--- a/libwild/src/symbol_db.rs
+++ b/libwild/src/symbol_db.rs
@@ -2,6 +2,7 @@
 //! information about where each symbol can be obtained.
 
 use crate::InputLinkerScript;
+use crate::args;
 use crate::args::Args;
 use crate::args::OutputKind;
 use crate::error::Result;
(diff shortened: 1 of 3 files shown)


lapla-cogito:
I verified this using [a test for this option in mold](https://github.com/rui314/mold/blob/main/test/bsymbolic-functions.sh). Specifically, I confirmed that when this option is passed to Wild for linking, the entry corresponding to `bar()` does not appear in the output of `readelf -r`.

libwild/src/symbol_db.rs
```diff
@@ -1025,7 +1026,13 @@ fn value_flags_from_elf_symbol(sym: &crate::elf::Symbol, args: &Args) -> ValueFl
         || args.output_kind().is_static_executable()
         // Symbols defined in an executable cannot be interposed since the executable is always the
         // first place checked for a symbol by the dynamic loader.
-        || (args.output_kind().is_executable() && !is_undefined);
+        || (args.output_kind().is_executable() && !is_undefined)
+        // `-Bsymbolic-functions`
+        || (
+            sym.st_visibility() == object::elf::STV_DEFAULT
```
davidlattimore: This line seems unnecessary. If the visibility isn't default, then the symbol will always be interposable.

davidlattimore (commented):
Thanks! Looks good

libwild/src/symbol_db.rs
lapla-cogito: Thanks! Fixed.

PR #782, opened 2025-05-27
Support `-Bsymbolic`, `-Bsymbolic-non-weak`, `-Bsymbolic-non-weak-functions` and `-Bno-symbolic`

lapla-cogito:
Since we previously added support for `-Bsymbolic-functions`, we could similarly support these options using the same approach.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -158,8 +158,7 @@ pub(crate) enum BSymbolicKind {
     All,
     Functions,
     NonWeakFunctions,
-    // TODO: Handle below cases
-    // NonWeak,
+    NonWeak,
 }
 
 pub const WILD_UNSUPPORTED_ENV: &str = "WILD_UNSUPPORTED";
@@ -205,7 +204,6 @@ const IGNORED_FLAGS: &[&str] = &[
     "fix-cortex-a53-835769",
     "fix-cortex-a53-843419",
     "no-export-dynamic",
-    "Bsymbolic-non-weak", // LLD specific
 ];
 
 // These flags map to the default behavior of the linker.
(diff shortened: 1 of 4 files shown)


lapla-cogito:
I'm not sure if we should include tests for these options in this patch. I checked these imeplementation locally using mold's corresponding tests (as I did when implementing `-Bsymbolic-functions`).
Since we're planning to run mold's test suite automatically on Wild soon, I'm wondering if any additional work is actually needed here.

libwild/src/args.rs
```diff
@@ -380,8 +378,17 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
...
             args.b_symbolic = BSymbolicKind::Functions;
+        } else if long_arg_eq("Bsymbolic-non-weak-functions") && !args.ignore_bsymbolic {
+            args.b_symbolic = BSymbolicKind::NonWeakFunctions;
+        } else if long_arg_eq("Bsymbolic-non-weak") && !args.ignore_bsymbolic {
+            args.b_symbolic = BSymbolicKind::NonWeak;
+        } else if long_arg_eq("Bsymbolic") && !args.ignore_bsymbolic {
+            args.b_symbolic = BSymbolicKind::All;
+        } else if long_arg_eq("Bno-symbolic") {
+            args.b_symbolic = BSymbolicKind::None;
+            args.ignore_bsymbolic = true;
```
davidlattimore: I'm unsure that the logic here matches the behaviour of GNU ld. What I'd expect, and what I seem to observe if I experiment with GNU ld, is that -Bno-symbolic cancels the existing symbolic setting, but doesn't prevent it from subsequently being reset. i.e. if I run GNU ld with `-Bno-symbolic -Bsymbolic` it behaves the same as if I ran it with just `-Bsymbolic`. If I reverse the arguments - `-Bsymbolic -Bno-symbolic`, then it behaves as if I just passed `-Bno-symbolic`. i.e., in both cases the later argument takes precedence.

lapla-cogito: I see. According to the GNU ld man page, it looks like `-Bno-symbolic` doesn't cancel the options that follow. I had implemented it that way in mold because it does cancel subsequent options there, but it makes more sense to follow GNU ld's behavior for consistency.

Now I'm going to update tests appropriately.

lapla-cogito:
As discussed on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Improve.20Wild.20linker.20test.20suites/near/521480146), adding tests for these options appears to require modifications to linker-diff. First, I disabled cross-compilation in `shared.c` to prevent test failures on QEMU. However, even in this change, skipping tests for `-Bsymbolic-non-weak` with lld causes linker-diff to panic (since there's no linker output to compare against wild's output). Therefore, for now I've only decided to skip testing for `-Bsymbolic-non-weak` and add a TODO comment. I will implement the new directive you proposed on Zulip once this patch is merged. Is this approach acceptable? @davidlattimore

PR #784, opened 2025-05-28
Refactor how internal symbols are stored

### LINKED ISSUE #773: TLS_MODULE_BASE_SYMBOL_NAME: refactoring
I noticed the current approach, where we define the symbol as both `start_symbol_name` and `end_symbol_name` (based on whether we build an executable or shared library), is fragile. The symbol won't be emitted if the section is not used!
Plus, we should get rid of `pub(crate) fn start_symbol_name` and `pub(crate) fn end_symbol_name` functions.

@davidlattimore Can you help me with that, please?

davidlattimore:
This refactoring mostly affects _TLS_MODULE_BASE_, but also changes how other internal symbols and how --undefined= is handled.

Fixes #773

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -31,7 +31,6 @@ use std::sync::atomic::Ordering;
 /// up file and memory offsets.
 pub const NON_PIE_START_MEM_ADDRESS: u64 = 0x400_000;
 
-pub(crate) const TLS_MODULE_BASE_SYMBOL_NAME: &str = "_TLS_MODULE_BASE_";
 pub(crate) const GLOBAL_POINTER_SYMBOL_NAME: &str = "__global_pointer$";
 
 pub(crate) type FileHeader = object::elf::FileHeader64<LittleEndian>;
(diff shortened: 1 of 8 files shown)


marxin:
> If we want to change that, we can do that separately.

Yes, I would recommend doing that.

davidlattimore:
> Yes, I would recommend doing that.

If we'd like to keep some symbols even if the section that they're contained within, could you raise a separate issue for that? It'd be good to have written down why we need to keep the symbol. e.g. is it just because GNU ld keeps it, or does something break without said symbol? That would also help when it comes to writing a test.

Somewhat relatedly, I noticed that in linker scripts, you can define a symbol in several different ways. e.g. `symname = .'` or `PROVIDE(symname = .);`. The latter apparently means that `symname` will only be defined if something else doesn't define it, whereas the former will always define it, so might result in a duplicate definition error. So it's a bit like weak, but not exactly the same.

> One related topic is the `GLOBAL_POINTER_SYMBOL_NAME` symbol that should be emitted conditionally only for the RISC-V target.

Yes, that should hopefully be easier now.

Issue #733, opened 2025-04-30
RISC-V: .dynamic.DT_PREINIT_ARRAY/.dynamic.DT_PREINIT_ARRAYSZ likely defined by BFD if empty


marxin:
It seems to me that the BFD linker defines the sections even if they are de facto empty.

PR #789, opened 2025-05-28
Support `.preinit_array` section by linker-diff

marxin:
Fixes: #733


linker-diff/src/lib.rs
```diff
@@ -273,9 +273,6 @@ impl Config {
                     // #732
                     ".dynamic.DT_RELA",
                     ".dynamic.DT_RELAENT",
-                    // #732
```
marxin, line 276: That comment here was wrong, it's supposed to be `#733`.

Issue #790, opened 2025-05-28
RISC-V: debug info support is not working


marxin:
I noticed we wrongly process the debug relocations for even a trivial C program:
```
Relocation section '.rela.debug_aranges' at offset 0x880 contains 4 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
...
0000000000000018  0000001b00000024 R_RISCV_ADD64          0000000000000014 .Letext0 + 0
0000000000000018  0000001a00000028 R_RISCV_SUB64          0000000000000000 .Ltext0 + 0
```

It's a relocation that needs to calculate `.Letext0` -  `.Ltext0`, but it seems to me the `resolution` in `apply_debug_relocation` contains the same address for both symbols:
```
2025-05-28T12:58:17.856725Z TRACE write_file{filename=a-main.o}:{section_name=".debug_aranges"}:relocation{address=24 address_hex=18}: libwild::elf_writer: debug relocation applied rel_info.kind=AbsoluteAddition value=8118 original_value=8118 value_hex=1fb6 symbol_name=.Letext0
2025-05-28T12:58:17.856944Z TRACE write_file{filename=a-main.o}:{section_name=".debug_aranges"}:relocation{address=24 address_hex=18}: libwild::elf_writer: debug relocation applied rel_info.kind=AbsoluteSubtraction value=0 original_value=8118 value_hex=0 symbol_name=.Ltext0
```

where there's a symbol table for the 2 involved symbols:
```
   26: 0000000000000000      0 NOTYPE  LOCAL  DEFAULT        1 .Ltext0
   27: 0000000000000014      0 NOTYPE  LOCAL  DEFAULT        1 .Letext0
```

Note the difference from x86_64 is related to the fact that all debug info uses section + offset format:
```
0000000000000054  000000070000000a R_X86_64_32            0000000000000000 .debug_str + 3e
0000000000000062  000000070000000a R_X86_64_32            0000000000000000 .debug_str + 56
```

@davidlattimore Do you know what can be responsible for the incorrect symbol resolution of the aforementioned labels?

marxin:
> It's a relocation that needs to calculate `.Letext0` - `.Ltext0`, but it seems to me the `resolution` in `apply_debug_relocation` contains the same address for both symbols:

Now I understand the issue, it's caused by the fact that we don't load relocations (in `ObjectLayoutState::load_debug_section`).

marxin:
[...]

PR #791, opened 2025-05-28
RISC-V: fix debug info relocations

marxin:
Necessary on RISC-V target where labels are referenced by relocations of the debug sections.

Should handle the first limitation mentioned in #790.

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -83,6 +83,10 @@ impl crate::arch::Arch for AArch64 {
         );
         Ok(())
     }
+
+    fn local_symbols_in_debug_info() -> bool {
+        false
+    }
 }
 
 #[derive(Debug, Clone)]
--- a/libwild/src/arch.rs
+++ b/libwild/src/arch.rs
@@ -36,6 +36,9 @@ pub(crate) trait Arch {
     fn get_dtv_offset() -> u64 {
         0
     }
+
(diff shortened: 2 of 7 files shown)


marxin:
I managed to build both tramp3d-v4 and Mold binaries with debug info, and it seems GDB is happy with the content of the debug info sections. I tried stepping into the Mold binary in gdb, and it looks fine!

davidlattimore:
Will this affect garbage collection of otherwise unreferenced global symbols?

davidlattimore:
I thought I'd check (https://github.com/davidlattimore/wild/pull/792) and it does indeed seem that this change means that debug info prevents --gc-sections from working.

davidlattimore:
It's fine by me to disable the failing test for riscv if you'd like to leave fixing it for a subsequent PR.

As for how it should work, I'm not sure I have a good enough understanding of how label references are supposed to work.

One possibly related thing currently in the linker is how eh-frames are handled. Basically there we keep frame data for sections that don't get garbage collected. So from the GC perspective, it's as though the section references the frame data, when in actual fact it's the frame data that references the section. Not sure if labels are a similar deal.

marxin:
> One possibly related thing currently in the linker is how eh-frames are handled. Basically there we keep frame data for sections that don't get garbage collected.

We can likely use a similar mechanism, where I described an example of how the symbols are referenced from debug info sections in the follow-up issue #795.

PR #793, opened 2025-05-28
CI: Remove ubuntu 24.10

davidlattimore:
There's both newer and older ubuntu releases in our matrix, so this one seems redundant.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,6 @@ jobs:
           - ubuntu-24.04-arm
         container:
           - 'ubuntu:24.04'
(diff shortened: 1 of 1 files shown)


davidlattimore:
> I'm also wondering if we could do without the non-QEMU variants of either `ubuntu:25.10` or `ubuntu:24.04`, since the qemu variants test everything that the non-QEMU variants test.

Although the qemu test isn't a superset of the -arm test. So it's only the x86-64 `ubuntu:25.10` or `ubuntu:24.04` that seem redundant.

marxin (approved):

PR #794, opened 2025-05-28
Integration tests: Support running multiple cross arch at once

davidlattimore:


davidlattimore:
I also changed the test config deserialisation to error on unknown fields.

PR #804, opened 2025-05-31
Implement --export-dynamic

mati865:
With this change, llvm-project configured with  `-G Ninja -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_PROJECTS=clang -DLLVM_BUILD_LLVM_DYLIB=ON` passes fully 🎉.
With `-DLLVM_LINK_LLVM_DYLIB=ON` there are still 2 failures left, caused by the segfault in `clang-offload-bundler`.

Fixes: https://github.com/davidlattimore/wild/issues/752

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -65,6 +65,7 @@ pub struct Args {
     pub(crate) undefined: Vec<String>,
     pub(crate) relro: bool,
     pub(crate) entry: Option<String>,
+    pub(crate) explicitly_export_dynamic: bool,
 
     /// If set, GC stats will be written to the specified filename.
     pub(crate) write_gc_stats: Option<PathBuf>,
@@ -191,7 +192,6 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "nostdlib",
     // TODO
     "no-undefined-version",
-    "export-dynamic",
     "fatal-warnings",
     "color-diagnostics",
     "undefined-version",
@@ -204,7 +204,6 @@ const IGNORED_FLAGS: &[&str] = &[
     "disable-new-dtags",
     "fix-cortex-a53-835769",
     "fix-cortex-a53-843419",
-    "no-export-dynamic",
 ];
 
 // These flags map to the default behavior of the linker.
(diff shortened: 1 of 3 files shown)


mati865:
Interesting, AArch64 runners passed but cross compilation ones failed on:
```
 dynsym.__bss_end__.section
  wild 
  ld .got.plt

dynsym.__bss_start.section
  wild 
  ld .got.plt

dynsym.__bss_start__.section
  wild 
  ld .got.plt

dynsym.__data_start.section
  wild 
  ld .got.plt

dynsym.__end__.section
  wild 
  ld .got.plt

dynsym._bss_end__.section
  wild 
  ld .got.plt

dynsym._edata.section
  wild 
  ld .got.plt

dynsym._end.section
  wild 
  ld .got.plt
```

davidlattimore:
I think I've seen GNU ld do that kind of thing before. It's putting symbols in sections that make no sense, most likely because the sections that it should be putting them don't exist. e.g. `_bss_end__` should be in `.bss`, but I'm guessing there is no `.bss`. It might be possible to "fix" this by making sure there's something in each of the relevant sections. e.g. declare a global variable with an initial value of 0 to put something in .bss. Alternatively, it'd be fine to just DiffIgnore those.

mati865:
Ah, native builders probably matched it with LLD.

I couldn't manage to make LD happy with global variables, but I believe we can just ignore it for cross compilation. There should be no real differences between different architectures here.
Explicitly ignoring all problematic symbols is a bit tedious and possibly fragile (linker-diff can handle only a single `*`), and ignoring all dynsyms would be even worse because it's a valuable check - without the fix the test fails with:
```
  stderr ───
    wild: /home/mateusz/Projects/wild/wild/tests/build/trivial-dynamic.c-export-dynamic-host.wild
    ld: /home/mateusz/Projects/wild/wild/tests/build/trivial-dynamic.c-export-dynamic-host.ld
    lld: /home/mateusz/Projects/wild/wild/tests/build/trivial-dynamic.c-export-dynamic-host.lld
    dynsym._start.section
      wild
      ld .text
      lld .text

    dynsym.exit_syscall.section
      wild
      ld .text
      lld .text

    dynsym.foo_ptr.section
      wild
      ld .data
      lld .data

    dynsym.runtime_init.section
      wild
      ld .text
      lld .text
```

PR #809, opened 2025-06-01
Don't apply -Bsymbolic* to undefined symbols

davidlattimore:


davidlattimore:
@lapla-cogito - FYI. I was seeing an issue where -Bsymbolic was affecting undefined symbols

PR #823, opened 2025-06-03
Parallelise opening of input files

### LINKED ISSUE #816: Parallelise opening of files
It took a while on my poor laptop, but I managed to build chromium. Linking performance with Wild is looking pretty good, but profiling shows a few opportunities for improvement. The largest is opening all the input files and identifying what kind of file they are. I'm pretty sure the type identification code isn't slow, but it's the first access to a page of the file, so is incurring a page fault. I'm going to try to parallelise this. It's made somewhat tricky since linker scripts and thin archives can cause additional files to be loaded, so this isn't just a straight `par_iter`. It's also complicated by the need for parsing linker scripts to borrow from the file that was just read. I think these challenges can be overcome though. Right now I'm thinking I'll probably use a channel to send work to workers and another to send results and requests for more files back.

davidlattimore:
Fixes #816

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,6 +294,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "colosseum"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "370c83b49aedf022ee27942e8ae1d9de1cf40dc9653ee6550e4455d08f6406f9"
+
+[[package]]
 name = "constant_time_eq"
 version = "0.3.1"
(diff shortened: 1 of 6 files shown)


davidlattimore:
This change mostly helps when there are lots of input files. In cases like clang, where the objects are packed into a relatively small number of archives, it doesn't seem to really make any difference. However, with projects like chromium, which passes about 13k separate object files to the linker, it makes quite a substantial difference. When linking chromium on my laptop, this reduces the time to open input files from about 314ms to about 93ms.

PR #827, opened 2025-06-03
Add flag to write symbols for PLT / GOT entries

### LINKED ISSUE #526: Add flag to write symbols for PLT / GOT entries
Split from #308.

The idea is to write symtab entries with `NAME_OF_SYMBOL$got` that points to the GOT entry for the symbol. Similarly, `NAME_OF_SYMBOL$plt` for the PLT entry of the symbol.

e.g. `__gmon_start__$got` should point to the GOT entry for `__gmon_start__`.

GNU ld and LLD as far as I know don't do this, but mold does. I'm not sure if we want to do this by default, at least not initially, so I'm inclined to say that we should put it behind a flag. For want of a better name, how about `--gen-got-plt-syms`, but open to alternative names.

The main places that will need changing are:

* Argument parsing in `args.rs` to parse the new flag.
* `SymbolRequestHandler::finalise_symbol_sizes`. This will need to allocate space in the `.symstr` and `.symtab` sections for the extra symbols. You can determine if there's GOT / PLT entries for the symbol by checking `current_res_flags`.
* `ObjectLayout::write_symbols` (in `elf_writer.rs`). This is where the symbols will actually get written. The variable `res` holds a `Resolution` which will have most of the info you'll need like the GOT and PLT addresses.
* The easiest way to test this is probably to add an extra config to an existing test where you enable the flag. I think the ifunc test (`ifunc.c`) is probably a reasonable choice, since ifuncs are the main place where Wild uses PLT entries. You can check for the symbol using a test directive  like `//#ExpectSym:foo$plt`.

If you'd like to work on this, please let us know so that we can assign it to you and we don't end up with multiple people working on it at the same time.

lapla-cogito:
This patch modifies the behavior by adding extra entries to the symbol table that point to GOT and PLT entries via the `--got-plt-syms` option. While this is the default behavior in mold, it represents an operation not present in other linkers like GNU ld, and therefore this option is disabled by default at present.

close #526


lapla-cogito:
When considering adding tests for this scenario, note that `--got-plt-syms` is a wild-specific option with no comparison target linker, causing linker-diff to panic (the same underlying [issue discussed previously on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/421156-gsoc/topic/Project.3A.20Improve.20Wild.20linker.20test.20suites/near/521482968)). Therefore, we need to determine how to address this linker-diff bug while testing this option.

libwild/src/layout.rs
```diff
@@ -704,6 +704,22 @@ trait SymbolRequestHandler<'data>: std::fmt::Display + HandlerData {
                 }
             }
 
+            if symbol_db.args.got_plt_syms {
+                let name = symbol_db.symbol_name(symbol_id)?;
+                let mut symbol_name = name.to_string() + "$";
```
davidlattimore: I'm not too worried about the performance of this flag given that it's off by default, but if you'd like to improve the performance a bit, you should be able to avoid the heap allocation here. All you need is the size that the string will be, you don't actually need the string itself. e.g. you can start with `let mut name_len = name.len();` then add to that based on what lengths you're going to add.

It should also be possible to avoid the heap allocation where you construct the name in elf_writer, although that's slightly harder.

libwild/src/elf_writer.rs
```diff
@@ -1650,6 +1650,68 @@ impl<'data> ObjectLayout<'data> {
                         format!("Failed to copy {}", layout.symbol_debug(symbol_id))
                     })?;
             }
+
+            if layout.symbol_db.args.got_plt_syms {
+                if let Some(info) = SymbolCopyInfo::new(
```
davidlattimore: This function is getting pretty long. Perhaps the contents of this if-block could be extracted out into a separate function?

davidlattimore: Does `SymbolCopyInfo::new` need to be called again? It was called just above and it looks like if you moved this code into the if-statement where it was originally called that you wouldn't need to call it again

libwild/src/elf_writer.rs
```diff
@@ -1650,6 +1650,68 @@ impl<'data> ObjectLayout<'data> {
...
+                    symbol_id,
+                    &layout.symbol_db,
+                    *sym_state,
+                    &self.sections,
+                ) {
+                    let current_res_flags = layout.symbol_resolution_flags[symbol_id.as_usize()];
+
+                    let mut symbol_name = info.name.to_vec();
+                    symbol_name.extend_from_slice(b"$");
+                    let suffix = if current_res_flags.needs_got() {
```
davidlattimore: It looks like this currently only supports generating a GOT symbol or a PLT symbol, but not both. When there's a PLT entry, there should also be a GOT entry, so I'm not sure if currently you'd ever get to see the plt symbol being produced.

davidlattimore (commented):
For testing, I'd suggest a test that does something like the following:

```
//#SkipLinker:ld
//#DiffEnabled:false
//#ExpectSym:foo$got .got
```

libwild/src/elf_writer.rs
lapla-cogito: Understood, thanks. So would changing the application order of these conditions be the correct approach? That is, while it's true that if there's a PLT entry, there must correspond to a GOT entry, I don't think the reverse is always true, so I'm considering the following code:
```rust
if current_res_flags.needs_plt() {
    b"plt"
} else if current_res_flags.needs_got() {
    b"got"
}
```

davidlattimore: If there's both a GOT and a PLT, then I suspect you'd kind of want a symbol for both

libwild/src/layout.rs
```diff
@@ -704,6 +704,24 @@ trait SymbolRequestHandler<'data>: std::fmt::Display + HandlerData {
                 }
             }
 
+            if symbol_db.args.got_plt_syms {
+                let name = symbol_db.symbol_name(symbol_id)?;
+                let mut name_len = name.len();
+                if current_res_flags.needs_got() || current_res_flags.needs_plt() {
+                    name_len += 4; // "$got" or "$plt" suffix
+                } else {
+                    continue;
```
davidlattimore: I'm not sure about this continue being here. It looks like it might skip stuff that it shouldn't - e.g. the call to `allocate_symbol_resolution` below. How about getting rid of the else-branch and just moving the allocation code into the if-branch? Actually, you could probably change your outer if-statement to `if symbol_db.args.got_plt_syms && current_res_flags.needs_got()`. This relies on `needs_got` always being set when `needs_plt` is set, but your code already relies on that below.

lapla-cogito: Removing the `continue` results in an "Insufficient `.symtab` entries" error (although this `continue` is root of bugs). I’m not sure why this happens yet, but I suspect that `allocate_symbol_resolution` is allocating extra resolution flags incorrectly for some symbols. If we remove the `continue`, the current tests fail with a `compute_unused$got` error, which I believe shouldn't occur.

davidlattimore: I'd suggest adding some print statements to the code both at the point where you allocate and the point where you write. You can then run the linker outside of the test by copy and pasting the `cargo run` command from the test output. Hopefully that'll give you a clue as to why allocation and writing are inconsistent. If get stuck, let me know and I can help.

lapla-cogito: Alright, we apparently needed to allocate the necessary space in the symbol table after the resolution was finalized. Now it's fixed.

marxin:
Noticed #834.

PR #829, opened 2025-06-03
Skip non-existant `-L` directories

Noratrieb:
```
$ ld.lld a.o -Lhi_david
$ ld.bfd a.o -Lhi_david
```

Both pass fine, while wild fails while trying to handle the save dir. I just ignored search paths that don't seem to exist, which doesn't feel _great_ (I don't really like `exists` because of TOCTOU) but in this case it seems fine.

I ran into this while trying to run the test suite on NixOS, whose GCC wrapper for some reason passes `-L` paths that don't exist. Seems like a NixOS bug, but a good excuse to make wild more compliant with BFD and LLD :).


Noratrieb:
Looks like this is the wrong approach, as this fails a unit test because of course the unit test files may not exist. Maybe I should just ignore it for the saving?

mati865:
Hmm, I cannot reproduce it on Arch Linux using any form of a non-existent path.
Is it failing on `std::path::absolute` or something else?

Noratrieb:
It's failing in the test suite while trying to create the `WILD_SAVE_DIR`. Set this environment variable and it should fail. I've adjusted the commit message to reflect that.

mati865:
Thanks, makes sense now,
I'm wondering if `handle_file` method shouldn't check the existence instead once it retrieves the absolute path. That would probably save `statx` call in the hot path.

davidlattimore:
Thanks!

davidlattimore:
I just added a test for this too - https://github.com/davidlattimore/wild/pull/833 - it fails without your change and passes with it.

PR #831, opened 2025-06-03
Handle empty `.got` in linker-diff

### LINKED ISSUE #830: linker-diff can't handle empty `.got`
When executing integration_test::program_name_03___trivial_dynamic_c__ locally (on NixOS), linker-diff crashed.
The cause was the binary produced by LLD, which had an empty `.got` section (instead putting its stuff in `.got.plit), I don't know whether that's good or bad. This was with LLD 19.

The crash was
```
thread 'main' panicked at linker-diff/src/asm_diff.rs:3147:14:
called `Result::unwrap()` on an `Err` value: ()
```

Why does it crash here?

```rust
object::slice_from_bytes(data, data.len() / entry_size)
```

seems innocent, except the pointer to `data` was unaligned. why was it unaligned? because `object` returns `&[]` for empty reads in `<&[u8] as ReadRef>::read_bytes_at`, which will not be aligned.

```rust
impl<'a> ReadRef<'a> for &'a [u8] {
    /*snip*/
    fn read_bytes_at(self, offset: u64, size: u64) -> Result<&'a [u8]> {
        if size == 0 {
            return Ok(&[]);
        }
```

`index_got_table` looks like the only questionable use of `slice_from_bytes` in linker-diff, so I think I'll just send a patch to check for 0 there.

Noratrieb:
If `.got` is empty, the slice may not be aligned since `object` will just return a `&[]` for empty sections.

Avoid depending on its alignment by just using the newly stabilized `as_chunks`.

This requires Rust 1.88 which isn't released yet. But I think we can wait a few weeks with merging this PR :upside_down_face:. We could also rewrite it to use `chunks_exact`, which would require using a vec instead in `GotIndex`.

fixes #830

### DIFF
--- a/linker-diff/src/asm_diff.rs
+++ b/linker-diff/src/asm_diff.rs
@@ -3140,11 +3140,13 @@ impl<'data> AddressIndex<'data> {
         };
 
         let data = got_section.data()?;
-        let entry_size = size_of::<u64>();
+        // Ensure we don't try to cast it to `[u64]`, as it may not be aligned.
+        let (raw_entries, tail) = data.as_chunks::<{ size_of::<u64>() }>();
 
-        let raw_entries: &[u64] = object::slice_from_bytes(data, data.len() / entry_size)
-            .unwrap()
-            .0;
+        ensure!(
+            tail.is_empty(),
+            "GOT table `{table_name}` does not have an aligned length"
+        );
 
         let base = got_section.address();
         Ok(Some(GotIndex {
@@ -3268,7 +3270,7 @@ struct GotIndex<'data> {
     /// The addresses covered by this table.
     address_range: Range<u64>,
 
-    entries: &'data [u64],
+    entries: &'data [[u8; size_of::<u64>()]],
 }
 
 impl<'data> GotIndex<'data> {
@@ -3365,12 +3367,13 @@ impl<'data> GotIndex<'data> {
                 .entries
                 .get((offset / entry_size) as usize)
                 .context("got_address past end of index range")?;
+            let raw_value = u64::from_le_bytes(*raw_value);
 
             match relocation_kind {
                 RelocationKind::GotTpOff
                 | RelocationKind::GotTpOffGot
                 | RelocationKind::GotTpOffGotBase => {
-                    Ok(Referent::UnmatchedTlsOffset(*raw_value as i64))
+                    Ok(Referent::UnmatchedTlsOffset(raw_value as i64))
                 }
                 RelocationKind::TlsDescCall => Ok(Referent::TlsDescCall),
                 RelocationKind::Absolute
@@ -3389,7 +3392,7 @@ impl<'data> GotIndex<'data> {
                 | RelocationKind::PltRelative
                 | RelocationKind::GotRelative
                 | RelocationKind::None
-                | RelocationKind::PairSubtraction => Ok(Referent::Absolute(*raw_value)),
+                | RelocationKind::PairSubtraction => Ok(Referent::Absolute(raw_value)),
                 RelocationKind::TlsGd
                 | RelocationKind::TlsGdGot
                 | RelocationKind::TlsGdGotBase


davidlattimore:
Thanks for the PR!  `as_chunks` looks like a neat API. There's still 3 weeks until 1.88. Any interest in merging this sooner? Two alternatives come to mind. Either we could allow unaligned bytes only for the empty case:

```patch
diff --git a/linker-diff/src/asm_diff.rs b/linker-diff/src/asm_diff.rs
index 291d000051..edc0c4361e 100644
--- a/linker-diff/src/asm_diff.rs
+++ b/linker-diff/src/asm_diff.rs
@@ -3140,13 +3140,15 @@
         };
 
         let data = got_section.data()?;
-        // Ensure we don't try to cast it to `[u64]`, as it may not be aligned.
-        let (raw_entries, tail) = data.as_chunks::<{ size_of::<u64>() }>();
-
-        ensure!(
-            tail.is_empty(),
-            "GOT table `{table_name}` does not have an aligned length"
-        );
+        let raw_entries: &[u64] = if data.is_empty() {
+            // An empty .got may not be aligned, so we avoid calling object::slice_from_bytes.
+            &[]
+        } else {
+            let entry_size = size_of::<u64>();
+            object::slice_from_bytes(data, data.len() / entry_size)
+                .unwrap()
+                .0
+        };
 
         let base = got_section.address();
         Ok(Some(GotIndex {
@@ -3270,7 +3272,7 @@
     /// The addresses covered by this table.
     address_range: Range<u64>,
 
-    entries: &'data [[u8; size_of::<u64>()]],
+    entries: &'data [u64],
 }
 
 impl<'data> GotIndex<'data> {
@@ -3363,11 +3365,10 @@
             }
         } else {
             // No dynamic relocation, just read from the original file data.
-            let raw_value = self
+            let raw_value = *self
                 .entries
                 .get((offset / entry_size) as usize)
                 .context("got_address past end of index range")?;
-            let raw_value = u64::from_le_bytes(*raw_value);
 
             match relocation_kind {
                 RelocationKind::GotTpOff


```

GOT entries should always be aligned, so an unaligned non-empty GOT likely means a linker bug. But perhaps we want Wild to handle it gracefully anyway, in that case we could keep the got as bytes until we go to read an entry:

```patch
diff --git a/linker-diff/src/asm_diff.rs b/linker-diff/src/asm_diff.rs
index 291d000051..feb41a2d70 100644
--- a/linker-diff/src/asm_diff.rs
+++ b/linker-diff/src/asm_diff.rs
@@ -3140,18 +3140,11 @@
         };
 
         let data = got_section.data()?;
-        // Ensure we don't try to cast it to `[u64]`, as it may not be aligned.
-        let (raw_entries, tail) = data.as_chunks::<{ size_of::<u64>() }>();
-
-        ensure!(
-            tail.is_empty(),
-            "GOT table `{table_name}` does not have an aligned length"
-        );
 
         let base = got_section.address();
         Ok(Some(GotIndex {
             address_range: base..base + data.len() as u64,
-            entries: raw_entries,
+            entries: data,
         }))
     }
 
@@ -3270,7 +3263,7 @@
     /// The addresses covered by this table.
     address_range: Range<u64>,
 
-    entries: &'data [[u8; size_of::<u64>()]],
+    entries: &'data [u8],
 }
 
 impl<'data> GotIndex<'data> {
@@ -3365,8 +3358,10 @@
             // No dynamic relocation, just read from the original file data.
             let raw_value = self
                 .entries
-                .get((offset / entry_size) as usize)
+                .get(offset as usize..)
+                .and_then(|bytes| bytes.first_chunk::<{ size_of::<u64>() }>())
                 .context("got_address past end of index range")?;
+
             let raw_value = u64::from_le_bytes(*raw_value);
 
             match relocation_kind {

```

davidlattimore:
Another PR (just merged) conflicted with this PR. I fixed the merge conflict and then applied the first of the two backport-to-stable options. I'll merge this now. We can always switch back to using `as_chunks` later once it's stable. Thanks again for the PR!

Noratrieb:
thanks, I didn't get to it earlier :)

PR #843, opened 2025-06-06
Relax TLSDESC and TLSDESC_CALL separately

### LINKED ISSUE #842: tlsdesc relaxations applied incorrectly in TLSDESC_CALL isn't immediately after
As reported in #832, the tlsdesc test fails when run on nixos.

The problem seems to be our handling of the following TLSDESC relocation.

```asm
  2e:	48 8d 05 00 00 00 00 	lea    0x0(%rip),%rax        # 35 <get_value+0x35>
			31: R_X86_64_GOTPC32_TLSDESC	g6-0x4
  35:	64 03 11             	add    %fs:(%rcx),%edx
  38:	ff 10                	call   *(%rax)
			38: R_X86_64_TLSDESC_CALL	g6
```

There's an extra instruction before the subsequent TLSDESC_CALL instruction, but we're currently assuming that lea and call instructions are adjacent.

I was able to reproduce the problem outside of nixos with the flags `-mtls-dialect=gnu2 -fPIC -O2`. The `-fPIC` was necessary to get the compiler to use TLSDESC at all, otherwise it decided it was unnecessary. The `-O2` was necessary for instruction reordering to insert the interleaving instruction.

davidlattimore:
These two relocations don't have to be adjacent.

To handle this, we now relax these two relocations separately. We also now assume that the instruction that the TLSDESC relocation is a REX-prefixed LEA instruction.

Fix #842

### DIFF
--- a/libwild/src/x86_64.rs
+++ b/libwild/src/x86_64.rs
@@ -308,25 +308,33 @@ impl crate::arch::Relaxation for Relaxation {
                     }
                 }
             }
+            // TODO: It looks to me like we should be able to optimise to local-exec for any
+            // executable, not just static executables. However, currently the test fails if we try
+            // this. Investigate why.
             object::elf::R_X86_64_GOTPC32_TLSDESC
                 if !interposable && output_kind.is_static_executable() =>
             {
-                // lea    0x0(%rip),%rax
-                if section_bytes.get(offset - 3..offset)? == [0x48, 0x8d, 0x05] {
-                    return create(
-                        RelaxationKind::TlsDescToLocalExec,
-                        object::elf::R_X86_64_TPOFF32,
-                    );
-                }
+                // We assume that the instruction that this relocation applies to is a REX-prefixed
+                // LEA instruction.
+                return create(
+                    RelaxationKind::TlsDescToLocalExec,
+                    object::elf::R_X86_64_TPOFF32,
+                );
             }
+            // Note, the conditions on this relaxation (is_executable) must match those on
+            // TLSDESC_CALL below.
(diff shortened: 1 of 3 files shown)


davidlattimore:
A more conservative fix for the linked bug would be to check the bytes of the subsequent instruction and not apply the relaxation if it's not the expected bytes for a TLSDESC_CALL. However, GNU ld seems to still apply relaxations even when there's an interleaved instruction, so it seems like we probably should too. Actually GNU ld looks to be applying a local-exec relaxation not initial-exec, but I figured I'd leave that for a separate change.

The reason why I removed the check of the instruction bytes for the TLSDESC relocation is because if that check resulted in us not relaxing, but then the subsequent TLSDESC_CALL got relaxed, we'd have done the wrong thing. So if we're going to do the wrong thing in such a case anyway, then we might as well do so without wasting memory access on the first check. From my limited googling, it seems that the instruction is expected to be a REX-prefixed lea instruction. I found a bug that got fixed in GCC where they were emitting an lea instruction without the REX prefix.

It'd actually be really nice if we didn't need to check any bytes for any relaxations - i.e. if we could decide whether to apply a relaxation based only on the properties of the relocation, the symbol, output type etc. That'd mean that we wouldn't need to access section bytes until we go to copy them to the output file. Whether that's realistic, or even possible with some of the older relocation types, I have no idea.

marxin:
Thanks for handling this limitation! Yeah, the current relaxation schema is not ideal. Can you please file an issue we can return to in the future? Maybe that's something that could be connected to the suggested simplification in the last paragraph.

PR #847, opened 2025-06-06
Add Nix package, overlay, stdenv adapter, and flake

RossSmyth:
Thanks for the great project! This PR is best review commit-by-commit.

This works as-is, but should can wait until after #831 and ~~#843~~ to remove the patches as carrying patches in-tree is sort of weird. 

Things done:

1. Add a default.nix that packages Wild using Nix
2. Add a stdenv adapter that allows one to use Wild as their linker for a Nix project
Example usage: https://github.com/RossSmyth/wild/blob/fa615bae64eb1999f2f11366d4904551bf1d646d/default.nix#L163-L175
4. Add a Flake to make it easier for people to try out upstream Wild
5. Add nix to CI

I'm not sure how much of this is desired, it can be reduced in scope if desired, but I was 90% of the way there so I just put this together. Big thanks to @Noratrieb and @dawnofmidnight for their great work, I just cleaned it up and did some API plumbing. 

I also do not know where the files should go, so they are just in the top-level.

Related: https://github.com/NixOS/nixpkgs/pull/414558


davidlattimore:
Thanks! I think I'm happy to accept all of those changes.

Github is showing here that all checks passed, but the [actions page](https://github.com/davidlattimore/wild/actions/runs/15498919824) is showing a couple of errors.

```
Error when evaluating 'strategy' for job 'nix-checks'. .github/workflows/ci.yml (Line: 166, Col: 16): Error from function 'fromJson': empty input, .github/workflows/ci.yml (Line: 166, Col: 16): Unexpected value ''
```

Not sure why the failure doesn't show here.

Also, "Find Nix Checks" seems to have succeeded, however if I look at [the logs](https://github.com/davidlattimore/wild/actions/runs/15498919824/job/43653873381#step:4:16), there are errors.

I know very little about Nix, so my review of your changes is likely not very comprehensive. If @Noratrieb or @dawnofmidnight have any comments on anything, please do add them :)

Is your intent to keep the commits separate when we merge or squash them? I'm happy to keep them separate if you prefer, so long as they each make sense by themselves and in the order they're committed - although it's a little hard for me to determine if this is the case.

RossSmyth:
Oh I forgot a semi-colon. I'll fix it when I get home.

RossSmyth:
It can be squashed, that's fine.

RossSmyth:
A sandboxed nix failure that cannot be replicated locally. Terrifying. I'll investigate tomorrow.

davidlattimore:
> A sandboxed nix failure that cannot be replicated locally. Terrifying. I'll investigate tomorrow.

I'm assuming you mean this error?

```
called `Result::unwrap()` on an `Err` value: Failed to symlink `/build/source/target/x86_64-unknown-linux-gnu/release/ld` to `/build/source/target/x86_64-unknown-linux-gnu/release/wild`
```

I see from the logs that nextest is being used. AFAIK, it runs multiple test processes (full processes, not just threads). Looking at the code in `integration_test.rs`, there's the potential for those separate processes to race creating that symlink. Probably the fix is to change the code that creates the symlink to ignore file-exists errors. I'll do that now, so then it'll hopefully just be a matter of rebasing your change to pick up the fix.

davidlattimore:
https://github.com/davidlattimore/wild/pull/851

RossSmyth:
Yeah I noticed that, so the changes I pushed after was disabling Nextest. I'll rebase tomorrow cause I'm in bed 🛏️.

RossSmyth:
Even after rebasing I still get an intermittent race occurring (this time on my laptop).

The test is `program_name_02___trivial_main_c__` and the failure is `File exists (os error 17)`

I just won't renable nextest.

RossSmyth:
Alright, after talking with @dawnofmidnight a bit I moved the files to their own directory. I also added how to use it to the README. This does change how using the devshell work if people use it right now. Rather than `nix-shell` it would be `nix-shell nix/shell.nix` Could also add a .envrc for direnv to pick up on.

RossSmyth:
Dawn reviewed my work on Discord, and it was implemented in 4f765facb2b8d540a4b83642b625d6bff77bfaf4 and the other commits around that time.

Nora said "I am very clueless about nix"

davidlattimore:
I'll merge this now. Thanks for your work on this and Dawn for reviewing. I notice that the actions are slower than our pre-existing actions - although our alpine action is close. Is caching a possibility? Our other actions (except alpine) use caching.

RossSmyth:
Yeah there's some caching and other tricks that can be experimented with.

purepani:
Instead of going to the nixpkgs rust overlay, it's probably better to just stay on rust-overlay if it has a cachix instance, or use [Fenix](https://github.com/nix-community/fenix) since it definitely has a cachix instance.

Issue #786, opened 2025-05-28
RISC-V: support -static


marxin:
Right now, we're missing a couple of the essential relaxations that are needed for a situation where the dynamic linker does not assist during the start of a program.

PR #848, opened 2025-06-06
RISC-V: support -static linking

marxin:
There's a PR that helps me to link a simple C program with `-static` successfully. 

Needed changes:
 - CALL_PLT -> make it a direct call without PLT entry (the same what we need for other archs)
 - TCB placement at the place where we emit TPOFF into a GOT entry - something which might be borked on AArch64 as well, but we don't hit it as it's handled by a relaxation
 - a surprise where I noticed the libc.a instructions (connected to a relocation) already contain an encoded immediate; I hit various crashes as the immediates were combined with a value generated by the linker

EDITTED: ~I'm going to investigate if we can enable some of the `libc-integration.c` tests using the `-static`~.

Fixes: #786

### DIFF
--- a/libwild/src/aarch64.rs
+++ b/libwild/src/aarch64.rs
@@ -90,7 +90,7 @@ impl crate::arch::Arch for AArch64 {
     }
 
     fn tcb_placement() -> TCBPlacement {
-        // TODO: double-check
+        // TODO: #864
         TCBPlacement::BeforeTP
     }
 }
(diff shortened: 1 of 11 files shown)


libwild/src/aarch64.rs
```diff
@@ -87,6 +88,11 @@ impl crate::arch::Arch for AArch64 {
     fn local_symbols_in_debug_info() -> bool {
         false
     }
+
+    fn tcb_placement() -> TCBPlacement {
+        // TODO: double-check
+        TCBPlacement::BeforeTP
```
marxin, line 94: Based on my hacky experiment (where I commented out all TLS-related relaxations), a static linking of a trivial main program fails! Will take a look tomorrow.

marxin: Alright, I've linked a GitHub issue, and I'd like to merge this PR.

PR #850, opened 2025-06-06
add: defs z option

### LINKED ISSUE #810: Add support for -z defs
The flag `-z defs` affects linking of shared objects. It should mean that we give an error if a symbol is referenced, but not defined, where normally, for shared objects, we don't do that.

The test can probably be something similar to this.

```c
//#LinkArgs:-Bshareable -z defs
//#Static:false
//#ExpectError:foo

int foo(void);

int call_foo(void) {
    return foo() + 2;
}
```

AadiWaghray:
Add defs option for z flag. Closes #810

### DIFF
--- a/wild/tests/sources/z-defs.c
+++ b/wild/tests/sources/z-defs.c
@@ -1,0 +1,7 @@
+// #LinkArgs:-Bshareable -z defs
+// #Static:false
+// #ExpectError:foo
+
+int foo(void);
+
+int call_foo(void) { return foo() + 2; }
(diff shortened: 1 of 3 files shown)


wild/tests/sources/z-defs.c
```diff
@@ -0,0 +1,7 @@
+// #LinkArgs:-Bshareable -z defs
```
davidlattimore: Oh, also, these comments need to start with '//#' (no space), otherwise they'll be ignored

PR #853, opened 2025-06-08
Nix CI time improvment

RossSmyth:
I'm going to try out different things to improve nix CI times

Baseline:
https://github.com/davidlattimore/wild/actions/runs/15513450877

- Nix/wild: 3m 12s
- Nix/adapterGcc: 2m 31s
- Nix/adapterClang: 2m 39s

Note that this isn't super scientific as runners can vary in hardware and load.

It seems most of the additional time is spent building the checkPhase for Nix/wild.

1. nix_on_tmpfs

This can improve speed a lot. Sometimes more than caching.

Downsides:
- Can OOM
- Limited to one builder

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,8 +158,30 @@ jobs:
           echo "Matrix: $matrix"
           echo "matrix=$matrix" >> "$GITHUB_OUTPUT"      
 
+  nix-cache:
+    name: Refresh Nix Cache
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: nixbuild/nix-quick-install-action@v30
+
+      - name: Build Wild
+        run: nix build -L --show-trace
+
+      - name: Cache Nix Store
+        uses: nix-community/cache-nix-action/save@v6
+        id: cache
+        with:
+          primary-key: ${{ runner.os }}-${{ hashFiles( 'flake.lock', 'Cargo.lock' ) }}
+
   nix-checks:
-    needs: calc-matrix
+    needs:
+      - calc-matrix
+      - nix-cache
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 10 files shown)


RossSmyth:
3efa9154446ed738251eb6002c3be446f88d28df
nix_on_tmpfs results:
https://github.com/davidlattimore/wild/actions/runs/15515040291/job/43680888132

- Nix/wild: 3m 13s
- Nix/adapterGcc: 2m 34s
- Nix/adapterClang: 2m 39s

Conclusion:

Not much difference.

RossSmyth:
bc220bd30ef15c18b9a549f565fbd410e9a9d3d4 Do all Nix checks in paralle

This can improve build times because the nix store is shared between all jobs. The downside is that GHA can't see what's being run.

RossSmyth:
all checks in one: Increased CI to 4m 56s
https://github.com/davidlattimore/wild/actions/runs/15518525905

RossSmyth:
Next I shall try caching the Nix Store. This can be finicky sometimes so I'm trying to do it smart.

RossSmyth:
bf5adf8c8bfd353544bc967f9e5b81e23bc3957c Restore Nix Store from cache

Conclusion:

Reduces CI time to 3m 9s

The main issue I've seen with this is that restoring the cache can take a while.

RossSmyth:
Oh it looks like the cache was missed. I'll investigate later.

https://github.com/davidlattimore/wild/actions/runs/15521101489/job/43694413416

davidlattimore:
It seems to me that there are potentially two different things we're testing here. (a) wild on nixos and (b) the nix flake and other nix-specific bits and pieces. I wonder if it might be possible to test test these separately. e.g. have a nixos container in which wild gets tested the same as how we currently test on ubuntu, opensuse and alpine. Then put the tests for the more nixos-specific stuff into a separate workflow and trigger that when any of the files in the nix directory change plus perhaps also weekly.

RossSmyth:
On cache hit: 3m 27s

RossSmyth:
I'm feeling ill and can't really think, but I will come back to this in a couple days. I think I know how to make it much better.

davidlattimore:
Hope you feel better soon

RossSmyth:
Ok, with a cache hit this reduces the builds to 1 minute and 19 seconds.

RossSmyth:
Why:

Nix caches at the derivation level. As it was setup previously, the entire Wild build was a single derivation from beginning to end. So changing any of Wild: the source code, whether tests are run or not, anything, resulted in the entire thing being rebuilt. This made caching not very effective.

What is different:

This now uses [Crane](https://github.com/ipetkov/crane). This make the build caching more fine-grained. Specifically it caches the Cargo.lock dependencies in their own derivation, so each time Wild is built they do not need to be rebuilt.

RossSmyth:
I'll change it around so that a cache refresh occurs when the flake.lock or the Cargo.lock file are updated. Adding new workflow files makes it not run in this repo so it had to be all in one file.

RossSmyth:
Also there are still races with Nextest. I will not be enabling it.

davidlattimore:
Nice improvement!

Issue #732, opened 2025-04-30
linker-diff: ignore missing .dynamic.DT_RELA/.dynamic.DT_RELAENT if DT_RELA is empty


marxin:
It seems to me bfd on RISC-V emits the following dynamic tags even when the size is zero:

```
readelf -d /home/ubuntu/Programming/wild/wild/tests/build/visibility-merging.c-default-host.ld

Dynamic section at offset 0xef8 contains 11 entries:
  Tag        Type                         Name/Value
 0x000000006ffffef5 (GNU_HASH)           0x190
 0x0000000000000005 (STRTAB)             0x238
 0x0000000000000006 (SYMTAB)             0x1c0
 0x000000000000000a (STRSZ)              23 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000007 (RELA)               0x0
 0x0000000000000008 (RELASZ)             0 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x0000000000000000 (NULL)               0x0
```

While Wild does not include these if there's no such entry:
```
readelf -d /home/ubuntu/Programming/wild/wild/tests/build/visibility-merging.c-default-host.wild

Dynamic section at offset 0x740 contains 10 entries:
  Tag        Type                         Name/Value
 0x0000000000000005 (STRTAB)             0x6b0
 0x000000000000000a (STRSZ)              35 (bytes)
 0x0000000000000006 (SYMTAB)             0x620
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x2900
 0x000000006ffffff9 (RELACOUNT)          0
 0x000000006ffffef5 (GNU_HASH)           0x5e8
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x0000000000000000 (NULL)               0x0
```

davidlattimore:
Two options come to mind:

1) Make Wild do the same. This would be relatively simple. I think it'd only require changing `has_rela_dyn` to add `|| inputs.args.arch == Architecture::Riscv64`. If you wanted, you could also rename `has_rela_dyn` to something more specific, like `needs_rela_dyn_tags` to make it clear that it's only used for that.

[...]

PR #859, opened 2025-06-08
linker-diff: handle situation where DT_RELA=0x0 with BFD

marxin:
Fixes: #732


marxin:
The failing test should become green once #848 is merged.

PR #879, opened 2025-06-13
Output dynamic executable when the executable depends on shared objects

### LINKED ISSUE #836: --export-dynamic: thread 'main' panicked at libwild/src/layout.rs:289:14:
@mati865 I noticed the following panic in one of the Mold tests after 0c5cd7e74bdcec6ee815fb5bff3fac8e601dd217:
```
$ cp ~/Programming/wild/ld mold && ../test/export-dynamic.sh
...
+ ./mold -o out/test/x86_64/export-dynamic/exe out/test/x86_64/export-dynamic/a.o out/test/x86_64/export-dynamic/b.so --export-dynamic

thread 'main' panicked at libwild/src/layout.rs:289:14:
Dynamic symbol definitions should start > 0: TryFromIntError(())
```

mati865:
Fixes #836

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -2386,7 +2386,8 @@ fn integration_test(
         "duplicate_strong_symbols.c",
         "preinit-array.c",
         "exception.cc",
-        "z-defs.c"
+        "z-defs.c",
+        "export-dynamic.c"
     )]
     program_name: &'static str,
     #[allow(unused_variables)] setup_symlink: (),
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -483,15 +483,18 @@ enum RustcChannel {
     Nightly,
 }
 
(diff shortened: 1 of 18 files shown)


wild/tests/integration_tests.rs
```diff
@@ -1574,10 +1574,11 @@ impl LinkCommand {
 
                     if direct_config.is_static {
                         command.arg("-static");
-                    } else {
-                        command
-                            .arg("-dynamic-linker")
-                            .arg(dynamic_linker_path(cross_arch));
+                    // TODO: `export-dynamic` test reproduces the problem only when using neither `-static` nor `-dynamic-linker`
```
davidlattimore: I guess one option here would be to replace `is_static` with an enum. e.g. `//#Mode:static`, `//#Mode:dynamic`, `//#Mode:unspecified`. The current default would be `static`, but I'm also open to changing that default to `unspecified` if we wanted - although we'd need to update a bunch of tests to explicitly specify `//#Mode:static`.

PR #880, opened 2025-06-13
Add missing write to symbol table for `--got-plt-syms`

### LINKED ISSUE #834: Didn't use up all allocated symtab/strtab space
Thank you @lapla-cogito for working on the feature, however, I noticed the following failures if enable the feature by default in the integration tests: `got_plt_syms: true,`:

```
failures:
    integration_test::program_name_02___trivial_main_c__
    integration_test::program_name_03___trivial_dynamic_c__
    integration_test::program_name_07___data_pointers_c__
    integration_test::program_name_16___tlsdesc_c__
    integration_test::program_name_17___tls_variant_c__
    integration_test::program_name_33___symbol_versions_c__
    integration_test::program_name_34___mixed_verdef_verneed_c__
    integration_test::program_name_35___copy_relocations_c__
    integration_test::program_name_39___libc_ifunc_c__
    integration_test::program_name_40___libc_integration_c__
    integration_test::program_name_41___rust_integration_rs__
    integration_test::program_name_42___rust_integration_dynamic_rs__
    integration_test::program_name_43___cpp_integration_cc__
    integration_test::program_name_47___ifunc2_c__
    integration_test::program_name_49___tls_local_exec_c__
    integration_test::program_name_50___undefined_symbols_c__
    integration_test::program_name_53___dynamic_bss_only_c__
    integration_test::program_name_54___shared_priority_c__
    integration_test::program_name_55___shared_c__
    integration_test::program_name_57___preinit_array_c__
    integration_test::program_name_58___exception_cc__

test result: FAILED. 37 passed; 21 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.34
```

Can you please take a look?

lapla-cogito:
close #834

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -1210,6 +1210,10 @@ fn write_symbols(
         .zip(&layout.symbol_resolution_flags[object.symbol_id_range.as_usize()])
     {
         let symbol_id = object.symbol_id_range.input_to_id(sym_index);
+
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, symbol_writer, symbol_id)?;
+        }
         if let Some(info) = SymbolCopyInfo::new(
             object.object,
             sym_index,
@@ -1267,15 +1271,6 @@ fn write_symbols(
             symbol_writer
                 .copy_symbol(sym, info.name, section_id, symbol_value)
                 .with_context(|| format!("Failed to copy {}", layout.symbol_debug(symbol_id)))?;
-
-            if layout.symbol_db.args.got_plt_syms {
-                write_got_plt_syms(layout, &info, symbol_writer, symbol_id).with_context(|| {
-                    format!(
-                        "Failed to write GOT/PLT symbol for {}",
-                        layout.symbol_debug(symbol_id)
-                    )
-                })?;
-            }
         }
     }
     Ok(())
@@ -1626,24 +1621,29 @@ fn get_resolution(
 
 fn write_got_plt_syms(
     layout: &Layout,
-    info: &SymbolCopyInfo<'_>,
     symbol_writer: &mut SymbolTableWriter<'_, '_>,
     symbol_id: SymbolId,
 ) -> Result {
-    let resolution = layout
-        .local_symbol_resolution(symbol_id)
-        .with_context(|| format!("Missing resolution for {}", layout.symbol_debug(symbol_id)))?;
-    let current_res_flags = resolution.resolution_flags;
-    if !current_res_flags.needs_got() && !current_res_flags.needs_plt() {
+    if !layout.symbol_db.is_canonical(symbol_id) {
         return Ok(());
     }
+
+    let Some(resolution) = layout.local_symbol_resolution(symbol_id) else {
+        return Ok(());
+    };
+
+    if !resolution.resolution_flags.needs_got() {
+        return Ok(());
+    }
+
+    let current_res_flags = resolution.resolution_flags;
 
     let mut write_sym = |suffix: &[u8],
                          section_id: OutputSectionId,
                          get_value: fn(&Resolution) -> Result<u64>|
      -> Result {
-        let mut symbol_name = info.name.to_vec();
-        symbol_name.extend_from_slice(suffix);
+        let mut symbol_name = layout.symbol_db.symbol_name(symbol_id)?.to_string();
+        symbol_name.push_str(std::str::from_utf8(suffix).unwrap_or("unknown"));
 
         let shndx = layout
             .output_sections
@@ -1655,7 +1655,7 @@ fn write_got_plt_syms(
         let value = get_value(resolution)?;
 
         symbol_writer
-            .define_symbol(true, shndx, value, 0, &symbol_name)
+            .define_symbol(true, shndx, value, 0, symbol_name.as_bytes())
             .with_context(|| {
                 format!(
                     "Failed to copy {} symbol for {}",
@@ -3416,6 +3416,10 @@ fn write_internal_symbols_plt_got_entries<A: Arch>(
                     format!("Failed to process `{}`", layout.symbol_debug(symbol_id))
                 })?;
         }
+
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, &mut table_writer.debug_symbol_writer, symbol_id)?;
+        }
     }
     Ok(())
 }
@@ -3433,6 +3437,9 @@ fn write_dynamic_file<A: Arch>(
         .resolutions_in_range(object.symbol_id_range)
         .zip(object.object.symbols.iter())
     {
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, &mut table_writer.debug_symbol_writer, symbol_id)?;
+        }
         if let Some(res) = resolution {
             let name = object.object.symbol_name(symbol)?;
 


libwild/src/elf_writer.rs
```diff
@@ -1652,10 +1657,17 @@ fn write_got_plt_syms(
...
+            "symbol_writer len: {:?}, {:?}, is_dynamic: {:?} ,symbol name: {:?}",
+            symbol_writer.local_entries.len(),
+            symbol_writer.global_entries.len(),
+            symbol_writer.is_dynamic,
+            symbol_name
+        );
 
         symbol_writer
-            .define_symbol(true, shndx, value, 0, &symbol_name)
+            .define_symbol(false, shndx, value, 0, symbol_name.as_bytes())
```
davidlattimore: This first argument (changed to false) is `is_local`. When you allocated space for the symbol, you allocated it in `SYMTAB_LOCAL`, so this should be left as `true` so that the two are consistent.

libwild/src/layout.rs
```diff
@@ -733,11 +733,14 @@ trait SymbolRequestHandler<'data>: std::fmt::Display + HandlerData {
                 symbol_db.args.output_kind(),
             );
 
-            if symbol_db.args.got_plt_syms && resolution_flags.get().needs_got() {
+            if symbol_db.args.got_plt_syms
+                && resolution_flags.get().needs_got()
+                && value_flags.is_dynamic()
```
davidlattimore: I'd suggest removing the `is_dynamic` check. If we have PLT/GOT entries for symbols that didn't come from shared objects, then we still want symbols for them.

libwild/src/elf_writer.rs
```diff
@@ -1267,15 +1267,15 @@ fn write_symbols(
...
 
-            if layout.symbol_db.args.got_plt_syms {
-                write_got_plt_syms(layout, &info, symbol_writer, symbol_id).with_context(|| {
-                    format!(
-                        "Failed to write GOT/PLT symbol for {}",
-                        layout.symbol_debug(symbol_id)
-                    )
-                })?;
-            }
+        if layout.symbol_db.args.got_plt_syms {
```
davidlattimore: One possible cause of this code not being reached is if the `continue` on line 1246 is hit (for absolute symbols). Undefined symbols end up as absolute (with the value 0) and undefined symbols can have GOT entries (in case they end up being defined at runtime). The easiest fix is probably to move this bit of code near the start of the loop (after `symbol_id` is computed). That way it won't be affected by the `continue`.

libwild/src/elf_writer.rs
```diff
@@ -1626,24 +1626,29 @@ fn get_resolution(
...
 fn write_got_plt_syms(
     layout: &Layout,
-    info: &SymbolCopyInfo<'_>,
     symbol_writer: &mut SymbolTableWriter<'_, '_>,
     symbol_id: SymbolId,
 ) -> Result {
-    let resolution = layout
-        .local_symbol_resolution(symbol_id)
-        .with_context(|| format!("Missing resolution for {}", layout.symbol_debug(symbol_id)))?;
+    let resolution = layout.merged_symbol_resolution(symbol_id);
```
davidlattimore: We only want to write got_plt_syms for symbols when they're canonical. I'd suggest calling `local_symbol_resolution` here instead, but then adding an `is_canonical` check beforehand and if the symbol isn't canonical, then skip it.

libwild/src/elf_writer.rs
```diff
@@ -1267,15 +1267,15 @@ fn write_symbols(
...
-            if layout.symbol_db.args.got_plt_syms {
-                write_got_plt_syms(layout, &info, symbol_writer, symbol_id).with_context(|| {
-                    format!(
-                        "Failed to write GOT/PLT symbol for {}",
-                        layout.symbol_debug(symbol_id)
-                    )
-                })?;
-            }
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, symbol_writer, symbol_id).with_context(|| {
```
davidlattimore: Currently you're only calling this function from here, which means it applies to regular objects and objects from archives. It isn't currently being called for shared objects or for the epilogue. So I think you'll need two additional calls to this function - e.g. in `write_internal_symbols_plt_got_entries` (epilogue) and `write_dynamic_file`.

If you wanted to not have three call-sites, then I think it would be possible to do it all from the epilogue, however you'd need to move the allocation code. Whatever allocates space, needs to use that space at write time. So you could have the epilogue during layout scan the entire symbol ID space and allocate space for all GOT/PLT symbols, then at write time it writes them all.

libwild/src/elf_writer.rs
lapla-cogito: Sorry, I just forgot to revert the changes I temporary made.

libwild/src/elf_writer.rs
lapla-cogito: Thanks, fixed.

libwild/src/elf_writer.rs
lapla-cogito: Indeed, such cases do seem plausible. However, the primary issue appears to be insufficient entries in the symbol table—error reports indicate this error is being encountered. Therefore, I suspect the problem may lie in the allocation process for the symbol table. While `readelf -sW` shows sufficient local symbol table entries are being allocated, I suspect some discrepancy occurs during the `SymbolTableWriter` creation process, though I haven't yet identified the exact problematic location.

libwild/src/elf_writer.rs
lapla-cogito: Thanks, however what I'm still unclear about is whether symbols like `__libc_start_main`—which allocate a GOT symbol in `ifunc2.c` but generate errors if there aren't enough entries when actually writing—are actually from shared objects.
Writing this symbol can occur without calling `write_got_plt_syms` within `write_internal_symbols_plt_got_entries` or `write_dynamic_files`

libwild/src/elf_writer.rs
```diff
@@ -1210,6 +1210,15 @@ fn write_symbols(
         .zip(&layout.symbol_resolution_flags[object.symbol_id_range.as_usize()])
     {
         let symbol_id = object.symbol_id_range.input_to_id(sym_index);
+
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, symbol_writer, symbol_id).with_context(|| {
+                format!(
```
davidlattimore: There's possibly more context here than is needed. For example, when there's a failure, it currently looks like this:

```
wild: error: Failed copying from /lib/x86_64-linux-gnu/libc.so.6 (61952 (242/0)) to output file
  Caused by:
    Failed to write GOT/PLT symbol for symbol `__cxa_finalize` (2200 local=223) in file #61952 (242/0) (/lib/x86_64-linux-gnu/libc.so.6) (DYNAMIC | FUNCTION)
    Failed to copy $got symbol for symbol `__cxa_finalize` (2200 local=223) in file #61952 (242/0) (/lib/x86_64-linux-gnu/libc.so.6) (DYNAMIC | FUNCTION)
    Insufficient .symtab local entries allocated for symbol `__cxa_finalize$got`
```

libwild/src/elf_writer.rs
```diff
@@ -3433,6 +3456,16 @@ fn write_dynamic_file<A: Arch>(
         .resolutions_in_range(object.symbol_id_range)
         .zip(object.object.symbols.iter())
     {
+        if layout.symbol_db.args.got_plt_syms {
+            write_got_plt_syms(layout, &mut table_writer.dynsym_writer, symbol_id).with_context(
```
davidlattimore: I think this is why you've got not enough space when writing. You allocated space for a debug symbol, but this is dynsym, which is a different symbol table. This should be `debug_symbol_writer` here and in the call above.

libwild/src/elf_writer.rs
davidlattimore: You can find where a symbol is from and also what references it by passing `--sym-info __libc_start_main` to the linker. In this case, the symbol is indeed from a shared object.

Other things that can help with debugging this sort of issue are setting `WILD_FILES_PER_GROUP=1` and passing the argument `--threads=1`. Allocation is done on a per-group basis, so having just one file in each group means that you don't have multiple files sharing allocation.

lapla-cogito:
Thank you very much for the review. With `--got-plt-syms` now applied by default, all tests are [now green](https://github.com/davidlattimore/wild/actions/runs/15725787611). I'll now revert `--got-plt-syms` to its default off state and clean up the commits.

libwild/src/elf_writer.rs
lapla-cogito: Changed to display only symbol name.

davidlattimore: I'd probably suggest going further and removing the entire `with_context` call. The symbol name is already in the original error as `__cxa_finalize$got` and in the middle context - "Failed to copy $got symbol for symbol `__cxa_finalize`". This bit of context is ideal to remove because it's repeated 3 times, once at each call to `write_got_plt_syms`.

davidlattimore:
Thanks for your work on this!

PR #887, opened 2025-06-16
Add rel_info_from_type macro for const evaluation

marxin:
By using the macro with `const` block, we prevent unsupported relocation constants from being used at the place where we build a Relaxation:

```
error[E0080]: evaluation panicked: called `Option::unwrap()` on a `None` value
   --> libwild/src/aarch64.rs:114:17
    |
114 |         const { relocation_type_from_raw($r_type).unwrap() }
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `<aarch64::Relaxation as arch::Relaxation>::new::{constant#5}` failed here
...
232 |                     rel_info: rel_info_from_type!(object::elf::R_X86_64_16),
    |                               --------------------------------------------- in this macro invocation
    |
    = note: this error originates in the macro `rel_info_from_type` (in Nightly builds, run with -Z macro-backtrace for more info)

note: erroneous constant encountered
   --> libwild/src/aarch64.rs:114:9
    |
114 |         const { relocation_type_from_raw($r_type).unwrap() }
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
232 |                     rel_info: rel_info_from_type!(object::elf::R_X86_64_16),
    |                               --------------------------------------------- in this macro invocation
```

Additionally, I unified the code structure across the targets.

### DIFF
--- a/libwild/src/x86_64.rs
+++ b/libwild/src/x86_64.rs
@@ -3,21 +3,19 @@
 //! work unless they're performed. e.g. it uses GOT relocations in _start, which cannot work in a
 //! static-PIE binary because dynamic relocations haven't yet been applied to the GOT yet.
 
-use crate::arch::Arch;
 use crate::args::OutputKind;
 use crate::elf::PLT_ENTRY_SIZE;
 use crate::error;
 use crate::error::Result;
 use crate::resolution::ValueFlags;
-use linker_utils::elf::AllowedRange;
 use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::RelocationKindInfo;
-use linker_utils::elf::RelocationSize;
 use linker_utils::elf::SectionFlags;
 use linker_utils::elf::shf;
 use linker_utils::elf::x86_64_rel_type_to_string;
 use linker_utils::relaxation::RelocationModifier;
 use linker_utils::x86_64::RelaxationKind;
+use linker_utils::x86_64::relocation_from_raw;
 
 pub(crate) struct X86_64;
 
@@ -40,20 +38,11 @@ impl crate::arch::Arch for X86_64 {
 
     #[inline(always)]
     fn relocation_from_raw(r_type: u32) -> Result<RelocationKindInfo> {
-        let (kind, size) =
-            linker_utils::x86_64::relocation_kind_and_size(r_type).ok_or_else(|| {
-                error!(
-                    "Unsupported relocation type {}",
-                    Self::rel_type_to_string(r_type)
-                )
-            })?;
-
-        Ok(RelocationKindInfo {
-            kind,
-            size: RelocationSize::ByteSize(size),
-            mask: None,
-            range: AllowedRange::no_check(),
-            alignment: 1,
+        linker_utils::x86_64::relocation_from_raw(r_type).ok_or_else(|| {
+            error!(
+                "Unsupported relocation type {}",
+                Self::rel_type_to_string(r_type)
+            )
         })
     }
 
@@ -87,6 +76,12 @@ impl crate::arch::Arch for X86_64 {
     }
 }
 
+macro_rules! rel_info_from_type {
+    ($r_type:expr) => {
+        const { relocation_from_raw($r_type).unwrap() }
+    };
+}
+
 #[derive(Debug, Clone)]
 pub(crate) struct Relaxation {
     kind: RelaxationKind,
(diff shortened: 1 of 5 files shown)


marxin:
@davidlattimore can you please explain the following comment:
https://github.com/davidlattimore/wild/blob/67d87a02f3eb536bfbd611982ccd648069322be8/libwild/src/x86_64.rs#L149-L151

Is it something this PR optimizes away?

davidlattimore:
I looked back at the commit (7fa243ddd52ac55ff39447ffb6ffcc1d947ba04d) where I introduced that comment. I'm not actually sure that the comment made a whole lot of sense even then. I suspect by "symbol kind", what I actually meant was the symbol's ValueKind, which later evolved into ValueFlags.

I do remember writing the comment and I think roughly what I was thinking was that for some relocation types, we needed to look up additional information, but for others we didn't, but that we were being slightly wasteful by always looking up that information before determining whether we needed it.

Looking at the code as it is now, we always check the ValueFlags to see if we're dealing with an ifunc. I'd say we can remove that TODO due to being no longer relevant. Note, the TODO is now present in 3 separate files.

libwild/src/x86_64.rs
```diff
@@ -87,6 +76,12 @@ impl crate::arch::Arch for X86_64 {
     }
 }
 
+macro_rules! rel_info_from_type {
```
davidlattimore, line 79: Does this need to be a macro? It looks like it could just be a regular function (possibly with a `#[track_caller]` annotation.

marxin: I tried this approach, and I cannot make it work. The only solution I came up with is:
```rust
   const fn rel_info_from_type<const r_type: u32>() -> RelocationKindInfo {
            const { relocation_from_raw(r_type).unwrap() }
        }
...
rel_info: rel_info_from_type::<{ object::elf::R_X86_64_PLT32 }>(),
```

which I don't like much, to be honest. If I make it just a normal function (with `#[track_caller]`), I get:
```
error[E0435]: attempt to use a non-constant value in a constant
  --> libwild/src/x86_64.rs:99:41
   |
98 |         const fn rel_info_from_type(r_type: const u32) -> RelocationKindInfo {
   |                                     ------ this would need to be a `const`
99 |             const { relocation_from_raw(r_type).unwrap() }
   |                                         ^^^^^^
```

davidlattimore: The arguments to a const function don't need to be const in the function definition. The error message the compiler gives you when you do that though is pretty confusing. To avoid potentially another day of back-and-forth, I merged, then changed it in #890

davidlattimore: I guess you had the const block still, which added to the confusion. A const function is just like a normal function but can be called either at compile time or at runtime. When called at compile time, all the arguments must be const, but that only affects the caller of the function, from the function's perspective, they're just normal arguments.

marxin: Thanks for finishing this up, however, I think the `const` function does not provide what I originally wanted - have a compile-time check that an unsupported constant is passed to `relocation_from_raw`:

```patch
diff --git a/libwild/src/x86_64.rs b/libwild/src/x86_64.rs
index 1e46fb2f..0d3b31ee 100644
--- a/libwild/src/x86_64.rs
+++ b/libwild/src/x86_64.rs
@@ -311,7 +311,7 @@ impl crate::arch::Relaxation for Relaxation {
                         Some(&[0x48, 0xb8]) => {
                             return Some(Relaxation {
                                 kind: RelaxationKind::TlsLdToLocalExec64,
-                                rel_info: rel_info_from_type(object::elf::R_X86_64_NONE),
+                                rel_info: rel_info_from_type(object::elf::R_AARCH64_ABS16),
                                 mandatory: false,
                             });
                         }
```

Builds just fine; however, it was rejected with macro:
```
❯ cargo b
   Compiling libwild v0.5.0 (/home/marxin/Programming/wild/libwild)
error[E0080]: evaluation panicked: called `Option::unwrap()` on a `None` value
   --> libwild/src/x86_64.rs:81:17
    |
81  |         const { relocation_from_raw($r_type).unwrap() }
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ evaluation of `<x86_64::Relaxation as arch::Relaxation>::new::{constant#16}` failed here
...
315 |                                 rel_info: rel_info_from_type!(object::elf::R_AARCH64_ABS16),
    |                                           ------------------------------------------------- in this macro invocation
    |
    = note: this error originates in the macro `rel_info_from_type` (in Nightly builds, run with -Z macro-backtrace for more info)

note: erroneous constant encountered
   --> libwild/src/x86_64.rs:81:9
    |
81  |         const { relocation_from_raw($r_type).unwrap() }
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
315 |                                 rel_info: rel_info_from_type!(object::elf::R_AARCH64_ABS16),
    |                                           ------------------------------------------------- in this macro invocation
    |
```

@davidlattimore  That said, I would prefer sticking with the compile-time checking macro?

davidlattimore:
Ah, good point. I've created a PR to revert it.

PR #903, opened 2025-06-19
Allow running the mold test suite

lapla-cogito:
By specifying the path to mold's test suite (`mold/test`) through the `external_test_suites` option, we can execute mold's test suite in `cargo test`. Future patches will automatically classify which outputs should be properly ignored, such as error messages.

As some preliminary statistics, the following tests currently pass as is:
<details><summary>click to expand</summary>

```
test/tls-gd-dlopen.sh
test/crel.sh
test/emit-relocs-dead-sections.sh
test/glibc-2.22-bug.sh
test/rpath.sh
test/dynamic-list4.sh
test/allow-multiple-definition.sh
test/arch-x86_64-gnu-retain.sh
test/lto-nostdlib.sh
test/trace.sh
test/arch-riscv64-variant-cc.sh
test/protected.sh
test/gnu-property.sh
test/whole-archive.sh
test/bno-symbolic.sh
test/gnu-warning.sh
test/tls-irregular-start-addr.sh
test/arch-x86_64-tls-gd-to-ie.sh
test/strip-debug.sh
test/dynamic-list3.sh
test/linker-script5.sh
test/arch-x86_64-z-ibtplt.sh
test/empty-version.sh
test/tls-small-alignment.sh
test/large-max-page-size.sh
test/tlsdesc-local-dynamic.sh
test/lto-gcc.sh
test/response-file2.sh
test/compress-debug-sections-zstd.sh
test/sysroot2.sh
test/disable-new-dtags.sh
test/arch-aarch64-variant-pcs.sh
test/dynamic-dt-debug.sh
test/export-from-exe.sh
test/dt-init.sh
test/version-script2.sh
test/ifunc-static.sh
test/dynamic-list-data.sh
test/arch-x86_64-z-text.sh
test/lto-llvm.sh
test/mcmodel-large.sh
test/version-script4.sh
test/version-script13.sh
test/gc-sections.sh
test/section-attributes.sh
test/link-order.sh
test/arch-s390x-got.sh
test/gnu-unique.sh
test/range-extension-thunk3.sh
test/version-script12.sh
test/plt-dso.sh
test/arch-x86_64-reloc-zero.sh
test/dynamic.sh
test/func-addr.sh
test/textrel2.sh
test/dt-needed.sh
test/emit-relocs-cpp.sh
test/ifunc-static-pie.sh
test/pltgot.sh
test/arch-x86_64-mergeable-records.sh
test/arch-x86_64-relax.sh
test/linker-script4.sh
test/arch-riscv64-attributes2.sh
test/tls-large-static-image.sh
test/ifunc-dso.sh
test/arch-x86_64-isa-level.sh
test/unresolved-symbols2.sh
test/arch-x86_64-tls-ld-mcmodel-large.sh
test/ifunc-dlopen.sh
test/arch-x86_64-incompatible-obj.sh
test/lto-archive.sh
test/symbol-version.sh
test/arch-x86_64-tls-large-tbss.sh
test/arch-x86_64-unique.sh
test/hidden-weak-undef.sh
test/arch-arm-range-extension-thunk.sh
test/main-in-dso.sh
test/hash-style-sysv.sh
test/tlsdesc-dlopen.sh
test/large-text.sh
test/tls-gd.sh
test/dynamic-list2.sh
test/arch-x86_64-init-array.sh
test/mergeable-strings.sh
test/hidden-archive.sh
test/compress-debug-sections.sh
test/demangle-rust.sh
test/large-alignment.sh
test/arch-x86_64-incompatible-libs-linker-script2.sh
test/lto-dso.sh
test/version-script3.sh
test/sysroot.sh
test/range-extension-thunk.sh
test/library.sh
test/audit.sh
test/push-pop-state.sh
test/soname.sh
test/z-pack-relative-relocs.sh
test/tlsdesc-import.sh
test/tlsdesc-static.sh
test/hello-dynamic.sh
test/arch-x86_64-z-ibt.sh
test/symbol-version-lto.sh
test/symbol-version4.sh
test/init-in-dso.sh
test/arch-arm-thumb-interwork.sh
test/pie.sh
test/z-nodump.sh
test/ar-alignment.sh
test/range-extension-thunk2.sh
test/defsym-lto.sh
test/z-dynamic-undefined-weak.sh
test/sysroot-linker-script.sh
test/lto-archive3.sh
test/ifunc-address-equality-exported.sh
test/range-extension-thunk4.sh
test/ifunc-address-equality.sh
test/init.sh
test/version-script21.sh
test/default-symver-version-script.sh
test/unresolved-symbols.sh
test/version-script23.sh
test/lto-version-script.sh
test/section-start.sh
test/reloc-rodata.sh
test/arch-x86_64-mergeable-strings.sh
test/visibility.sh
test/response-file.sh
test/ifunc-dynamic.sh
test/gnu-retain.sh
test/ifunc-alias.sh
test/large-alignment-dso.sh
test/exclude-libs2.sh
test/version-script10.sh
test/symbol-rank.sh
test/arch-x86_64-incompatible-libs-linker-script.sh
test/tls-alignment-multi.sh
test/depaudit.sh
test/version-script22.sh
test/common-ref.sh
test/relro.sh
test/interpose.sh
test/global-offset-table.sh
test/z-origin.sh
test/tls-ld.sh
test/arch-x86_64-apx-gotpcrelx.sh
test/exception.sh
test/initfirst.sh
test/arch-x86_64-incompatible-libs.sh
test/tls-large-alignment.sh
test/exclude-libs.sh
test/bsymbolic.sh
test/init-array-priorities.sh
test/debug-macro-section.sh
test/absolute-symbols.sh
test/ctors-in-init-array.sh
test/lto-comdat.sh
test/defsym2.sh
test/symbol-version2.sh
test/z-separate-code.sh
test/version-script8.sh
test/weak-undef4.sh
test/weak-undef-dso.sh
test/tls-pic.sh
test/ifunc-noplt.sh
test/bsymbolic-functions.sh
test/arch-x86_64-warn-execstack.sh
test/tls-gd-to-ie.sh
test/ifunc-funcptr.sh
test/linker-script3.sh
test/arch-x86_64-z-rewrite-endbr3.sh
test/dead-debug-sections.sh
test/version-script19.sh
test/tbss-only.sh
test/linker-script2.sh
test/shared.sh
test/tls-gd-noplt.sh
test/z-now.sh
test/gnu-hash.sh
test/weak-export-dso.sh
test/version-script11.sh
test/tail-call.sh
test/arch-x86_64-incompatible-libs2.sh
test/arch-x86_64-apx-tlsdesc.sh
test/version-script6.sh
test/copyrel-relro.sh
test/exclude-libs3.sh
test/dynamic-list.sh
test/hello-static.sh
test/arch-x86_64-apx-gottpoff.sh
test/tlsdesc.sh
test/arch-arm-tlsdesc.sh
test/arch-x86_64-large-bss.sh
test/arch-x86_64-exception-mcmodel-large.sh
test/verbose.sh
test/arch-riscv64-relax-align.sh
test/start-stop-symbol.sh
test/version-script7.sh
test/tls-df-static-tls.sh
test/ifunc-export.sh
test/arch-x86_64-z-shstk.sh
test/many-input-sections.sh
test/symbol-version3.sh
test/issue646.sh
test/protected-dynsym.sh
test/execstack.sh
test/fatal-warnings.sh
test/default-symver.sh
test/tls-nopic.sh
test/version-script20.sh
test/z-sectionheader.sh
test/z-stack-size.sh
test/linker-script6.sh
test/tls-ld-noplt.sh
test/large-max-page-size-strip.sh
test/emit-relocs.sh
test/version-script-search-paths.sh
test/compressed-debug-info.sh
test/arch-x86_64-init-array-readonly.sh
test/common-archive.sh
test/textrel.sh
test/version-script.sh
test/lto-archive2.sh
test/z-dynamic-undefined-weak2.sh
test/relax-got-load.sh
test/arch-x86_64-tls-gd-mcmodel-large.sh
test/z-max-page-size.sh
test/entry.sh
test/arch-x86_64-tlsdesc.sh
test/weak-export-dso2.sh
test/copyrel-relro2.sh
test/depaudit2.sh
test/weak-undef.sh
test/z-nodefaultlib.sh
test/version-script18.sh
test/arch-x86_64-apx-gottpoff2.sh
test/as-needed.sh
test/arch-x86_64-tls-module-base.sh
test/arch-x86_64-section-alignment.sh
test/symbol-version5.sh
test/arch-x86_64-preinit-array.sh
test/tls-le.sh
test/arch-x86_64-ifunc-alias.sh
test/linker-script-defsym.sh
```
</details>

All output of `cargo test` has been consolidated into [this](https://gist.github.com/lapla-cogito/fbe80dee61abe2f94d8e0cb51e66e875)


lapla-cogito:
Since the architecture being tested should be injectable via environment variables from outside, expanding this support to all architectures currently supported by wild should be achievable when running these tests via GitHub Actions.

davidlattimore:
Thanks for working on this PR! Don't worry about those two test failures on ubuntu:25.10, they're not related to your change and have been happening all day. We might need to disable 25.10 if they continue.

My first question is how does this use wild as the linker when running the tests?

lapla-cogito:
@davidlattimore Sorry, I simply forgot to add the `PATH` modification 😅 and have now corrected it (I've also updated the list of passed tests and the output of `cargo test`).

wild/tests/integration_tests.rs
```diff
@@ -464,6 +465,9 @@ struct TestConfig {
     /// version of GNU ld that doesn't perform certain optimisations.
     #[serde(default)]
     diff_ignore: Vec<String>,
+
+    #[serde(default)]
+    external_test_suites: String,
```
davidlattimore: Given that this option is called "external_test_suits" (plural), should this be a `Vec<String>`? That'll also make it easier if we want to have more than one later on.

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2429,66 @@ fn integration_test(
...
+            external_test_suites_path.display()
+        );
+    }
+
+    exec_mold_tests(external_test_suites_path)?;
+
+    Ok(())
+}
+
+fn exec_mold_tests(mold_test_path: PathBuf) -> Result {
```
davidlattimore: Performance here doesn't matter at all, but in general, it's best if you can accept non-owned arguments unless you need ownership. e.g. `&Path` here and `&str` in the previous function.

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2429,66 @@ fn integration_test(
     Ok(())
 }
 
+fn external_test_suites(external_test_suites_path: String) -> Result {
```
davidlattimore: From the name, I'd assume that this function is going to return the external test suits, but it actually runs a test suit. Perhaps `run_external_test_suite`?

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2429,66 @@ fn integration_test(
...
+}
+
+fn exec_mold_tests(mold_test_path: PathBuf) -> Result {
+    let mut failed_tests = Vec::new();
+
+    let path = env::var("PATH").unwrap();
+    let current_dir = env::current_dir()?;
+    let wild_dir = current_dir.parent().unwrap();
+
+    for entry in std::fs::read_dir(&mold_test_path)? {
```
davidlattimore: If the configuration file is written with the wrong directory for the external test suite - say one that doesn't contain any .sh files, it'd be good to have the test detect this and give an appropriate error.

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2429,66 @@ fn integration_test(
...
+        let entry = entry?;
+        if entry.file_type()?.is_file() && entry.path().extension() == Some("sh".as_ref()) {
+            let output = Command::new("bash")
+                .arg("-c")
+                .arg(entry.path())
+                .env("PATH", format!("{}:{path}", wild_dir.display()))
```
davidlattimore: Path::display is only intended for displaying paths. I'd probably suggest failing (return an error) if the path isn't valid UTF-8.

davidlattimore: I thought that Mold's tests relied on the current working directory containing a binary called "ld". I guess if it can't find `ld` in the current directory, it would look on the path, so this might work, although I think currently it'd find the release build of wild. Wild's repository root contains a symlink called "ld", but it's a symlink to `target/release/wild` which wouldn't be suitable if the tests were being run without `--release`. The `fakes-debug` directory contains similar symlinks, but for debug builds.

One useful manual test for your tests is to introduce some bug - say change wild so that it always panics, then run `cargo test test_external_test_suits` and verify that all of the tests now fail.

wild/tests/integration_tests.rs
lapla-cogito: In fact, I believe making this field `Vec<String>` wouldn't significantly reduce future changes. That is, each linker's test suite execution methods vary greatly, so at least when multiple test suites are provided, `integration_test` should be able to identify which test suite it is dealing with. In such cases, `Vec<String>` would be unsuitable, and while using a key-value combination might seem reasonable, I'm not convinced there's sufficient need to do so when the exact scope of other test suites remains unclear.

davidlattimore: If you have a list of paths, you can auto-detect what kind of test suite is located at that path, then error if it's not recognised. One way this could be done is to require that the path is not the path to the tests directory, but to the root of the relevant repository. e.g. the path to the `mold` repository. You could then "detect" that the mold test suite should be run by the presence of a `test/elf` directory.

lapla-cogito: Thank you! While I initially considered specifying the repository root, the fact that the config option name clearly indicates it's a test suite, and given that mold's test suite only requires verifying the presence of `test/common.inc`, I'm inclined to specify the test suite path. Therefore, I implemented the `detect_external_test_suites` function to identify which test suite is being referenced. (BTW: I think the "detection" method for mold's test suite you mentioned is probably the one for lld.)

wild/tests/integration_tests.rs
lapla-cogito: I think this has been resolved by [this](https://github.com/davidlattimore/wild/pull/903#discussion_r2156981793).

marxin:
Have a couple of general comments:
1. I would really like direct integration of the individual tests from the Mold's test suite using the following compile-time glob listing: https://docs.rs/rstest/0.25.0/rstest/attr.rstest.html#files-path-as-input-arguments.
2. Having that, the test harness will run the tests in parallel, as we won't have to implement our own parallelism.
3. It would also be great to list all the known failing Mold tests and have `XFAIL` (fail the test if the test is actually passing).
4. As a follow-up, I know Mold's test suite supports running cross-arch tests using QEMU; it would be nice to support it.
5. Filtering of `tests/arch-*` should depend on the host system.
6. To prevent the Mold test-suite output, I would consider suppressing `set -x`, which is set by common.inc.

marxin:
One potential integration opportunity would be adding Mold as a submodule of the project.
@davidlattimore: What do you think about it?

davidlattimore:
> One potential integration opportunity would be adding Mold as a submodule of the project.
> @davidlattimore: What do you think about it?

I guess that should be OK. Perhaps in a subdirectory. e.g. `external_test_suites/mold`

lapla-cogito:
> > One potential integration opportunity would be adding Mold as a submodule of the project.
> > @davidlattimore: What do you think about it?
> 
> I guess that should be OK. Perhaps in a subdirectory. e.g. `external_test_suites/mold`

I'm generally reluctant to use external test suites as submodules:
1. Submodules require the entire repository. Since we only need the test suite itself, this is excessively redundant.
2. If there are updates to the external test suite and we want to incorporate those changes, we'd need to explicitly update the submodule (this is just one example - we could maintain the mold test suite itself as a GitHub Actions cache and periodically update the cache at fixed intervals like weekly via cron jobs, allowing CI to automatically obtain the latest results without manual intervention).

marxin:
> * Submodules require the entire repository. Since we only need the test suite itself, this is excessively redundant.

Well, yes, but we're speaking about a rather small git repository.

> * If there are updates to the external test suite and we want to incorporate those changes, we'd need to explicitly update the submodule (this is just one example - we could maintain the mold test suite itself as a GitHub Actions cache and periodically update the cache at fixed intervals like weekly via cron jobs, allowing CI to automatically obtain the latest results without manual intervention).

That's, from my perspective, a desired feature. Every time we pull Mold's changes, one needs to triage new tests and mark them as passing or failing. Otherwise, everyone from Wild's developers will stick to a different revision of Mold, and then we'll see test suite pass rate inconsistency.

Plus, having the test suite at a fixed location will simplify any current code that deals with the proper finding of the tests.

mati865:
You also do not always have to initialize the submodules. This is how the Rust compiler repository works. It has some necessary submodules, like docs, but you can skip others depending on your build configuration.

davidlattimore:
> That's, from my perspective, a desired feature. Every time we pull Mold's changes, one needs to triage new tests and mark them as passing or failing. Otherwise, everyone from Wild's developers will stick to a different revision of Mold, and then we'll see test suite pass rate inconsistency.

I think this argument is a pretty compelling reason to want either a submodule, or some other mechanism that involves having the commit hash of mold to use checked into our repo.

> 1. I would really like direct integration of the individual tests from the Mold's test suite using the following compile-time glob listing: https://docs.rs/rstest/0.25.0/rstest/attr.rstest.html#files-path-as-input-arguments.

This would definitely be very nice and from the look of things, I think shouldn't be too hard.

It's also fine to leave changes until later PRs if you prefer - just let us know which things you'd like to leave for later.

RossSmyth:
Submodules and nix flakes interact extremely poorly for very silly reasons. I have no clue how to get them to work together after messing with it for a few minutes and googling around. If it is holding you up, you can comment out these lines and it should work 
https://github.com/davidlattimore/wild/blob/c2e0af40fd040a556d9ebf27e93d7f1ed4893f8e/flake.nix#L49-L57

davidlattimore:
It looks as though rstest is giving an error at build time if the file pattern doesn't exist. That's unfortunate. The only solution I can think of at the moment would be to put that test behind a `cfg`. e.g. `#[cfg(feature = "mold-tests")]`

davidlattimore:
Or, if we wanted it to be driven by `test-config.toml` then we could have a `build.rs` that parses `test-config.toml` and if the mold tests are enabled (which could be a boolean in the config) prints `cargo::rustc-cfg=mold_tests`. It'd also need to print `cargo::rustc-check-cfg=mold-tests` regardless of whether they're enabled. It should also print `cargo::rerun-if-changed=test-confiig.toml`. Then in the code, you can use `#[cfg(mold-tests)]`.

marxin:
I'm curious if we could simply remove the `build.rs` file (and the integration into `config.toml`) and manually drive the testing of Mold's test suite using a `cfg` feature? Doing that, we'll end up with a much smaller change?

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2435,43 @@ fn integration_test(
     Ok(())
 }
 
+#[rstest]
+#[cfg(mold_tests)]
+fn exec_mold_tests(
+    #[files("../external_test_suites/mold/test/*.sh")] mold_test: PathBuf,
```
marxin, line 2432: To be honest, I am pretty excited about how elegant the listing is and how the tests run in parallel, great!
Are you planning to come up with a list of excluded tests via `#exclude()` directive:
https://docs.rs/rstest/latest/rstest/attr.rstest.html#files-path-as-input-arguments
?

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2435,43 @@ fn integration_test(
...
+    }
+
+    let path = env::var("PATH").unwrap();
+    let current_dir = env::current_dir()?;
+    let wild_dir = current_dir.parent().unwrap().join("fakes-debug");
+
+    let output = Command::new("bash")
+        .arg("-c")
+        .arg(&mold_test)
+        .env("PATH", format!("{wild_dir:?}:{path}"))
```
marxin, line 2442: I noticed many of the tests fail because of the following error:
```
+ ./mold --relocatable -o out/test/x86_64/many-input-sections2/b.o out/test/x86_64/many-input-sections2/a.o
/home/marxin/Programming/wild/external_test_suites/mold/test/many-input-sections2.sh: line 11: ./mold: No such file or directory
++ on_error 11
++ code=127
```

That means, we might consider adding a `mold` symlink to `fakes-debug` as well?

marxin (changes requested):

lapla-cogito:
> I'm curious if we could simply remove the build.rs file (and the integration into config.toml) and manually drive the testing of Mold's test suite using a cfg feature? Doing that, we'll end up with a much smaller change?

I was exactly struggling with the best way to centrally manage `ExternalTestSuites` when your suggestion about using feature flags seemed particularly promising. Thanks!

davidlattimore:
> > I'm curious if we could simply remove the build.rs file (and the integration into config.toml) and manually drive the testing of Mold's test suite using a cfg feature? Doing that, we'll end up with a much smaller change?
> 
> I was exactly struggling with the best way to centrally manage `ExternalTestSuites` when your suggestion about using feature flags seemed particularly promising. Thanks!

I guess that means that we don't need anything in `test-config.toml` - whether we look for and run mold's test suite is controlled entire by `--features mold-tests` or something like that. That is nice and simple :) It also has the advantage that we can choose from run to run whether we want to run those tests.

fakes/mold
```diff
@@ -0,0 +1 @@
+fakes/ld
```
davidlattimore: Can you try running `fakes/mold --version` and confirm that it works and reports that it's Wild? Just that this symlink target doesn't look right to me

wild/Cargo.toml
```diff
@@ -37,7 +37,13 @@ toml = { workspace = true }
 wait-timeout = { workspace = true }
 which = { workspace = true }
 
+[build-dependencies]
```
davidlattimore: Looks like you missed these?

wild/tests/integration_tests.rs
lapla-cogito: Thank you! I hadn't noticed that.

fakes/mold
lapla-cogito: It was a broken symlink, fixed.

wild/tests/integration_tests.rs
lapla-cogito: Yes, I agree that filtering tests using something like the `exclude` attribute would be necessary. However, this wasn't originally planned as part of this PR. For the future of this feature, current plan is to automatically categorize test execution results from mold tests based on error messages, aggregate them into CSV files, and integrate them into CI. Therefore, deciding what mold test cases should be excluded when implementing this categorization feature as a patch would be most effective when determining which test cases should be excluded.

davidlattimore: I know generation of CSV files was something I suggested, but at the time, I was imagining running the tests with a separate tool rather than running them as part of the test suite. If running as part of the test suite, then CSV perhaps doesn't make as much sense. So I'd probably suggest only doing CSV if we can see a genuine use for it.

Related to automating the classification of the failures... I have no idea how practical that will be. Manually classifying each failure might be easier. It'll partly depend on how many failures there are and in what way they fail. I guess unknown flags should be easy to identify. But if the binary just misbehaves, then that could be one of many different causes.

Rather than excluding the failing tests, I'd probably suggest doing something that Martin suggested earlier - having a list of tests that are expected to fail. If we do that, then we run those tests anyway and if they pass, we fail the test. That way when those tests start passing, we're forced to remove them from the list of tests that are expected to fail.

I know this is deviating a bit from what I put in the original project proposal and from what you put in your proposal, but it still meets the same underlying goals and represents a potential significant simplification, which would give you more time to work on extension work.

As for this PR, I'm happy for it to be merged as-is. Handling of failing tests can be left for a later PR.

lapla-cogito: Got it, let's discuss the details of how we'll aggregate test results or classifying failures in Zulip or during meetings.

davidlattimore (approved):

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2426,35 @@ fn integration_test(
...
 }
 
+#[rstest]
+#[cfg(feature = "mold_tests")]
+fn exec_mold_tests(
+    #[files("../external_test_suites/mold/test/*.sh")] mold_test: PathBuf,
+) -> Result {
+    let path = env::var("PATH").unwrap();
+    let current_dir = env::current_dir()?;
+    let wild_dir = current_dir.parent().unwrap().join("fakes-debug");
```
marxin, line 2436: Seems to me, various tests that execute `./mold` fail:
```
Error: Mold test `/home/marxin/Programming/wild/external_test_suites/mold/test/relocatable-many-sections.sh` failed with status: exit status: 127
stdout: Testing relocatable-many-sections ... command failed: 19: ./mold -r -o $t/c.o $t/a.o $t/b.o

stderr: + nm mold
+ grep __tsan_init
nm: 'mold': No such file
+ seq 1 80000
+ sed 's/.*/.section .data.&,"aw"\n.word 0\n/g'
+ cc -c -xassembler -o out/test/x86_64/relocatable-many-sections/a.o -
+ cc -c -xc -o out/test/x86_64/relocatable-many-sections/b.o -
+ cat
+ ./mold -r -o out/test/x86_64/relocatable-many-sections/c.o out/test/x86_64/relocatable-many-sections/a.o out/test/x86_64/relocatable-many-sections/b.o
/home/marxin/Programming/wild/external_test_suites/mold/test/relocatable-many-sections.sh: line 19: ./mold: No such file or directory
++ on_error 19
++ code=127
++ echo 'command failed: 19: ./mold -r -o $t/c.o $t/a.o $t/b.o'
++ trap - EXIT
++ exit 127
```

As seen, many tests expect to be executed from a location where `./mold` is present. That said, I would probably change working directory to the root of the tests, before running the script.

wild/tests/integration_tests.rs
marxin: Fully agree with @davidlattimore where we should make a manual investigation of the failing tests, fill-up the corresponding issues, and list all of them into a known-to-fail list. Based on my experiments, on x86_64, the pass rate is pretty poor: `41% tests passed, 250 tests failed out of 423`, but still, there are many tests with a common root of failure:

```
	368 - x86_64-version-script10 (Failed)
	369 - x86_64-version-script11 (Failed)
	370 - x86_64-version-script12 (Failed)
	371 - x86_64-version-script13 (Failed)
	372 - x86_64-version-script14 (Failed)
	373 - x86_64-version-script15 (Failed)
	374 - x86_64-version-script16 (Failed)
	375 - x86_64-version-script17 (Failed)
	376 - x86_64-version-script18 (Failed)
	377 - x86_64-version-script19 (Failed)
	378 - x86_64-version-script2 (Failed)
	379 - x86_64-version-script20 (Failed)
	381 - x86_64-version-script22 (Failed)
	382 - x86_64-version-script23 (Failed)
	383 - x86_64-version-script3 (Failed)
	384 - x86_64-version-script4 (Failed)
	385 - x86_64-version-script5 (Failed)
	386 - x86_64-version-script6 (Failed)
	388 - x86_64-version-script8 (Failed)
...
	263 - x86_64-relocatable-archive (Failed)
	264 - x86_64-relocatable-c++ (Failed)
	266 - x86_64-relocatable-debug-info (Failed)
	267 - x86_64-relocatable-exception (Failed)
	268 - x86_64-relocatable-many-sections (Failed)
	269 - x86_64-relocatable-merge-sections (Failed)
	270 - x86_64-relocatable-mergeable-sections (Failed)
...
	148 - x86_64-gdb-index-compress-output (Failed)
	149 - x86_64-gdb-index-dwarf2 (Failed)
	150 - x86_64-gdb-index-dwarf3 (Failed)
	151 - x86_64-gdb-index-dwarf4 (Failed)
	152 - x86_64-gdb-index-dwarf5 (Failed)
	153 - x86_64-gdb-index-dwarf64 (Failed)
	155 - x86_64-gdb-index-split-dwarf (Failed)
```

.github/workflows/ci.yml
```diff
@@ -59,17 +59,18 @@ jobs:
...
+      - run: apt-get update && apt-get -y install gcc g++ clang lld curl bubblewrap binutils-aarch64-linux-gnu binutils-riscv64-linux-gnu git
         if: ${{ contains(matrix.container, 'ubuntu') }}
-      - run: apt-get update && apt-get -y install qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-riscv64-linux-gnu g++-riscv64-linux-gnu
+      - run: apt-get update && apt-get -y install qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-riscv64-linux-gnu g++-riscv64-linux-gnu git
         if: ${{ matrix.test-qemu }}
-      - run: zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap
+      - run: zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap git
         if: ${{ contains(matrix.container, 'opensuse') }}
-      - run: apk add build-base lld clang bash curl
+      - run: apk add build-base lld clang bash curl git
```
marxin, line 68: Do you need a git binary? How does `actions/checkout@v4` action work if you don't have a `git` binary, and how is it different from `submodules: recursive`?

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2426,35 @@ fn integration_test(
     Ok(())
 }
 
+#[rstest]
+#[cfg(feature = "mold_tests")]
+fn exec_mold_tests(
+    #[files("../external_test_suites/mold/test/*.sh")] mold_test: PathBuf,
+) -> Result {
+    let path = env::var("PATH").unwrap();
```
marxin: I know it's very unlikely the `env` is not set, but I would rather add `.context` and use `?` (instead of the unwrap);

marxin (changes requested):

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2426,35 @@ fn integration_test(
...
+    let current_dir = env::current_dir()?;
+    let wild_dir = current_dir.parent().unwrap().join("fakes-debug");
+
+    let output = Command::new("bash")
+        .arg("-c")
+        .arg(&mold_test)
+        .env("PATH", format!("{wild_dir:?}:{path}"))
```
marxin, line 2442: One limitation here is that you basically rely on the fact that the wild binary is already built (and up-to-date), but if you just run `cargo test`, the binary is not built. I am not sure if one can add such a dependency on a binary target.
@davidlattimore  ?

.github/workflows/ci.yml
lapla-cogito: While `actions/checkout@v4` itself can function without installing a new git binary, it appears to use the installed git version, which is apparently outdated. The workflow attempts to update repositories via the GitHub REST API, which seems to be causing the failure ([example here](https://github.com/davidlattimore/wild/actions/runs/15852585930/job/44689652692)). Therefore, we explicitly install git.

wild/tests/integration_tests.rs
lapla-cogito: I've updated the list of currently passing tests the test execution results presented at the beginning of this PR. Based on this, I'm going to begin manual investigation.

.github/workflows/ci.yml
marxin: Thanks for the explanation!

wild/tests/integration_tests.rs
marxin: Sounds good!

wild/tests/integration_tests.rs
```diff
@@ -2425,6 +2426,35 @@ fn integration_test(
...
+#[rstest]
+#[cfg(feature = "mold_tests")]
+fn exec_mold_tests(
+    #[files("../external_test_suites/mold/test/*.sh")] mold_test: PathBuf,
+) -> Result {
+    let path = env::var("PATH")?;
+    let current_dir = env::current_dir()?;
+    let wild_dir = current_dir.parent().unwrap().join("fakes-debug");
+
+    let output = Command::new("bash")
```
marxin, line 2438: One last nit: can you please combine stdout+stderr directly into one stream? Having that, we'll better see the interleaved messages like Wild's warnings directly at the place they happen.

lapla-cogito: Right, that does sound more convenient. Fixed, thanks!

wild/tests/integration_tests.rs
davidlattimore: Cargo will build binaries before it runs integration tests. You can find the location of those binaries at build time via the environment variable `CARGO_BIN_EXE_<name>` e.g. `env!("CARGO_BIN_EXE_wild")`. So for maximum correctness, the test would need to create a temporary directory and put symlinks into it that pointed to the actual location of the wild binary. It would be somewhat inconvenient however if that temporary directory got deleted after the tests finished, because it'd make it harder to do manual checks of what happened - e.g. the out directory would vanish. I suppose one option would be to always use the same directory and update the symlinks when the test runs if they're not correct, then .gitignore the whole directory.

davidlattimore: But I think using `fakes-debug` should work OK so long as you don't `cargo test --release`, specify some other profile or override the target directory.

lapla-cogito: Since we plan to run external tests via CI, and the `wild` binary should be present during integration test execution (except when running only `cargo test --release --features mold_test`), I don't think this is a major issue.

marxin (approved):

PR #912, opened 2025-06-22
Refactor Format impl for SectionFlags

marxin:


davidlattimore:
Nice! Your change picked up a bug. Looks like we're propagating the retain bit to the output section when we probably shouldn't be.

marxin:
Haha, cool! Note, I used Copilot to transform, and it exhaustively converted the code for all `shf` constants.

@davidlattimore can you try working on the underlying issue?

davidlattimore:
> can you try working on the underlying issue?

done - https://github.com/davidlattimore/wild/pull/914

Hopefully if you rebase or merge, it should now pass. I did try to reproduce on my RPi first, but the problem didn't show up there. Perhaps because it's running Debian not Ubuntu.

marxin:
@davidlattimore  Thanks for the fix, however, it seems there's still a fallout.

davidlattimore:
Hmm. Seems that sometimes GNU ld does propagate the retain bit. Maybe just comment out the printing of the retain bit for now.

davidlattimore (approved):

Issue #700, opened 2025-04-24
Support merging of .riscv.attributes sections


marxin:
Documentation of the section format:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#layout-of-riscvattributes-section

Example of the section content:
```
Attribute Section: riscv
File Attributes
  Tag_RISCV_stack_align: 16-bytes
  Tag_RISCV_arch: "rv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_zicsr2p0_zifencei2p0_zmmul1p0_zaamo1p0_zalrsc1p0"
```

Issue #916, opened 2025-06-22
Advanced support of .riscv.attributes sections


marxin:
The basic support is addressed in #700, and some features will be missing:
1. linker-diff support -> #1535 
2. correct merging of the `Tag_RISCV_arch` ISA extensions: verification of the conflicting extensions -> #1526
3. direct write to the output file: `write_riscv_attributes`
4. support for vendor-specific sub-sections

lapla-cogito:
I'm not fully clear on the advantages of writing `write_riscv_attributes` directly without using an intermediate buffer. `.riscv.attributes` is typically very small, and mixing direct file I/O would likely introduce overhead from file handle synchronization with other threads. While it would eliminate the need to allocate buffer memory for `.riscv.attributes`, I don't think this advantage outweighs the overhead concerns. Or am I missing some crucial point here?

marxin:
> Or am I missing some crucial point here?

No, you are right - it's probably not worth implementing.
Btw. are you planning to work on point 4), or can we close this issue? Btw. that you for working on the issue, good work.

lapla-cogito:
Thanks for your response! Regarding point 4, I'm not currently working on it, so implementing it when needed should be sufficient.

PR #917, opened 2025-06-22
Initial support for .riscv.attributes

marxin:
The implementation encompasses all the necessary components for running tests on my RISC-V board. The missing pieces are listed in the #916.

Fixes: #700

### DIFF
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -30,6 +30,7 @@ libc = { workspace = true }
 linker-layout = { path = "../linker-layout", version = "0.5.0" }
 linker-trace = { path = "../linker-trace", version = "0.5.0" }
 linker-utils = { path = "../linker-utils", version = "0.5.0" }
+leb128 = {workspace = true}
 memchr = { workspace = true }
 memmap2 = { workspace = true }
 object = { workspace = true }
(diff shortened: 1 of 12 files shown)


libwild/Cargo.toml
```diff
@@ -43,6 +44,7 @@ tracing-subscriber = { workspace = true }
 uuid = { workspace = true }
 winnow = { workspace = true }
 zstd = { workspace = true }
+indexmap = "2.9.0"
```
davidlattimore: It'd be good to make this a workspace dependency like all the others

libwild/src/elf_writer.rs
```diff
@@ -2655,6 +2670,65 @@ fn write_gnu_property_notes(
     Ok(())
 }
 
+fn write_riscv_attributes(
+    epilogue: &EpilogueLayout,
+    buffers: &mut OutputSectionPartMap<&mut [u8]>,
+) -> Result {
+    // TODO: write the output directly to the output buffer
```
davidlattimore: I think you can create a Cursor directly from the `&mut [u8]`. e.g.

```rust
let mut writer = Cursor::new(buffers.get_mut(part_id::RISCV_ATTRIBUTES));
```

marxin: Oh, neat, I had to reborrow it, but it works correctly.

PR #918, opened 2025-06-22
Run cargo-update

marxin:
Seems some of the released versions are >2 weeks old, which is weird given the fact `dependabot` was run last time a week ago.


mati865:
Yeah, the tooling is not quite there yet: https://github.com/davidlattimore/wild/pull/751
I've tried with Renovate later on, and it updated a few more packages but neither upgraded all of them.

PR #919, opened 2025-06-22
 change the description for `-Bsymbolic-non-weak` tests

lapla-cogito:
Previously, `linker-diff` would panic and prevent adding tests for this option, but now it seems possible to do so.


lapla-cogito:
Ah, it appears the Ubuntu 22.04 lld version doesn't implement this option, which is causing the CI to fail. In any case, since the actual phenomenon differs from what's documented in the TODO, I'd like to limit this PR to just updating the comment.

PR #920, opened 2025-06-22
linker-diff for RISC-V

### LINKED ISSUE #776: RISC-V linker-diff support
Initial support for RISC-V is close to landing in Wild. We'd like to support RISC-V in linker-diff too, since it makes it easier to diagnose problems.

The starting point is to look for calls to `report_arch_specific_diffs` and add an equivalent call for RISC-V. This will involve implementing linker-diff's `Arch` trait for RISC-V. The code in `linker-diff/src/aarch64.rs` can be used as a bit of a guide.

The function `decode_insn_with_objdump` in `aarch64` should probably be moved somewhere common so that it can be reused for risc-v - that's assuming there isn't a lightweight, fast to build risc-v instruction decoder. For context, on x86-64, we use a Rust crate to decode instructions. For a short time, we did the same on aarch64, however the disassembler for aarch64 was very slow to compile, so we switched to running objdump to disassemble instructions. My guess is that we should probably use objdump on ~~aarch64~~ risc-v as well. I even wouldn't be opposed to just using objdump everywhere for consistency.

lapla-cogito:
close #776

### DIFF
--- a/linker-diff/src/aarch64.rs
+++ b/linker-diff/src/aarch64.rs
@@ -1,11 +1,10 @@
+use crate::ArchKind;
 use crate::arch::Arch;
 use crate::arch::Instruction;
 use crate::arch::Relaxation;
 use crate::arch::RelaxationByteRange;
 use crate::asm_diff::BasicValueKind;
-use anyhow::Context;
-use anyhow::Result;
-use itertools::Itertools;
+use crate::utils::decode_insn_with_objdump;
 use linker_utils::aarch64::DEFAULT_AARCH64_PAGE_IGNORED_MASK;
 use linker_utils::aarch64::DEFAULT_AARCH64_PAGE_SIZE_BITS;
 use linker_utils::aarch64::RelaxationKind;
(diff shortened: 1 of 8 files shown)


marxin:
I would like to continue working on the RISC-V port somewhen in September. @lapla-cogito Are you planning to work on this PR, please? It's definitely something that will make my life easier.

lapla-cogito:
Sorry for the delay, I'll work on this in this week.

marxin:
> Sorry for the delay, I'll work on this in this week.

Don't hurry, it's great you've been working on it and don't hesitate to ask me about anything if you feel you're stuck.

lapla-cogito:
I'm not sure why some tests hang and never finish when I run them with `WILD_TEST_CROSS=riscv64`. The same thing happens both locally and in the CI environment...

marxin:
> I'm not sure why some tests hang and never finish when I run them with `WILD_TEST_CROSS=riscv64`.

Based on my initial investigation, it's connected to your changes where you decode the instructions with objdump (`linker_diff::utils::decode_insn_with_objdump`), So I guess you're either doing too many invocations of the tool, or there's some glitch connected to read/write with the sub-process. Please add some verbose logging that will help you.

lapla-cogito:
This PR didn't implement relaxation itself, which caused some tests to hang due to repeated calls to the relatively heavy `add_diff` function within `report_section_diffs`. I addressed this by checking which sections should be ignored before calling `add_diff`, preventing unnecessary calls where possible.

marxin:
Thank you for working on that @lapla-cogito!

PR #922, opened 2025-06-22
CI: enable 25.10 again

marxin:


lapla-cogito:
I've been testing a minimal repro of this CI failure in a private repository. Based on that, it looks like the issue has been resolved, so I've re-run the workflow. Merging should now be possible.

Issue #508, opened 2025-03-01
Jobserver integration


davidlattimore:
Wild by default uses as many threads as there are CPUs (or hyperthreads) on the system. When the system is heavily loaded because we're running `cargo build` or similar, this might be suboptimal. We might be better off using less threads and allowing the rest of the build process to use some of those resources instead.

We should investigate options for connecting to a jobserver and using that to limit parallelism.

See [GNU make jobserver docs](https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html)

See [docs for how rustc uses this](https://doc.rust-lang.org/rustc/jobserver.html).

The  [jobserver](https://crates.io/crates/jobserver) crate looks like it'd probably work well enough.

The simplest integration is that we acquire as many tokens as we can on startup, use that to set the size of our threadpool and then release them on shutdown. More complicated integrations would involve releasing tokens when we go single-threaded (but we only do that for very short periods of time) and attempting to acquire even more tokens throughout the linking process. I'm not sure either of those would be worthwhile though.

If `--threads` is specified on the commandline, then I think that should probably cause us to not use the jobserver.

PR #923, opened 2025-06-22
Jobserver integration

marxin:
Resolves: #508

### DIFF
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -27,6 +27,7 @@ hashbrown = { workspace = true }
 hex = { workspace = true }
 indexmap = { workspace = true }
 itertools = { workspace = true }
+jobserver = "0.1.33"
 libc = { workspace = true }
 linker-layout = { path = "../linker-layout", version = "0.5.0" }
 linker-trace = { path = "../linker-trace", version = "0.5.0" }
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -27,7 +27,7 @@ hashbrown = { workspace = true }
 hex = { workspace = true }
 indexmap = { workspace = true }
 itertools = { workspace = true }
-jobserver = "0.1.33"
+jobserver = { workspace = true }
 libc = { workspace = true }
 linker-layout = { path = "../linker-layout", version = "0.5.0" }
 linker-trace = { path = "../linker-trace", version = "0.5.0" }
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -851,9 +851,10 @@ impl Args {
             }
         });
 
-        ThreadPoolBuilder::new()
+        // The pool might be already initialized, supress the error intentionally.
+        let _ = ThreadPoolBuilder::new()
             .num_threads(self.available_threads.get())
-            .build_global()?;
+            .build_global();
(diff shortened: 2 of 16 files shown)


libwild/src/args.rs
```diff
@@ -716,11 +722,33 @@ impl Args {
...
-    pub(crate) fn setup_thread_pool(&self) -> Result {
+    /// Sets up the thread pool, using the explicit number of threads if specified,
+    /// or falling back to the jobserver protocol if available.
+    /// and return the number of token which will be returned at drop place.
+    ///
+    /// https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html
+    pub(crate) fn setup_thread_pool(&mut self) -> Result<Vec<Acquired>> {
+        let mut tokens = Vec::new();
+        self.available_threads = self.num_threads.unwrap_or_else(|| {
+            // SAFETY: should be called early before other descriptors are opened
```
davidlattimore: How sure do you think we can be that we haven't opened any files at this point? I guess one other option would be to create the client earlier, perhaps even before argument parsing, then pass it in. I'm happy to make this change if you don't want to.

libwild/Cargo.toml
```diff
@@ -26,6 +26,7 @@ gimli = { workspace = true }
 hashbrown = { workspace = true }
 hex = { workspace = true }
 itertools = { workspace = true }
+jobserver = "0.1.33"
```
davidlattimore: Declare in workspace?

davidlattimore (approved):

libwild/src/args.rs
marxin: I recall various unusual issues from pre-4.4 GNU make (released in October 2022), where named pipes were not utilized, and a pair of file descriptors was passed to the child process. With named pipes, it does not matter if any file descriptors are opened before `Client::from_env`, and thus I would like to stick with the setup in `setup_thread_pool` function. It belongs there logically, I guess.

davidlattimore: I tried checked what environment variables cargo is passing to rustc. For me at least, it looks like `CARGO_MAKEFLAGS=-j --jobserver-fds=8,9 --jobserver-auth=8,9`. So, unfortunately it seems to be using file descriptors, not named pipes.

davidlattimore: I guess one slight mitigation is that it looks like the implementation tries to verify that the file descriptors are actually pipes and any files that we've opened already are most likely not pipes. I'm not sure what would happen if used from a program that already had multiple threads running. This made me think to check what integration_test does. Looks like it doesn't call `setup_thread_pool`, which I guess means that rayon will default-initiialise it based on the number of CPU cores / hyperthreads. There are various bits of code that check how many threads we're running with, which before would have see the value returned by `available_parallelism`, but now will see 1, so will likely take slightly different code paths.

marxin: Thanks for the testing effort. I decided to create an abstraction where `Args` need to be promoted to `ActivatedArgs`, allowing `Client` to be constructed early but activated at a specific point.
What do you think about it?

davidlattimore (approved):
Looks good to me. We can always revisit the API later if we decide something else would fit better. Thanks for your work on this!

PR #936, opened 2025-06-27
Filter mold tests by architecture

lapla-cogito:
Some mold tests specify their target architecture in the filename. This PR adds functionality to skip tests for architectures that shouldn't run, based on the host architecture and the `WILD_TEST_CROSS` setting.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -2584,6 +2584,15 @@ fn integration_test(
     Ok(())
 }
 
+fn get_wild_test_cross() -> Option<Vec<Architecture>> {
+    std::env::var("WILD_TEST_CROSS").ok().map(|cross_arch| {
+        cross_arch
+            .split(',')
+            .filter_map(|s| Architecture::from_str(s).ok())
+            .collect()
+    })
+}
+
 #[rstest]
 #[cfg(feature = "mold_tests")]
 fn exec_mold_tests(
@@ -2592,6 +2601,10 @@ fn exec_mold_tests(
     let path = env::var("PATH")?;
     let current_dir = env::current_dir()?;
     let wild_dir = current_dir.parent().unwrap().join("fakes-debug");
+
+    if should_skip_mold_test(&mold_test) {
+        return Ok(());
+    }
 
     let output = Command::new("bash")
         .current_dir("../fakes-debug")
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -2613,6 +2617,50 @@ fn exec_mold_tests(
     Ok(())
 }
 
+#[cfg(feature = "mold_tests")]
```
davidlattimore, line 2629: Do you think we should move the mold test support to a module. e.g. have a file `mold_tests.rs`. Then we can just have:

```rust
#[cfg(feature = "mold_tests")]
mod mold_tests;
```

So we wouldn't need to repeat the cfg for each helper function that's needed. Fine to do so in a separate PR if you prefer.

wild/tests/integration_tests.rs
```diff
@@ -2613,6 +2617,50 @@ fn exec_mold_tests(
...
+        return false;
+    }
+
+    let test_arch = file_name["arch-".len()..]
+        .split('-')
+        .next()
+        .unwrap_or_default();
+
+    let current_arch = get_host_architecture();
+    let cross_archs = std::env::var("WILD_TEST_CROSS")
```
davidlattimore: It'd be good to not repeat the logic for parsing this environment variable. Could we extract the existing logic out into a function that we can then call here?

lapla-cogito: Thanks, done.

wild/tests/integration_tests.rs
lapla-cogito: To achieve this, creating files like `wild/tests/common/mod.rs` and moving shared functions, variables or more there seems like the logical approach. However, this would require significant changes, so I'd prefer to implement this in a separate pull request.

wild/tests/integration_tests.rs
```diff
@@ -2639,18 +2692,8 @@ fn read_test_config() -> Result<TestConfig> {
...
-            .map(|s| s.trim())
-            .filter(|s| !s.is_empty())
-            .map(|s| {
-                Architecture::from_str(s)
-                    .with_context(|| format!("Unknown WILD_TEST_CROSS value `{s}`"))
-            })
-            .collect::<Result<Vec<Architecture>>>()?
-    }
+    // The environment variable `WILD_TEST_CROSS` can override the config file setting.
+    config.qemu_arch = get_wild_test_cross();
```
davidlattimore: It looks like the value of qemu_arch from the config file will now always be ignored. Perhaps return an `Option<Vec<Architecture>>` from `get_wild_test_cross`. Then the other caller can use `unwrap_or_default` on the result, while this call-site can conditionally assign.

wild/tests/integration_tests.rs
davidlattimore: That's fine by me to do the move in a separate PR. Note though, that I'm not suggesting having a common module - just a module for mold tests.

lapla-cogito: What I wanted to say was that while we'll create a module specifically for mold tests just like you're suggesting, some functions like `get_host_architecture` need to be shared, and these should be placed in a common module instead (I believe you can't share items between modules without doing this under `tests/`. If there's a more convenient way to do that, that might be better).

davidlattimore: For now, I'd suggest leaving the rest of `integration_test.rs` where it is and just moving the mold stuff to a submodule. That submodule can access functions in the parent module - e.g. `super::get_host_architecture`

PR #937, opened 2025-06-27
Bump minimal Rust version to 1.88

lapla-cogito:
The second commit is primarily due to stabilizing `let_chains`.


lapla-cogito:
Will Nix-related errors disappear once rustc 1.88.0 is added to nixpkgs (and we run CI again)?
(edit: It seems such a PR has already been submitted.)

davidlattimore:
I've just updated the nix `flake.lock` in #940. You could rebase your change on top of that and see if that fixes it?

lapla-cogito:
Thanks, but it seems that updating `flake.lock` after [this PR](https://github.com/NixOS/nixpkgs/pull/420390) is merged will still be necessary.

davidlattimore:
> Thanks, but it seems that updating `flake.lock` after [this PR](https://github.com/NixOS/nixpkgs/pull/420390) is merged will still be necessary.

~~I'd be surprised if that made a difference. I suspect we'll need to wait until Nix updates rustc. I'm not sure how to tell when that has happened though.~~ edit: I didn't notice the link in your previous comment.

mati865:
Perhaps we should avoid using Nix provided Rust and instead use rustup to avoid such problems?

davidlattimore:
> Thanks, but it seems that updating `flake.lock` after this PR is merged will still be necessary.

I just realised that I hadn't noticed the `this PR` link. I thought you were referring to this PR that we're commenting on. Yes, the PR you linked to looks like one that we should try again after it's merged.

> Perhaps we should avoid using Nix provided Rust and instead use rustup to avoid such problems?

Maybe. Eventually though, I think we'll want to slow a little with bumping our MSRV to make it easier for distros to package.

davidlattimore:
The PR was merged yesterday, but it looks like that only updates "staging", which AFAICT is a step before "unstable", which it looks like our `flake.nix` uses. From what I can tell, we can look in [this directory](https://github.com/NixOS/nixpkgs/tree/nixos-unstable/pkgs/development/compilers/rust) to see what version of rustc is currently available on unstable.

RossSmyth:
Yeah, staging is usually merged every few weeks. Using rustup isn't an option for Nix because software built with Nix uses a sandbox without network access.

RossSmyth:
If bumped to 1.88, the only real option is using https://github.com/oxalica/rust-overlay again
Let me know if this happens and I'll do it.

RossSmyth:
https://nixpk.gs/pr-tracker.html?pr=420390

PR #943, opened 2025-06-27
Version script glob support

marxin:
Based on the discussion with David, I tested the bevy with dynamic linking (500K symbols in a version script file), and I decided to keep the exact match implementation. The only change is that `glob::Pattern` is used for a general glob pattern; having that, the link time is equal for `bevy`. For a reasonably sized shared library that uses a couple of general globs, the speed should be good!

Plus the PR support if a space(s) is present after a symbol name.

### DIFF
--- a/libwild/src/version_script.rs
+++ b/libwild/src/version_script.rs
@@ -46,10 +46,7 @@ impl<'data> MatchRules<'data> {
     fn push(&mut self, pattern: SymbolMatcher<'data>) {
         match pattern {
             SymbolMatcher::All => self.matches_all = true,
-            SymbolMatcher::Glob(glob) => self.globs.push(
-                // TODO
-                Pattern::new(str::from_utf8(glob).unwrap()).unwrap(),
-            ),
+            SymbolMatcher::Glob(glob) => self.globs.push(glob),
             SymbolMatcher::Exact(exact) => {
                 self.exact.insert(UnversionedSymbolName::prehashed(exact));
             }
@@ -59,11 +56,9 @@ impl<'data> MatchRules<'data> {
     fn matches(&self, name: &PreHashed<UnversionedSymbolName>) -> bool {
         self.matches_all
             || self.exact.contains(name)
-            || self
-                .globs
-                .iter()
-                // TODO
-                .any(|glob| glob.matches(str::from_utf8(name.bytes()).unwrap()))
+            || self.globs.iter().any(|glob| {
+                glob.matches(str::from_utf8(name.bytes()).expect("Valid utf-8 identifier expected"))
+            })
     }
 
     fn merge(&mut self, other: &MatchRules<'data>) {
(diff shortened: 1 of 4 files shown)


libwild/src/version_script.rs
```diff
@@ -55,10 +56,9 @@ impl<'data> MatchRules<'data> {
     fn matches(&self, name: &PreHashed<UnversionedSymbolName>) -> bool {
         self.matches_all
             || self.exact.contains(name)
-            || self
-                .prefixes
-                .iter()
-                .any(|prefix| name.bytes().starts_with(prefix))
+            || self.globs.iter().any(|glob| {
+                glob.matches(str::from_utf8(name.bytes()).expect("Valid utf-8 identifier expected"))
```
davidlattimore: If someone is using a version with a glob pattern and any of their symbols aren't valid UTF-8, then this will panic and won't tell them what symbol contained invalid UTF-8. It'd be good to at least say which symbol was the problem. We can't easily print the non-UTF-8 character, but we can report the surrounding characters using `String::from_utf8_lossy`.

Also, if we have a few glob patterns, we're currently repeating utf-8 validation for each pattern. i.e. we'll incur a cost of N*M where N in the total length of all symbols and M is the number of patterns.

davidlattimore (commented):
Thanks! Definitely good to have support for more patterns. The utf-8 string thing does make me a bit uneasy, but I guess we can go with it and potentially change it to something byte-based later.

Out of curiosity, what kinds of patterns have you encountered that weren't `something*`? Is it just patterns that have `*` somewhere in the middle, or have you found use of `?` or character sets?

marxin:
> Out of curiosity, what kinds of patterns have you encountered that weren't `something*`? Is it just patterns that have `*` somewhere in the middle, or have you found use of `?` or character sets?

One example could be `alsa-lib`, where one of the versions is defined as:
```
$ wc -l ./work/alsa-lib-1.2.14-abi_x86_64.amd64/src/Versions
222 ./work/alsa-lib-1.2.14-abi_x86_64.amd64/src/Versions
...
$ grep '\*' ./work/alsa-lib-1.2.14-abi_x86_64.amd64/src/Versions
    snd_*;
    _snd_*_open;
    _snd_*_dlsym_*;
    _snd_*_poll_descriptor;
    _snd_pcm_hook_*;
    __snd_pcm_hw_params_*;
    __snd_pcm_sw_params_*;
    __snd_*_dlsym_*;
    *;
    _snd_safe_strto*;
    snd_ump_*;
    snd_seq_ump_*;
    snd_ump_endpoint_info_set_*;
    snd_ump_block_info_set_*;
```

libwild/src/version_script.rs
marxin: Sure, the error message should be improved.

Ah, good point, I reduced to N as I parse (to utf-8 string) the input symbol name only once.

marxin:
> The utf-8 string thing does make me a bit uneasy, but I guess we can go with it and potentially change it to something byte-based later.

Yep, my expectation is the glob patterns are not so common, and if so, then the total number of symbols is reasonably small.

marxin:
Well, in the context of #952, I believe, the symbol identifiers should be valid UTF-8 strings :)

PR #957, opened 2025-06-29
Don't emit non-standard .phdr and .shdr section headers

davidlattimore:
Other linkers don't emit sections for these parts of the file. Also, using SHT_NULL for sections other than section 0 causes problems for some tools, such as GNU strip.


marxin (approved):
I can confirm it fixed the `gpm` shared library, where `strip` command now makes the library unbroken.

marxin:
@davidlattimore Should we close the reported issue (https://sourceware.org/bugzilla/show_bug.cgi?id=33119) as invalid, or is the patch still worth being accepted? Note that llvm-strip works fine.

davidlattimore:
> @davidlattimore Should we close the reported issue (https://sourceware.org/bugzilla/show_bug.cgi?id=33119) as invalid, or is the patch still worth being accepted? Note that llvm-strip works fine.

I'd say comment on the issue that the linker in question has been updated to not emit those sections. They may still decide that they want to make the change.

marxin:
Yep, I've just sent the comment, leaving the decision to them.

PR #958, opened 2025-06-29
Unify parsing of `--option val` and `--option=val`

lapla-cogito:
Some options can be specified in either the `--option val` or `--option=val` format. Right now, they’re handled by separate branches in the code, but since both formats behave the same, we can merge them.


davidlattimore:
Nice!

PR #961, opened 2025-06-29
Reject .tcommon section

marxin:
It is probably very exotic construct, where right now, we panic in one of the Mold's tests (arch-x86_64-tls-large-tbss.sh):
```
thread '<unnamed>' panicked at libwild/src/elf_writer.rs:1280:17:
attempt to subtract with overflow
```

### DIFF
--- a/libwild/src/resolution.rs
+++ b/libwild/src/resolution.rs
@@ -44,6 +44,8 @@ use crossbeam_queue::ArrayQueue;
 use crossbeam_queue::SegQueue;
 use linker_utils::elf::SectionFlags;
 use linker_utils::elf::SectionType;
+use linker_utils::elf::secnames::TCOMMON_SECTION_NAME;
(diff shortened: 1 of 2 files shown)


davidlattimore:
Is there anything special about the name ".tcommon"? It seems to me like it's possibly a custom section with the TLS bit set and that we're not handling that case properly. I'll have a bit of a look to see if I can fix this another way.

davidlattimore:
It looks like the default linker script for GNU ld does list `.tcommon`. It puts it at the end of the TLS segment as part of .tbss. I looked into custom TLS sections and we weren't handling them correctly. I fixed that in #965. With that change, we no longer panic for that test case, although we still fail the test. I'm not sure why as yet.

marxin:
Thanks for taking a look. Yes, supporting the feature would be nicer than ignoring it. Let's open a proper issue: #966.

PR #967, opened 2025-06-30
Change behaviour of --no-allow-shlib-undefined to match lld

### LINKED ISSUE #897: Undefined reference error when linking executable depending on shared object with an undefined symbol
Right now, Wild reports any undefined symbol in shared objects when linking an executable. I have found this issue by building LLVM with `-DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON` CMake options:
```
❯ ninja bugpoint
[1/1] Linking CXX executable bin/bugpoint
FAILED: bin/bugpoint
: && /home/mateusz/Projects/wild-proxy/wild-g++ -fPIC -fno-semantic-interposition -fvisibility-inlines-hidden -Werror=date-time -Wall -Wextra -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wno-missing-field-initializers -pedantic -Wno-long-long -Wimplicit-fallthrough -Wno-maybe-uninitialized -Wno-nonnull -Wno-class-memaccess -Wno-redundant-move -Wno-pessimizing-move -Wno-noexcept-type -Wdelete-non-virtual-dtor -Wsuggest-override -Wno-comment -Wno-misleading-indentation -Wctad-maybe-unsupported -fdiagnostics-color -ffunction-sections -fdata-sections -O3 -DNDEBUG -Wl,--color-diagnostics -Wl,--export-dynamic tools/bugpoint/CMakeFiles/bugpoint.dir/BugDriver.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/CrashDebugger.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/ExecutionDriver.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/ExtractFunction.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/FindBugs.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/Miscompilation.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/OptimizerDriver.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/ToolRunner.cpp.o tools/bugpoint/CMakeFiles/bugpoint.dir/bugpoint.cpp.o -o bin/bugpoint  -Wl,-rpath,"\$ORIGIN/../lib:/home/mateusz/Projects/llvm-project/build/lib:"  lib/libLLVM.so.21.0git && :
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/15.1.1/liblto_plugin.so is not yet supported
Error: Failed to activate lib/libLLVM.so.21.0git (2816 (11/0))
  Caused by:
    undefined reference to `history` from lib/libLLVM.so.21.0git (2816 (11/0))

ninja: build stopped: subcommand failed.
```

Some more info about this symbol:
```
❯ ./run-with ~/Projects/wild/target/debug/wild --sym-info history
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/15.1.1/liblto_plugin.so is not yet supported
Global name `history` refers to: None
Definitions / references with name `history`:
  Global Undefined: symbol_id=2945 -> 2945 ABSOLUTE | DYNAMIC | FUNCTION res=[]
    #6 in File #2816 (11/0) ./home/mateusz/Projects/llvm-project/build/lib/libLLVM.so.21.0git (LOADED)
wild: error: Failed to activate ./home/mateusz/Projects/llvm-project/build/lib/libLLVM.so.21.0git (2816 (11/0))
  Caused by:
    undefined reference to `history` from ./home/mateusz/Projects/llvm-project/build/lib/libLLVM.so.21.0git (2816 (11/0))


~/tmp/llvm
❯ readelf -Ws ./home/mateusz/Projects/llvm-project/build/lib/libLLVM.so.21.0git | rg 'history$|Symbol table'
Symbol table '.dynsym' contains 38280 entries:
     6: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND history
Symbol table '.symtab' contains 163860 entries:
```

LD's description of this flag is not very obvious:
> Allows or disallows undefined symbols in shared libraries.

Indeed, sounds like it should prevent any undefined symbols, but digging deeper into `man` there is:
> The default behaviour is to report errors for any undefined symbols referenced in shared libraries if the linker is being used to create an executable

To me, it should not report any undefined symbol, just those the executable would use.

---

Additionally, this flag is completely ignored when creating shared objects, instead it should be disabled for them by default.

davidlattimore:
It turns out that LLD is more relaxed about undefined symbols in shared objects compared to GNU ld. We now hopefully do the same relaxed behaviour as LLD. That is, we only report an error if the shared object has all of its direct dependencies in the link.

Also, change the default back to --no-allow-shlib-undefined.

Fixes #897

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -288,8 +288,7 @@ impl Default for Args {
             files_per_group: None,
             exclude_libs: false,
             no_undefined: false,
-            // TODO: Change default to false once #897 is fixed.
-            allow_shlib_undefined: true,
+            allow_shlib_undefined: false,
             should_print_version: false,
             sysroot: None,
             save_dir: Default::default(),
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -22,8 +22,10 @@ use foldhash::fast::RandomState;
 use memmap2::Mmap;
 use rayon::iter::IntoParallelRefIterator;
 use rayon::iter::ParallelIterator;
+use std::ffi::OsStr;
 use std::fmt::Display;
 use std::ops::Deref;
+use std::os::unix::ffi::OsStrExt;
 use std::path::Path;
 use std::path::PathBuf;
 use std::time::Duration;
@@ -38,6 +40,10 @@ pub(crate) struct InputData<'data> {
 
     pub(crate) version_script_data: Option<VersionScriptData<'data>>,
     pub(crate) linker_scripts: Vec<InputLinkerScript<'data>>,
+
+    /// Which files we loaded. The keys aren't relevant anymore, but we keep them to avoid
+    /// regenerating the map.
+    path_to_load_index: HashMap<PathBuf, FileLoadIndex>,
 }
 
 pub(crate) struct InputBytes<'data> {
@@ -232,12 +238,12 @@ impl<'data> InputData<'data> {
             inputs: Vec::new(),
             linker_scripts: Vec::new(),
             version_script_data,
+            path_to_load_index: temporary_state.path_to_load_index,
         };
 
         inputs.extract_all(&mut temporary_state.files)?;
 
-        args.save_dir
-            .finish(temporary_state.path_to_load_index.keys())?;
+        args.save_dir.finish(inputs.path_to_load_index.keys())?;
 
         Ok(inputs)
     }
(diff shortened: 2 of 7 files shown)


marxin:
I can confirm that I successfully built around 300 Gentoo packages with the change and haven't encountered any issues related to this behaviour.

PR #972, opened 2025-06-30
Implement proper symbol lookup in a collections of versions

marxin:
Apparently, the matching logic for symbols in versions in more complex than we've got:
https://maskray.me/blog/2020-11-26-all-about-symbol-versioning#version-script


libwild/src/version_script.rs
```diff
@@ -90,63 +101,138 @@ impl<'data> MatchRules<'data> {
...
         }
     }
+}
 
-    fn matches(&self, name: &PreHashed<UnversionedSymbolName>) -> bool {
-        if self.general.matches_all || self.general.exact.contains(name) {
-            return true;
-        }
+impl<'data> VersionScript<'data> {
+    fn find_match(
```
marxin: @davidlattimore @mati865  The current data structures and the lookup in this function is pretty chatty, do you have any tips&tricks how to make it nicer? A feedback would be appreciated :)

libwild/src/version_script.rs
```diff
@@ -90,63 +101,138 @@ impl<'data> MatchRules<'data> {
...
+        &self,
+        name: &PreHashed<UnversionedSymbolName>,
+    ) -> Option<(usize, VersionRuleSection)> {
+        // Perform symbol lookup the same was as descried for the LLD linker:
+        // https://maskray.me/blog/2020-11-26-all-about-symbol-versioning#version-script
+
+        let symbol_name = LazyCell::new(|| {
```
davidlattimore: I'd perhaps create a struct that wraps the symbol name and caches the bits that need to be lazily computed. That would avoid the need for LazyCell, which has thread-safety characteristics that we don't need.

libwild/src/version_script.rs
```diff
@@ -90,63 +101,138 @@ impl<'data> MatchRules<'data> {
...
         });
-        if self
-            .general
-            .globs
-            .iter()
-            .any(|glob| glob.matches(symbol_name))
-        {
-            return true;
+
+        // 1) The first version tag with an exact pattern wins.
```
davidlattimore, line 204: I wonder if exact (non-glob) based matching should be done completely separately. Maskray mentioned in the blog post -- "Most patterns are exact so gold and ld.lld iterate patterns instead of symbols to improve performance". This makes a lot of sense and I think is probably something we may want to do at some point - not necessarily in this PR, since I imagine it'd be a bit of a change.

libwild/src/version_script.rs
```diff
@@ -90,63 +101,138 @@ impl<'data> MatchRules<'data> {
...
+        for (i, version) in self.versions.iter().enumerate() {
+            let body = &version.version_body;
+
+            if body.globals.general.exact.contains(name)
+                || demangled_name.as_deref().is_some_and(|demangled| {
```
davidlattimore: Does this mean that we demangle every symbol that doesn't match a globals.general.exact entry? That sounds pretty expensive. If the version script doesn't say `extern "C++"` or similar, do we actually need to do any demangling. If it does say `extern "C++"`, does that mean that we can just mangle the symbol names listed in that part of the version script, then match those mangled names against the mangled names in the object files?

libwild/src/version_script.rs
```diff
@@ -90,63 +101,138 @@ impl<'data> MatchRules<'data> {
...
+                    body.globals
+                        .cxx
+                        .nonstar_globs
+                        .iter()
+                        .any(|g| g.matches(demangled))
+                })
+            {
+                return Some((i, VersionRuleSection::Global));
+            }
+            if body
```
davidlattimore: From what I can see, this condition is the same as the one in the previous if-statement, except it's `body.locals` instead of `body.globals`. Perhaps this should be a method on `BasicMatchRules`? This would be helped by first creating the struct that wraps the symbol name and caches stuff, since then you can pass that to the method. The method can return whether the supplied symbol matches.

libwild/src/version_script.rs
marxin: What a great idea and it seems pretty nice right now, the only missing cached item is the `PreHashed<UnversionedSymbolName<'data>>` for the demangled name. It's something I was fighting a lot with the borrow checker.

libwild/src/version_script.rs
marxin: Yeah, we might return to it. On the other hand, if most of the symbols are exactly listed, then the hash set lookup should be pretty fast.

libwild/src/version_script.rs
marxin: Yep, reworked.

libwild/src/version_script.rs
marxin: Right now, I am more satisfied with the content of the `find_match` function!

libwild/src/version_script.rs
```diff
@@ -22,61 +22,103 @@ use winnow::error::FromExternalError;
...
-        }
-
-        if self.matches_all {
-            self.exact.clear();
-            self.globs.clear();
-            return;
+    #[inline]
+    fn matches_exact(&self, lookup: &mut SymbolLookupNameWrapper, mangled: bool) -> bool {
+        if mangled {
+            let demangled_name = lookup.get_demangled_name();
```
davidlattimore, line 90: Does it still demangle the name when there are no C++ globals/locals? Just wondering if adding something like the following would prevent that cost unless the user has actually put `extern "C++"` in their version script.

```rust
if self.exact.is_empty() {
    return false;
}
```

libwild/src/version_script.rs
marxin: > If it does say extern "C++", does that mean that we can just mangle the symbol names listed in that part of the version script.

That would be a great idea, but do have a crate that can do the mangling (opposite of the demangling)? I haven't found one :)

libwild/src/version_script.rs
marxin: Fixed.

marxin:
Thanks to the nice integration done by @lapla-cogito, I can say the PR fixes **10** linker script related tests in Mold.

lapla-cogito:
Then this would allow us to remove those tests from `skip_tests.toml`, right? (I'm still figuring out how best to maintain the list of skipped tests—I plan to gather feedback on this in Zulip later.)

mati865:
Sorry for being late to the party.
In overall, it looks quite good to me. How does it stand on performance side?

Not related to this PR, but maybe we should utf8-parse whole file at once with SIMD? For example, using https://github.com/rusticstuff/simdutf8
I imagine parsing short strings separately isn't the world's fastest thing.

davidlattimore (approved):
From a performance perspective, what's most important to me is that we only pay costs if features are used. Validating that all symbols are valid UTF-8 in a single function call could be an interesting thing to experiment, however, we'd only really want to do that for links where we need UTF-8 validity, which I think is only when we have a version script with `extern "C++"`. From some very limited performance testing that I just did, this PR fixes #978 - which makes sense, since we're now skipping demangling and UTF-8 validation unless we have match rules that need them. I'm happy for this to be merged, but will leave it to you.

That's great to have a few more tests passing and to have the builds that originally motivated this change working as well!

mati865:
There is also https://github.com/marxin/wild/blob/5d1b3303fc8500a82cea9be51cde7bb76c6116ee/libwild/src/version_script.rs#L474
So the only cases where UTF-8 is not parsed is when names are not mangled and are either an asterisk or plain text - adding globs requires parsing. Naturally, this would be a pessimisation for Rust case.

davidlattimore:
Good point, I forgot about the globbing. I'm somewhat tempted to make a create that matches glob patterns against bytes

PR #974, opened 2025-07-01
Implement more arguments for exporting symbols

### LINKED ISSUE #955: Add support for `-dynamic-list`
Noticed while building `qtbase` package on Gentoo:
https://github.com/qt/qtbase/blob/9e05e86f973fc820f8fba4ea6edbfb210eeae1b5/src/gui/CMakeLists.txt#L352

mati865:
For example used by deno:
https://github.com/denoland/deno/blob/45dfae18eec45c91146231aa05f4b249378691b5/ext/napi/generated_symbol_exports_list_linux.def

Fixes #955


libwild/src/input_data.rs
```diff
@@ -58,6 +60,11 @@ pub(crate) struct VersionScriptData<'data> {
     pub(crate) raw: &'data [u8],
 }
 
+#[derive(Clone, Copy)]
+pub(crate) struct ExportListData<'data> {
+    pub(crate) raw: &'data [u8],
+}
```
mati865: This is exactly the same as the struct above and differs only by the name.

libwild/src/input_data.rs
```diff
@@ -613,6 +626,23 @@ fn read_version_script<'data>(
...
+    let file = inputs_arena.alloc(InputFile {
+        filename: path.to_owned(),
+        original_filename: path.to_owned(),
+        kind: FileKind::Text,
+        modifiers: Default::default(),
+        data: Some(data),
+    });
+
+    Ok(ExportListData { raw: file.data() })
+}
```
mati865: Also basically the same as function above and differs only by names.

libwild/src/layout.rs
```diff
@@ -4630,6 +4638,16 @@ fn can_export_symbol(
         return false;
     }
 
+    if !export_all_dynamic
+        && !&symbol_db
+            .explicitly_explicitly_export_symbol_list_data_list
+            .contains(&UnversionedSymbolName::prehashed(
```
mati865: `prehashed` is called quite often now:
```
❯ rg "::prehashed\("
libwild/src/version_script.rs
77:                self.exact.insert(UnversionedSymbolName::prehashed(exact));
98:                .contains(&UnversionedSymbolName::prehashed(demangled_name.as_bytes()))
532:        let Some(m) = script.find_match(&UnversionedSymbolName::prehashed(name.as_bytes())) else {

libwild/src/diagnostics.rs
64:            .get_unversioned(&UnversionedSymbolName::prehashed(name.as_bytes()));

libwild/src/symbol_db.rs
1119:            let name = UnversionedSymbolName::prehashed(info.name_bytes);
1231:            let name = UnversionedSymbolName::prehashed(name_bytes);
1553:        Self::from_prehashed(symbol_id, UnversionedSymbolName::prehashed(name))
1572:            name: VersionedSymbolName::prehashed(name, version),

libwild/src/symbol.rs
86:        let name = UnversionedSymbolName::prehashed(name_info.name_bytes);
88:            PreHashedSymbolName::Versioned(VersionedSymbolName::prehashed(name, version))

libwild/src/layout.rs
3173:                .get_unversioned(&UnversionedSymbolName::prehashed(
4644:            .contains(&UnversionedSymbolName::prehashed(

libwild/src/resolution.rs
627:        .get_unversioned(&UnversionedSymbolName::prehashed(name))

libwild/src/elf_writer.rs
2807:                            .version_for_symbol(&UnversionedSymbolName::prehashed(sym_def.name))
```
but I don't know if this is avoidable.

libwild/src/export_symbol_list.rs
```diff
@@ -0,0 +1,220 @@
...
+        let mut lookup_name = SymbolLookupNameWrapper::from_name(name);
+
+        if self.general.matches_exact(&mut lookup_name, false)
+            || self.cxx.matches_exact(&mut lookup_name, true)
+        {
+            return true;
+        }
+
+        for &non_star in &[true, false] {
+            if self.general.matches_glob(&mut lookup_name, non_star, false)
```
marxin: Thinking loudly: shouldn't we make the last 2 arguments of `matches_exact`  function of an enum type?
What do you think?

libwild/src/layout.rs
marxin: Might be something that can be included into refactoring as mentioned in #994.

libwild/src/export_symbol_list.rs
mati865: Sorry, forgot to reply here.
That would probably improve the readability, but personally I don't mind either way.

mati865:
This should be probably ready for the review, but I want to test it with Qt first.

If somebody has ideas for better naming of these things, don't hesitate to leave a comment.

libwild/src/input_data.rs
davidlattimore: I assume that you've decided that it's worth having a separate type rather than just reusing the existing type?

libwild/src/export_symbol_list.rs
```diff
@@ -0,0 +1,264 @@
...
+                ));
+            };
+
+            matchers.push(matcher);
+        }
+
+        return Ok(ParsedSymbolMatcher::CxxMatchers(matchers));
+    }
+
+    // TODO: This is the only significant divergence between `parse_matcher` from this file and version_script.rs
```
davidlattimore: Is this a TODO that you intend to resolve before merging, or something for later?

libwild/src/symbol_db.rs
```diff
@@ -279,11 +282,20 @@ impl<'data> SymbolDb<'data> {
         version_script_data: Option<VersionScriptData<'data>>,
         args: &'data Args,
         linker_scripts: &[InputLinkerScript<'data>],
+        explicitly_explicitly_export_symbol_list_data_list_data: Option<ExportListData<'data>>,
```
davidlattimore: This name repeats "explicitly" and "list_data" twice. The field above also has "explicitly" twice. I haven't thought deeply about the name, so feel free to disregard, but what about just `export_list`, `explicit_export_list`, `explicit_symbol_exports` or some similar variant?

libwild/src/layout.rs
davidlattimore: While multiple calls for the same symbol could be bad for performance, I'd say what perhaps matters most here is whether we actually make that call most of the time. I'd assume that in the vast majority of linker usages, the list of explicit exports will be empty. So, if we can avoid calling `prehashed` in that case, then we're probably OK.

libwild/src/symbol_db.rs
```diff
@@ -1203,6 +1226,11 @@ impl<'data> SymbolLoader<'data> for RegularObjectSymbolLoader<'_, 'data> {
     fn compute_value_flags(&self, sym: &crate::elf::Symbol) -> ValueFlags {
         let is_undefined = sym.is_undefined(LittleEndian);
 
+        let symbol_is_exported = || {
+            let name_bytes = self.object.symbol_name(sym).expect("TODO");
```
davidlattimore: If we get an error reading the symbol name, then I think it'd be reasonable to return false. The symbol name will almost certainly get read somewhere else such that the error will be reported properly.

libwild/src/symbol_db.rs
```diff
@@ -1203,6 +1226,11 @@ impl<'data> SymbolLoader<'data> for RegularObjectSymbolLoader<'_, 'data> {
...
+            let name = UnversionedSymbolName::prehashed(name_bytes);
+            self.export_symbol_list.contains(&name)
```
davidlattimore: It'd be good to avoid reading and hashing the name if this is empty.

libwild/src/input_data.rs
mati865: I didn't though much about it. Maybe we should just call it `ScriptData` and reuse?

mati865:
Turns out Qt also uses `extern "C"` syntax, so I have added it here but I'll probably split it out to a new PR.
https://github.com/qt/qtbase/blob/3babea74b2acfae64c9db2a42f33ce2c4cfad445/src/corelib/QtCore.dynlist

libwild/src/input_data.rs
davidlattimore: Sounds reasonable to me

libwild/src/layout.rs
mati865: I was wondering if it would be possible and sensible to cache symbol_id -> symbol name lookups in additional hashmap/BTreeMap.
All the calls of `prehashed` in this PR should be avoidable unless we are given names of the symbols to export.

libwild/src/symbol_db.rs
mati865: Yeah, I had misclicked something in Helix and didn't notice the change (yay, feels like the early days in Vim). I have fixed it before seeing your comment, I think I'll go ahead with `export_list`.

libwild/src/symbol_db.rs
mati865: Yeah, I missed the condition here. Although I'll probably refactor export list to be an option.

libwild/src/layout.rs
davidlattimore: We could. I'm unsure if it would be a performance win or not though. How many times do we hash a symbol name in the common case? I'd kind hope that it would still be 1. For the extra calls that we make in less common cases, I guess we could run some benchmarks that hit those cases and see how much of a problem it is.

It may be possible to get rid of some unnecessary calls via other means. e.g. the call you're adding in `compute_value_flags` - that function is called from `load_symbols` which is the primary place where symbol name hashes are computed. So in that case, I could imagine passing a `&mut SymbolInfo` to `compute_value_flags` where `SymbolInfo` contains the `&crate::elf::Symbol` that we currently pass, but also an `Option<Prehashed<UnversionedSymbolName>>`.

libwild/src/export_symbol_list.rs
mati865: This was kind of "loud thinking" in case I or somebody who does the review gets any idea on how to improve it.

davidlattimore: Would changing the version in `version_script.rs` to accept `from_arg` be an option?

mati865: Yeah, while not useful for version scripts, it deduplicates things a lot. So, probably worth the tradeoff.

libwild/src/layout.rs
mati865: Based on `samply` profile (with 10 000 Hz) when linking Clang using main branch with a single thread, `prehashed()` is called from:
- `libwild::symbol::UnversionedSymbolName::prehashed` inside `libwild::symbol::PreHashedSymbolName::from_raw` called by `libwild::resolution::resolve_symbol` inside `libwild::resolution::resolve_symbols` - 7.9% of samples
- `libwild::symbol::UnversionedSymbolName::prehashed` called by `libwild::symbol_db::SymbolLoader::load_symbols` - 3.2% of samples
- `libwild::symbol::UnversionedSymbolName::prehashed` called by `libwild::symbol_db::SymbolLoader::load_symbols` (again, looks like a different object given to Rayon?) - 1.6% of samples
- - `libwild::symbol::UnversionedSymbolName::prehashed` called by `libwild::symbol_db::SymbolLoader::load_symbols` (again, looks like a different object given to Rayon?) - 0.8% of samples
- ... and so on with `libwild::symbol_db::SymbolLoader::load_symbols` 

So, looks like 2 calls sites. One of them is parallel already and the other one will be after https://github.com/davidlattimore/wild/pull/1003

davidlattimore: Thanks for looking at that. Those two call sites should in theory amount to calling once per symtab entry. The call in `resolve_symbol` should only be for undefined symbols (references) and the call in `load_symbols` should only be for definitions.

libwild/src/export_symbol_list.rs
marxin: Ok, let's keep it as it is for now.

libwild/src/version_script.rs
```diff
@@ -23,9 +23,9 @@ use winnow::token::take_until;
 use winnow::token::take_while;
 
 #[derive(Debug, Default)]
-struct MatchRules<'data> {
-    general: BasicMatchRules<'data>,
-    cxx: BasicMatchRules<'data>,
+pub(crate) struct MatchRules<'data> {
```
mati865, line 26: Dunno if we should somehow reorganise or perhaps extract the code that is shared between version scripts and export list.

libwild/src/version_script.rs
```diff
@@ -423,7 +432,10 @@ fn parse_version_section<'data>(input: &mut &'data BStr) -> winnow::Result<Versi
     Ok(out)
 }
 
-fn parse_matcher<'data>(input: &mut &'data BStr) -> winnow::Result<ParsedSymbolMatcher<'data>> {
+pub(crate) fn parse_matcher<'data>(
+    input: &mut &'data BStr,
+    without_colon: bool, // e.g. symbol to export passed via CLI arg
```
davidlattimore: nit: This should perhaps be called `without_semicolon`. A colon is ':', but the character you're skipping is ';'

mati865: Good catch! There is no way I'd ignore this commend.

libwild/src/version_script.rs
mati865: Since there is other work around this area that is right now blocked by this PR, let's leave this discussion for later.

mati865:
Thanks for the reviews.

Issue #966, opened 2025-06-30
Support `.tcommon` TLS section


marxin:
It's part of the Mold's test: `test/arch-x86_64-tls-large-tbss.sh`.

PR #975, opened 2025-07-01
Use segment rather than section layout for TLS addresses

davidlattimore:


davidlattimore:
This helps, but is not sufficient for #966.

PR #979, opened 2025-07-02
Merge LTO detection into section loading

davidlattimore:
Also add a test for that code.


libwild/src/resolution.rs
```diff
@@ -786,6 +786,11 @@ fn resolve_sections_for_object<'data>(
         .enumerate()
         .map(|(input_section_index, input_section)| {
             let section_name = obj.object.section_name(input_section).unwrap_or_default();
+
+            if section_name.starts_with(b".gnu.lto_.symtab") {
```
marxin: Maybe a new constant in `secnames` ?

davidlattimore: Done

PR #980, opened 2025-07-02
Allow skipping of mold tests

lapla-cogito:
Skips tests listed in `skip_tests.toml` when running `cargo test --features mold_tests`. These tests are roughly grouped for easier troubleshooting, with those whose failure causes remain unknown placed in the "misc" group.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -2,3 +2,4 @@
 flate = "flate"
 rela = "rela"
 ot = "ot"
+unkown = "unkown"
--- a/wild/tests/external_tests/mold_tests.rs
+++ b/wild/tests/external_tests/mold_tests.rs
@@ -1,13 +1,30 @@
 use crate::Architecture;
 use crate::Result;
+use crate::external_tests::should_not_ignore_tests;
 use crate::get_host_architecture;
 use crate::get_wild_test_cross;
 use rstest::rstest;
+use serde::Deserialize;
+use std::collections::HashMap;
 use std::env;
+use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
 use std::str::FromStr;
+use std::sync::OnceLock;
+
+#[derive(Deserialize)]
+struct Config {
+    skipped_groups: HashMap<String, SkippedGroup>,
+}
+
+#[derive(Deserialize)]
+struct SkippedGroup {
+    tests: Vec<String>,
+}
+
+static SKIP_TESTS_NAME: OnceLock<Option<Vec<String>>> = OnceLock::new();
 
 #[rstest]
 fn exec_mold_tests(
@@ -41,12 +58,44 @@ fn exec_mold_tests(
     Ok(())
 }
 
+fn load_skip_tests_config() -> &'static Option<Vec<String>> {
+    SKIP_TESTS_NAME.get_or_init(|| {
+        let skip_tests_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("tests")
+            .join("external_tests")
+            .join("skip_tests.toml");
+
+        fs::read_to_string(&skip_tests_path)
+            .map(|content| {
+                let config: Config =
+                    toml::from_str(&content).expect("Failed to parse skip_tests.toml");
+
+                config
+                    .skipped_groups
+                    .into_iter()
+                    .flat_map(|(_, group)| group.tests)
+                    .collect()
+            })
+            .ok()
+    })
+}
+
 fn should_skip_mold_test(path: &Path) -> bool {
     let file_name = path
         .file_name()
         .expect("Must be a valid filename")
         .to_str()
         .expect("Expected valid string name");
+
+    if should_not_ignore_tests("mold") {
+        return false;
+    }
+
+    if let Some(skip_list) = load_skip_tests_config() {
+        if skip_list.contains(&file_name.to_string()) {
+            return true;
+        }
+    }
 
     let Some(name) = file_name.strip_prefix("arch-") else {
         return false;
--- a/wild/tests/external_tests/skip_tests.toml
+++ b/wild/tests/external_tests/skip_tests.toml
@@ -1,0 +1,324 @@
+[skipped_groups.unsupported_options]
+reason = "these tests use unsupported options"
+tests = [
+  "allow-multiple-definition.sh",
+  "audit.sh",
+  "color-diagnostics.sh",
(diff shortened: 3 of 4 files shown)


.typos.toml
```diff
@@ -2,3 +2,4 @@
 flate = "flate"
 rela = "rela"
 ot = "ot"
+unkown = "unkown"
```
lapla-cogito, line 5: There seems to be a typo in the mold test name.
https://github.com/rui314/mold/blob/main/test/unkown-section-type.sh

wild/tests/external_tests/mold_tests.rs
```diff
@@ -41,13 +57,44 @@ fn exec_mold_tests(
...
+
 fn should_skip_mold_test(path: &Path) -> bool {
     let file_name = path
         .file_name()
         .expect("Must be a valid filename")
         .to_str()
         .expect("Expected valid string name");
 
+    if let Some(skip_list) = load_skip_tests_config() {
+        if skip_list.contains(&file_name.to_string()) {
```
lapla-cogito, line 95: One possible option would be to apply this skip during CI but exclude it for other scenarios (such as when developers run tests locally). This could be implemented by checking the value of `$CI`, though whether this implementation is appropriate remains debatable I think.

wild/tests/external_tests/mold_tests.rs
```diff
@@ -41,13 +57,44 @@ fn exec_mold_tests(
...
 }
 
+fn load_skip_tests_config() -> &'static Option<Vec<String>> {
+    SKIP_TESTS_NAME.get_or_init(|| {
+        let skip_tests_path = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .join("tests")
+            .join("external_tests")
+            .join("skip_tests.toml");
+
+        match fs::read_to_string(&skip_tests_path) {
```
marxin: `fs::read_to_string(&skip_tests_path).map(|..| ...)`

wild/tests/external_tests/skip_tests.toml
```diff
@@ -0,0 +1,317 @@
...
+  "no-allow-shlib-undefined2.sh",
+  "no-allow-shlib-undefined3.sh",
+  "no-allow-shlib-undefined4.sh",
+  "rpath.sh",
+  "undefined.sh",
+  "undefined2.sh",
+]
+
+[skipped_groups.gdb_index]
+reason = "GDB index support #811"
```
marxin: What about putting the tracking issue to `bug = 811` field, rather than having it as part of the comment?

wild/tests/external_tests/skip_tests.toml
```diff
@@ -0,0 +1,317 @@
+[skipped_groups.unsupported]
```
marxin: I would include `options` in the short name (`skipped_groups.unsupported_options`)?

wild/tests/external_tests/skip_tests.toml
lapla-cogito: Thanks, added `tracking_issue` field.

wild/tests/external_tests/mold_tests.rs
davidlattimore: My preference would be to skip by default, but have some mechanism to not skip. An environment variable seems like a reasonable way to control that, but probably not the `CI` variable. Something not too long and not too hard to remember would be good. e.g. `WILD_IGNORE_SKIP` or `NO_SKIP_TESTS`. But feel free to use your imagination as to what you think is good.

lapla-cogito: Thanks, added support for not skipping external test suites by listing their names separated by commas in `WILD_IGNORE_SKIP`.

marxin: ~Seems to me the environment variable does not work, I get the same test results with and w/o the option.~

Oh, can we please add a special option value, that will disable the skipping at all?

marxin:
Seems there are multiple unlisted tests:
```
failures:
    external_tests::mold_tests::exec_mold_tests::mold_test_116__UP_external_test_suites_mold_test_compress_debug_sections_zstd_
    external_tests::mold_tests::exec_mold_tests::mold_test_125__UP_external_test_suites_mold_test_crel_sh
    external_tests::mold_tests::exec_mold_tests::mold_test_433__UP_external_test_suites_mold_test_weak_export_dso2_sh
    external_tests::mold_tests::exec_mold_tests::mold_test_438__UP_external_test_suites_mold_test_weak_undef4_sh
```

marxin:
And suggestion (can be done in a separate PR) would be running all Mold's tests, and verifying the failing ones are the once listed in the TOML file? Having that, we'll catch cases where a test passes, but it's ignored.

wild/tests/external_tests/mold_tests.rs
lapla-cogito: Added support for `WILD_IGNORE_SKIP=all`.

lapla-cogito:
> Seems there are multiple unlisted tests:

This seems strange because in my local environment, all these tests pass (without having written them in `skip_tests.toml`).

edit: I also made it skip these tests.

> And suggestion (can be done in a separate PR) would be running all Mold's tests, and verifying the failing ones are the once listed in the TOML file? Having that, we'll catch cases where a test passes, but it's ignored.

This approach seems reasonable. Regardless, operationally it's easier to assume GitHub Actions results are accurate, so in a separate PR we should temporarily run `env WILD_IGNORE_SKIP=mold WILD_TEST_CROSS=aarch64,riscv64 cargo test --features mold_tests` and `WILD_TEST_CROSS=aarch64,riscv64 cargo test --features mold_tests` on each runner, then update the list based on these results.

wild/tests/external_tests/mold_tests.rs
marxin: Thank you!

marxin (approved):

marxin:
> This seems strange because in my local environment, all these tests pass (without having written them in `skip_tests.toml`).

Now it's all green for me as well. In an uncomming PR we should also handle the test skipping mechanism in the Mold's tests:
```
skip() {
  echo skipped
  trap - EXIT
  exit 0
}
```

It's something the `ctest` harness can handle and properly display:
```
The following tests did not run:
	  2 - x86_64-arch-x86_64-apx-gotpcrelx (Skipped)
	  3 - x86_64-arch-x86_64-apx-gottpoff (Skipped)
	  4 - x86_64-arch-x86_64-apx-gottpoff2 (Skipped)
```

PR #983, opened 2025-07-02
 Enable cross tests with LLD

mati865:


mati865:
That wasn't supposed to pass everywhere with `tmp: test multiple GCC versions on CI` commit...

mati865:
I don't have a clue why it works now, but it does...

However, `program_name_42___libc_integration_c__` test on my machine goes from 4.064s to 5.646s. Should I roll it back?

davidlattimore:
> I don't have a clue why it works now, but it does...
> 
> However, `program_name_42___libc_integration_c__` test on my machine goes from 4.064s to 5.646s. Should I roll it back?

I think that's fine. It's to be expected if we're running more variants of the test. I guess at some point we might need to think about some way to run the variants in parallel.

davidlattimore (approved):

PR #985, opened 2025-07-03
Fix Nix Docker

lapla-cogito:
I’ve noticed that building the Nix Docker image isn’t currently possible for several reasons.

1. `shell.nix` has been moved from its previous location in root to `nix/shell.nix`. This causes `COPY shell.nix shell.nix` to fail.
2. Even after correcting this, the current `nix/shell.nix` uses craneLib, which appears to be unsuitable for handling by nix-shell. From my understanding, craneLib was introduced primarily to reduce CI build times, so by creating another `shell.nix` in `docker/shell.nix` that doesn't use craneLib, we should be able to perform the build successfully.

### DIFF
--- a/docker/nix.Dockerfile
+++ b/docker/nix.Dockerfile
@@ -6,7 +6,7 @@
 
 FROM nixos/nix AS chef
 
-COPY shell.nix shell.nix
+COPY docker/shell.nix shell.nix
 RUN nix-shell --run "rustup toolchain install nightly"
 
 WORKDIR /wild
@@ -17,7 +17,7 @@ RUN nix-shell --run "cargo chef prepare --recipe-path recipe.json"
 
 FROM chef AS builder
 COPY --from=planner /wild/recipe.json recipe.json
-COPY shell.nix shell.nix
+COPY docker/shell.nix shell.nix
 RUN nix-shell --run "cargo chef cook --all-targets --recipe-path recipe.json"
 COPY . .
 
--- a/docker/shell.nix
+++ b/docker/shell.nix
@@ -1,0 +1,24 @@
+{
+  pkgs ? import <nixpkgs> { },
+}:
+pkgs.mkShell {
+  nativeBuildInputs = [
+    (pkgs.writeShellApplication {
+      name = "gcc";
+      text = ''${pkgs.lib.getExe pkgs.gcc} "$@" -B${pkgs.binutils-unwrapped-all-targets}/bin '';
+    })
+    (pkgs.writeShellApplication {
+      name = "g++";
+      text = ''${pkgs.lib.getExe' pkgs.gcc "g++"} "$@" -B${pkgs.binutils-unwrapped-all-targets}/bin '';
+    })
+    pkgs.binutils-unwrapped-all-targets
+    pkgs.cargo-chef
+    pkgs.llvmPackages_20.clang
+    pkgs.lld
+    pkgs.glibc.out
+    pkgs.glibc.static
+    pkgs.rustup
+  ];
+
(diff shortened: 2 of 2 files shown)


RossSmyth:
I don't really understand the point of the Nix docker container so I didn't bother fixing it. But it can just reuse the same shell.nix.

lapla-cogito:
The Dockerfiles located in the `docker` directory are mainly used for verifying compatibility across different operating systems during development.
Perhaps my understanding of Nix is limited, but while you say we can reuse `shell.nix`, when I attempt to build using `nix/shell.nix` in `nix.Dockerfile`, the following error occurs in my environment (and I believe this is because we're trying to use craneLib from nix-shell):
<details>
  <summary>click to view</summary>

```
（▰╹◡╹）❯  docker build --progress=plain -t wild-dev-nix . -f docker/nix.Dockerfile
#0 building with "default" instance using docker driver

#1 [internal] load build definition from nix.Dockerfile
#1 transferring dockerfile: 614B done
#1 DONE 0.0s

#2 [internal] load metadata for docker.io/nixos/nix:latest
#2 DONE 0.2s

#3 [internal] load .dockerignore
#3 transferring context: 75B done
#3 DONE 0.0s

#4 [chef 1/4] FROM docker.io/nixos/nix:latest@sha256:2c8d8095d21720e64d025661caca856118505d17964e39cb7b9c49faed57defd
#4 DONE 0.0s

#5 [internal] load build context
#5 transferring context: 149.12MB 0.9s done
#5 DONE 0.9s

#6 [chef 2/4] COPY nix/shell.nix shell.nix
#6 CACHED

#7 [chef 3/4] RUN nix-shell --run "rustup toolchain install nightly"
#7 0.511 error: cannot evaluate a function that has an argument without a value ('craneLib')
#7 0.511        Nix attempted to evaluate a function as a top level expression; in
#7 0.511        this case it must have its arguments supplied either by default
#7 0.511        values, or passed explicitly with '--arg' or '--argstr'. See
#7 0.511        https://nixos.org/manual/nix/stable/language/constructs.html#functions.
#7 0.511        at /shell.nix:3:3:
#7 0.511             2|   pkgs ? import <nixpkgs> { },
#7 0.511             3|   craneLib,
#7 0.511              |   ^
#7 0.511             4| }:
#7 ERROR: process "/bin/sh -c nix-shell --run \"rustup toolchain install nightly\"" did not complete successfully: exit code: 1
------
 > [chef 3/4] RUN nix-shell --run "rustup toolchain install nightly":
0.511 error: cannot evaluate a function that has an argument without a value ('craneLib')
0.511        Nix attempted to evaluate a function as a top level expression; in
0.511        this case it must have its arguments supplied either by default
0.511        values, or passed explicitly with '--arg' or '--argstr'. See
0.511        https://nixos.org/manual/nix/stable/language/constructs.html#functions.
0.511        at /shell.nix:3:3:
0.511             2|   pkgs ? import <nixpkgs> { },
0.511             3|   craneLib,
0.511              |   ^
0.511             4| }:
------
nix.Dockerfile:10
--------------------
   8 |     
   9 |     COPY nix/shell.nix shell.nix
  10 | >>> RUN nix-shell --run "rustup toolchain install nightly"
  11 |     
  12 |     WORKDIR /wild
--------------------
ERROR: failed to solve: process "/bin/sh -c nix-shell --run \"rustup toolchain install nightly\"" did not complete successfully: exit code: 1
```
</details>

P.S.: BTW, even if we could reuse `shell.nix`, as mentioned earlier its purpose differs from what we would use in CI environments, so it might be better to keep it separated in the `docker` directory IMO.

PR #989, opened 2025-07-03
Add doc on steps to run external tests

lapla-cogito:
Now would be an appropriate time to document this feature. Some of the content aligns with #987 and #988, so it's preferable for this PR to be merged after those are merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,6 +102,39 @@ file, specify its location using the `WILD_TEST_CONFIG` environment variable as 
 WILD_TEST_CONFIG=path_to_config cargo test
 ```
 
+## Running external tests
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -102,6 +102,39 @@ file, specify its location using the `WILD_TEST_CONFIG` environment variable as
...
 ```
 
+## Running external tests
+
+Wild can run some external test suites. Currently only the test suite of mold is supported.
+
+You can run the mold tests as follows:
+
+```sh
+# Clone the mold repository (only needed once)
```
mati865, line 112: Just a nitpick: it will be required each time we update mold to a newer commit.

PR #992, opened 2025-07-07
Run external tests in CI

lapla-cogito:
I’m keeping this as a draft because I need to run some experiments.

There are several considerations regarding this feature. Specifically,:
1. Should we write external tests in a separate workflow?
2. At what stage should these tests be executed?
3. For how many targets and distributions should they be run?

In any case, I'd like to conduct some experimentation before reaching my conclusion. For example, integrating this feature into the current `ci.yml` would be easy.


mati865:
Mold tests unlike Wild tests do not compare the outputs against other linkers and instead just perform assertions. So, I think there is no reason to run them across different distributions.

For now, I'd focus on getting mold's x86_64 tests to run, then either get AArch64 to run or further refine existing ones.

I don't have a strong preference regarding workflow/job setup, but having a separate workflow would be nice. It'd give us more flexibility in regard to configuring them. This not a crucial part though.

lapla-cogito:
I think the basic implementation has been completed. The CI process consists of the following five steps:
1. Execute all mold tests, including those explicitly marked to be skipped in advance
2. Analyze the test results and save both successful and failed tests
3. Load tests specified in `mold_skip_tests.toml` to be skipped
4. Detect regressions
5. Identify which tests can be removed from `mold_skip_tests.toml`

Considering that other CI processes take longer, running this workflow for each PR and on main pushes should be sufficient. ([sample run](https://github.com/davidlattimore/wild/actions/runs/16312217468/job/46070000521))

However, whether to fail the CI when a regression occurs remains an open question for consideration.

.github/scripts/analyze_test_results.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
mati865, line 1: Let's use `#!/usr/bin/env bash` like in `save-dir-prelude.sh`
```suggestion
#!/usr/bin/env bash
```

.github/scripts/analyze_test_results.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
```
mati865, line 6: These variables are unused.

.github/scripts/analyze_test_results.sh
```diff
@@ -0,0 +1,45 @@
...
+with open('${FAILED_TESTS}', 'w') as f:
+    for test in failed_tests:
+        f.write(test + '\n')
+
+with open('${PASSED_TESTS}', 'w') as f:
+    for test in passed_tests:
+        f.write(test + '\n')
+
+print(f'\\nFound {len(failed_tests)} failed tests and {len(passed_tests)} passed tests')
+"
```
mati865, line 45: Maybe this whole file should be a Python script, or a Python script called from this shell script? I think that would make it easier to work on (syntax highlighting, completions, formatting, ...).

.github/scripts/check_regressions.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
mati865, line 1:
```suggestion
#!/usr/bin/env bash
```

.github/scripts/check_regressions.sh
```diff
@@ -0,0 +1,43 @@
...
+# Find tests that are not in skip list but failed (these are regressions)
+while IFS= read -r failed_test; do
+    # Check if any skip list entry is a substring of this failed test
+    is_skipped=false
+    failed_test_name=$(echo "${failed_test}" | sed 's/.*test_suites_mold_test_//')
+    while IFS= read -r skip_test; do
+        # Replace '+', '.' and '-' with '_' in skip_test to align with rstest behavior
+        skip_test_converted="${skip_test//+/_}"
+        skip_test_converted="${skip_test_converted//./_}"
+        skip_test_converted="${skip_test_converted//-/_}"
```
mati865, line 18:
```suggestion
        skip_test_converted="${skip_test//[+.-]/_}"
```
This works for me with Bash 5.3.0 and Brush 0.2.20, is it not supported by the older versions?

.github/scripts/check_regressions.sh
```diff
@@ -0,0 +1,43 @@
...
+
```
mati865, line 19: Trailing whitespaces
```suggestion

```

.github/scripts/identify_removable_tests.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
mati865, line 1:
```suggestion
#!/usr/bin/env bash
```

.github/scripts/load_skip_list.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
mati865, line 1:
```suggestion
#!/usr/bin/env bash
```

.github/scripts/identify_removable_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+PASSED_TESTS="$(pwd)/passed_tests.txt"
+CURRENT_SKIP_LIST="$(pwd)/current_skip_list.txt"
+REMOVABLE_TESTS="$(pwd)/removable_tests.txt"
```
mati865, line 7: I'm not convinced by the naming here. This skip list works as XFAIL list - when XFAIL test passes an error is returned, skipped test should pass no matter what.

This doesn't have to be addressed at this time - we can discuss it later.

.github/scripts/identify_removable_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+
+while IFS= read -r skip_test; do
+    while IFS= read -r passed_test; do
+        if [[ "${passed_test}"* == "${skip_test}_sh" ]]; then
```
mati865, line 11: This won't work: https://www.shellcheck.net/wiki/SC2203
Globs work only at right-hand side of `==`.

.github/scripts/load_skip_list.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
```
mati865, line 6: These variables are unused.

.github/scripts/load_skip_list.sh
```diff
@@ -0,0 +1,35 @@
...
+        test_name = test
+        if test_name.endswith('.sh'):
+            test_name = test_name[:-3]
+        # Convert special characters to match rstest behavior
+        converted_name = test_name.replace('+', '_').replace('.', '_').replace('-', '_')
+        print(converted_name)
+except Exception as e:
+    print(f'Error parsing TOML: {e}', file=sys.stderr)
+    sys.exit(1)
+" > "${CURRENT_SKIP_LIST}"
```
mati865, line 33: Again, I think it'd be easier to work with as a proper Python script.

davidlattimore: I haven't looked through this change in any detail yet, but another possibility other than python would be Rust. It should be pretty quick to compile.

mati865: Cargo script is also an option but still requires nightly toolchain (or `RUST_BOOTSTRAP` hack): https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#script

marxin (commented):
To be honest, I don't really like introduction of the another test harness made of additional Python and Bash script. I would really prefer a direct implementation in Rust (as part of `mold_tests.rs`).

Generally speaking, can't we just add another `rstest` next to `exec_mold_tests` that will run if `!exec_mold_tests` and you will make it fail if the underlying Mold didn't fail. Having that, you'll just add another `cargo test --feature external_tests` and you're done. Or do I miss something?

lapla-cogito:
Got it. I'll close this PR and try a different approach in a separate PR.

PR #993, opened 2025-07-07
Add `--unresolved-symbols=` options support

### LINKED ISSUE #954: Add support for `--unresolved-symbols=ignore-in-object-files`
Used in the `pulseaudio` package:
https://github.com/pulseaudio/pulseaudio/blob/98c7c9eafb148c6e66e5fe178fc156b00f3bf51a/src/modules/echo-cancel/meson.build#L19

lapla-cogito:
This PR adds `--unresolved-symbols={report-all, ignore-in-shared-libs, ignore-all, ignore-in-object-files}` support.

Close #954.


libwild/src/args.rs
```diff
@@ -634,6 +634,12 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
             args.allow_shlib_undefined = true;
         } else if long_arg_eq("no-allow-shlib-undefined") {
             args.allow_shlib_undefined = false;
+        } else if let Some(rest) = get_option_value("unresolved-symbols") {
+            match rest.as_str() {
+                "report-all" | "ignore-in-shared-libs" => args.no_undefined = true,
+                "ignore-all" | "ignore-in-object-files" => args.no_undefined = false,
```
marxin: Seems to me `ignore-in-shared-libs` and `ignore-in-object-files` options won't precisely match what the ld.bfd manual says, right?

mati865:
We probably will want at least a minimal test for it.

libwild/src/layout.rs
```diff
@@ -3608,9 +3608,11 @@ fn is_symbol_undefined(
         return false;
     }
 
-    sym_file_id == sym_def_file_id
+    let is_undefined = sym_file_id == sym_def_file_id
         && symbol.is_undefined(LittleEndian)
-        && symbol_value_flags.is_absolute()
+        && symbol_value_flags.is_absolute();
+
+    is_undefined && args.no_undefined
```
davidlattimore: The symbol is still undefined even if `no_undefined`is false. So I feel like the change to this line is changing the meaning of this function. Perhaps the function should be renamed? `should_emit_undefined_error`, `is_undefined_error`, `is_undefined_symbol_error` or something along those lines.

libwild/src/args.rs
```diff
@@ -711,6 +745,25 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
         args.allow_copy_relocations = false;
     }
 
+    let should_set_no_undefined = match args.output_kind() {
```
davidlattimore: There are other ways this could be written that might be simpler. Up to you if you think it's easier to read.

```rust
args.no_undefined |= matches!((args.output_kind(), args.unresolved_symbols),
  (OutputKind::SharedObject, UnresolvedSymbols::IgnoreInObjectFiles) |
  (_, UnresolvedSymbols::IgnoreInSharedLibs | UnresolvedSymbols::ReportAll)
);
```

davidlattimore (commented):
I still need to do a little more review of this. I'll hopefully be able to do later today.

wild/tests/external_tests/mold_skip_tests.toml
```diff
@@ -73,8 +73,6 @@ tests = [
   "trace.sh",
   "undefined-glob-gc-sections.sh",
   "undefined-glob.sh",
-  "unresolved-symbols.sh",
-  "unresolved-symbols2.sh",
```
lapla-cogito, line 77: What I'm unclear about is the behavior of these options when the output is a shared object. For example, looking at [this failing test](https://github.com/rui314/mold/blob/8eafdfbd0ac28e7e1ce3c52f9c49f036e9608022/test/unresolved-symbols2.sh) in [this CI](https://github.com/davidlattimore/wild/actions/runs/16545833836/job/46793320324?pr=993), the test is passing `--unresolved-symbols=ignore-in-object-files` when creating the shared library. Therefore, I believe specifying this option doesn't actually have any effect, but apparently that's not the case. Am I missing something?

davidlattimore: From the GNU ld manpage, it sounds like `--unresolved-symbols=ignore-in-object-files` should ignore undefined symbols in regular object files. I'm not sure if the output kind matters. It's the kind of input file that contains the undefined symbol that matters. For the linked test, the undefined symbol is from a regular object file, `a.o`.

In your test, the undefined symbols are always from a regular object file. So, you're perhaps testing `ignore-in-object-files`, but not properly testing `ignore-in-shared-libs`

lapla-cogito: Thank you! Now I understand what I was misunderstanding.

libwild/src/args.rs
```diff
@@ -638,6 +657,30 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
...
         } else if long_arg_eq("no-allow-shlib-undefined") {
             args.allow_shlib_undefined = false;
+        } else if let Some(rest) = get_option_value("unresolved-symbols") {
+            match rest.as_str() {
+                "report-all" => {
+                    args.no_undefined = true;
+                    args.unresolved_symbols = UnresolvedSymbols::ReportAll;
+                }
+                "ignore-in-shared-libs" => {
+                    args.no_undefined = true;
```
davidlattimore: I'm a little unsure why we need this boolean in addition to `args.unresolved_symbols`. It kind of feels like it's doing the same thing, but perhaps I'm missing something.

libwild/src/args.rs
```diff
@@ -91,6 +91,8 @@ pub struct Args {
     pub(crate) got_plt_syms: bool,
     pub(crate) b_symbolic: BSymbolicKind,
     pub(crate) relax: bool,
+    pub(crate) unresolved_symbols: UnresolvedSymbols,
+    pub(crate) error_unresolved_symbols: bool,
```
davidlattimore, line 95: Do you think the code that uses this variable might be easier to read if this were an enum instead of a bool? e.g. `UnresolvedSymbolAction::Warn`, `UnresolvedSymbolAction::Error`.

libwild/src/layout.rs
```diff
@@ -3032,10 +3032,17 @@ fn process_relocation<A: Arch>(
...
-                    "Undefined symbol {symbol_name}, referenced by {}\n    {}",
-                    source_info, object.input,
-                ));
+                if args.error_unresolved_symbols {
+                    resources.report_error(error!(
+                        "Undefined symbol {symbol_name}, referenced by {}\n    {}",
+                        source_info, object.input,
+                    ));
+                } else {
+                    crate::error::report_error(&error!(
```
davidlattimore: It's a bit confusing calling `report_error` in order to issue a warning. Perhaps use `crate::error::warning(&format!(...))` instead?

libwild/src/layout.rs
```diff
@@ -3608,9 +3615,20 @@ fn is_symbol_undefined(
...
+    let is_symbol_undefined = sym_file_id == sym_def_file_id
         && symbol.is_undefined(LittleEndian)
-        && symbol_value_flags.is_absolute()
+        && symbol_value_flags.is_absolute();
+
+    match args.unresolved_symbols {
+        crate::args::UnresolvedSymbols::IgnoreAll => false,
+        crate::args::UnresolvedSymbols::ReportAll => is_symbol_undefined,
+        crate::args::UnresolvedSymbols::IgnoreInObjectFiles => {
+            is_symbol_undefined && symbol_value_flags.is_dynamic()
```
davidlattimore: I'm not sure that checking the value flags will always do what you want. It looks like the function `should_emit_undefined_error` is only called from `ObjectLayoutState`, so the reference to the symbol will always be from a regular object file.

References to symbols from shared objects are handled separately in `DynamicLayoutState::request_all_undefined_symbols`

libwild/src/args.rs
lapla-cogito: When both the `--unresolved-symbols` and `-z defs` options are specified, I think wild needs to adjust the `no_undefined` value to accommodate their compatibility. This realization came while running [unresolved-symbols2.sh](https://github.com/rui314/mold/blob/main/test/unresolved-symbols2.sh).

libwild/src/layout.rs
lapla-cogito: I didn't know things like `crate::error::warning()` existed. Thanks!

libwild/src/layout.rs
lapla-cogito: Thank you! Indeed, this seems like it could cause issues when outputting shared objects. In [this commit](https://github.com/davidlattimore/wild/pull/993/commits/66ef693578b4761b8cce871e27496b9a8e7070c7), I implemented error reporting for `DynamicLayoutState::request_all_undefined_symbols` and test cases, but when running the tests, the message format appears to be behaving as if it were being invoked with `--warn-unresolved-symbols`. This is puzzling - could the configuration syntax be incorrect?
```
（▰╹◡╹）❯  cargo test integration_test::program_name_70___unresolved_symbols_c__
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running unittests src/lib.rs (target/debug/deps/libwild-93114e9367a38fe3)

...

running 1 test
test integration_test::program_name_70___unresolved_symbols_c__ ... FAILED

failures:

---- integration_test::program_name_70___unresolved_symbols_c__ stdout ----
WARNING: wild: Undefined symbol foo, referenced by 
    /home/lapla/repos/wild/wild/tests/build/unresolved-symbols.warn-unresolved-symbols-host-3c2e1be6e14a3fee.o
Error: Failed to build program `unresolved-symbols.c` with linker `wild` config `ignore-in-object-files-in-shared-object`
  Caused by:
    Linker returned exit status of 0, when an error was expected. Command:
        cargo run --bin wild -- -static --gc-sections --unresolved-symbols=ignore-in-object-files -shared -z now -z now -shared -o /home/lapla/repos/wild/wild/tests/build/unresolved-symbols1.ignore-in-object-files-in-shared-object-host-18efd7613709aedf.wild.so /home/lapla/repos/wild/wild/tests/build/unresolved-symbols1.ignore-in-object-files-in-shared-object-host-18efd7613709aedf.o --validate-output
        cargo run --bin wild -- -static --gc-sections --unresolved-symbols=ignore-in-object-files -shared -z now -o /home/lapla/repos/wild/wild/tests/build/unresolved-symbols.c-ignore-in-object-files-in-shared-object-host.wild /home/lapla/repos/wild/wild/tests/build/unresolved-symbols.ignore-in-object-files-in-shared-object-host-3c2e1be6e14a3fee.o /home/lapla/repos/wild/wild/tests/build/unresolved-symbols1.ignore-in-object-files-in-shared-object-host-18efd7613709aedf.wild.so --validate-output
```

libwild/src/args.rs
lapla-cogito: If there are other possible actions when unresolved symbols exist, an enum might be preferable, but otherwise, a bool works fine IMO.

wild/tests/sources/unresolved-symbols.c
```diff
@@ -0,0 +1,74 @@
+//#AbstractConfig:default
+
+//#Config:report-all:default
+//#Object:unresolved-symbols1.c
+//#SkipLinker:ld
```
davidlattimore: Why skip GNU ld? It's good if we can verify that our tests pass with GNU ld too, that way we have more confidence that our tests are correct. If you do enable GNU ld, then it should be possible to enable diffing too.

wild/tests/sources/unresolved-symbols.c
```diff
@@ -0,0 +1,74 @@
...
+//#DiffEnabled:false
+//#LinkArgs:--unresolved-symbols=report-all
+//#ExpectError:Undefined symbol foo
+
+//#Config:ignore-all:default
+//#Object:unresolved-symbols1.c
+//#SkipLinker:ld
+//#DiffEnabled:false
+//#RunEnabled:false
```
davidlattimore: Consider moving anything common into the abstract config.

libwild/src/layout.rs
```diff
@@ -5291,10 +5310,28 @@ impl<'data> DynamicLayoutState<'data> {
...
                         .symbol(self.symbol_id_range.id_to_input(symbol_id))?;
                     if !symbol.is_weak() {
-                        bail!(
-                            "undefined reference to `{}` from {self}",
-                            resources.symbol_db.symbol_name_for_display(symbol_id)
-                        );
+                        let should_report = match args.unresolved_symbols {
+                            crate::args::UnresolvedSymbols::ReportAll => true,
+                            crate::args::UnresolvedSymbols::IgnoreInObjectFiles => {
+                                value_flags.is_dynamic()
```
davidlattimore: I'm not sure that checking `value_flags.is_dynamic()` is needed. The fact that we're running this bit of code means that the reference is from a shared object, so this can probably just be `true`. Similarly, the one below can just be `false`.

libwild/src/layout.rs
davidlattimore: If you want to test different flags when outputting a shared object, I'd suggest having a separate test file, since you only care about the shared object produced and whether linking it fails. You don't actually need to link an executable against that shared object.

libwild/src/layout.rs
```diff
@@ -3608,9 +3616,20 @@ fn is_symbol_undefined(
...
+        && symbol_value_flags.is_absolute();
+
+    match args.unresolved_symbols {
+        crate::args::UnresolvedSymbols::IgnoreAll => false,
+        crate::args::UnresolvedSymbols::ReportAll => is_symbol_undefined,
+        crate::args::UnresolvedSymbols::IgnoreInObjectFiles => {
+            is_symbol_undefined && symbol_value_flags.is_dynamic()
+        }
+        crate::args::UnresolvedSymbols::IgnoreInSharedLibs => {
+            is_symbol_undefined && !symbol_value_flags.is_dynamic()
```
davidlattimore: It shouldn't be necessary to check `symbol_value_flags.is_dynamic()`. The fact that we're here means that the reference is from a regular object. The `DYNAMIC` flag will often be set on undefined symbols, since if the symbol remains undefined, we can in some circumstances resolve it dynamically. See `compute_value_flags`.

wild/tests/sources/unresolved-symbols.c
lapla-cogito: When specifying `--unresolved-symbols`, it appears that GNU ld and wild handle AArch64 relocation differently (see [this CI](https://github.com/davidlattimore/wild/actions/runs/16586858564/job/46913799878)).
So I think setting `SkipLinker:ld` globally in `unresolved-symbols-object.c` makes sense.

wild/tests/sources/unresolved-symbols-shared.c
```diff
@@ -0,0 +1,21 @@
+//#AbstractConfig:default
+//#SkipLinker:ld
+//#DiffEnabled:false
+//#RunEnabled:false
+//#Shared:unresolved-symbols-shared1.c
```
davidlattimore: This will still produce an executable - at least if linking the shared object succeeds. What I was thinking was to not produce an executable at all. i.e. pass `-shared` as a link argument. For example, see `shared.c`.

wild/tests/sources/unresolved-symbols.c
davidlattimore: Disabling both diffing and running with GNU ld is a pretty heavy solution. It'd be good to verify that we're reporting errors and succeeding the link in the scenarios as GNU ld. Perhaps just `DiffIgnore:rel.extra-opt.R_AARCH64_CALL26.ReplaceWithNop.static-non-pie`?

wild/tests/sources/unresolved-symbols-shared.c
lapla-cogito: Thanks, updated config of this file to create shared objects in each steps.

davidlattimore: I think it should be OK to delete this line and `unresolved-symbols-shared1.c`. This file (`unresolved-symbols-shared.c`) is being built into a shared object already.

davidlattimore (approved):

Issue #968, opened 2025-06-30
Undefined symbol provided as a versioned name only


marxin:
Isolated from the `alsa-lib` Gentoo package:

```
$ cat script.txt 
VERS_2.0 {
};
$ cat symver.c 
void implementation(void) {
    __builtin_printf("new_foo called\n");
}

__asm__(".symver implementation,foo@@VERS_2.0");
$ gcc symverc -shared -o libx.so -fPIC -B ~/Programming/wild
$ cat main.c
void foo();
int main()
{
  foo();
  return 0;
}
$ gcc main.c -lx -L. -fuse-ld=bfd
/usr/bin/ld.bfd: a.out: no symbol version section for versioned symbol `foo@@VERS_2.0'
/usr/bin/ld.bfd: final link failed
collect2: error: ld returned 1 exit status
```

Wild emits:
```
$ readelf -a libx.so
...
Symbol table '.dynsym' contains 8 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0:      6: 0000000000001ed0    22 FUNC    GLOBAL DEFAULT   13 foo@@VERS_2.0
...
Symbol table '.symtab' contains 28 entries:
    26: 0000000000001ed0    22 FUNC    GLOBAL DEFAULT   13 foo@@VERS_2.0
```

while other linkers emit the following entry into '.symtab':
```
    25: 00000000000016b9    22 FUNC    GLOBAL DEFAULT   14 foo
```

marxin:
CC: @mati865

mati865:
You've got a typo at `$ gcc symverc`, it should be `symver.c` and I think you are missing `-Wl,--version-script=script.txt`.

ld.bfd 2.44.0:
```
# Without version script
❯ gcc symver.c -shared -o libx.so -fPIC
/usr/bin/ld: libx.so: version node not found for symbol foo@@VERS_2.0
/usr/bin/ld: failed to set dynamic section sizes: bad value
collect2: error: ld returned 1 exit status

# With version script
❯ gcc symver.c -shared -o libx.so -fPIC -Wl,--version-script=script.txt
❯ readelf -Ws libx.so | rg 'Symbol table|foo'
Symbol table '.dynsym' contains 9 entries:
     6: 0000000000001109    22 FUNC    GLOBAL DEFAULT   13 foo@@VERS_2.0
Symbol table '.symtab' contains 18 entries:
[...]

PR #994, opened 2025-07-07
Add handling of the .symver asm directives

marxin:
The following WIP patch "fixes" #968, however, it's more about papering over the various limitation of the symbol name handling. To be honest, a proper fix would require a bigger refactoring where something like `VersionedSymbolName` will become a first class citizen when it comes to handling of the names.

@davidlattimore Can you please help with that? Feel free to close this PR at any time ;)

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -74,6 +74,7 @@ use crate::slice::slice_take_prefix_mut;
 use crate::slice::take_first_mut;
 use crate::string_merging::get_merged_string_output_address;
 use crate::symbol::UnversionedSymbolName;
+use crate::symbol_db::RawSymbolName;
 use crate::symbol_db::SymbolDb;
 use crate::symbol_db::SymbolId;
 use foldhash::HashMap as FoldHashMap;
@@ -1043,7 +1044,10 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
             })?
         };
         let e = LittleEndian;
-        let string_offset = self.strtab_writer.write_str(name);
+
+        // Always save the name without the symbol version (e.g. foo@@VER_1).
+        let RawSymbolName { name_bytes, .. } = RawSymbolName::parse(name);
+        let string_offset = self.strtab_writer.write_str(name_bytes);
         entry.st_name.set(e, string_offset);
         entry.st_other = 0;
         entry.st_shndx.set(e, shndx);
@@ -1055,6 +1059,8 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
     /// Verifies that we've used up all the space allocated to this writer. i.e. checks that we
     /// didn't allocate too much or missed writing something that we were supposed to write.
     fn check_exhausted(&self) -> Result {
+        // TODO
+        return Ok(());
         if !self.local_entries.is_empty()
             || !self.global_entries.is_empty()
             || !self.strtab_writer.out.is_empty()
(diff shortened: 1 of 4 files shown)


davidlattimore (approved):
Awesome! I had assumed that this might have a slight performance impact, but it seems fine.

mati865:
I'm back from hiking and upon @marxin request I've run his Gentoo test build.

As the first step I have removed `RUN emerge alsa-lib elfutils` from https://github.com/marxin/gentoo-docker-testing/blob/main/Dockerfile and it works much better than before but it's still incomplete. Great work though!
Lvm2 build failed with:
```
make[1]: Entering directory '/var/tmp/portage/sys-fs/lvm2-2.03.22-r7/work/LVM2.2.03.22/libdm'
if x86_64-pc-linux-gnu-readelf -Ws ioctl/libdevmapper.so | grep -q dm_stats_create_region@@; then \
SYM=$(x86_64-pc-linux-gnu-readelf -Ws ioctl/libdevmapper.so | grep dm_stats_create_region@DM_1_02_106); \
if test -n "$SYM"; then touch .symver_check; \
else echo >&2 "Missing versioned symbols in ioctl/libdevmapper.so"; false; fi; \
else touch .symver_check; fi
Missing versioned symbols in ioctl/libdevmapper.so
```
The symbols inside the shared object:
```
cd /var/tmp/portage/sys-fs/lvm2-2.03.22-r7/work/LVM2.2.03.22/libdm
readelf -Ws ioctl/libdevmapper.so | grep dm_stats_create_region@
   359: 0000000000027330   322 FUNC    GLOBAL DEFAULT   13 dm_stats_create_region@@DM_1_02_107
   360: 000000000002bac0    28 FUNC    GLOBAL DEFAULT   13 dm_stats_create_region@@DM_1_02_106
   361: 000000000002bae0    29 FUNC    GLOBAL DEFAULT   13 dm_stats_create_region@@DM_1_02_104
```
Notice `@` vs `@@` difference.

According to https://maskray.me/blog/2020-11-26-all-about-symbol-versioning we should create symbol with only one `@` if the symbol is defined.

marxin:
Thanks @mati865 for the testing effort and I think I have a fix: #1129. Feel free to continue with the gentoo-docker if you have spare cycles.

mati865:
@marxin  it doesn't fix lvm2 build.
The part that I was referring to, is "If foo is defined, produce foo and foo@v1 with the same binding and st_other value.".

marxin:
@mati865 I've just re-run locally the build of `lvm2` and the build process finishes correctly (with #1129) with the following symbol listing:
```
readelf -Ws libdm/ioctl/libdevmapper.so | grep dm_stats_create_region@
   354: 0000000000049400    28 FUNC    GLOBAL DEFAULT   14 dm_stats_create_region@DM_1_02_106
   355: 0000000000049420    29 FUNC    GLOBAL DEFAULT   14 dm_stats_create_region@DM_1_02_104
   356: 0000000000044d30   317 FUNC    GLOBAL DEFAULT   14 dm_stats_create_region@@DM_1_02_107
```

While running on the `main` branch, I hit the aforementioned issue:
```
    [CHECK] ioctl/libdevmapper.so
Missing versioned symbols in ioctl/libdevmapper.so
```

mati865:
Hmm, I'm sure I've copied the binary after building it. Perhaps I've built the debug one by mistake, I'll try later.

mati865:
🤦🏻‍♂️ I had run `cp target/release/wild /usr/bin` in Gentoo container ...

Sorry for the confusion.

marxin:
No worries, that happens. Anyways, thanks for the testing effort.

PR #1005, opened 2025-07-24
Run external tests in CI

lapla-cogito:
This aims to detect regressions by running external tests within the CI pipeline itself.
`check_mold_tests_regression` detects regressions, while `verify_skipped_mold_tests_still_fail` checks whether any tests marked for skipping in `mold_skip_tests.toml` have stopped failing.


.github/workflows/external_tests.yml
```diff
@@ -0,0 +1,44 @@
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-externaltests-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
+      - name: Run external tests
+        continue-on-error: true
+        run: WILD_TEST_CROSS=aarch64,riscv64 cargo test --features mold_tests
```
davidlattimore: If running the mold tests doesn't slow down our CI substantially, then I'd be OK with just enabling this feature in the main `ci.yml` for one of the configurations.

If we decide to run them as a separate job or even a separate workflow like you've got here, then do you think we should limit testing to just the mold tests?

lapla-cogito: Implementing this directly in `ci.yml` seems like one viable option. However, setup procedures differ slightly between running external tests versus executing integration tests that are already present in wild. For instance, during implementing this feature, I encountered a case where some mold tests require some dependencies such as `gdb` to be installed. Therefore, unless there's a compelling reason to combine them, running them as separate workflows seems appropriate.
However, since running only external tests in such cases seemed reasonable, I made the additional modification to implement this behavior, thanks!

.github/workflows/external_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-externaltests-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
+      - name: Check regressions
+        run: WILD_TEST_CROSS=aarch64,riscv64 cargo test --features mold_tests external_tests::mold_tests::check_mold_tests_regression
+      - name: Check tests that should fail still fail
+        continue-on-error: true
+        run: WILD_TEST_CROSS=aarch64,riscv64 cargo test --features mold_tests external_tests::mold_tests::verify_skipped_mold_tests_still_fail
```
lapla-cogito, line 43: If a test that was previously skipped according to mold_skip_tests.toml now passes—meaning its name can be removed from the list—the CI will simply report this without causing the entire process to fail.

.github/workflows/external_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: Run mold tests
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          submodules: recursive
+      - run: sudo apt-get update && sudo apt-get -y install gcc g++ clang-18 lld curl bubblewrap binutils-aarch64-linux-gnu binutils-riscv64-linux-gnu qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-riscv64-linux-gnu g++-riscv64-linux-gnu gdb
+      - uses: dtolnay/rust-toolchain@nightly
```
davidlattimore: If we're not running Wild's tests, then we can probably run on stable instead of nightly. That will give better cache hits, since the rust toolchain won't be changing every day.

marxin (approved):

davidlattimore (approved):

PR #1006, opened 2025-07-24
Format C/C++ code in the tests

mati865:
Different commits show some of different styles (there are more styles), what do you think about this idea?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
           # Ubuntu 22.04 contains an old lld linker which cannot do a relaxation on AArch64 (required by linker-diff)
           - runs-on: ubuntu-24.04
             container: 'ubuntu:22.04'
+            # clang-format in Ubuntu 22.04 is too old and tries to format differently.
+            ignore-format: true
         exclude: # TODO: https://github.com/actions/runner/issues/1637
           - runs-on: ubuntu-24.04-arm
             container: 'rust:1.88-alpine'
@@ -59,14 +61,16 @@ jobs:
         if: ${{ env.CI == 'true' }}
       - run: echo "WILD_TEST_CROSS=aarch64,riscv64" >> $GITHUB_ENV
         if: ${{ matrix.test-qemu }}
-      - run: apt-get update && apt-get -y install gcc g++ clang lld curl bubblewrap binutils-aarch64-linux-gnu binutils-riscv64-linux-gnu
+      - run: apt-get update && apt-get -y install gcc g++ clang clang-format lld curl bubblewrap binutils-aarch64-linux-gnu binutils-riscv64-linux-gnu
         if: ${{ contains(matrix.container, 'ubuntu') }}
       - run: apt-get update && apt-get -y install qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-riscv64-linux-gnu g++-riscv64-linux-gnu
         if: ${{ matrix.test-qemu }}
       - run: zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap
         if: ${{ contains(matrix.container, 'opensuse') }}
-      - run: apk add build-base lld clang bash curl
+      - run: apk add build-base lld clang clang-extra-tools bash curl
         if: ${{ contains(matrix.container, 'alpine') }}
+      - run: echo "WILD_TEST_IGNORE_FORMAT=1" >> $GITHUB_ENV
+        if: ${{ matrix.ignore-format }}
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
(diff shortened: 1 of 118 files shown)


davidlattimore:
Autoformatting seems like a good idea to me. I don't have any particular preference between the LLVM style and the Google style. The GNU style looks pretty weird to me and is very different to the way our Rust code is styled, so I'd prefer not that.

We should probably enforce the style, either via a CI check like we currently have for our Rust code, or plausibly via a test, which has the advantage that it's harder for a new contributor to miss.

mati865:
> I don't have any particular preference between the LLVM style and the Google style. The GNU style looks pretty weird to me and is very different to the way our Rust code is styled, so I'd prefer not that.

There are things that I prefer in Google style over LLVM and vice versa, so I have no strong opinion on these two. I don't like GNU style, but it's a huge OSS community, therefore I included it.

> or plausibly via a test

Which we probably should call `tidy`, like in Rust's repo.

marxin:
I welcome the suggested change and I don't have any style preference ;)

mati865:
Example from Tumbleweed job:
obraz

Looking deeper into it, I slightly prefer Google version, so went for it.
The way it works currently, it creates a new test for every file that we want to format with `clang-format` which obviously has some overhead (especially for nextest with spawns each test in a separate process):
```
# with hot cache
❯ time cargo t clang_format_check
...
________________________________________________________
Executed in  202.20 millis    fish           external
   usr time    1.15 secs      0.00 micros    1.15 secs
   sys time    1.66 secs    714.00 micros    1.66 secs


❯ time cargo nextest run clang_format_check
...
________________________________________________________
Executed in  345.40 millis    fish           external
   usr time    1.33 secs      0.00 micros    1.33 secs
   sys time    2.25 secs    836.00 micros    2.24 secs
```

Open questions:
- should we do it as a single test? With hot cache, this brings the time down to 100 milliseconds and 200 milliseconds respectively.
- do we want to run it by default or make it configurable
	- if so, which agents should run it?
- do we want some kind of env variable or something to automatically bless (format) these files?
	- alternatively, we could document how to configure popular editors

davidlattimore:
> * should we do it as a single test? With hot cache, this brings the time down to 100 milliseconds and 200 milliseconds respectively.

Given the time difference, I'd be in favour of running it all in a single process - not just the rust test, but also a single invocation of `clang-format`. At least on my system, that makes quite a difference.

Running a separate clang-format for each file takes 0.94 seconds on my laptop.

```bash
time bash -c 'for i in *.c *.h *.cc; do clang-format --dry-run $i; done'
```

Running all in one process takes only 0.11 seconds.

```bash
time bash -c 'clang-format --dry-run *.c *.h *.cc'
```

> * do we want to run it by default or make it configurable

I think just running it by default makes sense. If we didn't run it by default, but then did run it in CI, then sometimes people would send a PR and have it fail CI. Better if they know before they send their PR that there's stuff that needs fixing. Having a way to turn it off seems reasonable though. e.g. setting `WILD_SKIP_FORMAT_TESTS=1` or something like that.

> * do we want some kind of env variable or something to automatically bless (format) these files?

I'd be fine with that. The other option would be to have the test print out a command that you can run that would fix the formatting. e.g. something like `clang-format wild/tests/*.{c,h,cc}`

If you wanted to submit the formatting changes of the existing tests in this PR and do the formatting test in a separate PR, that would be fine by me.

mati865:
> Running a separate clang-format for each file takes 0.94 seconds on my laptop.
> 
> ```shell
> time bash -c 'for i in *.c *.h *.cc; do clang-format --dry-run $i; done'
> ```
> 
> Running all in one process takes only 0.11 seconds.
> 
> ```shell
> time bash -c 'clang-format --dry-run *.c *.h *.cc'
> ```

This is not fully representative of how this test works because you run all files serially. The test runs them with all CPU threads, so it's way faster, but obviously it will always be slower than a single invocation.

> If you wanted to submit the formatting changes of the existing tests in this PR and do the formatting test in a separate PR, that would be fine by me.

That was my initial idea but went ahead and repurposed the PR, unless you think it'd be beneficial for the review to split it up.

davidlattimore:
> This is not fully representative of how this test works because you run all files serially. The test runs them with all CPU threads, so it's way faster, but obviously it will always be slower than a single invocation.

That's true, however generally you'd be running the entire test suite, so even if we only use a single thread for checking formatting, it'll run in parallel with other tests. In any case, the speed of either approach probably uses sufficiently little CPU / time that it doesn't matter, so I'd say go with whatever approach is easiest.

> That was my initial idea but went ahead and repurposed the PR, unless you think it'd be beneficial for the review to split it up.

That's fine. Whatever is easiest.

mati865:
obraz

This is how it looks right now, obviously the second time tests were run they passed.

wild/tests/tidy.rs
davidlattimore, line 1: I just noticed that the test is in the top level of the `tests` directory. That means that it'll get compiled and linked separately, which does incur some extra cost. It would work as a unit test. The tricky part would be figuring out where to put it, since it doesn't belong in any existing module. One option would be to create a `tests` module. i.e. `wild/src/tests` as opposed to `wild/tests` where it is now. It'd also be fine to make it a submodule of `integration_tests`.

mati865: I thought that was the expected solution.

> One option would be to create a `tests` module. i.e. `wild/src/tests` as opposed to `wild/tests` where it is now. It'd also be fine to make it a submodule of `integration_tests`.

`wild/src/tests` seems a bit out of place considering this formats only the files for integration tests. Went for a submodule of integration tests, but let me know (or feel free to do it yourself) if that doesn't suit you.

PR #1015, opened 2025-07-30
Implement `--allow-multiple-definition` and `-z muldefs`

lapla-cogito:
e.g., used by relibc in Redox OS.
https://github.com/redox-os/relibc/blob/4f9267be5941a008b3cf5e0c1228c3df8bb72b5a/Makefile#L184

### DIFF
--- a/libwild/src/symbol_db.rs
+++ b/libwild/src/symbol_db.rs
@@ -901,8 +901,9 @@ fn select_symbol(
                     // We don't implement full COMDAT logic, however if we encounter duplicate
                     // strong definitions, then we don't emit errors if all the strong definitions
                     // are defined in COMDAT group sections.
-                    if !symbol_db.is_in_comdat_group(existing, resolved)
-                        || !symbol_db.is_in_comdat_group(id, resolved)
+                    if (!symbol_db.is_in_comdat_group(existing, resolved)
+                        || !symbol_db.is_in_comdat_group(id, resolved))
+                        && !symbol_db.db.args.allow_multiple_definitions
                     {
                         bail!(
                             "{}, defined in {} and {}",
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -93,6 +93,7 @@ pub struct Args {
     pub(crate) relax: bool,
     pub(crate) unresolved_symbols: UnresolvedSymbols,
     pub(crate) error_unresolved_symbols: bool,
+    pub(crate) allow_multiple_definitions: bool,
 
     output_kind: Option<OutputKind>,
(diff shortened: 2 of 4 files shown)


lapla-cogito:
One point of concern is the handling of duplicate symbols. For example, when linking [allow-multiple-definition.sh](https://github.com/rui314/mold/blob/main/test/allow-multiple-definition.sh) using mold or GNU ld, the `main` function appears as follows:
```
0000000000001679 <main>:
    1679:       f3 0f 1e fa             endbr64
    167d:       55                      push   %rbp
    167e:       48 89 e5                mov    %rsp,%rbp
    1681:       b8 00 00 00 00          mov    $0x0,%eax
    1686:       5d                      pop    %rbp
    1687:       c3                      ret
    1688:       f3 0f 1e fa             endbr64
    168c:       55                      push   %rbp
    168d:       48 89 e5                mov    %rsp,%rbp
    1690:       b8 01 00 00 00          mov    $0x1,%eax
    1695:       5d                      pop    %rbp
    1696:       c3                      ret
```
In contrast, the current implementation in this PR is as follows:
```
0000000000001df0 <main>:
    1df0:       f3 0f 1e fa             endbr64
    1df4:       55                      push   %rbp
    1df5:       48 89 e5                mov    %rsp,%rbp
    1df8:       b8 00 00 00 00          mov    $0x0,%eax
    1dfd:       5d                      pop    %rbp
    1dfe:       c3                      ret
```
Both implementations use the first defined symbol, and since this aligns with the description in the GNU ld manual page, I believe this is acceptable.

> Normally, when the linker encounters multiple definitions of the same symbol, it will report a fatal error. These options allow multiple definitions and use the first one encountered.

libwild/src/symbol_db.rs
```diff
@@ -806,17 +806,19 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data>(
...
-
-    if !duplicate_errors.is_empty() {
-        let error_details = duplicate_errors
-            .iter()
-            .map(|e| e.to_string())
-            .collect::<Vec<_>>()
-            .join("\n");
-
-        bail!("Duplicate symbols detected: {error_details}");
+    if !symbol_db.args.allow_multiple_definitions {
```
davidlattimore: Rather than doing the check here, I'd suggest checking in `select_symbol` at the point where we bail with the error. Currently, the only error emitted by `select_symbol` is if it encounters a duplicate symbol, however it'd be easy to change that an not realise that setting `--allow-multiple-definitions` would suppress the newly added error. Although, if a new error is added, then the message below would also be wrong, so maybe doing the check here is OK for now, but I think some changes might be needed here at some point.

PR #1019, opened 2025-08-03
nix: fix invalid path specification

lapla-cogito:
The changes to `lib.makeBinPath` introduced in #1006 appear to be incorrect and are causing the CI to fail.
e.g. https://github.com/davidlattimore/wild/actions/runs/16699919696/job/47269321983 (At first glance, the CI appears to be successful, but an error occurs in the "Calculate check" step, causing the subsequent checks to not be performed as intended.)

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -118,7 +118,7 @@ let
         lib.makeBinPath [
           binutils-unwrapped-all-targets
           clang
-          clang-format
+          clang-tools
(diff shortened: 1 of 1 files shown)


mati865:
Oh, that's really unfortunate. Opened https://github.com/davidlattimore/wild/pull/1020 to detect this problem.

mati865 (approved):

PR #1022, opened 2025-08-04
version scripts: Allow parsing of `extern` blocks without a trailing ";" after the last symbol

lapla-cogito:
In version scripts, within the `extern` block, the last symbol doesn't necessarily need to be followed by a semicolon (see [its definition in Yacc](https://gnu.googlesource.com/binutils-gdb/+/refs/tags/binutils-2_45/ld/ldgram.y#1471) and [version-script5.sh](https://github.com/rui314/mold/blob/8eafdfbd0ac28e7e1ce3c52f9c49f036e9608022/test/version-script5.sh)).

In fact, we can check whether such a script is valid by following:
```
（▰╹◡╹）❯  cat <<EOF > a.scr
foo
EOF

（▰╹◡╹）❯  ld --version-script a.scr
ld:a.scr:0: syntax error in VERSION script

（▰╹^╹）❯  cat <<EOF > b.scr
{extern "C" {foo};};
EOF

（▰╹◡╹）❯  ld --version-script b.scr
ld: no input files
```

### DIFF
--- a/libwild/src/version_script.rs
+++ b/libwild/src/version_script.rs
@@ -464,7 +464,17 @@ pub(crate) fn parse_matcher<'data>(
                 break;
             }
 
-            let matcher = parse_matcher(input, without_semicolon)?;
+            // Symbols at the end of `extern` blocks may omit semicolons
+            let expect_semicolon = {
+                let remaining = &**input;
+                if let Some(close_pos) = remaining.windows(2).position(|w| w == b"};") {
+                    remaining[..close_pos].contains(&b';')
+                } else {
+                    without_semicolon
+                }
+            };
+
+            let matcher = parse_matcher(input, !expect_semicolon)?;
             let ParsedSymbolMatcher::Single(matcher) = matcher else {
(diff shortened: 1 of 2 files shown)


libwild/src/version_script.rs
```diff
@@ -487,9 +497,13 @@ pub(crate) fn parse_matcher<'data>(
...
-        // TODO: Clippy bug
-        #[allow(clippy::needless_borrow)]
-        &input
+        if input.contains(&b'}') {
+            take_until(1.., b'}').parse_next(input)?
+        } else {
+            // TODO: Clippy bug
+            #[allow(clippy::needless_borrow)]
+            &input
+        }
```
lapla-cogito, line 506: `parse_matcher` is also used from [`export_list.rs`](https://github.com/davidlattimore/wild/blob/2cb866ac17616b215c28f3df134be8c771ab3920/libwild/src/export_list.rs#L52), which requires this conditional branching.

PR #1029, opened 2025-08-09
Implement new declarative argument parser

lapla-cogito:
This represents a new proposal for the current argument parser of wild. While the current naive `if-else`-based argument handling is simpler to implement, it presents several issues—for example, making available options difficult to inspect. This has consequently created barriers to implementing options like `--help`. After discussing with David, I decided to create a new declarative parser. 

When defining options, you can specify long option names, short option names, help text, and the actual processing logic as follows:

```rust
parser
    .declare()
    .long("static")
    .long("Bstatic")
    .help("Disallow linking of shared libraries")
    .execute(|_args, modifier_stack| {
        modifier_stack.last_mut().unwrap().allow_shared = false;
        Ok(())
    });
```

In addition to implementing this parser framework, I've also implemented `--help`. The key approach here is to only perform minimal `--help` processing during each option definitions themselves, instead gathering all necessary information only when `--help` is specified to construct the help message. This design accounts for potential overhead in typical usage scenarios.

When running `--help`, the resulting output looks like [this](https://gist.github.com/lapla-cogito/e0bd23787ecd17fb42038cc316a26e0f).

The issues I’m currently aware of are as follows. I’ll address these in this PR before marking it “Ready for review,” but I’m opening it now to check if CI tests pass across various environments.
- ~`@file` performs ad-hoc processing (not defined declaratively) and thus doesn't appear in the `--help` message.~ I don't think this is an issue that absolutely requires urgent resolution. After all, no other option requires hyphens like `@file` does (except for specifying input files). Therefore, I'd like to temporarily disregard this and proceed with accepting the review.
- [x] Option descriptions are insufficiently detailed.

Additionally, while the following point may not be a major concern, I'm including it for reference:
- Short options are specified using char. While this works fine in most cases, GNU ld appears to have an option `-Ur`. However, our current wild argument parser would interpret `--Ur` as `--Ur` even if `-Ur` were implemented, so compatibility with GNU ld would be maintained.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -34,6 +34,8 @@ use std::sync::atomic::AtomicI64;
 use std::sync::atomic::Ordering;
 
 pub struct Args {
+    pub(crate) unrecognized_options: Vec<String>,
+
     pub(crate) arch: Architecture,
     pub(crate) lib_search_path: Vec<Box<Path>>,
     pub(crate) inputs: Vec<Input>,
@@ -164,7 +166,7 @@ pub(crate) enum FileWriteMode {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
-pub(crate) struct Modifiers {
+pub struct Modifiers {
     /// Whether shared objects should only be linked if they're referenced.
     pub(crate) as_needed: bool,
 
@@ -240,8 +242,6 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     // Just like other modern linkers, we don't need groups in order to resolve cycles.
     "start-group",
     "end-group",
-    "(",
-    ")",
     // TODO: This is supposed to suppress built-in search paths, but I don't think we have any
     // built-in search paths. Perhaps we should?
     "nostdlib",
@@ -253,6 +253,7 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "sort-common",
     "stats",
 ];
+const SILENTLY_IGNORED_SHORT_FLAGS: &[&str] = &["(", ")"];
 
 const IGNORED_FLAGS: &[&str] = &[
     "gdb-index",
@@ -267,15 +268,18 @@ const DEFAULT_FLAGS: &[&str] = &[
     "no-copy-dt-needed-entries",
     "no-add-needed",
     "discard-locals",
+    "enable-new-dtags",
+];
+const DEFAULT_SHORT_FLAGS: &[&str] = &[
     "X",  // alias for --discard-locals
     "EL", // little endian
-    "enable-new-dtags",
 ];
 
 impl Default for Args {
     fn default() -> Self {
         Args {
             arch: default_target_arch(),
+            unrecognized_options: Vec::new(),
 
             lib_search_path: Vec::new(),
             inputs: Vec::new(),
@@ -375,8 +379,6 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
         ..Default::default()
     };
 
-    let mut unrecognised = Vec::new();
-
     args.save_dir = SaveDir::new(&input)?;
 
     let mut input = input();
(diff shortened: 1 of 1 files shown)


libwild/src/args.rs
```diff
@@ -387,403 +388,26 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
...
-            args.sysroot = Some(Box::from(sysroot.as_path()));
-            for path in &mut args.lib_search_path {
-                if let Some(new_path) = maybe_forced_sysroot(path, &sysroot) {
-                    *path = new_path;
-                }
-            }
-        } else if arg.starts_with('-') {
-            unrecognised.push(format!("`{arg}`"));
-        } else {
-            args.save_dir.handle_file(arg)?;
```
davidlattimore, line 770: The test failure looks consistent with the save-dir mechanism not working correctly. To verify locally, one way would be to `rm -rf wild/tests/build` before you test. At a guess, I'd say perhaps that this line here has been omitted from the new code.

libwild/src/args.rs
```diff
@@ -1090,6 +714,1414 @@ fn warn_unsupported(opt: &str) -> Result {
...
+        .sub_option("elf64lriscv", "RISC-V 64-bit ELF target")
+        .execute(|args, _modifier_stack, value| {
+            args.arch = Architecture::from_str(value)?;
+            Ok(())
+        });
+
+    parser
+        .declare_prefix("z")
+        .help("Linker option")
+        .sub_option("now", "Set BIND_NOW for object")
```
davidlattimore: It could be good if `sub_option` took an execute callback as an argument. That way we can be sure that the help and actually supported options won't get out-of-sync.

libwild/src/args.rs
```diff
@@ -1090,6 +714,1414 @@ fn warn_unsupported(opt: &str) -> Result {
...
+        declaration = declaration.long(flag);
+        declaration.execute(|_args, _modifier_stack| Ok(()));
+    }
+}
+
+fn add_ignored_flags(parser: &mut ArgumentParser) {
+    for flag in IGNORED_FLAGS {
+        let mut declaration = parser.declare();
+        declaration = declaration.long(flag);
+        declaration.execute(|_args, _modifier_stack| Ok(()));
```
davidlattimore, line 2206: Previously, flags listed in `IGNORED_FLAGS` issued a warning. However, now it looks like they're treated the same as `SILENTLY_IGNORED_FLAGS`.

davidlattimore (commented):
Nice work! Looking forward to having --help!

libwild/src/args.rs
lapla-cogito: Thanks! After making this fix, all integration tests except those using QEMU now pass. I believe the failures with QEMU are due to these changes, so I'll work on resolving both issues together.

lapla-cogito:
It seems that only the mold test suite is failing in CI ([sample](https://github.com/davidlattimore/wild/actions/runs/16866873054/job/47774976612)), but I'm not entirely sure why this is happening... I believe specifying input files between other arguments should be properly parsed, and it doesn't appear to be related to the `@` option either...

lapla-cogito:
With this implementation, we can also define options that don't have help messages. In such cases, the help message won't display the option's name itself. This is intended for debugging options like `debug-fuel`. I'd like to get reviews from core contributors about this new argument parser, such as how to define options, if possible.

libwild/src/args.rs
```diff
@@ -1091,6 +726,1534 @@ fn warn_unsupported(opt: &str) -> Result {
...
+
+#[derive(Clone)]
+struct OptionHandler {
+    help_text: &'static str,
+    handler: OptionHandlerFn,
+    short_names: Vec<char>,
+}
+
+/// Handler for sub-options
+type SubOptionHandler = Rc<dyn Fn(&mut Args, &mut Vec<Modifiers>, &str) -> Result<()>>;
```
davidlattimore: It looks like this could be a Box instead of an Rc.

libwild/src/args.rs
```diff
@@ -1091,6 +725,1534 @@ fn warn_unsupported(opt: &str) -> Result {
...
+    help_text: &'static str,
+}
+
+/// Builder for declaring options with optional parameters
+pub struct OptionWithOptionalParamDeclaration<'a> {
+    parser: &'a mut ArgumentParser,
+    long_names: Vec<&'static str>,
+    short_names: Vec<char>,
+    help_text: &'static str,
+}
```
mati865: This is qute a bit of duplication but doesn't have to be addressed in this PR.

libwild/src/args.rs
```diff
@@ -1091,6 +725,1534 @@ fn warn_unsupported(opt: &str) -> Result {
...
+        let mut declaration = parser.declare();
+        declaration = declaration.long(flag);
+        declaration.execute(|_args, _modifier_stack| Ok(()));
+    }
+}
+
+fn add_default_flags(parser: &mut ArgumentParser) {
+    for flag in DEFAULT_FLAGS {
+        let mut declaration = parser.declare();
+        declaration = declaration.long(flag);
```
mati865, line 2213: Some options like `X`, `x`, `EL` should use short format.

libwild/src/args.rs
```diff
@@ -1091,6 +725,1534 @@ fn warn_unsupported(opt: &str) -> Result {
...
+    add_silently_ignored_flags(&mut parser);
+    add_default_flags(&mut parser);
+
+    parser
+}
+
+fn add_silently_ignored_flags(parser: &mut ArgumentParser) {
+    for flag in SILENTLY_IGNORED_FLAGS {
+        let mut declaration = parser.declare();
+        declaration = declaration.long(flag);
```
mati865, line 2200: Similarly `(` and `)` should use short format, although this one is probably irrelevant.

libwild/src/args.rs
```diff
@@ -1091,6 +725,1500 @@ fn warn_unsupported(opt: &str) -> Result {
...
+    parser
+}
+
+fn add_silently_ignored_flags(parser: &mut ArgumentParser) {
+    for flag in SILENTLY_IGNORED_FLAGS {
+        let mut declaration = parser.declare();
+        match *flag {
+            "(" | ")" => declaration = declaration.short(flag),
+            _ => declaration = declaration.long(flag),
+        }
```
mati865: Alternatively, we could create separate lists for long and short flags or use a tuple to differentiate between them, like this (wrote this on my mobile so it might not work):
```
let test = vec![ (Some("s"), None), (None, Some("long")) ];
```
I don't have a special preference here, just slightly don't like having to keep the arguments in sync in two distinct locations. So it's up to you and @davidlattimore

mati865 (approved):

libwild/src/args.rs
davidlattimore: I agree that it would be nice to avoid having to keep two separate things in sync. If each flag needs to only be either a short flag or a long flag, then splitting the constant into two separate constants sounds good. e.g. `SILENTLY_IGNORED_SHORT_FLAGS`.

marxin:
Great work, thanks for working on that @lapla-cogito!

PR #865, opened 2025-06-10
-z undefs

AadiWaghray:
add undefs option for z flag following further discussion in issue #810.

PR #1030, opened 2025-08-10
added: z-undefs flag with test

AadiWaghray:
add undefs option for z flag following further discussion in issue #810. PR is recreated after rebase issue on previous PR (#865). The "other two" issues mentioned in previous PR persist. I want to confirm how to proceed.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -436,6 +436,7 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
                 "nocopyreloc" => args.allow_copy_relocations = false,
                 "nodelete" => args.needs_nodelete_handling = true,
                 "defs" => args.no_undefined = true,
+                "undefs" => args.no_undefined = false,
                 "muldefs" => args.allow_multiple_definitions = true,
                 _ => {
                     warn_unsupported(&format!("-z {arg}"))?;
--- a/wild/tests/sources/z-defs.c
+++ b/wild/tests/sources/z-defs.c
@@ -1,6 +1,14 @@
-//#LinkArgs:-Bshareable -z defs
+//#Config:z-defs
+//#LinkArgs:-Bshareable -z now -z defs
 //#Mode:dynamic
 //#ExpectError:foo
+
+//#Config:z-undefs
+//#LinkArgs:-Bshareable -z now -z undefs
+//#Mode:dynamic
+//#RunEnabled:false
+//#DiffIgnore:.dynamic.DT_RELA
+//#DiffIgnore:.dynamic.DT_RELAENT
 
 int foo(void);
 


davidlattimore:
From the test failures, it looks like you need to add a couple of `DiffIgnore`s. e.g. `//#DiffIgnore:.dynamic.DT_RELA`.

Don't worry about running into git issues. From the look of things, you probably merged to head and there had just been a lot of commits since you did the original version of the PR. Different people have different workflows. Personally, I always rebase and amend commits. Some people prefer to merge and create fixup commits. Either is fine and we can always squash and rebase when we merge the PR. The git book is a good resource to learn more - https://git-scm.com/book/en/v2. The jj version control system is also good. That's what I've been using for the last year or so.

wild/tests/sources/z-undefs.c
lapla-cogito, line 1: Wild includes integration tests, and the current CI failure is caused by these tests. The integration tests compare linking results with other linkers (specifically GNU ld in this case) and generate errors if any discrepancies are found.
However, not all differences need to be resolved—you can explicitly mark certain parts as acceptable to ignore differences without triggering errors.
Examining the CI output (or by running `cargo test` locally) reveals that two sections differ: `.dynamic.DT_RELA` and `.dynamic.DT_RELAENT`. These differences can be safely ignored. To suppress these differences, you can add lines like `//#DiffIgnore:.dynamic.DT_RELA` or `//#DiffIgnore:.dynamic.DT_RELAENT` to the test source files.

wild/tests/integration_tests.rs
```diff
@@ -2534,6 +2534,7 @@ fn integration_test(
         "preinit-array.c",
         "exception.cc",
         "z-defs.c",
+        "z-undefs.c",
```
lapla-cogito: There doesn’t seem to be much point in creating a new source file here, since it contains the same C code as `z-defs.c`. I’d suggest adding another option for `-z undefs` directly in `z-defs.c`, similar to how `shared.c` does it.
This would make it possible to remove `z-undefs.c`.

AadiWaghray:
I appreciate the leniency and advice. I'll be sure to review the git book.

wild/tests/sources/z-undefs.c
AadiWaghray: I've added these `//#DiffIgnore:` statements along with the `//#Config:` statements to allow for both tests to be done from a single file.  However, in local testing, the z-undefs config fails stating the "Binary ran for too long." This occurs after adding the `//#DiffIgnore:` statements to the z-undefs config. Is there a way you recommend tracing this problem?

lapla-cogito: You can set `//#RunEnabled:false` to disable executing the binary, which will make the test pass.

AadiWaghray: Thank you. That solved the issue. Was the problem that the linker produced a bad binary by ignoring the undefined symbol error, so when it tried to run the binary, it wasn't able to? If that's the case, I'm sorry for not realizing it earlier.

davidlattimore: That might be a problem too, but the main problem is that the binary isn't an executable, it's a shared object, which can't be directly executed. When we get `Mode:dynamic`, we should probably default to `RunEnabled:false`.

davidlattimore (approved):

lapla-cogito (approved):

PR #1041, opened 2025-08-12
obtain verdefnum from verdef section header

mati865:
Similarly to #611 Rust's ELF raw-dylibs do notcontain the information about verdef count.
This goes against LSB, but all the other linkers can handle it, presumably by using `st_info` of`.gnu.version_d` section.

I'm not sure if this is the right thing to do, maybe `sh_info` `should be used just a fallback when `.dynamic` doesn't have `DT_VERDEFNUM` tag?

On current main `tests/ui/linkage-attr/raw-dylib/elf/glibc-x86_64.rs` fails with:
```
wild: error: Failed to load symbols from `/tmp/rustcnCgdWG/raw-dylibs/lib105m6f543m4qw1t7khnttg751.so`
  Caused by:
    Invalid version index 1
```


libwild/src/layout.rs
```diff
@@ -5282,9 +5282,16 @@ impl<'data> DynamicLayoutState<'data> {
         resources: &GraphResources<'data, '_>,
         queue: &mut LocalWorkQueue,
     ) -> Result {
-        let dt_info = DynamicTagValues::read(self.object)?;
-        self.symbol_versions_needed = vec![false; dt_info.verdefnum as usize];
+        let mut verdefnum = 0;
```
davidlattimore: Would it work to just use `self.object.verdefnum` here instead?

mati865: Yeah, that works fine.

davidlattimore (approved):

mati865:
FWIW this has been fixed in upstream Rust: https://github.com/rust-lang/rust/pull/145820

PR #1044, opened 2025-08-13
Run external tests on Ubuntu 25.04

lapla-cogito:
Some external tests seem to be yielding undesirable successes or failures due to the tool being outdated.


lapla-cogito:
The regression tests for CREL reveal [failures](https://github.com/davidlattimore/wild/actions/runs/16926322445/job/47962584365). This is the correct behavior, as previously this test was incorrectly passing due to outdated toolchains.

lapla-cogito:
Additionally, some tests now [pass](https://github.com/davidlattimore/wild/actions/runs/16926404836/job/47962807947) (see the "Check tests that should fail still fail"). This also represents correct behavior.

PR #1049, opened 2025-08-13
args: Allow prefixes to unify with long options in help

davidlattimore:
For now, all this affects is that -T and --script are shown on one line instead of two.


lapla-cogito (approved):
This also fixes an issue where `-O` would be recognized as an argument but actually have no effect while still appearing in the help output, thanks!

Issue #464, opened 2025-02-22
Improve undefined symbol error when the symbol exists only as LTO bytecode


mati865:
Upstream issue: https://github.com/rust-lang/rust/issues/126280

Linking library that contains only LTO bytecode fails with errors like `Error: Undefined symbol _Z8checksumPKvi`. Although this is correct since the symbol is missing in provided libraries, there might be LTO bytecode containing it.
It would be nice to let the user know they should recompile a library in a way that preserves the "normal" symbols, or use a different linker.

We should defer reading LTO sections until wild finds undefined symbol to avoid regressing the performance.

Not marking as a good first issue because I don't know if we can easily read enough of LTO sections to figure whether the symbol is present.

davidlattimore:
I had a look at the files in the repo linked from the upstream issue and it looks like getting the symbol names out isn't too hard. e.g.

```
❯ readelf -p .gnu.lto_.symtab.8ae6ee4a5d4d180b libs/libexample.a | c++filt 
readelf: Warning: Section '.gnu.lto_.symtab.8ae6ee4a5d4d180b' was not dumped because it does not exist
readelf: Warning: Section '.gnu.lto_.symtab.8ae6ee4a5d4d180b' was not dumped because it does not exist

File: libs/libexample.a(interfaces.o)

File: libs/libexample.a(Net.o)

File: libs/libexample.a(export.o)

String dump of section '.gnu.lto_.symtab.8ae6ee4a5d4d180b':
  [     0]  CRC32_ProcessSingleBuffer(void const*, int)
  [    22]  CRC32_ProcessSingleBuffer(void const*, int)
  [    52]  checksum(void const*, int)
  [    71]  CRC32_Final(unsigned int*)
  [    92]  CRC32_ProcessBuffer(unsigned int*, void const*, int)
  [    bf]  CRC32_Init(unsigned int*)
```

Although even if we couldn't get the symbol names, just seeing the presence of `.gnu.lto_` sections could be enough to justify giving a different error message.

But it looks like we could find all such symbols by iterating through all input objects, iterating through all the sections in each object, then if the section name starts with `.gnu.lto_.symtab.`, reading the section data, splitting the null-terminated strings and then checking if that matches the symbol that was undefined. Hopefully that's enough detail to mark this as a good-first-issue. If anyone working on this would like more detail, please reach out. Even if you don't need more detail, please reach out so that we know you're working on it and others don't duplicate your efforts.

AadiWaghray:
[...]

PR #1050, opened 2025-08-14
LTO archive test

AadiWaghray:
PRs #91, #744, and #979 should be sufficient to resolve issue #464. But, I think this test highlights the need for more work on the issue. However, I'm not entirely sure if the test is erroneously written. It gets past the LTO guard, but it doesn't error on an undefined symbol as one would expect. Instead, it errors on "Binary ran for too long."

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3055,7 +3055,9 @@ fn process_relocation<A: Arch>(
                 )
                 .context("Failed to get source info")?;
 
-                let raw_symbol_name = symbol_db.symbol_name(symbol_id).unwrap();
+                let raw_symbol_name = symbol_db
+                    .symbol_name(symbol_id)
+                    .expect("Errored on symbol {symbol_name} so it exists");
                 let mut lto_file: Option<&PathBuf> = None;
                 let lto_unresolved = resources.symbol_db.groups.iter().any(|group| match group {
                     Group::Objects(data) => data.iter().any(|input| {
@@ -3070,7 +3072,7 @@ fn process_relocation<A: Arch>(
                             if has_lto {
                                 let lto_contains_undef = file
                                     .section_data_cow(section_header)
-                                    .unwrap()
+                                    .unwrap_or_default()
                                     .split(|datum| *datum == 0)
                                     .any(|symbol| symbol == raw_symbol_name.bytes());
                                 if lto_contains_undef {
@@ -3085,10 +3087,19 @@ fn process_relocation<A: Arch>(
                 });
 
                 if lto_unresolved {
-                    resources.report_error(error!(
-                        "undefined reference to `{symbol_name}` found in LTO section of {}",
-                        lto_file.unwrap().file_name().unwrap().display()
-                    ));
+                    if let Some(file) = lto_file {
+                        let file = file.canonicalize().unwrap_or(PathBuf::new());
+                        resources.report_error(error!(
+                            "undefined reference to `{symbol_name}` found in LTO section of {}",
+                            file.file_name()
+                                .expect("Canonicalized path can't have ../ at end")
+                                .to_str()
+                                .expect("File name should be valid UTF-8")
+                                .split(".")
+                                .next()
+                                .unwrap()
+                        ));
+                    }
                 } else if args.error_unresolved_symbols {
                     resources.report_error(error!(
                         "Undefined symbol {symbol_name}, referenced by {}\n    {}",
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -33,6 +33,7 @@ use crate::error::Error;
 use crate::error::Result;
 use crate::error::warning;
 use crate::file_writer;
+use crate::grouping::Group;
 use crate::input_data::FileId;
 use crate::input_data::InputData;
 use crate::input_data::InputRef;
@@ -91,6 +92,7 @@ use linker_utils::elf::riscvattr::TAG_RISCV_STACK_ALIGN;
 use linker_utils::elf::riscvattr::TAG_RISCV_UNALIGNED_ACCESS;
 use linker_utils::elf::riscvattr::TAG_RISCV_WHOLE_FILE;
 use linker_utils::elf::riscvattr::TAG_RISCV_X3_REG_USAGE;
+use linker_utils::elf::secnames;
 use linker_utils::elf::shf;
 use linker_utils::elf::sht;
 use linker_utils::relaxation::RelocationModifier;
@@ -120,6 +122,7 @@ use std::mem::swap;
 use std::mem::take;
 use std::num::NonZeroU32;
 use std::num::NonZeroU64;
+use std::path::PathBuf;
 use std::sync::Mutex;
 use std::sync::atomic;
 use std::sync::atomic::AtomicBool;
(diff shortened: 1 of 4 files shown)


wild/tests/sources/lto-undefined.c
```diff
@@ -0,0 +1,5 @@
+//#LinkArgs:-L./ -lexample
+
+extern int CRC32_ProcessSingleBuffer(void const*, int);
```
davidlattimore: Declaring an extern function won't actually cause a reference to the symbol to be emitted. The code needs to actually refer to the symbol - e.g. from main.

wild/tests/sources/libexample.a
davidlattimore, line 1: It's fine to add this file to experimentation, but I think if we wanted to merge this PR, we'd need to figure out how to build a similar file from source.

AadiWaghray:
Apologies for pushing prematurely. There is an issue I need to fix. But, the changes mentioned above have been added.

wild/tests/sources/lto-undefined.c
```diff
@@ -0,0 +1,11 @@
+//#Archive:archive_lto.c:-flto
+
+extern int addition(int, int);
+
+int main(int argc, char **argv) {
```
davidlattimore, line 9: Most of our tests use `_start` rather than `main`. `_start` is the default entry point for the binary. `main` is called sometime later by the libc runtime. A few of our tests do use `main`, but we only do that when necessary, since it requires linking against libc, which substantially slows down linking. I suspect the reason you're seeing a failure is because you haven't defined `_start` and also haven't linked against libc. I'd suggest looking at a test like `trivial.c` to see how the `_start`-based tests are written.

AadiWaghray: That makes sense. I incorrectly assumed libc was linked automatically for testing since I used main in my z-defs testing. I saw the `Requires` set of flags but didn't put two and two together. Pushing fixes now, the error is now, correctly, the undefined symbol error showing archives avoid the LTO guard. I still have the issue we discussed earlier of accessing the archive's sections at the point we emit the undefined error to emit an LTO error rather than an undefined error.

davidlattimore:
Looking at one of the test failures:

```
Error: Failed to build program `lto-undefined.c` with linker `ld` config `default`
  Caused by:
    Linker failed:
/usr/bin/ld.bfd: /__w/wild/wild/wild/tests/build/archive_lto.default-host-2d02e06e5b1d047f.a(archive_lto.default-host-2d02e06e5b1d047f.o): plugin needed to handle lto object
/usr/bin/ld.bfd: /__w/wild/wild/wild/tests/build/archive_lto.default-host-2d02e06e5b1d047f.a(archive_lto.default-host-2d02e06e5b1d047f.o): plugin needed to handle lto object
/usr/bin/ld.bfd: /__w/wild/wild/wild/tests/build/lto-undefined.default-host-c78a7f268027252.o: in function `_start':
lto-undefined.c:(.text+0x24): undefined reference to `addition'
```

This is failing to link with ld. i.e. it hasn't actually gotten to linking with wild, which happens after GNU ld. GNU ld looks to be failing because the linker plugin wasn't passed as an argument. I guess that suggests that for this test, we do need to use the C compiler to drive the linker rather than invoking the linker directly. If you look at `trivial-main.c`, you can see how we set up to use GCC as the linker driver - pretty much set `//#LinkerDriver:gcc`.

AadiWaghray:
The test should be good now!

davidlattimore:
Cool, yep that failure does indeed show that we're still giving an undefined symbol error rather than a more appropriate error. I suspect #979 might have actually regressed that, but I don't think we want to undo that change, since that would hurt performance. The issue is that we now only look for LTO sections in archive entries that we activate, but we never activate the relevant archive entry because it doesn't define any any needed symbols in its symtab.

One possibility would be that if we detect that we've got an undefined symbol error, we go back and scan all archived objects that we didn't already process. If any turn out to have LTO bitcode, then make that be the error instead of the undefined symbol error.

AadiWaghray:
There was something about the tests that was confusing me. If you run `cargo test` multiple times, then ld will fail every time on the first run. But, every `cargo test` after that will succeed...As I'm writing this, I realize its because the test assumes the error will be for both linkers. I'll fix that and the formatting in the morning.

AadiWaghray:
I'm not quite sure why the tests aren't passing. They pass locally and are giving the error message I expect and catch in the test header.

Edit: I realize what the problem is. I'll push a fix.

libwild/src/layout.rs
```diff
@@ -3052,7 +3055,52 @@ fn process_relocation<A: Arch>(
...
+                if lto_unresolved {
+                    if let Some(file) = lto_file {
+                        let file = file.canonicalize().unwrap_or(PathBuf::new());
+                        resources.report_error(error!(
+                            "undefined reference to `{symbol_name}` found in LTO section of {}",
+                            file.file_name()
+                                .expect("Canonicalized path can't have ../ at end")
+                                .to_str()
+                                .expect("File name should be valid UTF-8")
+                                .split(".")
```
davidlattimore: Why are you only taking the filename up to '.'?

libwild/src/layout.rs
```diff
@@ -3052,7 +3055,52 @@ fn process_relocation<A: Arch>(
                 )
                 .context("Failed to get source info")?;
 
-                if args.error_unresolved_symbols {
+                let raw_symbol_name = symbol_db
+                    .symbol_name(symbol_id)
+                    .expect("Errored on symbol {symbol_name} so it exists");
+                let mut lto_file: Option<&PathBuf> = None;
+                let lto_unresolved = resources.symbol_db.groups.iter().any(|group| match group {
```
davidlattimore: Is `lto_unresolved` actually needed? It looks like it's always `true` when `lto_file` is `Some` and `false` when `lto_file` is `None`. Perhaps just replace the two `any` calls with `find_map`.

libwild/src/layout.rs
```diff
@@ -3052,7 +3055,52 @@ fn process_relocation<A: Arch>(
                 )
                 .context("Failed to get source info")?;
 
-                if args.error_unresolved_symbols {
+                let raw_symbol_name = symbol_db
+                    .symbol_name(symbol_id)
+                    .expect("Errored on symbol {symbol_name} so it exists");
```
davidlattimore: Note that the `{symbol_name}` won't substitute the variable `symbol_name`.

libwild/src/layout.rs
```diff
@@ -3052,7 +3055,52 @@ fn process_relocation<A: Arch>(
...
+                let mut lto_file: Option<&PathBuf> = None;
```
davidlattimore: This method is already pretty large and adding this makes it ever larger. Could something be moved out to a separate function? Perhaps the code to compute `lto_file`?

libwild/src/layout.rs
AadiWaghray: The file generated is `archive_lto.a`. I believe that the linker adds something to the end of the object file extracted into memory when parsing the archive. I thought it would better to give the name of the actual archive file and avoid the random values. This is the error I mentioned fixing above where the test would pass locally but not in github actions. I never checked where these values are being added onto the name. If we are able to know what it is and add the appropriate value to the `Expected` header value, it is possible that this is entirely unnecessary.

libwild/src/layout.rs
AadiWaghray: That's fair. I was originally just checking if the undefined symbol came from a file with lto, but I then realized I needed to add the file name in the error message.

libwild/src/layout.rs
AadiWaghray: That makes sense. Will work on it.

libwild/src/layout.rs
davidlattimore: I suspect the extra stuff in the filename is added by the integration testing framework, so it's actually part of the filename.

libwild/src/layout.rs
AadiWaghray: I didn't know that expect doesn't format strings. Thanks for pointing it out!

libwild/src/layout.rs
AadiWaghray: After reading `integration_tests.rs` and reviewing the commits, I see you're 100% right. The command is hashed as part of the file name, and I added the `//#SkipLinker:ld` directive between commits. I'll change that and everything should be good.

Issue #948, opened 2025-06-28
Wrong R_X86_64_GOTPC32_TLSDESC relaxation


marxin:
Noticed that while debugging a crashing binary on Gentoo Linux. Since db10f0fee2d4475b89c8cf7c9704b6fd9458de50, we make the following relaxation where we ignore the destination register (in this example `%r12` and `%rbx`):
https://github.com/davidlattimore/wild/blob/3bf6aa26469caf4b0d0d586c880c8aa38bbc79ea/linker-utils/src/x86_64.rs#L164-L176

```
Disassembly of section .text.clc_initialize_llvm:

0000000000000000 <clc_initialize_llvm>:
   0:   f3 0f 1e fa             endbr64
   4:   41 54                   push   %r12
   6:   4c 8d 25 00 00 00 00    lea    0x0(%rip),%r12        # d <clc_initialize_llvm+0xd>      9: R_X86_64_GOTPC32_TLSDESC     _ZSt11__once_call-0x4
   d:   48 8d 0d 00 00 00 00    lea    0x0(%rip),%rcx        # 14 <clc_initialize_llvm+0x14>    10: R_X86_64_PC32       .text._ZZNSt9once_flag18_Prepare_executionC4IZSt9call_onceIZ19clc_initialize_llvmEUlvE_JEEvRS_OT_DpOT0_EUlvE_EERS5_ENUlvE_4_FUNEv-0x4
  14:   55                      push   %rbp
  15:   53                      push   %rbx
  16:   48 8d 1d 00 00 00 00    lea    0x0(%rip),%rbx        # 1d <clc_initialize_llvm+0x1d>    19: R_X86_64_GOTPC32_TLSDESC    _ZSt15__once_callable-0x4
```

```
0000000000015d80 <clc_initialize_llvm>:
   15d80:       f3 0f 1e fa             endbr64
   15d84:       41 54                   push   %r12
   15d86:       48 8b 05 53 66 00 00    mov    0x6653(%rip),%rax        # 1c3e0 <_ZSt11__once_call@GLIBCXX_3.4.11>
   15d8d:       48 8d 0d dc ff ff ff    lea    -0x24(%rip),%rcx        # 15d70 <_ZZNSt9once_flag18_Prepare_executionC4IZSt9call_onceIZ19clc_initialize_llvmEUlvE_JEEvRS_OT_DpOT0_EUlvE_EERS5_ENUlvE_4_FUNEv>
   15d94:       55                      push   %rbp
[...]

PR #1051, opened 2025-08-15
Fix R_X86_64_GOTPC32_TLSDESC relaxations

marxin:
Properly detect the destination register and encode the information into the `mov` replacement instruction.

Both variants are tested for `%rax` in our test-suite (one via a Mold test). @mati865 can you please test the suggested fix?

Fixes: #948

### DIFF
--- a/linker-utils/src/x86_64.rs
+++ b/linker-utils/src/x86_64.rs
@@ -162,16 +162,44 @@ impl RelaxationKind {
                 *offset_in_section += 15;
             }
             RelaxationKind::TlsDescToLocalExec => {
+                let rex = section_bytes[offset - 3];
+                let modrm = section_bytes[offset - 1];
+
+                // Extract REX.R (bit 2 of rex) and reg field (bits 3-5 of modrm)
+                let rex_r = (rex >> 2) & 1;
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):

linker-utils/src/x86_64.rs
```diff
@@ -162,16 +162,44 @@ impl RelaxationKind {
...
+
+                // Extract REX.R (bit 2 of rex) and reg field (bits 3-5 of modrm)
+                let rex_r = (rex >> 2) & 1;
+                let reg = (modrm >> 3) & 0x7;
+
+                let rex = if rex_r == 0 { 0x48 } else { 0x49 };
                 section_bytes[offset - 3..offset + 4].copy_from_slice(&[
-                    // mov {offset},%rax
-                    0x48, 0xc7, 0xc0, 0, 0, 0, 0,
+                    // mov {offset},%{dest}
```
mati865: Nit: do we want to use `dest` here? It might be unclear without the context. Also applies to the chunk below.

marxin: Yep, renamed to `%{reg}`, hope it's better?

Issue #990, opened 2025-07-04
Version script: minor improvements


marxin:
Right now, there are 2 minor limitations that will enable more of the Mold's tests:
- Escaping - a glob pattern with an escape char (`\`) is rejected; one needs to properly identify if `*` is present in a glob or not (`StarGlob`, `NonstarGlob`).
- The currently used glob crate (`glob`) uses the exclamation mark for negation (`!`) of a group, while the spec says '^' is the char that negates characters in a set.
- `extern "C" {..}` can be parsed and skipped by default; it's a no-op.

lapla-cogito:
Using `^` to represent negation in version scripts seems to be non-standard for GNU. In fact, GNU ld treats scripts like [`version-script15.sh`](https://github.com/rui314/mold/blob/main/test/version-script15.sh) as invalid:

```
（▰╹◡╹）❯  cat <<EOF > a.scr
{
local:
  *;
global:
  [abc][^abc][^\]a-zABC];
};
EOF

（▰╹◡╹）❯  ld --version-script a.scr
ld:a.scr:4: syntax error in VERSION script
```

However, since both lld and mold tolerate such scripts, supporting this feature in wild could offer some convenience. This might also be a good opportunity to implement functionalities that were previously achieved using the `glob` crate without relying on it. I'd like to proceed with this implementation and investigate how it affects performance.

PR #1053, opened 2025-08-15
Support escaping in version scripts

lapla-cogito:
part of #990 

Ensure that characters used in glob patterns in version scripts are correctly parsed even when escaped.

### DIFF
--- a/libwild/src/version_script.rs
+++ b/libwild/src/version_script.rs
@@ -51,6 +51,8 @@ pub(crate) struct VersionScript<'data> {
 pub(crate) enum SymbolMatcher<'data> {
     // Exact match.
     Exact(&'data [u8]),
+    // Exact match with escape sequences that need unescaping.
+    EscapedExact(&'data [u8]),
     // A glob pattern with a '*' token.
     StarGlob(Pattern),
     // A glob pattern without any '*' token.
@@ -62,7 +64,7 @@ pub(crate) enum SymbolMatcher<'data> {
 #[derive(Debug, Default)]
 pub(crate) struct BasicMatchRules<'data> {
     exact: HashSet<PreHashed<UnversionedSymbolName<'data>>, PassThroughHasher>,
-    escaped_exact: Vec<&'data [u8]>,
+    escaped_exact: HashSet<Vec<u8>>,
     star_globs: Vec<Pattern>,
     nonstar_globs: Vec<Pattern>,
     matches_all: bool,
@@ -75,11 +77,11 @@ impl<'data> BasicMatchRules<'data> {
             SymbolMatcher::StarGlob(glob) => self.star_globs.push(glob),
             SymbolMatcher::NonstarGlob(glob) => self.nonstar_globs.push(glob),
             SymbolMatcher::Exact(exact) => {
-                if has_escape_chars(exact) {
-                    self.escaped_exact.push(exact);
-                } else {
-                    self.exact.insert(UnversionedSymbolName::prehashed(exact));
-                }
+                self.exact.insert(UnversionedSymbolName::prehashed(exact));
+            }
+            SymbolMatcher::EscapedExact(escaped) => {
+                let unescaped = unescape_pattern(escaped);
+                self.escaped_exact.insert(unescaped);
             }
         }
     }
@@ -90,11 +92,6 @@ impl<'data> BasicMatchRules<'data> {
         lookup: &mut SymbolLookupNameWrapper,
         mangled: bool,
     ) -> bool {
-        // Early exit before we actually demangle the name.
-        if self.exact.is_empty() && self.escaped_exact.is_empty() {
-            return false;
-        }
-
         // Check normal exact matches first
         if !self.exact.is_empty() {
             if mangled {
@@ -121,10 +118,8 @@ impl<'data> BasicMatchRules<'data> {
                 lookup.name.bytes()
             };
 
-            for &escaped_pattern in &self.escaped_exact {
-                if matches_escaped_pattern(symbol_bytes, escaped_pattern) {
-                    return true;
-                }
+            if self.escaped_exact.contains(symbol_bytes) {
+                return true;
             }
         }
 
@@ -568,6 +563,7 @@ pub(crate) fn parse_matcher<'data>(
 
             match glob_type {
                 GlobPatternType::Exact => SymbolMatcher::Exact(token),
+                GlobPatternType::EscapedExact => SymbolMatcher::EscapedExact(token),
                 GlobPatternType::Star => SymbolMatcher::StarGlob(create_pattern(token)?),
                 GlobPatternType::NonStar => SymbolMatcher::NonstarGlob(create_pattern(token)?),
             }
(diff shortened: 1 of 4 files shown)


libwild/src/version_script.rs
```diff
@@ -552,12 +551,41 @@ fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
     take_while(1.., |b| !b" (){}\n\t".contains(&b)).parse_next(input)
 }
 
+/// Check if a pattern contains unescaped glob characters.
+/// Returns (has_unescaped_star, has_unescaped_nonstar_glob)
+fn analyze_glob_pattern(pattern: &[u8]) -> (bool, bool) {
```
marxin: I would rather return an enum that a pair of boolean types.

libwild/src/version_script.rs
```diff
@@ -552,12 +551,41 @@ fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
...
+    let mut i = 0;
+
+    while i < pattern.len() {
+        match pattern[i] {
+            b'\\' => {
+                // Skip the escaped character
+                i += 2;
+            }
+            b'*' => {
+                has_unescaped_star = true;
```
marxin: You can probably do an early exit if unescaped star is found, right?

libwild/src/version_script.rs
```diff
@@ -552,12 +551,41 @@ fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
...
 }
 
+/// Check if a pattern contains unescaped glob characters.
+/// Returns (has_unescaped_star, has_unescaped_nonstar_glob)
+fn analyze_glob_pattern(pattern: &[u8]) -> (bool, bool) {
+    let mut has_unescaped_star = false;
+    let mut has_unescaped_nonstar_glob = false;
+    let mut i = 0;
+
+    while i < pattern.len() {
```
marxin: Please use rather `while let Some(c) = it.next()` iterator pattern (more idiomatic in Rust).

libwild/src/version_script.rs
```diff
@@ -552,12 +558,43 @@ fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
...
 }
 
+enum GlobPatternType {
+    Exact,
+    Star,
+    NonStar,
+}
+
+fn analyze_glob_pattern(pattern: &[u8]) -> GlobPatternType {
+    let mut has_unescaped_nonstar_glob = false;
```
marxin: I would keep the state directly in a var as `GlobPatternType`.

libwild/src/version_script.rs
```diff
@@ -522,28 +522,34 @@ pub(crate) fn parse_matcher<'data>(
...
-            } else {
-                SymbolMatcher::NonstarGlob(pattern)
-            }
         } else {
-            SymbolMatcher::Exact(token)
+            let glob_type = analyze_glob_pattern(token);
+
+            match glob_type {
+                GlobPatternType::Star | GlobPatternType::NonStar => {
+                    let pattern = Pattern::new(str::from_utf8(token).map_err(|_| {
```
marxin: I would factor out the `Pattern` creation to a closure so that you can have just one level `match glob_type`.

marxin (approved):

libwild/src/version_script.rs
lapla-cogito: Makes sense, that does make the code simpler. Thanks!

davidlattimore:
Does this mean that a matcher with 'foo\*bar' will match a symbol called 'foo*bar' or is this just the parsing side of things and matching will be done separately? I'd expect that matching with escape sequences would be slower than matching without escapes, so there might be something for handling them separately.

marxin:
We've got a code above the changed one that handles the quoted strings that are subject to exact matches only.
The entire logic responsible for the matching is done via `glob::Pattern`, and this PR is only responsible for classification of the patterns for the purpose of the priorities (non-glob patterns, non-start blogs, star globs).

davidlattimore:
It looked to me like `foo\*bar` would be classified as exact, so would be put into the the hashset, which would result in only symbols named `foo\*bar` matching. But I've only had time for a quick look, so I'd defer to you.

davidlattimore:
With this PR checked out, I just tried editing `wild/tests/sources/symbol-versions-script.map` to change `bar_local` to `bar\_local`. This caused the `symbol_versions` test to fail because wild stopped treating `bar_local` as a local, emitting a dynamic relocation for it, whereas the GNU ld output didn't change from what it was producing when the version script listed `bar_local`.

libwild/src/version_script.rs
```diff
@@ -86,19 +91,44 @@ impl<'data> BasicMatchRules<'data> {
         mangled: bool,
     ) -> bool {
         // Early exit before we actually demangle the name.
-        if self.exact.is_empty() {
+        if self.exact.is_empty() && self.escaped_exact.is_empty() {
```
davidlattimore: This early exit seems to now be redundant because the demangling is now guarded below by `if !self.exact.is_empty()` and also by `if !self.escaped_exact.is_empty()`. Although depending on what you do with the next comment below, you might not need the changes to this method at all, so look at this comment last.

libwild/src/version_script.rs
```diff
@@ -552,12 +579,74 @@ fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
...
+}
+
+fn analyze_glob_pattern(pattern: &[u8]) -> GlobPatternType {
+    let mut pattern_type = GlobPatternType::Exact;
+    let mut it = pattern.iter();
+
+    while let Some(&c) = it.next() {
+        match c {
+            b'\\' => {
+                it.next();
```
davidlattimore, line 596: If we get here and don't subsequently encounter a `*`, we know that the pattern has an escape sequence. Consider adding another variant to `GlobPatternType` - e.g. `EscapedExact`. Then you wouldn't need the `has_escape_chars` function.

libwild/src/version_script.rs
```diff
@@ -86,19 +91,44 @@ impl<'data> BasicMatchRules<'data> {
...
+        if !self.escaped_exact.is_empty() {
+            let symbol_bytes = if mangled {
+                let demangled_name = lookup.get_demangled_name();
+                demangled_name.as_bytes()
+            } else {
+                lookup.name.bytes()
+            };
+
+            for &escaped_pattern in &self.escaped_exact {
+                if matches_escaped_pattern(symbol_bytes, escaped_pattern) {
```
davidlattimore: It looks as though every time we check if a symbol matches one of the `escaped_exact` patterns, we de-escape all the patterns. It'd likely be more efficient to just de-escape them once and store those de-escaped patterns in a `HashSet`. In fact, if you wanted, you could store the de-escaped pattern in the main `HashSet` - `exact` - although you'd need to copy the de-escaped bytes into an arena in order for the lifetimes to work out.

lapla-cogito: Modified to use `unescaped_exact` HashSet for storing patterns.

davidlattimore (approved):

PR #1058, opened 2025-08-20
Do not lookup files from args in search paths

mati865:
Fixes a regression from #610 that manifests in cups source code.
Minimised reproducer:
```sh
$ mkdir bad_foo
$ touch bad_foo/foo.o
$ echo 'int main() { return 0; }' > main.c
$ gcc -c main.c -o foo.o
❯ gcc -B ~/Projects/wild/fakes-debug/ -pie foo.o -Lbad_foo/
wild: error: Undefined symbol main, referenced by
    /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../lib/Scrt1.o
```
The problem was due to `bad_foo/foo.o` being preferred over `./foo.o`.

Created on top of https://github.com/davidlattimore/wild/pull/1057

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -194,8 +194,10 @@ pub(crate) struct Input {
 
 #[derive(Debug, Eq, PartialEq)]
 pub(crate) enum InputSpec {
+    /// Path (possibly just a filename) to the file.
     File(Box<Path>),
-    Lib(Box<str>),
+    /// Name of the library and whether it already has suffix and prefix.
+    Lib(Box<str>, bool),
 }
 
 #[derive(Debug, Eq, PartialEq)]
@@ -1198,7 +1200,7 @@ fn setup_argument_parser() -> ArgumentParser {
             "libname",
             "Link with library libname.so or libname.a",
             |args, modifier_stack, value| {
-                let spec = InputSpec::Lib(Box::from(value));
+                let spec = InputSpec::Lib(Box::from(value), false);
                 args.inputs.push(Input {
                     spec,
                     search_first: None,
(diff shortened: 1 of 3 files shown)


wild/tests/integration_tests.rs
```diff
@@ -74,7 +74,7 @@
 //!
 //! Compiler:gcc|g++|clang|clang++ Specifies what compiler should be used to compile C/C++ code.
 //!
-//! Arch:{arch1}[,{arch2}...] Specifies which architectures this test should be run with. Defaults
```
davidlattimore: I think the square brackets here were intended to designate optionality. So one item needs to be specified, additional items are optional. I'd be fine with just removing the square brackets, but putting them around all the arguments seems odd to me.

libwild/src/args.rs
```diff
@@ -194,8 +194,10 @@ pub(crate) struct Input {
 
 #[derive(Debug, Eq, PartialEq)]
 pub(crate) enum InputSpec {
+    /// Path (possibly just a filename) to the file.
     File(Box<Path>),
-    Lib(Box<str>),
+    /// Name of the library and whether it already has suffix and prefix.
+    Lib(Box<str>, bool),
```
davidlattimore: I wonder how it might look if instead of adding a bool to this variant, we had a separate variant. e.g. `Search(Box<str>)` which indicates to search for a file with the indicated name.

davidlattimore (approved):

wild/tests/integration_tests.rs
mati865: Ah, I thought the intention was to make it an array. That coma threw me off.
Anyway, I'll abandon that commit.

libwild/src/args.rs
mati865: Sure, I'll prepare it.

PR #1065, opened 2025-08-26
Bump minimal Rust version to 1.89

lapla-cogito:


RossSmyth:
Already merged and in the channel, just need the lock file to be updated.
NixOS/nixpkgs#431899
https://nixpk.gs/pr-tracker.html?pr=431899

nix/default.nix
```diff
@@ -18,8 +18,8 @@
   glibc,
   lld,
 }:
-assert lib.assertMsg (lib.versionAtLeast "1.88.0" pkgs.rustc.version)
-  "Wild requires at least Rust 1.88.0, this instance of nixpkgs has Rust ${pkgs.rustc.version}";
+assert lib.assertMsg (lib.versionAtLeast "1.89.0" pkgs.rustc.version)
```
whispersofthedawn:
```suggestion
assert lib.assertMsg (lib.versionAtLeast pkgs.rustc.version "1.89.0")
```
This call was incorrect, even before this PR. `versionAtLeast` checks that the first parameter is at least the second, so it would require that the rustc version was *less than or equal to* 1.88.0 (prior to this PR), instead of the reverse.

Relevant docs: https://ryantm.github.io/nixpkgs/functions/library/strings/#function-library-lib.strings.versionAtLeast

RossSmyth: Oops, that's my fault
https://github.com/davidlattimore/wild/commit/815a96ba1b720d10c420e75db8a566ff95d0d66d#diff-4e3bd328e4c998f53b19da7639df45b2db63601be4b4343534880adfbeaa5840R20-R22

RossSmyth:
Personally the fact that the MSRV is bumped doesn't effect me (nixpkgs), we're just waiting for 0.6.0 to be released to add it.

Issue #1073, opened 2025-09-08
Too much is kept when run with --no-gc-sections


davidlattimore:
Wild, unlike the other linkers defaults to --gc-sections, at least for the time being. Wild does however support turning off of gc-sections via the `--no-gc-sections` flag. When doing so, a hello-world Rust program ends up around 8MiB in size, while GNU ld with the same flags produces a 1.3MiB binary. I suspect this is because we're embedding the .meta sections into the executable. These sections are marked with the `exclude` flag, so shouldn't be linked.

The fix should be pretty straightforward - adding an extra check to `SectionRules::lookup`.

We should also add a test for this. To create a section with the exclude bit set, it'll probably be necessary to write at least part of the test in assembly. Look at existing '.s' files that make use of the '.section' directive. If you put some string in a section with the exclude bit set, then you can verify that the resulting binary doesn't contain that string using the "DoesNotContain" test directive.

Tests with assembly are often architecture-specific. It should be possible to make this one architecture-neutral by building only a shared object rather than an executable. e.g. make a shared object with a couple of sections, one that is exclude and one that isn't. Assert that the non-exclude section is present in the resulting shared object and that the exclude section is absent.

anupamcdot:
@davidlattimore can i work on this issue?

davidlattimore:
Sure! Let me know if you get stuck or run into any problems.

anupamcdot:
Hi @davidlattimore , unfortunately I won’t be able to give this task the time it requires, so I’d like to step back and let someone else take it over. Please feel free to reassign it. Apologies for the inconvenience, and thank you for understanding.

davidlattimore:
No worries. Thanks for letting me know!

PR #1077, opened 2025-09-10
Support excluded sections

lqd:
This adds support for sections to exclude, flagged "e", as described in #1073. 

As expected from the description, with this PR a rust helloworld is indeed 1.3MB, and 8.1MB without it as the `.llvmbc` sections are not discarded.

I'm not sure about the test. I don't really know what I'm doing and wasn't able to test how the issue describes: I kept hitting errors about loaded sections while trying make a shared object. I'm also not sure if it should be using `--no-gc-sections`. To be clear, I have verified it fails without the last commit discarding excluded sections.

### DIFF
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -389,6 +389,10 @@ impl<'data> SectionRules<'data> {
         section_flags: SectionFlags,
         sh_type: SectionType,
     ) -> SectionRuleOutcome {
+        if section_flags.should_exclude() {
+            return SectionRuleOutcome::Discard;
+        }
+
         if let Some(hash) = section_name_prefix_hash(section_name)
             && let Some(rule) = self.rules.find(hash, |rule| rule.matches(section_name))
         {


mati865:
Hello there!

Could you run `clang-format -i /__w/wild/wild/wild/tests/sources/*.{c,cc,h}` so the CI shows the real issue?

> I'm also not sure if it should be using `--no-gc-sections`

Yeah, Wild unlike other linkers performs sections GC by default. For this test we need to disable it, to make sure the section is discarded even without GC.
I can look into it tomorrow.

wild/tests/sources/exclude-section.c
```diff
@@ -0,0 +1,22 @@
+// Defines strings in two sections, but one of the sections is marked with the `exclude` flag.
+
+//#Object: exclude-section.s
+//#Object: runtime.c
+//#Arch: x86_64
+
```
davidlattimore: You can pass `--no-gc-sections` by adding:

```
//#LinkArgs:--no-gc-sections
```

wild/tests/sources/exclude-section.s
```diff
@@ -0,0 +1,11 @@
+.section .keep, "aSM", @progbits, 1
```
davidlattimore: At the moment both sections are string-merge sections ("S" and "M" flags). I'd probably keep this file somewhat minimal. Perhaps even get rid of the symbols so that the only difference is the exclude bit (at the moment one of the sections has a symbol that is referenced, which prevents GC if gc-sections is enabled). Something like this:

```
.section .keep
.ascii "ShouldBeKept\0"

.section .exclude, "e"
.ascii "ShouldBeExcluded\0"
```

davidlattimore (commented):
Thanks for working on this!

wild/tests/sources/exclude-section.s
lqd: Ah thanks, that's much simpler and focused on the actual feature!

wild/tests/sources/exclude-section.c
lqd: Done.

lqd:
> Could you run `clang-format`

Done.

I've also pushed forcing `--no-gc-sections` and the more minimal test structure David suggested. I imagine turning it into a shared object would also remove the need to set up the runtime, have a `_start` symbol and some `.text`, etc?

davidlattimore:
Yep, if you wanted to go the shared object route, I think you'd add `-shared` as a link argument, then add `//#RunEnabled:false`. You could also move all the test directives into the `.s` file and delete the `.c` file. There are some other shared-object-only tests if you need something to reference. e.g. `dynamic-bss-only.c`. I'm not sure there are any others that are assembly-only, so that might be a first.

I think it might also be possible to remove the `Arch` line, since I think the assembly you've got is likely portable to other architectures as it doesn't actually use any assembly instructions.

mati865:
The closest assembly only test would be `wild/tests/sources/basic-comdat-1.s` (found using `rg "//#" wild/tests/sources/*.s`).

lqd:
Thanks! I've adapted the test to copy `basic-comdat.s`'s structure, which looked very much like exactly what was needed.

davidlattimore:
section.got
  wild Section missing
  ld OK

Looks like GNU ld is generating a GOT section on aarch64. You could add //#DiffIgnore:section.got

lqd:
Done. Hopefully this will pass CI now, but it doesn't like me enough to run automatically.

lqd:
(thanks mati :)

davidlattimore:
> Done. Hopefully this will pass CI now, but it doesn't like me enough to run automatically.

Now that you've had a PR merged, I think any subsequent PRs won't need approval for actions to run. 

Thanks for your contribution!

Issue #1086, opened 2025-09-19
perf-event-open-sys cannot be built on riscv64gc target


marxin:
The cargo build command fails on my RISC-V board with:
```
error[E0432]: unresolved import `crate::bindings`
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/perf-event-open-sys-4.0.0/src/functions.rs:15:5
    |
 15 | use crate::bindings;
    |     ^^^^^^^^^^^^^^^ no `bindings` in the root
    |
note: found an item that was configured out
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/perf-event-open-sys-4.0.0/src/lib.rs:183:9
    |
181 | #[cfg(target_arch = "aarch64")]
    |       ----------------------- the item is gated behind the `aarch64` feature
182 | #[path = "bindings_aarch64.rs"]
183 | pub mod bindings;
    |         ^^^^^^^^
note: found an item that was configured out
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/perf-event-open-sys-4.0.0/src/lib.rs:187:9
    |
185 | #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
    |          --------------------------------------------- the item is gated here
186 | #[path = "bindings_x86_64.rs"]
187 | pub mod bindings;
    |         ^^^^^^^^

error[E0432]: unresolved imports `crate::bindings`, `crate::bindings`
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/perf-event-open-sys-4.0.0/src/functions.rs:64:16
    |
 64 |     use crate::bindings::{self, perf_event_attr, perf_event_query_bpf};
    |                ^^^^^^^^   ^^^^ no `bindings` in the root
    |                |
    |                could not find `bindings` in the crate root
    |
    = help: consider importing this unresolved item through its public re-export instead:
            crate::functions::bindings
note: found an item that was configured out
   --> /home/ubuntu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/perf-event-open-sys-4.0.0/src/lib.rs:183:9
    |
181 | #[cfg(target_arch = "aarch64")]
    |       ----------------------- the item is gated behind the `aarch64` feature
182 | #[path = "bindings_aarch64.rs"]
183 | pub mod bindings;
    |         ^^^^^^^^
[...]

PR #1087, opened 2025-09-19
disable perf.rs on riscv64 target

marxin:
Fixes: #1086


marxin:
The reason it was not caught by CI is that we don't actually target `riscv64gc`, we just use `WILD_TEST_CROSS` for the integration tests where the host system (x86_86) targets `riscv64gc` and the tests are run under qemu.

davidlattimore (approved):

PR #1090, opened 2025-09-21
Bump MSVC to 1.90

marxin:


RossSmyth:
Nxipkgs doesn't even have a PR open for 1.90 yet cause there are other fires being put out right now (CMake 4, LLVM 21, and GCC 15)

marxin:
Sure, no problem. Let's keep the PR opened and use it once the Nix package is available.

mati865:
We are already halfway to 1.91. Was there any progress on the Nix front?

lapla-cogito:
https://nixpk.gs/pr-tracker.html?pr=451179

RossSmyth:
Unfortunately this cycle has been really slow due to the aforementioned fires and also dealing with the new lld behavior. If desired I can put up a PR that changes to use rust-overlay again instead temporarily.

davidlattimore:
Personally, I don't mind waiting a few more weeks, but others might be more keen

marxin:
Same for me, we don't have to hurry :)

lapla-cogito:
Finally, rustc 1.90.0 has landed in nixpkgs-unstable.: https://github.com/NixOS/nixpkgs/tree/nixpkgs-unstable/pkgs/development/compilers/rust

PR #1091, opened 2025-09-22
Use wild for build in CI

davidlattimore:


davidlattimore:
I think this will need to wait until after the 0.6.0 release, so that we have releases in gzip format rather than xz, otherwise I'd have to install xz-utils from the github action, which would be tricky without making the action specific to say ubuntu.

marxin:
I welcome that! Btw. could we please include the `wild-action` in the `README.md` under the Installation section?

davidlattimore:
> I welcome that! Btw. could we please include the `wild-action` in the `README.md` under the Installation section?

Done

PR #1093, opened 2025-09-22
Switch to handwritten release.yml

davidlattimore:
cargo-dist is not getting much maintenance. Doing our own thing gives us more flexibility.

The main loss of feature is that we now don't have an installer script.

We have switched from using xz compression to just plain gzip. This should make it easier to install on minimal systems that might not have xz-utils installed.

We also now link our binary with wild.


RoloEdits:
Sorry for the noise, but as far as I can tell, `cargo-dist` is seeing updates? I use `cargo-dist` for my own hobby stuff, but seeing as you believe there to be an issue, could you extrapolate on what you mean? I might switch off to something manually as well if the need calls for it.

davidlattimore:
It's possible that the situation has improved with cargo-dist. I was mostly thinking of a few months ago when the github 20.04 runners were being turned down and cargo-dist was still using them.

PR #1123, opened 2025-09-28
use depfile in integrationtests

### LINKED ISSUE #649: Stale test build dir causes test failure after rustc update
When I updated to rust 1.86, I noticed that the tests that use the cranelift backend were failing. After some investigation, I realised that this was because some of the input files had changed and the test framework wasn't aware, so it didn't rerun the link for GNU ld. Once I realised, the fix was very quick - just `rm -rf wild/tests/build`.

I was thinking about how best to prevent this from happening. One possibility would be to pass `--dependency-file=depfile` to GNU ld. That'll give us a list of all the files that it used during the link. When we run again and we're trying to decide whether to rerun the linker or not, we can parse this file, then stat all the input files that it lists and if any are newer than the existing output file, rerun the link.

It's worth noting that Wild doesn't yet support the `--dependency-file` flag, so we shouldn't pass it to wild. That's OK, since we always rerun wild every time we run the tests. It's only the other linkers that we cache the outputs of.

The output of `--dependency-file` is pretty simple, so should be no trouble to parse. It looks something like this:

```
hello-world: \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/Scrt1.o \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/crti.o \
  /usr/lib/gcc/x86_64-linux-gnu/11/crtbeginS.o \
  a.o \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libgcc_s.so.1 \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so \
  /lib/x86_64-linux-gnu/libc.so.6 \
  /usr/lib/x86_64-linux-gnu/libc_nonshared.a \
  /lib64/ld-linux-x86-64.so.2 \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libgcc_s.so.1 \
  /usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a \
  /usr/lib/gcc/x86_64-linux-gnu/11/crtendS.o \
  /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/crtn.o \
  /lib64/ld-linux-x86-64.so.2

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/Scrt1.o:

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/crti.o:

/usr/lib/gcc/x86_64-linux-gnu/11/crtbeginS.o:

a.o:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so:

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libgcc_s.so.1:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a:

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so:

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so:

/usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libc.so:

/lib/x86_64-linux-gnu/libc.so.6:

/usr/lib/x86_64-linux-gnu/libc_nonshared.a:

/lib64/ld-linux-x86-64.so.2:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc.a:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so:

/usr/lib/gcc/x86_64-linux-gnu/11/libgcc_s.so:
```

All the dependencies are listed twice. Once as a dependency of the binary ("hello-world" in this case) and then a second time presumably because this is intended as input to `make` that that makes it happy. I'd probably parse the dependencies of the binary and ignore the rest of the file.

YamasouA:
closes #649 

This PR try to create dependencyfile and parse the file.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1850,7 +1850,7 @@ impl LinkCommand {
         let depfile_path = output_path.with_extension("d");
         if !linker.is_wild() {
             match invocation_mode {
-                LinkerInvocationMode::Direct => {
+                LinkerInvocationMode::Direct | LinkerInvocationMode::Script => {
                     link_command
                         .command
                         .arg(format!("--dependency-file={}", depfile_path.display()));
@@ -1860,22 +1860,17 @@ impl LinkCommand {
                         .command
                         .arg(format!("-Wl,--dependency-file={}", depfile_path.display()));
                 }
-                LinkerInvocationMode::Script => {
-                    link_command
-                        .command
-                        .arg(format!("--dependency-file={}", depfile_path.display()));
-                }
             }
         }
 
-        let base_ok = !matches!(linker, Linker::Wild)
-            && cmd_file_is_current(output_path, &link_command.to_string());
+        let base_ok =
+            !linker.is_wild() && cmd_file_is_current(output_path, &link_command.to_string());
 
         // We allow skipping linking if all the object and otherwise tracked files
         // are unchanged and are older than our output file, but not if we're linking
         // with our linker, since we're always changing that. We also require that the
         // command we're going to run hasn't changed.
-        let can_skip = if matches!(linker, Linker::ThirdParty(_)) && depfile_path.is_file() {
+        let can_skip = if !linker.is_wild() && depfile_path.is_file() {
             match parse_ld_dependency_file(&depfile_path) {
                 Ok(deps) if !deps.is_empty() => {
                     base_ok


YamasouA:
check script
```
#!/usr/bin/env bash
set -euo pipefail

# Target to watch
LD_OUT="wild/tests/build/rust-integration.rs-llvm-dynamic-host.ld"
SRC="wild/tests/sources/rust-integration.rs"

# mtime (nanoseconds)
mtime_ns() { python3 - "$1" <<'PY'
import os,sys; print(os.stat(sys.argv[1]).st_mtime_ns)
PY
}

# 1st run: expect full link
cargo clean
cargo test --test integration_tests -- --nocapture
t1=$(mtime_ns "$LD_OUT"); echo "t1=$t1"

# 2nd run: no changes -> expect ld skip (mtime unchanged)
cargo test --test integration_tests -- --nocapture
t2=$(mtime_ns "$LD_OUT"); echo "t2=$t2"

if [[ "$t1" != "$t2" ]]; then
  echo "NG: mtime changed on 2nd run (= ld ran)"
  exit 1
else
  echo "OK: mtime unchanged on 2nd run (= ld skipped)"
fi

# 3rd run: touch only the relevant source -> expect relink (mtime changes)
sleep 1                       # cross the second boundary as a safety margin
touch "$SRC"
sleep 1                       # avoid running immediately

cargo test --test integration_tests -- --nocapture
t3=$(mtime_ns "$LD_OUT"); echo "t3=$t3"

if [[ "$t2" == "$t3" ]]; then
  echo "NG: mtime unchanged after touching source (= ld did not run)"
  exit 1
else
  echo "OK: mtime changed after touching source (= ld relinked)"
fi
```

wild/tests/integration_tests.rs
```diff
@@ -1798,14 +1847,48 @@ impl LinkCommand {
...
+                    link_command
+                        .command
+                        .arg(format!("-Wl,--dependency-file={}", depfile_path.display()));
+                }
+                LinkerInvocationMode::Script => {
+                    link_command
+                        .command
+                        .arg(format!("--dependency-file={}", depfile_path.display()));
+                }
+            }
```
lapla-cogito, line 1863:
```suggestion
            match invocation_mode {
                LinkerInvocationMode::Direct | LinkerInvocationMode::Script => {
                    link_command
                        .command
                        .arg(format!("--dependency-file={}", depfile_path.display()));
                }
                LinkerInvocationMode::Cc => {
                    link_command
                        .command
                        .arg(format!("-Wl,--dependency-file={}", depfile_path.display()));
                }
            }
```

wild/tests/integration_tests.rs
```diff
@@ -1798,14 +1847,48 @@ impl LinkCommand {
...
+        }
+
+        let base_ok = !matches!(linker, Linker::Wild)
+            && cmd_file_is_current(output_path, &link_command.to_string());
```
lapla-cogito:
```suggestion
        let base_ok =
            !linker.is_wild() && cmd_file_is_current(output_path, &link_command.to_string());
```

wild/tests/integration_tests.rs
```diff
@@ -1798,14 +1847,48 @@ impl LinkCommand {
...
+
         // We allow skipping linking if all the object and otherwise tracked files
         // are unchanged and are older than our output file, but not if we're linking
         // with our linker, since we're always changing that. We also require that the
         // command we're going to run hasn't changed.
-        let can_skip = !matches!(linker, Linker::Wild)
-            && is_newer(output_path, inputs.iter().map(|i| i.path.as_path()))
-            && is_newer(output_path, config.tracked_files.iter())
-            && cmd_file_is_current(output_path, &link_command.to_string());
+        let can_skip = if matches!(linker, Linker::ThirdParty(_)) && depfile_path.is_file() {
```
lapla-cogito: ditto
```suggestion
        let can_skip = if !linker.is_wild() && depfile_path.is_file() {
```

lapla-cogito (commented):
Thanks for working on this!

YamasouA:
@lapla-cogito 
Thank you for your quick review.
I fix your comment.

davidlattimore:
Thanks for doing this!

PR #1124, opened 2025-09-28
Use std's HashMap/HashSet instead of hashbrown.

daniel-levin:
This change doesn't add an awful lot of value but it does make the code a bit less surprising to read (you don't see hashbrown that often in new code since std adopted hashbrown's code), removes a third-party dependency from linker-diff, and it confines the usage of hashbrown to the one place where it's still needed. There are also clippy lints which are specific to std's collection types which we were missing before. You'll notice I force pushed [c9645f4](https://github.com/davidlattimore/wild/pull/1124/commits/c9645f482805d50d78de39486b2a55dc4be29ccf) after realizing this and fixing one such lint.


libwild/src/symbol_db.rs
```diff
@@ -851,7 +850,7 @@ fn process_alternatives(
     symbol_db: &AtomicSymbolDb,
     resolved: &[ResolvedGroup],
 ) {
-    for (first, alternatives) in replace(alternative_definitions, HashMap::new()) {
+    for (first, alternatives) in std::mem::take(alternative_definitions) {
```
daniel-levin, line 853: Here's the clippy lint that it automatically fixed.

mati865:
Hashbrown has some benefits over std, like newly implemented improvements: https://github.com/rust-lang/hashbrown/commit/64bd7db1d1b148594edfde112cdb6d6260e2cfc3 and better default hasher.
In my tests it gave small but reproducible improvements for Wild >1% and for linker-diff 3-6% of wall time.

I didn't benchmark your changes but I'd expect couple percentage points regression due to using SipHash.

> This change doesn't add an awful lot of value but it does make the code a bit less surprising to read (you don't see hashbrown that often in new code since std adopted hashbrown's code)

I disagree, it's much cleaner to use `hashbrown::HashMap::new()` rather than `std::collections::HashMap::with_hasher(foldhash::RandomState::default()`. We did that before to benefit from the faster hasher.

daniel-levin:
> `std::collections::HashMap::with_hasher(foldhash::RandomState::default()`

Could also write `foldhash::HashMap::default()`. And then change that to ahash/fxhash etc for trying out other hash functions. In superproprietaryland I got an unreasonable performance increase finding and replacing ahash with fxhash.

daniel-levin:
>Hashbrown has some benefits over std, like newly implemented improvements: https://github.com/rust-lang/hashbrown/commit/64bd7db1d1b148594edfde112cdb6d6260e2cfc3 and better default hasher.

I suppose the trade off is between clippy lints and hashbrown improvements not downstreamed yet (which AFAICT only work with a nightly compiler). Also, the hash function in use should probably be whatever benchmarks best, not what's used by default in std/hashbrown.

daniel-levin:
Eh, I don't really care too much. So I'm closing this in favor of #1125 which does actually prevent a needless allocation thanks to the clippy lint.

mati865:
> > `std::collections::HashMap::with_hasher(foldhash::RandomState::default()`
> 
> Could also write `foldhash::HashMap::default()`.

Which is the same as using hashbrown, except hashbrown gave me small performance improvements.

> And then change that to ahash/fxhash etc for trying out other hash functions. In superproprietaryland I got an unreasonable performance increase finding and replacing ahash with fxhash.

Hashbrown uses foldhash these days, which also is faster than ahash. If you find fxhash faster for Wild don't hesitate to create PR, so we can consider it.

> I suppose the trade off is between clippy lints and hashbrown improvements not downstreamed yet (which AFAICT only work with a nightly compiler)

This is the lint's issue, it should fire regardless if the structure is provided by std or not, and it'd be great if it was improved.

> Also, the hash function in use should probably be whatever benchmarks best, not what's used by default in std/hashbrown.

I have tried various hashers previously and foldhash was fastest. I first learned about it when hashbrown changed the default hasher.
I think hashbrown is trustworthy in performance regards to follow its choices, Amanieu does really great job there. That said, if you can show another hasher that is an improvement across the board for Wild, I'm open to the change.

mati865:
> Eh, I don't really care too much. So I'm closing this in favor of #1125 which does actually prevent a needless allocation thanks to the clippy lint.

As I said in https://github.com/davidlattimore/wild/pull/1125#issuecomment-3343699693 it does not prevent allocation.
Also in Rust Vector and HashMap with size == 0 do not allocate, so `HashMap::new()` (which is equivalent to `Default::default()`) did not create an allocation.

daniel-levin:
> I have tried various hashers previously and foldhash was fastest.

Indeed, experiment wins. There are a few crates I work on in proprietary-land where fxhash is consistently better.

PR #1125, opened 2025-09-28
Apply clippy lint not detected due to use of hashbrown

daniel-levin:
See #1124 for some context. This change applies a clippy lint that is relevant but not picked up on due to use of hashbrown instead of std's hash tables.


mati865:
The code is equivalent, but `take` indeed looks better.
```
#[stable(feature = "mem_take", since = "1.40.0")]
pub fn take<T: Default>(dest: &mut T) -> T {
    replace(dest, T::default())
}
```

PR #1126, opened 2025-09-29
chore: updated some cargo dependencies

dougg0k:


davidlattimore:
Thanks. Looks like you've reduced the minimal versions of some crates, can you verify that things still build with the new minimal versions? e.g. using cargo-minimal-versions

dougg0k:
`cargo build --all` works.

However `cargo minimal-versions build --workspace`, return errors from code that are not from wild, but from some package which I did not upgrade anything `pkg-config`.

davidlattimore:
Might need the --direct flag

dougg0k:
`cargo minimal-versions build --direct` seems to be working now.

dougg0k:
Perhaps, something like https://github.com/casey/just or https://github.com/sagiegurari/cargo-make would be useful to the project. As well as having certain libraries like minimal-versions as dev deps.

davidlattimore:
Looking over the change again, it looks like several crates are being changed so that their minimal version is now the latest version. We want Cargo.toml to list our minimal supported versions, since that can help when using libwild from other crates that might have more restrictive versioning requirements.

I think if you'd just like to update the versions of the crates that were using, a cargo update should be sufficient. We also have dependabot doing that, but it sometimes misses stuff, so there can be value in doing it manually from time to time. 

Sorry for not noticing this sooner, I was paying too much attention to the version decreases and didn't realise what the version increases were doing. 

We should add some docs, perhaps to Cargo.toml that describe how we do versioning of dependencies.

If the versions we've got in Cargo.toml at the moment don't build with minimal versions, then bumping those so that they do would definitely be good. 

I agree that we should automate checking of minimal versions, possiblity by adding something to CI that uses cargo-minimal-versions.

dougg0k:
No worries, feel free to close the PR.

PR #1129, opened 2025-09-29
symver: distinguish default version (@@) from non-default (@).

marxin:


davidlattimore (approved):

mati865:
FWIW with this change I only had to skip NetworkManager (doesn't have the fix yet) and Spidermonkey (build system expects ld.bfd output) and libs with x32 code. Everything else linked fine.

marxin:
@mati865 That are great news! I'm going to give it also a try building as many Gentoo packages as possible.

PR #1140, opened 2025-10-01
Fix a couple of crate minimum versions and document version policy

davidlattimore:
This fixes a couple of build failures when running cargo minimum-versions --direct.


mati865 (approved):

marxin:
What about setting a new CI job that will verify the `minimum-versions` policy?

davidlattimore:
> What about setting a new CI job that will verify the `minimum-versions` policy?

https://github.com/davidlattimore/wild/pull/1142

PR #1141, opened 2025-10-01
Enable more Mold's tests that already pass

mati865:


mati865:
I have no idea why there are differences between Arch and Ubuntu in regard to Mold tests, but it is what it is.

PR #1142, opened 2025-10-01
ci: Verify minimal versions

davidlattimore:


.github/workflows/ci.yml
```diff
@@ -93,6 +93,30 @@ jobs:
       - run: cargo build --profile ci --workspace --no-default-features
       - run: WILD_TEST_CROSS=$WILD_TEST_CROSS cargo test --profile ci --workspace
 
+  minimal-versions:
+    runs-on:
+      - ubuntu-24.04
+    container: 'ubuntu:24.04'
```
lapla-cogito: I'm not sure whether this job needs to be run inside a container. Isn't simply `runs-on: ubuntu-24.04` sufficient?

.github/workflows/ci.yml
```diff
@@ -93,6 +93,30 @@ jobs:
...
+    steps:
+      - run: apt-get update && apt-get -y install gcc g++ clang clang-format lld curl
+      - uses: actions/checkout@v4
```
lapla-cogito:
```suggestion
      - uses: actions/checkout@v5
```

davidlattimore: Thanks for the review. I've made both changes.

lapla-cogito (approved):

Issue #1144, opened 2025-10-02
Hot reloading with dioxus `subsecond` is broken with string merging


jakobhellermann:
`subsecond` is a crate/technique enabling hot reloading of rust code: https://docs.rs/subsecond/0.7.0-rc.0/subsecond/
I honestly have no idea how it [works](https://github.com/DioxusLabs/dioxus/blob/ad40f816073f91da67c0287a5512a5111e5a1617/packages/subsecond/subsecond/src/lib.rs#L498), so this issue might be on them for doing something wonky.

This is the setup:

```toml
[package]
name = "subsecond-testing"
version = "0.1.0"
edition = "2024"

[dependencies]
subsecond = { version = "0.7.0-rc.0" }
dioxus-devtools = { version = "0.7.0-rc.0" }
```

```rs
use std::{thread::sleep, time::Duration};

fn main() {
    dioxus_devtools::connect_subsecond();
    loop {
        subsecond::call(function);
        sleep(Duration::from_secs(1));
    }
}

fn function() {
    println!("Hello world!");
}
```

```
cargo install dioxus-cli@0.7.0-rc.0 --locked
dx serve --hot-patch
# builds the code, which works

# two linker invocations:
# Original output file: /home/jakob/.cargo/build/43/f778200ec67c69/x86_64-unknown-linux-gnu/desktop-dev/deps/subsecond_testing-27239bf71a8e452f
# Original output file: /home/jakob/dev/rust/subsecond-testing/target/x86_64-unknown-linux-gnu/desktop-dev/subsecond-testing

# now I edit the string in the function

# two more linker invocations
# Original output file: /home/jakob/.cargo/build/43/f778200ec67c69/x86_64-unknown-linux-gnu/desktop-dev/deps/subsecond_testing-27239bf71a8e452f
# Original output file: /home/jakob/dev/rust/subsecond-testing/target/dx/subsecond-testing/debug/linux/app/libsubsecond-testing-patch-1759396203728.so

# the second one breaks
```

```
wild: error: Failed copying from <epilogue> to output file
  Caused by:
    Internal error: Defined symbols should always be for a loaded section
```

```sh
#!/usr/bin/env bash
D=$(dirname $BASH_SOURCE)
if [ -z "$OUT" ]; then
  OUT=$D/bin${S}
fi
exec "$@" \
 \
  --hash-style=gnu \
  --build-id \
  --eh-frame-hdr \
  -m \
  elf_x86_64 \
  -shared \
[...]

PR #1147, opened 2025-10-03
Allow string-merge sections with `write_regular_object_dynamic_symbol_definition`

lapla-cogito:
When creating a dynamic symbol that refers to merged-string, `write_regular_object_dynamic_symbol_definition` was expecting only loaded sections and would cause a panic.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -3024,10 +3024,13 @@ fn write_regular_object_dynamic_symbol_definition(
     let sym = object.object.symbol(sym_index)?;
     let name = sym_def.name;
     if let Some(section_index) = object.object.symbol_section(sym, sym_index)? {
-        let SectionSlot::Loaded(section) = &object.sections[section_index.0] else {
-            bail!("Internal error: Defined symbols should always be for a loaded section");
+        let output_section_id = match &object.sections[section_index.0] {
+            SectionSlot::Loaded(section) => section.output_section_id(),
+            SectionSlot::MergeStrings(merge_section) => merge_section.part_id.output_section_id(),
+            _ => bail!(
+                "Internal error: Defined symbols should always be for a loaded or merge-strings section"
+            ),
         };
-        let output_section_id = section.output_section_id();
         let symbol_id = sym_def.symbol_id;
         let resolution = layout.local_symbol_resolution(symbol_id).with_context(|| {
             format!(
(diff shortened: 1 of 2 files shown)


lapla-cogito:
This issue appears to be related to #1144, but when I attempted to reproduce the steps described here locally, the build failed at [an unrelated point](https://github.com/davidlattimore/wild/blob/f17f3dc5a24f555d7146ec41722ebea1e806a1e1/libwild/src/file_kind.rs#L46), so I'm not completely certain this issue has been fully resolved (at least, I can confirm that the tests I added in this PR were previously causing panics but now successfully link). I examined the relevant files using `file` and `xxd` on my local machine but didn't find any obvious issues, so further investigation seems necessary to determine whether this problem has been completely resolved...

```
（▰╹◡╹）❯  PATH=.:$PATH dx serve --hot-patch
10:52:47 [dev] -----------------------------------------------------------------
                Serving your app: dio! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
               ---------------------------------------------------------------- 
10:52:48 [dev] Failed to generate fat binary: wild: error: Failed to parse archive `/home/lapla/zatsu/wild_playground/dio/target/x86_64-unknown-linux-gnu/desktop-dev/libdeps-e887ab2d.a @ #1/8`
  Caused by:
    Couldn't identify file type

10:52:48 [dev] Build failed: No such file or directory (os error 2) 
10:52:48 [dev] Verbose logging is now on 
╭──────────────────────────────────────────────────────────────────────────────── /:more ╮
│  App:     ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           Platform: Linux                     │
│  Bundle:  ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           App features: []                    │
│  Status:  Failed                                   Server at: no server address        │
╰────────────────────────────────────────────────────────────────────────────────────────╯
```

davidlattimore:
Regarding #1144. It looks like you're not getting as far into linking as @jakobhellermann. My guess is that this PR will fix the issue they were encountering. The issue you're hitting is probably a different one that Jakob doesn't get for some reason. From the error, it looks like the problematic entry in the archive has an empty name, so perhaps the other linkers ignore entries with an empty name. That's just a guess though.

jakobhellermann:
I can confirm that this fixes #1144 for me

lapla-cogito:
@jakobhellermann Thank you for confirming!

PR #1148, opened 2025-10-03
Wild 1143 - combined inline and @file options parsing and tests

### LINKED ISSUE #1143: Allow mixed use of @filename and inline command line options
# Current Behaviour

`wild @opts -o foo.exe`

will produce the following error:

`wild: error: Mixing of @{filename} and regular arguments isn't supported``

# Desired behaviour

command line options passed to wild can be a mixture of:
* "inline" options passed directly on the command line
* "file" options that are stored in a file and referenced by a command line option that is the filename prefixed with "@"
 
# Options processing
* command line options will be processed in the order they appear on the command line. A later option may override or nullify an earlier option, independently of the source of the option.
* when an option consists of multiple parts (e.g. `"-L", "/usr/lib"`), they must be contiguous and from the same source (command line, or the same file). 
    * It is not permitted to have the first option before a @file option and the later option(s) after the @file
    * A multiple part option cannot have the first option inline and a later option picked up from the file
    * A multiple part option cannot have the first option in a file (it would have to be the last one) and a later option picked up from subsequent inline options
* Options from a @file will be expanded and the resulting list of options processed in the order they appear in the file
* multiple @file options are permitted, in any order, interspersed with inline options provided other rules here are followed
* @file options within a file referenced by a @file option will be expanded as they are found, recursively. The above rules about sets of options spanning a file apply.

andrewdavidmackenzie:
Fixes #1143 by implementing parsing of combined inline and @file options on the command line, as described in #1143

### DIFF
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -53,8 +53,8 @@ glob = "0.3.2"
 perf-event = { workspace = true }
 
 [dev-dependencies]
-ar = "0.9.0"
-tempfile = "3.0.2"
+ar = { workspace = true }
+tempfile = { workspace = true }
 
 [features]
 # Support for running the linker as a subprocess.
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -54,6 +54,7 @@ perf-event = { workspace = true }
 
 [dev-dependencies]
 ar = "0.9.0"
+tempfile = "3.0.2"
 
 [features]
 # Support for running the linker as a subprocess.
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -830,12 +830,7 @@ impl ArgumentParser {
             let file_args = read_args_from_file(Path::new(path))?;
             let mut file_arg_iter = file_args.iter();
             while let Some(file_arg) = file_arg_iter.next() {
-                self.handle_argument(
-                    args,
-                    modifier_stack,
-                    file_arg,
-                    &mut file_arg_iter,
-                )?;
+                self.handle_argument(args, modifier_stack, file_arg, &mut file_arg_iter)?;
             }
         }
 
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -2347,6 +2347,15 @@ mod tests {
         "somewhere",
     ];
 
+    const FILE_OPTIONS: &[&str] = &[
+        "-pie",
+    ];
+
+    const INLINE_OPTIONS: &[&str] = &[
+        "-L",
+        "/lib",
+    ];
+
     fn write_options_to_file(file: &File, options: &[&str]) {
         let mut writer = BufWriter::new(file);
         for option in options {
@@ -2395,6 +2404,11 @@ mod tests {
         assert_eq!(args.rpath.as_deref(), Some("foo/:bar/:baz:somewhere"));
     }
 
+    fn inline_and_file_options_assertions(args: &Args) {
+        assert!(args.is_relocatable());
+        assert_contains(&args.lib_search_path, "/lib");
+    }
+
     #[test]
     fn test_parse_inline_only_options() {
         let args = super::parse(|| INPUT1.iter()).unwrap();
@@ -2415,47 +2429,41 @@ mod tests {
 
     #[test]
     fn test_parse_mixed_file_and_inline_options() {
-        // start with the usual INPUT1 set of command line options
-        let mut inline_options = INPUT1.to_vec();
-
-        // split the options set approximately in the middle
-        let file_options = inline_options.split_off(inline_options.len() / 2);
-
-        // Create a temporary file containing the same options (one per line) as INPUT1
+        // Create a temporary file containing some options
         let file = NamedTempFile::new().expect("Could not create temp file");
-        write_options_to_file(file.as_file(), &file_options);
+        write_options_to_file(file.as_file(), FILE_OPTIONS);
 
-        // pass the name of the file where options are, as an inline option "@filename"
+        // create an inline option referring to "@filename"
         let file_option = format!("@{}", file.path().to_str().unwrap());
+        // start with the set of inline options
+        let mut inline_options = INLINE_OPTIONS.to_vec();
+        // and extend with the "@filename" option
         inline_options.push(&file_option);
 
         // confirm that this works and the resulting set of options is correct
         let args = super::parse(|| inline_options.iter()).unwrap();
-        input1_assertions(&args);
+        inline_and_file_options_assertions(&args);
     }
 
     #[test]
     fn test_parse_overlapping_file_and_inline_options() {
-        // start with the usual INPUT1 set of command line options
-        let mut inline_options = INPUT1.to_vec();
-
-        // split the options set approximately in the middle
-        let file_options = inline_options.split_off(inline_options.len() / 2);
-
-        // Take the last option from `file_options` and repeat it in `inline_options`
-        inline_options.push(file_options.last().unwrap());
-
-        // Create a temporary file containing the same options (one per line) as INPUT1
+        // Create a set of file options that has a duplicate of an inline option
+        let mut file_options = FILE_OPTIONS.to_vec();
+        file_options.append(&mut INLINE_OPTIONS.to_vec());
+        // and save them to a file
         let file = NamedTempFile::new().expect("Could not create temp file");
         write_options_to_file(file.as_file(), &file_options);
 
         // pass the name of the file where options are, as an inline option "@filename"
         let file_option = format!("@{}", file.path().to_str().unwrap());
+        // start with the set of inline options
+        let mut inline_options = INLINE_OPTIONS.to_vec();
+        // and extend with the "@filename" option
         inline_options.push(&file_option);
 
         // confirm that this works and the resulting set of options is correct
         let args = super::parse(|| inline_options.iter()).unwrap();
-        input1_assertions(&args);
+        inline_and_file_options_assertions(&args);
     }
 
     #[test]
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -826,6 +826,7 @@ impl ArgumentParser {
         arg: &str,
         input: &mut I,
     ) -> Result<()> {
+        // TODO @lapla-cogito standardize the interface. @file doesn't use a leading hyphen.
         // Handle `@file`option (recursively) - merging in the options contained in the file
         if let Some(path) = arg.strip_prefix('@') {
             let file_args = read_args_from_file(Path::new(path))?;
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -16,7 +16,7 @@ use crate::alignment::Alignment;
 use crate::arch::Architecture;
 use crate::bail;
 use crate::ensure;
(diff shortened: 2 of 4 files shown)


libwild/src/args.rs
```diff
@@ -399,19 +399,22 @@ pub(crate) fn parse<F: Fn() -> I, S: AsRef<str>, I: Iterator<Item = S>>(input: F
     }
 
     let arg_parser = setup_argument_parser();
-    let mut arg_num = 0;
     while let Some(arg) = input.next() {
         let arg = arg.as_ref();
 
-        // Handle `@file`
-        // TODO: This is ad-hoc.
```
lapla-cogito, line 407: I wrote this TODO when I fundamentally rewrote the argument parser because the `@file` option differs from other options in its definition method. Since this hasn't changed, I believe this comment should remain.

libwild/Cargo.toml
```diff
@@ -54,6 +54,7 @@ perf-event = { workspace = true }
 
 [dev-dependencies]
 ar = "0.9.0"
+tempfile = "3.0.2"
```
lapla-cogito: I think it's fine to manage this at the workspace level.

libwild/src/args.rs
```diff
@@ -2386,6 +2398,83 @@ mod tests {
...
+        input1_assertions(&args);
+    }
+
+    #[test]
+    fn test_parse_mixed_file_and_inline_options() {
+        // start with the usual INPUT1 set of command line options
+        let mut inline_options = INPUT1.to_vec();
+
+        // split the options set approximately in the middle
+        let file_options = inline_options.split_off(inline_options.len() / 2);
```
lapla-cogito: Could this potentially split options in unexpected places? For example, if `INPUT1` is as follows: 
```
["-O", "1"]
```
splitting it into two parts results in `"-O"` and `1`, which individually would be invalid options.

lapla-cogito (changes requested):

libwild/src/args.rs
andrewdavidmackenzie: When reading it, I understood, or inferred, that the TODO was related to combining @file options with inline options, and that attempts to combine them would result in the error below.

Now, that parsing of combined @file and inline options is implemented in this PR (as I described in issue #1143) can you explain more what is the remaining TODO. Is it _how_ the @file options parsing is done?

libwild/src/args.rs
lapla-cogito: Considering that `INPUT1`'s contents may change due to future additions of options, I recommend minimizing the use of `INPUT1` in the additional tests for this change.

libwild/Cargo.toml
andrewdavidmackenzie: I considered that. Currently only used by this and linker-diff.
@davidlattimore  suggested here. 
But if you both agree, I can easily bump this up to workspace level.

libwild/src/args.rs
andrewdavidmackenzie: Yes, this makes assumptions that INPUTS1 is more than 2 elements. Seemed a reasonable assumption looking at the existing options test. 

I could force it to be at least 1, not 0.

If INPUT1 were just one option, nothing would work. So, not sure it's worth the effort to make this bullet-proof to radical changes in INPUT1?

libwild/src/args.rs
lapla-cogito: All options except `@file` follow a declarative interface.
https://github.com/davidlattimore/wild/blob/085840b14da842575b0179d19917a8f9c5b19526/libwild/src/args.rs#L1508-L1516

Since `@file` doesn't use a leading hyphen in its options, unifying this interface proved challenging, so I deferred it. Therefore, the remaining task is to standardize the interface. While this isn't a major issue due to the option's uniqueness, keeping it as a TODO comment remains useful I think.

edit: Indeed, I realize I didn't clearly specify what the task was that needed to be marked as a TODO, so comments that address this aspect would be greatly appreciated.

libwild/src/args.rs
lapla-cogito: Let me explain in more detail what I was trying to demonstrate with the previous example.

For instance, suppose `INPUT1` was something like `["-O", "1"]` (in fact, such split examples do exist in actual INPUT1 data). When processed by this logic using `split_off`, `inline_options` would contain only `-O`, while the temporary file would contain `1`. For explanation purposes, let's assume the file containing just `1` is named `foo` (In practice, this should be any suitable file under `/tmp`.).

In this test, `file_option`—specifically `@foo`—is concatenated after `inline_options`. The resulting `args` would then be `["-O", "@foo"]`. This seems undesirable when considering how invalid options might slip into `foo` (although in practice, `1` would actually be written to `foo` and `-O @foo` would be equivalent to `-O1`).

lapla-cogito: In other words, I believe the current behavior of `INPUT1` merely happens to work correctly for this specific case. Considering the possibility that future `INPUT1` contents might change and cause such unintended splitting, if we're going to adopt this broadly two-partition logic, we should create dedicated mock argument data specifically for this test.

andrewdavidmackenzie:
I've just tried running the external_tests locally (copying invocation from workflow) on main branch, and they fail their for me too. Do they usually pass for others?

libwild/src/args.rs
lapla-cogito: That said, I'm only considering this because even a file by itself could potentially be an invalid option—I'm not strongly opposed to it, so if you want to keep this logic and resolve this thread, that's perfectly fine.

mati865:
Yes, this is a regression.

andrewdavidmackenzie:
@mati865 can you clarify.

This most recent build of main in GH Actions passes
https://github.com/davidlattimore/wild/actions/runs/18217963477

Is it expected that this fails in local on HEAD of main?

libwild/src/args.rs
andrewdavidmackenzie: Basically, I was being lazy as I didn't want to hav eto become an expert on options and the set of assertions to confirm them, and piggy-backed of INPUT1 and the set of assertions in the original test_parse() test.

I guess I can hand craft some new sets of options and matching assertions to b eable to test this without depending on INPUT1, will just take a bit more effort.

libwild/src/args.rs
andrewdavidmackenzie: Is it worth our while to try and write an issue to address your points?

I find that TODOs in the code tend to get overlooked for a very long time, tending to infinity, and only ever get addressed by the original TODO writer when they eventually have spare time to scartch that itch.

mati865:
Depending on the distro some of the tests might fail but most of them passes. Response files tests work fine for me.

andrewdavidmackenzie:
Thanks, that helps me understand.

Do you know what it might depend on?

On main branch, I have 1 test out of 458 fail :-( 

```
failures:

---- external_tests::mold_tests::check_mold_tests_regression::mold_test_062__UP_external_test_suites_mold_test_arch_x86_64_large_bss_sh stdout ----
Error: Mold test `/home/andrew/workspace/wild/external_test_suites/mold/test/arch-x86_64-large-bss.sh` failed with status: exit status: 1
Output:
Testing arch-x86_64-large-bss ... + cat
+ cc -o out/test/x86_64/arch-x86_64-large-bss/a.o -c -xc - -mcmodel=large
+ cc -o out/test/x86_64/arch-x86_64-large-bss/b.o -c -xc - -mcmodel=large
+ cat
+ cc -B. -o out/test/x86_64/arch-x86_64-large-bss/exe out/test/x86_64/arch-x86_64-large-bss/a.o out/test/x86_64/arch-x86_64-large-bss/b.o
WARNING: wild: --plugin /usr/libexec/gcc/x86_64-linux-gnu/13/liblto_plugin.so is not yet supported
+ grep -E '^1 c0000000$'
+ out/test/x86_64/arch-x86_64-large-bss/exe
++ on_error 18
++ code=1
++ echo 'command failed: 18: grep -E '\''^1 c0000000$'\'''
command failed: 18: grep -E '^1 c0000000$'
++ trap - EXIT
++ exit 1



failures:
    external_tests::mold_tests::check_mold_tests_regression::mold_test_062__UP_external_test_suites_mold_test_arch_x86_64_large_bss_sh
```

mati865:
No idea, this one passes on my machine.

andrewdavidmackenzie:
It's weird, when I execute the exe in the output directory manually it always runs well - but when run from the test script it segfaults.

davidlattimore:
> It's weird, when I execute the exe in the output directory manually it always runs well - but when run from the test script it segfaults.

I'm able to reproduce that one failing. I'll have a look later and see if I can figure out what's going on.

libwild/src/args.rs
```diff
@@ -838,6 +825,15 @@ impl ArgumentParser {
...
         input: &mut I,
     ) -> Result<()> {
+        // Handle `@file`option (recursively) - merging in the options contained in the file
+        if let Some(path) = arg.strip_prefix('@') {
+            let file_args = read_args_from_file(Path::new(path))?;
+            let mut file_arg_iter = file_args.iter();
+            while let Some(file_arg) = file_arg_iter.next() {
+                self.handle_argument(args, modifier_stack, file_arg, &mut file_arg_iter)?;
+            }
+        }
```
davidlattimore, line 838: It looks odd to me that there's not a return here. Won't that mean that we'll do something else with `arg` even after we read the response file?

libwild/Cargo.toml
davidlattimore: Yep, ideally all our crate versions are selected by the workspace, otherwise it's too easy to get them out of sync. `ar` should really be a workspace dependency too.

libwild/src/args.rs
lapla-cogito: Even if that were the case, would there be any positive reason not to explicitly state those concerns? In this specific case, I would be the original creator of the TODO item. However, since there's some possibility that someone else might address this issue, I wanted to keep this statement intact (though the original phrasing was somewhat unclear about what needed to be done).

libwild/Cargo.toml
andrewdavidmackenzie: Done

libwild/src/args.rs
andrewdavidmackenzie: Fixed, now it matches the other branches

libwild/src/args.rs
andrewdavidmackenzie: I hav eadded a TODO based on your comments here where the @file option is no whandled, i.e.
in handle_argument().

If you would like different text for it, please tell me here what you would like it to so.

libwild/src/args.rs
andrewdavidmackenzie: Added specific sets of options for @file options and inline options and used them in the corresponding tests

andrewdavidmackenzie:
OK, so I have tried to address all the review comments - please re-review.

The external test CI workflow now passes. 
They also pass locally (on PopOS) now. The issue was resolved by updating to GCC14, and making sure GCC indeed invokes GCC-14.

Let me know if you'd like me to squash commit history, or you will do that when it's eventually merged.

davidlattimore:
Are you happy with the change now @lapla-cogito ?

lapla-cogito (approved):
Thanks! Look good.

PR #1151, opened 2025-10-04
ci(release): build for `(aarch|x86_|riscv)64(|gc)-unknown-linux-(musl|gnu)`

jarjk:
see title, maybe commit messages

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,7 @@ jobs:
       run: cargo build --release
 
     - name: Switch to linking with wild
+      if: matrix.build != 'zigbuild' # zigbuild ignores this config
       run: |
         mkdir -p "$HOME/.cargo"
         WILD="${PWD}/target/release/wild"


.github/workflows/release.yml
```diff
@@ -29,22 +35,27 @@ jobs:
...
+    - uses: mlugg/setup-zig@v2
 
     - name: Extract release notes
       shell: bash
       run: |
         awk "/# ${GITHUB_REF_NAME}/{flag=1; next} /^$/{flag=0} flag" CHANGELOG.md >REL.md
 
     - name: Build stage1
-      run: cargo build --release
+      run: cargo zigbuild --release --target ${{ matrix.target }}
```
davidlattimore: Do we need stage1 to be built with matrix.target or could it still just use the host target? It probably doesn't matter either way for now, but I was intending to build for riscv at some point and github doesn't AFAIK have riscv runners, so we'd probably need to build on x86_64.

.github/workflows/release.yml
```diff
@@ -29,22 +35,27 @@ jobs:
...
     - name: Switch to linking with wild
       run: |
-        mkdir .cargo
-        echo -e "[target.${{matrix.target}}]\nlinker = \"clang\"\nrustflags = [\"-Clink-arg=--ld-path=${PWD}/target/release/wild\"]\n" >>"$HOME/.cargo/config.toml"
+        mkdir -p "$HOME/.cargo"
+        echo -e "[target.${{matrix.target}}]\nlinker = \"clang\"\nrustflags = [\"-Clink-arg=--ld-path=${PWD}/target/${{matrix.target}}/release/wild\"]\n" >>"$HOME/.cargo/config.toml"
 
     - name: Build release binary
-      run: cargo build --profile dist --target ${{ matrix.target }}
+      run: cargo zigbuild --profile dist --target ${{ matrix.target }}
```
davidlattimore: The musl allocator is pretty slow. I'd suggest adding `--features mimalloc`, but only for the musl builds. I just benchmarked and a musl build with mimalloc was about 20% faster than a musl build without.

.github/workflows/release.yml
jarjk: they could be of course, didn't even occur to me. should I change it back, for the sake of speed?
and yes, there are no `riscv` github action runners according to the [runner images repo](https://github.com/actions/runner-images)

.github/workflows/release.yml
jarjk: impressive, I'll add that right away

.github/workflows/release.yml
```diff
@@ -47,15 +47,19 @@ jobs:
...
     - name: Switch to linking with wild
       run: |
         mkdir -p "$HOME/.cargo"
         echo -e "[target.${{matrix.target}}]\nlinker = \"clang\"\nrustflags = [\"-Clink-arg=--ld-path=${PWD}/target/${{matrix.target}}/release/wild\"]\n" >>"$HOME/.cargo/config.toml"
 
-    - name: Build release binary
+    - name: Build release binary for gnu-libc
+      if: contains(matrix.target, 'gnu')
       run: cargo zigbuild --profile dist --target ${{ matrix.target }}
+    - name: Build release binary for musl-libc
```
davidlattimore: Would it work to add features to the matrix? e.g. `--features "${{ matrix.features }}"`. Then we wouldn't need separate conditional steps

jarjk: definitely

.github/workflows/release.yml
jarjk: should I try adding `riscv64gc-unknown-linux-gnu` and `riscv64gc-unknown-linux-musl` as well?

davidlattimore: Sure, if you felt like giving that a go while you were at it, that'd be awesome :)

mati865:
Is there are real benefit for using zigbuild for Glibc target? It'll reduce targeted Glibc version from 2.35, down to 2.28 which might hurt performance.

jarjk:
> Is there are real benefit for using zigbuild for Glibc target? It'll reduce targeted Glibc version from 2.35, down to 2.28 which might hurt performance.

probably no, then, I'm changing it back

.github/workflows/release.yml
jarjk: I guess it's done: https://github.com/jarjk/wild/actions/runs/18250421023

jarjk:
one question though: how accurate is this `strings wild | grep "Linker: Wild"`?

I have no idea how these linkers and friends work, but I have concerns: 
in the wild repo, running `cargo zigbuild --target x86_64-unknown-linux-musl` on `aarch64-apple-darwin`, with no `wild` on path,
`strings target/x86_64-unknown-linux-musl/debug/wild | grep 'Linker: Wild'` gives me: `Linker: Wild version 0.6.0`.

although I do have a `~/.cargo/config.toml` with `wild` configured as linker for this target, but it shouldn't find it.

is `cargo-zigbuild` perhaps configured to use `wild` internally and so ignores the config, or how is this possible?

.github/workflows/release.yml
```diff
@@ -40,19 +51,22 @@ jobs:
...
 
     - name: Create tarballs
       run: |
         n="wild-linker-${{ github.ref_name }}-${{ matrix.target }}"
         mkdir "$n"
         cp "target/${{ matrix.target }}/dist/wild" "$n"
         cp README.md LICENSE* CHANGELOG.md "$n"
-        strip "$n/wild"
+        strings "$n/wild" | grep "Linker: Wild"
+        strip "$n/wild" || riscv64-linux-gnu-strip "$n/wild" || echo "::warning::couldn't strip binary"
```
davidlattimore: One alternative that might be simpler would be to add `strip = true` to `[profile.dist]` is `Cargo.toml`. Then the linker can just not write the symbol table.

.github/workflows/release.yml
```diff
@@ -40,19 +51,22 @@ jobs:
...
+      run: cargo ${{matrix.build || 'build' }} ${{matrix.args}} --profile dist --target ${{ matrix.target }}
 
     - name: Create tarballs
       run: |
         n="wild-linker-${{ github.ref_name }}-${{ matrix.target }}"
         mkdir "$n"
         cp "target/${{ matrix.target }}/dist/wild" "$n"
         cp README.md LICENSE* CHANGELOG.md "$n"
-        strip "$n/wild"
+        strings "$n/wild" | grep "Linker: Wild"
```
davidlattimore: That's very strange that you're seeing a binary linked by wild when using zigbuild. I just tried zigbuild on my laptop and it produced a binary that was linked with LLD. i.e. it ignored my `~/.cargo/config.toml` as you suggested it would.

davidlattimore: BTW, it should be possible to build the riscv binary without zigbuild if that makes things any simpler. See the riscv-build job in `ci.yml` for an example. I'm not too worried either way.

mati865:
The most reliable way is to run `readelf -p .comment <path>`, I think `strings` could have picked the string that is output by `--version` or something similar.

davidlattimore:
Ah, of course, that makes sense. Using `strings` to determine the linker is likely a reliable way for any binary that isn't the wild linker itself.

mati865:
Exactly 😉

.github/workflows/release.yml
mati865: If zigbuild doesn't use Wild at all, why should we use it for any build? Cross compiling for musl is easy as well.

jarjk:
sure, but there's no such thing for macos :(

jarjk:
but indeed, I ran `readelf` in an alpine container, and indeed it says `Linker: Homebrew LLD 20.1.8`

.github/workflows/release.yml
jarjk: well, cross-compiling should be easy, but I wouldn't say it is.
even with `apt install musl-tools`, the build process fails with: `error occurred in cc-rs: failed to find tool "x86_64-linux-musl-g++": No such file or directory`

I couldn't find any information on actually solving this, so went with `cargo-zigbuild`

jarjk: of course a `rust:alpine` image could build it

davidlattimore:
If it ends up too annoying trying to not use zigbuild, feel free to just use zigbuild. We can always try to get rid of it later.

jarjk:
yes, exactly.
it's especially true for the `riscv64gc`, as that (as far as I can tell) can't be linked with `wild` anyway
the others (amd64|aarch64 musl) could be fairly easily built in an alpine container, but that'd either result in loads of boilerplate, huge refactors, or every target running in containers (unnecessarily)

.github/workflows/release.yml
mati865: Ah, I didn't realise the demangler needs C++ compiler. With https://github.com/davidlattimore/wild/pull/1155 cross-compilation to Musl is easy once again, but I won't block on using Wild for Musl.

mati865:
Switching to Alpine would be an overkill, we can keep zigbuild for now. Alternatively, wait for https://github.com/davidlattimore/wild/pull/1155 to make Musl easy to cross-compile again.

jarjk:
excellent, #1155 would be great!
should I keep it on draft till then?

mati865:
I don't mind either way.

jarjk:
> I don't mind either way.

I'll wait then, that's pretty neat.

should the `riscv...-gnu` build use the `.cargo/config.toml.cross-riscv` or `zigbuild`?

also, how come that config isn't simply called `.cargo/config.toml`, would `riscv` be used elsewhere with a different config?

mati865:
Sorry, I have no experience with building for RISC-V from other systems. Arch Linux makes it easy by providing `riscv64-linux-gnu-gcc` package.
I'd be fine with whatever you and others think is the best or the most convenient for CI.

davidlattimore (commented):
I'm happy to just go with what you've got here and we can look to maybe move away from zigbuild later. If the action is succeeding when run, if you mark the PR as ready-for-review, I'll merge it.

jarjk:
it runs indeed: https://github.com/jarjk/wild/actions/runs/18258381845

davidlattimore (approved):

PR #1153, opened 2025-10-05
Ignore -C flag

daniel-levin:
For some reason, a spurious -C flag is passed to wild on illumos when using rustflags to delegate linking to wild. Ignoring it is necessary for wild to link Rust programs on illumos.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -260,7 +260,7 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "sort-common",
     "stats",
 ];
-const SILENTLY_IGNORED_SHORT_FLAGS: &[&str] = &["(", ")"];
+const SILENTLY_IGNORED_SHORT_FLAGS: &[&str] = &["(", ")", "C"];
 
 const IGNORED_FLAGS: &[&str] = &[
     "gdb-index",


davidlattimore:
On Linux at least, GNU ld rejects the `-C` flag. Does it accept it on Illumos? It'd be good to understand what's adding the flag and if it's supposed to do something.

daniel-levin:
Agreed. I don’t know where it comes from. The Sun linker accepts -C to mean
“demangle symbols”. Annoyingly I couldn’t get the Rust toolchain to pass
the aberrant -C to GNU ld. Perhaps we can selectively ignore the flag on
illumos while other, bigger blockers are resolved? In any event I’ll keep
it around and searchable as the “stupid illumos hack”.

daniel-levin:
Ah, found it. It turns out that both `gcc` and `clang` are aware of and prefer to use the Sun linker (`bin/ld` / `/usr/bin/ld`) on Illumos.

```
omnios@ip-10-0-12-140:~/exp$ sudo dtrace -qn 'proc:::exec-success /execname == "ld"/ {
    printf("LD called with: %s\n", curpsinfo->pr_psargs);
  }'
LD called with: /bin/ld -Y P,/usr/gcc/14/lib/amd64:/lib/amd64:/usr/lib/amd64 -L /usr/gcc/14/lib
^C

omnios@ip-10-0-12-140:~/exp$ sudo dtrace -qn 'proc:::exec-success /execname == "ld"/ {
    printf("LD called with: %s\n", curpsinfo->pr_psargs);
  }'
LD called with: /usr/bin/ld -C -e _start -o build/prog /usr/lib/amd64/crt1.o /usr/lib/amd64/crt
```

daniel-levin:
https://github.com/llvm/llvm-project/blob/4c8275470465528c469436d37b9aaf71d6598899/clang/lib/Driver/ToolChains/Solaris.cpp#L95

PR #1158, opened 2025-10-06
Added test config setting for skipping diffing.

daniel-levin:
As described in the title. Now, when a test fails because of only a diff, text is printed indicating as much, with a pointer to how to disable diffing. This is useful for packagers and bundlers who want to check that wild, as configured, works sanely. It is also invaluable for porting wild to other unixes, as one single test setting can be used to isolate truly broken tests from differences in opinion between linkers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,9 @@ Currently, the behavior for the following test options can be configured using t
 - `diff_ignore`: Adds extra rules to ignore certain diffs. This can be useful if you're developing
   on a system with an older version of GNU ld that doesn't perform certain optimisations.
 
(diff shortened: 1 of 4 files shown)


davidlattimore (commented):
Thanks for doing this. I'm wondering if we should go further and default to not diffing, then allow people to opt-in to diffing if they're running on a system where it's supported. If we do that, then we could potentially rename the option to something like `enable_diff`. My reason for opt-in rather than opt-out is that people packaging wild are often doing so in unusual and minimal environments and they're really just wanting to know if the way that they packaged it broke something.

Obviously for CI, we'd enable it and we'd probably recommend enabling it for people doing active development on wild.

daniel-levin:
Good point. I'll invert the flag.

davidlattimore (approved):

PR #1165, opened 2025-10-07
Support for handling BSD-style archives

lapla-cogito:
context: https://github.com/davidlattimore/wild/pull/1147#issuecomment-3364013506

While conducting the experiment mentioned in the above comment, I was working on Ubuntu 24.04.

I found that Dioxus was using BSD-style archive entries for some files (for reasons unknown). This is indicated by the `#1/8` part in the following error message. Since Wild currently doesn’t support handling BSD-style archives, this causes “No such file or directory” errors.

```
（▰╹◡╹）❯  PATH=.:$PATH dx serve --hot-patch
10:52:47 [dev] -----------------------------------------------------------------
                Serving your app: dio! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
               ---------------------------------------------------------------- 
10:52:48 [dev] Failed to generate fat binary: wild: error: Failed to parse archive `/home/lapla/zatsu/wild_playground/dio/target/x86_64-unknown-linux-gnu/desktop-dev/libdeps-e887ab2d.a @ #1/8`
  Caused by:
    Couldn't identify file type

10:52:48 [dev] Build failed: No such file or directory (os error 2) 
10:52:48 [dev] Verbose logging is now on 
╭──────────────────────────────────────────────────────────────────────────────── /:more ╮
│  App:     ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           Platform: Linux                     │
│  Bundle:  ━━━━━━━━━━━━━━━━━━━━━━━━━━  ❌           App features: []                    │
│  Status:  Failed                                   Server at: no server address        │
╰────────────────────────────────────────────────────────────────────────────────────────╯
```

Although BSD are not currently supported, the fact that this behavior occurred on Ubuntu (even if it’s a somewhat niche case) — combined with the [recent discussions](https://wild.zulipchat.com/#narrow/channel/508469-general/topic/.28future.29.20FreeBSD.20support.3F/with/542491047) around BSD support — suggests that adding support for BSD-style archive entries here would be worthwhile.

### DIFF
--- a/wild/Cargo.toml
+++ b/wild/Cargo.toml
@@ -22,6 +22,7 @@ mimalloc = { workspace = true, optional = true }
 dhat = { workspace = true, optional = true }
 
 [dev-dependencies]
+ar = { workspace = true }
 fd-lock = { workspace = true }
 itertools = { workspace = true }
 linker-diff = { path = "../linker-diff" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1697,6 +1697,7 @@ dependencies = [
 name = "wild-linker"
 version = "0.6.0"
 dependencies = [
+ "ar",
  "dhat",
  "fd-lock",
  "itertools",
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -46,6 +46,9 @@ pub(crate) struct ArchiveContent<'data> {
 
     /// The offset in the archive at which the data is from.
     pub(crate) data_offset: usize,
+
+    /// For BSD-style extended filenames, contains the filename data
+    pub(crate) bsd_filename: Option<&'data [u8]>,
 }
 
 pub(crate) struct ThinEntry<'data> {
@@ -81,6 +84,7 @@ enum IdentifierKind {
     FileReference, // Files in thin archives
     Filenames,
     SymbolTable,
+    BsdExtended, // BSD-style extended filename (#1/len)
 }
 
 impl<'data> ArchiveIterator<'data> {
@@ -128,6 +132,7 @@ impl<'data> ArchiveIterator<'data> {
         let ident_kind = match ident {
             "/" => IdentifierKind::SymbolTable,
             "//" => IdentifierKind::Filenames,
+            _ if ident.starts_with("#1/") => IdentifierKind::BsdExtended,
             _ => match self.is_thin {
                 false => IdentifierKind::InlineContent,
                 true => IdentifierKind::FileReference,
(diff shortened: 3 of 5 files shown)


lapla-cogito:
After applying this patch, I confirmed that Dioxus applications could update correctly via hot reload using the following code.

```rust
use std::{thread::sleep, time::Duration};

fn main() {
    dioxus_devtools::connect_subsecond();
    loop {
        subsecond::call(function);
        sleep(Duration::from_secs(1));
    }
}

fn function() {
    println!("Hello world!");
}
```

```toml
[package]
name = "dio"
version = "0.1.0"
edition = "2024"

[dependencies]
subsecond = { version = "0.7.0-rc.0" }
dioxus-devtools = { version = "0.7.0-rc.0" }
```

```
（▰╹◡╹）❯  PATH=.:$PATH dx serve --hot-patch
11:40:00 [dev] -----------------------------------------------------------------
                Serving your app: dio! 🚀
                • Press `ctrl+c` to exit the server
                • Press `r` to rebuild the app
                • Press `p` to toggle automatic rebuilds
                • Press `v` to toggle verbose logging
                • Press `/` for more commands and shortcuts
               ----------------------------------------------------------------
11:40:09 [dev] Build completed successfully in 9045ms, launching app! 💫
11:40:09 [linux] Hello world!
11:40:10 [linux] Hello world!
11:40:11 [linux] Hello world!
11:40:12 [linux] Hello world!
11:40:13 [linux] Hello world!
11:40:14 [linux] Hello world!
11:40:15 [linux] Hello world!
11:40:16 [linux] Hello world!
11:40:17 [linux] Hello world!
╭──────────────────────────────────────────────────────────────────────────────── /:more ╮
│  App:     ━━━━━━━━━━━━━━━━━━━━━━━━━━  🎉 9.1s      Platform: Linux                     │
│  Bundle:  ━━━━━━━━━━━━━━━━━━━━━━━━━━  🎉 0.0s      App features: []                    │
│  Status:  Serving dio 🚀 9.1s                      Server at: no server address        │
╰────────────────────────────────────────────────────────────────────────────────────────╯
```

lapla-cogito:
One issue I'm facing is testing it—I'm unsure how to actually create BSD-style archive entries on Linux, so I'm struggling to figure out the best way to implement this verification.

davidlattimore:
Nice! Thanks for figuring out that this was the issue and fixing it.

Regarding testing, one option would be to use the `ar` crate, which I think we might already have a dev dependency on. It can read and write the various different archive formats. As a starting point, here's some code that I just happened to write the other day that uses it to read an archive, process the files within, then write a new archive back out again.

```rust
fn process_archive(input_bytes: &[u8]) -> Result<Vec<u8>> {
    let identifiers = get_archive_identifiers(input_bytes)?;
    let mut archive = ar::Archive::new(input_bytes);
    let mut out_bytes = Vec::new();
    let mut builder = ar::GnuBuilder::new(&mut out_bytes, identifiers);

    while let Some(entry) = archive.next_entry() {
        let mut entry = entry?;
        let mut entry_bytes_in = Vec::new();
        entry.read_to_end(&mut entry_bytes_in)?;

        let entry_bytes_out =
            process_object_bytes(&entry_bytes_in).context("Failed to process archive entry")?;

        let mut header = entry.header().clone();
        header.set_size(entry_bytes_out.len() as u64);

        builder.append(&header, entry_bytes_out.as_slice())?;
    }

    Ok(out_bytes)
}

fn get_archive_identifiers(input_bytes: &[u8]) -> Result<Vec<Vec<u8>>> {
    let mut archive = ar::Archive::new(input_bytes);
    let mut identifiers = Vec::new();

    while let Some(entry) = archive.next_entry() {
        let entry = entry?;
        identifiers.push(entry.header().identifier().to_owned());
    }

    Ok(identifiers)
}
```

It'd need some adapting, since I was both reading and writing GNU style archives. You can just delete the call to `process_object_bytes`, since you don't need to do anything with the entries.

Anyway, if you put code like that into `integration_test.rs`, then you could add a new directive `//#BsdArchive:file.c` that could first create a regular archive, then convert it to a BSD archive. I guess you could skip the regular archive step if that was easier.

lapla-cogito:
Thanks! I didn't realize `ar` was so useful :)

wild/Cargo.toml
```diff
@@ -22,6 +22,7 @@ mimalloc = { workspace = true, optional = true }
 dhat = { workspace = true, optional = true }
 
 [dev-dependencies]
+ar = { workspace=true }
```
davidlattimore: nit: Could you add spaces around the `=` for consistency with the lines below?

davidlattimore (approved):

wild/Cargo.toml
lapla-cogito: Done

PR #1170, opened 2025-10-08
Add wild-test-files as a submodule.

daniel-levin:
As discussed on Zulip - a submodule of static binaries. Intended to be orthogonal to Wild and Wild's primary testing mechanisms. This is an experiment to add test coverage to non-Linux platforms. This change also _excludes_ the `selftest` Rust crate in wild-test-files because that way it can still be run from within the `static` directory, and the enclosing workspace need not be aware of `selftest`, which would necessitate checking out the submodule before running any Cargo commands.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "external_test_suites/mold"]
 	path = external_test_suites/mold
 	url = https://github.com/rui314/mold
+[submodule "wild/tests/bins"]
+	path = wild/tests/bins
+	url = git@github.com:davidlattimore/wild-test-files.git


davidlattimore:
My only slight concern is that the name "static" might be confusing, since it's used for so many other things, including in linkers - e.g. static linking. Perhaps "test-bins" or "test-files"? I also wonder if we should put it in a subdirectory. Probably under `wild/tests` so that it's with the rest of the integration testing stuff. I guess if it's there, then it's clear it's for testing, which does open up other naming possibilities. e.g. `wild/tests/bins` or `wild/tests/binaries`. "static" is perhaps less of a problem in that location too, so `wild/tests/static-bins` or even without the `-bins` suffix might be OK.

daniel-levin:
Ah yes, you make a good point. I'm going to slide them under `wild/test/bins`.

marxin:
@daniel-levin Please fix the submodule URL so that `https` is used rather than `git` protocol.

daniel-levin:
@marxin https://github.com/davidlattimore/wild/pull/1174

PR #1171, opened 2025-10-08
Add instructions for using Wild with Rust on Illumos.

daniel-levin:
The actual, correct resolution to #1153. It turns out that Clang is [very picky](https://github.com/llvm/llvm-project/blob/4c8275470465528c469436d37b9aaf71d6598899/clang/lib/Driver/ToolChains/Solaris.cpp#L95) and platform-dependent about delegating to arbitrary linkers. Mercifully, Clang _will_ still provide GNU-ld compatible arguments to Wild. As such there is no need to understand Sun ld's argument parser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,23 @@ See [nix/nix.md](nix/nix.md)
 If you'd like to use Wild as your default linker for building Rust code, you can put the following
 in `~/.cargo/config.toml`.
 
+On Linux:
 ```toml
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
 rustflags = ["-C", "link-arg=--ld-path=wild"]
+```
+
+On Illumos:
+```
+[target.x86_64-unknown-illumos]
+# Absolute path to clang - on OmniOS this is likely something like /opt/ooce/bin/clang.
+linker = "/usr/bin/clang"
+
+rustflags = [
+    # Will silently delegate to GNU ld or Sun ld unless the absolute path to Wild is provided.
+    "-C", "link-arg=-fuse-ld=/absolute/path/to/wild"
+]
 ```
 
 ## Using wild in CI


daniel-levin:
For those interested in this tooth-pulling exercise, it's worth noting that Clang and GCC's treatment of `-fuse-ld` is even more different and platform-specific than the comments in `integration_tests.rs` suggest. Concretely, On Illumos, `gcc` (`gcc (OmniOS 151054/14.2.0-il-1) 14.2.0 `) will joyfully but erroneously pass Sun ld flags to `lld`, `gold`, and even GNU `bfd` (go figure) if you use the `-fuse-ld` argument.

|Driver|Argument|Result|
|:-|:-:|-:|
|clang | None | Uses Sun ld with valid and correct arguments |
|clang|`-fuse-ld=/absolute/path/wild`|Uses Wild at the absolute path, with GNU ld-compatible arguments :partying_face:   (but one aberrant `-C`)|
|clang|`-fuse-ld=wild`|Fails, moaning about invalid linker name |
|clang|`-fuse-ld=lld`|Fails, moaning about invalid linker name despite it being LLVM's linker |
|clang | `-fuse-ld=ld` | Uses GNU ld |
|clang | `-fuse-ld=gld` | Uses GNU ld; not gold |
| gcc | None | Uses Sun ld with valid and correct arguments, but that are different from what Clang passes to Sun ld ceteris paribus  |
| gcc | `-fuse-ld=ld` | Uses GNU ld with Sun ld arguments (somehow passes) |
| gcc | `-fuse-ld=lld` | Uses lld with Sun ld arguments (explodes) |
| gcc | `-fuse-ld=wild` | Refuses to recognize argument; explodes|
| gcc |  `-fuse-ld=/abs/path/to/wild/` | `gcc: error: unrecognized command-line option '-fuse-ld=/home/omnios/.cargo/bin/wild'`|

PR #1173, opened 2025-10-09
Add Illumos-specific defaults.

daniel-levin:
1. Ignore the -C flag Clang passes only on Illumos
2. Provide a default argument to --dynamic-linker

These changes are sufficient for Wild to link ripgrep (and a few small programs I tried too)

I don't really like the _look_ of this platform-specific code but I think it will suffice for now.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -260,7 +260,13 @@ const SILENTLY_IGNORED_FLAGS: &[&str] = &[
     "sort-common",
     "stats",
 ];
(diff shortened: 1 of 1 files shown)


mati865:
Won't that also set interpreter for shared objects?

mati865:
> Won't that also set interpreter for shared objects?

Yeah, that would have done so. It won't be the case after https://github.com/davidlattimore/wild/pull/1178

PR #1175, opened 2025-10-09
check for overflow on 32bit relocations

### LINKED ISSUE #1149: Wild doesn't report error if a 32 bit relocation overflows
To test for this, you'll likely need two C files, each defining a large uninitialised array and referencing the large array defined by the other. Provided the code is compiled with default flags, you should be able to get a linker error.

Once you've got a test that shows GNU ld producing an error and Wild not, a good place to start looking at a fix is `write_relocation_to_buffer`. There's already some range checking there that should be able to be reused. Most likely it's just x86_64 that needs fixing.

karolzwolak:
Fixes #1149

### DIFF
--- a/wild/tests/sources/relocation-overflow.c
+++ b/wild/tests/sources/relocation-overflow.c
@@ -15,8 +15,8 @@ asm(".section .data\n"
     ".text\n");
 
 extern int reloc_a_to_b;
-void *keep_a_ref(void) { return &reloc_a_to_b; }
-extern void *keep_b_ref(void);
+void* keep_a_ref(void) { return &reloc_a_to_b; }
+extern void* keep_b_ref(void);
 
 void _start() {
   keep_a_ref();
--- a/wild/tests/sources/relocation-overflow-b.c
+++ b/wild/tests/sources/relocation-overflow-b.c
@@ -7,4 +7,4 @@ asm(".section .data\n"
     ".text\n");
 
 extern int reloc_b_to_a;
-void *keep_b_ref(void) { return &reloc_b_to_a; }
+void* keep_b_ref(void) { return &reloc_b_to_a; }


karolzwolak:
Don't mind the additional .gitignore changes and .c files and test.sh.
I couldn't reproduce the error in the actual test infrastructure so I temporarily included in in the PR so someone can tell me how to port it to the test infra.

@davidlattimore what test attributes should I add/remove? The test unexpectedly passes when run via cargo. I didn't see any docs about making new tests. However when compiling the same files (minus the _start shenanigan) using test.sh (assumes PWD is project root and wild is compiled) -- everything works as expected, so the linker reports the error.

wild/tests/sources/relocation-overflow.c
```diff
@@ -0,0 +1,20 @@
...
+    ".global reloc_a_to_b\n"
+    "reloc_a_to_b:\n"
+    "    .long big_b - .\n"  //  R_X86_64_PC32 relocation
+    ".text\n");
+
+extern int reloc_a_to_b;
+void *keep_a_ref(void) { return &reloc_a_to_b; }
+
+void _start() {}
+int main() {}
```
davidlattimore: `main` isn't needed

wild/tests/sources/relocation-overflow.c
```diff
@@ -0,0 +1,20 @@
...
+asm(".section .data\n"
+    ".global reloc_a_to_b\n"
+    "reloc_a_to_b:\n"
+    "    .long big_b - .\n"  //  R_X86_64_PC32 relocation
+    ".text\n");
+
+extern int reloc_a_to_b;
+void *keep_a_ref(void) { return &reloc_a_to_b; }
+
+void _start() {}
```
davidlattimore: I'm pretty sure that the reason the test isn't working is because the functions above are getting garbage collected. Wild GCs by default. If you call the functions from here, then that should fix it.

linker-utils/src/elf.rs
```diff
@@ -1197,6 +1197,17 @@ impl AllowedRange {
     pub const fn no_check() -> Self {
         Self::new(i64::MIN, i64::MAX)
     }
+
+    #[must_use]
+    pub const fn from_byte_number(n_bytes: usize) -> Self {
+        if n_bytes == 0 || n_bytes > 8 {
```
davidlattimore: This `>` could be a `>=` right? i.e. if we've got an 8 byte relocation, we don't need a check.

karolzwolak: The end result should be the same, but I realized more glaring issue -- this code assumes each relocation is a signed integer. The match should also return whether it should be signed or unsigned.

davidlattimore: Does it matter? I'm not sure it does. We do our computation using u64s with wrapping_add, wrapping_sub etc, so it's possible that it might work fine as-is anyway. I'm not 100% sure that it doesn't matter, but perhaps see if you can come up with a concrete example where it breaks?

wild/tests/sources/relocation-overflow.c
karolzwolak: Yes indeed, thank you!

linker-utils/src/elf.rs
karolzwolak: Hm okay I think you're right. I was thinking about 32 bit signed vs unsigned relocations, but I don't even see the distinction in the match. This function was computing wrong values though because my silly mistake with shift op precedence, which I fixed and also added tests for.

karolzwolak:
CI on tumbleweed seems to be failing because of formatting -- maybe it has different version of the tool?
```
 thread 'tidy::check_sources_format' (6386) panicked at wild/tests/integration_tests.rs:2856:13:
clang-format check failed:
/__w/wild/wild/wild/tests/sources/relocation-overflow-b.c:10:5: error: code should be clang-formatted [-Wclang-format-violations]
void *keep_b_ref(void) { return &reloc_b_to_a; }
```

karolzwolak:
I created #1183 to help debug such weird cases in CI.

karolzwolak:
I temporarily added commit to check clang version to test if the version in CI is the same.

EDIT: the version on tumbleweed is `21.1.2`, but on other jobs it varies from `14.x` to `20.x`.

PR #1178, opened 2025-10-09
Only emit `PT_INTERP` for shared executables

mati865:
Looks to be already well covered by the tests.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -613,6 +613,10 @@ impl OutputKind {
         matches!(self, OutputKind::SharedObject)
     }
 
+    pub(crate) fn is_dynamic_executable(self) -> bool {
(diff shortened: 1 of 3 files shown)


libwild/src/args.rs
```diff
@@ -1549,7 +1553,6 @@ fn setup_argument_parser() -> ArgumentParser {
         .long("dynamic-linker")
         .help("Set dynamic linker path")
         .execute(|args, _modifier_stack, value| {
-            args.is_dynamic_executable.store(true, Ordering::Relaxed);
```
mati865, line 1552: FWIW this was introduced in https://github.com/davidlattimore/wild/commit/53d7ee001025361e49bc6588478a28ae18140562 but ld.bfd doesn't seem to care about it when deciding output kind, as found by string merging test where ld.bfd generates static library despite `--dynamic-linker` argument.

PR #1187, opened 2025-10-11
Linker script: support `VERSION` command

lapla-cogito:
As noted in [this discussion](https://wild.zulipchat.com/#narrow/channel/508469-general/topic/Gentoo.20testing.20with.20Wild/with/543727426), `emerge mariadb-connector-c` currently fails on Gentoo, producing the following error message:
```
wild: error: Failed to parse linker script `/var/tmp/portage/dev-db/mariadb-connector-c-3.4.5-r1/work/mariadb-connector-c-3.4.5-src_build-abi_x86_64.amd64/libmariadb/mariadbclient.def`:
parse error at line 1, column 9
  |
1 | VERSION {
  |         ^
```

This occurs because Wild currently lacks the capability to set symbol visibility in linker scripts in the same way as version scripts.

The PR addresses this issue by using the parser for version scripts to also handle `VERSION` commands in linker scripts, allowing them to be parsed and applied. While this is presumably a rare case, builds that involve both `VERSION` commands and version scripts now prioritize the later-defined definitions.

### DIFF
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -60,6 +60,7 @@ pub(crate) enum Command<'a> {
     Ignored,
     Sections(Sections<'a>),
     Entry(&'a [u8]),
+    Version(&'a [u8]),
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -126,6 +127,13 @@ impl<'data> LinkerScript<'data> {
         foreach_input(&self.commands, starting_modifiers, &mut cb)?;
         Ok(())
     }
+
+    pub(crate) fn get_version_script_content(&self) -> Option<&'data [u8]> {
+        self.commands.iter().find_map(|cmd| match cmd {
+            Command::Version(content) => Some(*content),
+            _ => None,
+        })
+    }
 }
 
 fn parse_token<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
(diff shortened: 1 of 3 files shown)


libwild/src/linker_script.rs
```diff
@@ -200,6 +221,37 @@ fn parse_entry<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]>
     Ok(symbol_name)
 }
 
+fn parse_version<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
```
davidlattimore, line 212: I'd expect that with a bit of refactoring, we could actually parse the version script here, rather than getting a slice that we'll parse later. If we did that, then errors in the version-script part could more easily be reported with correct line numbers. That said, I'm fine with leaving that until later, especially since I'm not sure how much refactoring would be required.

libwild/src/linker_script.rs
```diff
@@ -159,6 +167,18 @@ fn parse_command<'input>(input: &mut &'input BStr) -> winnow::Result<Command<'in
 
     skip_comments_and_whitespace(input)?;
 
+    // Ignore symbol alias definitions like `"symbol@version" = symbol;`
```
davidlattimore: Is it safe to ignore this or might it result in a binary that does the wrong thing? I think that for the most part, we'd rather fail to link than produce an incorrect binary.

lapla-cogito: Understood. By reverting this change, linking will now fail if symbol aliases are defined in the linker script.

libwild/src/linker_script.rs
lapla-cogito: Yes, I also believe we could implement a more appropriate integration between version scripts and linker scripts, but after reviewing several proposed refactorings in some existing issues, I've determined that waiting to implement them would result in fewer rework issues.

marxin:
Great @lapla-cogito, the PR fixes building of `mysql-connector-c` Gentoo package, good job.

PR #1189, opened 2025-10-11
Delete ValueFlags::ADDRESS

davidlattimore:
Turned out that it was able to be computed from other bits.

### DIFF
--- a/libwild/src/value_flags.rs
+++ b/libwild/src/value_flags.rs
@@ -29,62 +29,59 @@ bitflags! {
     /// encounter to it.
     #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
     pub(crate) struct ValueFlags: u16 {
-        /// Something with an address. e.g. a regular symbol, a section etc.
-        const ADDRESS = 1 << 0;
-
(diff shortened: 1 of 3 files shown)


libwild/src/value_flags.rs
```diff
@@ -29,8 +29,7 @@ bitflags! {
     /// encounter to it.
     #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
     pub(crate) struct ValueFlags: u16 {
-        /// Something with an address. e.g. a regular symbol, a section etc.
-        const ADDRESS = 1 << 0;
+        // Bit 0 is currently used.
```
marxin: What about using a Copilot for re-numbering :P ?

davidlattimore: I don't have copilot set up in my IDE. I did try briefly at one point, but it wanted me to sign in, which due to my IDE running in a sandbox was possibly going to be tricky, so I gave up. But I renumbered them anyway.

marxin: Thanks for the change.

PR #1197, opened 2025-10-12
Support -m elf_x86_64_sol2

daniel-levin:
Currently, we have a somewhat janky workaround for getting Wild to work on Illumos: https://github.com/davidlattimore/wild/pull/1173

I've opened a PR in Clang to bring the Solaris (Illumos) driver up to parity with the others: https://github.com/llvm/llvm-project/pull/163000. Once this lands, it will be possible to use the `--ld-path` argument identically to other platforms. This benefits all linkers - including lld, Gold, Mold and Wild. That is, Clang will start to supply GNU ld-compatible flags if an unknown, non-Solaris Link Editor is supplied.

Of course, we need to then support the "emulation" `elf_x86_64_sol2`. Testing Wild with this change, and Clang with my change above, I have been able to link a couple of Rust programs, including ripgrep. Excitingly, it also links LLVM itself. Finally, I confirmed with DTrace that the arguments were expected.

```
❯ readelf -p .comment $(which rg)

String dump of section '.comment':
  [     0]  GCC: (OmniOS 151054/14.2.0-il-1) 14.2.0
  [    29]  @(#)illumos  May 2025
  [    3f]  rustc version 1.90.0 (1159e78c4 2025-09-14)
  [    6b]  Linker: Wild version 0.6.0
```

```
❯ readelf -p .comment /home/omnios/llvm-project/build/bin/clang

String dump of section '.comment':
  [     0]  GCC: (OmniOS 151054/14.2.0-il-1) 14.2.0
  [    28]  OmniOS/151054 clang version 20.1.8
  [    4c]  @(#)illumos  May 2025
  [    62]  Linker: Wild version 0.6.0
```

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -306,9 +306,6 @@ impl Default for Args {
             inputs: Vec::new(),
             output: Arc::from(Path::new("a.out")),
             is_dynamic_executable: AtomicBool::new(false),
-            #[cfg(target_os = "illumos")]
-            dynamic_linker: Some(Path::new("/lib/amd64/ld.so.1").into()),
-            #[cfg(not(target_os = "illumos"))]
             dynamic_linker: None,
             output_kind: None,
             time_phase_options: None,
@@ -1253,7 +1250,18 @@ fn setup_argument_parser() -> ArgumentParser {
             "elf_x86_64",
             "x86-64 ELF target",
             |args, _modifier_stack, _value| {
-                args.arch = Architecture::from_str("elf_x86_64")?;
+                args.arch = Architecture::X86_64;
+                Ok(())
+            },
+        )
+        .sub_option(
+            "elf_x86_64_sol2",
+            "x86-64 ELF target (Solaris)",
+            |args, _modifier_stack, _value| {
+                if args.dynamic_linker.is_none() {
+                    args.dynamic_linker = Some(Path::new("/lib/amd64/ld.so.1").into());
+                }
+                args.arch = Architecture::X86_64;
                 Ok(())
             },
         )
@@ -1261,7 +1269,15 @@ fn setup_argument_parser() -> ArgumentParser {
             "aarch64elf",
             "AArch64 ELF target",
             |args, _modifier_stack, _value| {
-                args.arch = Architecture::from_str("aarch64elf")?;
+                args.arch = Architecture::AArch64;
+                Ok(())
+            },
+        )
+        .sub_option(
+            "aarch64linux",
+            "AArch64 ELF target (Linux)",
+            |args, _modifier_stack, _value| {
+                args.arch = Architecture::AArch64;
                 Ok(())
             },
         )
@@ -1269,13 +1285,12 @@ fn setup_argument_parser() -> ArgumentParser {
             "elf64lriscv",
             "RISC-V 64-bit ELF target",
             |args, _modifier_stack, _value| {
-                args.arch = Architecture::from_str("elf64lriscv")?;
+                args.arch = Architecture::RISCV64;
                 Ok(())
             },
         )
-        .execute(|args, _modifier_stack, value| {
-            args.arch = Architecture::from_str(value)?;
-            Ok(())
+        .execute(|_args, _modifier_stack, value| {
+            bail!("-m {value} is not yet supported");
         });
 
     parser
--- a/libwild/src/arch.rs
+++ b/libwild/src/arch.rs
@@ -15,7 +15,6 @@ use object::elf::EM_RISCV;
 use object::elf::EM_X86_64;
 use std::borrow::Cow;
 use std::fmt::Display;
-use std::str::FromStr;
 
 pub(crate) trait Arch {
     type Relaxation: Relaxation;
@@ -61,19 +60,6 @@ pub(crate) enum Architecture {
     X86_64,
     AArch64,
     RISCV64,
-}
-
-impl FromStr for Architecture {
-    type Err = crate::error::Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "elf_x86_64" => Ok(Architecture::X86_64),
-            "aarch64elf" | "aarch64linux" => Ok(Architecture::AArch64),
-            "elf64lriscv" => Ok(Architecture::RISCV64),
-            _ => bail!("-m {s} is not yet supported"),
-        }
-    }
 }
 
 impl TryFrom<u16> for Architecture {


mati865:
Should `elf_x86_64_sol2` set the defaults from #1173 or are they supposed to be also set for `elf_x86_64` on Solaris/Illumos?

daniel-levin:
No, it need not set those defaults because those defaults are there to work
around the bugs and FIXME’s in Clang’s Solaris driver. #1173 should be
reverted once my fix lands in Clang because it will no longer be necessary.

mati865:
Ignoring `-C` is necessary because you interpose ld which is the Solaris linker, instead of interposing ld.bfd binary. So that one will indeed be fixed by `--ld-path`.
But the change to Clang doesn't seem to affect `--dynamic-linker`, so wouldn't that one be still necessary?

daniel-levin:
Ah, you are correct. `--dynamic-linker` would have to stay.

davidlattimore:
Do you think we should make it so that `-m elf_x86_64_sol2` sets the default dynamic linker? From a quick look at the code, it might be simpler if the `FromStr` implementation for `Architecture` got deleted. The `sub_option`s in `args.rs` look like they could handle the different architectures well.

mati865:
FWIW, I made this change:
```
❯ jj show
Commit ID: c3f87297d75106c5ea91d3c51b74647dde2a096b
Change ID: yoxyrupzktlzkrpqvmxzswuupwokynmv
Author   : Mateusz Mikuła <oss@mateuszmikula.dev> (2025-10-12 12:11:30)
Committer: Mateusz Mikuła <oss@mateuszmikula.dev> (2025-10-12 12:11:30)

    (no description set)

Modified regular file libwild/src/args.rs:
    ...
 309  309:             #[cfg(target_os = "illumos")]
 310  310:             dynamic_linker: Some(Path::new("/lib/amd64/ld.so.1").into()),
 311  311:             #[cfg(not(target_os = "illumos"))]
 312  312:             dynamic_linker: NoneSome(Path::new("/lib/ld-linux-x86-64.so.2").into()),
 313  313:             output_kind: None,
 314  314:             time_phase_options: None,
 315  315:             num_threads: None,
    ...
```
And all our tests and mold tests still pass, so maybe there should be a default value for all platforms?

I have checked Clang binary linked on Linux with ld with `--dynamic-linker` removed and ld sets interpreter to a non-existent file, while Wild doesn't set it at all:
```
❯ WILD_REFERENCE_LINKER=/usr/bin/ld.bfd ./run-with ~/Projects/wild/ld
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/liblto_plugin.so is not yet supported
wild: ./bin
ref: ./bin.ref-linker
.dynamic.DT_FLAGS_1.NOW
  wild 1
  ref

section.tdata.alignment
  wild 0x4
  ref 0x8

section.interp
  wild Section missing
  ref OK

dynsym.__bss_start.section
  wild
  ref .bss

dynsym._edata.section
  wild
  ref .data

dynsym._end.section
  wild
  ref .bss

segment.PT_INTERP.alignment
  wild
  ref 0x1

segment.PT_INTERP.flags
  wild
  ref 0x4

[..]

❯ readelf -W -p .interp bin.ref-linker

String dump of section '.interp':
  [     0]  /lib/ld64.so.1
```

EDIT: In any case, I can align the Linux one and add a test later. I can also refactor it if necessary.

daniel-levin:
> Do you think we should make it so that -m elf_x86_64_sol2 sets the default dynamic linker? 

Yes, I do. While @mati865 has identified a viable alternative for `--dynamic-linker`, I am certain that there will be _some_ other behavior induced by the `-m` flag that's worth memorializing in an obviously-discoverable way, like a `match` somewhere.

daniel-levin:
Mold makes inferences if `-m` is missing: https://github.com/rui314/mold/blob/5164da8ad2b6bed328c219b178082ff0e850be81/src/main.cc#L128 (solitary callsite: https://github.com/rui314/mold/blob/5164da8ad2b6bed328c219b178082ff0e850be81/src/main.cc#L283)

daniel-levin:
So does lld: https://github.com/llvm/llvm-project/blob/4c8275470465528c469436d37b9aaf71d6598899/lld/ELF/Driver.cpp#L2209. For the record, I'm not interested in cargo culting the behavior of existing linkers but it does corroborate @davidlattimore instinct that `-m` is a useful point of departure for all kinds of machine-specific code.

daniel-levin:
Aside: 67d998c9ee76d0d8f26b0558e7a4712d887cad27 says that printing supported emulations in `--help` is necessary for autoconf. Also: https://stackoverflow.com/questions/38951492/linkers-emulation-vs-output-format

daniel-levin:
https://github.com/bminor/binutils-gdb/blob/def8befb9db26ac20a9b8ebd258b20751f6deccc/ld/emulparams/elf_x86_64.sh#L49

Gah, this is horrible. We will do this more neatly in Wild.

davidlattimore:
Do you want to make the changes in this PR? I don't think it should be an especially big change - just delete the `FromStr` implementation, fix up the code that was using it, then when `elf_x86_64_sol2` is requested, check if there's already a dynamic-linker set and if not, set the appropriate value.

daniel-levin:
> Do you want to make the changes in this PR?

Yes, made the small change.

libwild/src/args.rs
```diff
@@ -1253,29 +1250,37 @@ fn setup_argument_parser() -> ArgumentParser {
...
-                args.arch = Architecture::from_str("elf_x86_64")?;
+                args.arch = Architecture::X86_64;
+                Ok(())
+            },
+        )
+        .sub_option(
+            "elf_x86_64_sol2",
+            "x86-64 ELF target (Solaris)",
+            |args, _modifier_stack, _value| {
+                args.dynamic_linker = Some(Path::new("/lib/amd64/ld.so.1").into());
```
davidlattimore: If we always set args.dynamic linker then that means that --dynamic-linker=/some/path -m elf_x86_64_sol2 will give a different outcome to -m elf_x86_64_sol2 --dynamic-linker=/some/path. i.e. the -m being later will override the earlier --dynamic-linker flag, which I don't think is what we want. Easiest fix is probably to check if args.dynamic_linker is already set and leave it as-is if so.

daniel-levin: D'oh. Good catch.

daniel-levin:
Unbroke the test (I missed a branch in the old `FromStr` implementation) and ensured the supplied `--dynamic-linker` argument is always respected.

mati865:
I'm not convinced whether it's sensible to complicate the code by setting the default dynamic linker for each architecture. On Linux, passing `--dynamic-linker` to the linker is crucial for dynamic executables anyway.

Without having a proper interpreter configured, the binary won't run:
```
❯ ./bin
fish: Job 1, './bin' terminated by signal SIGSEGV (Address boundary error)

❯ ./bin.ref-linker
exec: Failed to execute process './bin.ref-linker': The file exists and is executable. Check the interpreter or linker?

❯ readelf -p .interp bin bin.ref-linker

File: bin
readelf: Warning: Section '.interp' was not dumped because it does not exist

File: bin.ref-linker

String dump of section '.interp':
  [     0]  /lib/ld64.so.1
```

daniel-levin:
> I'm not convinced whether it's sensible to complicate the code by setting the default dynamic linker for each **architecture**

Emphasis mine. I agree. But, we are _not_ setting the default dynamic linker based on **architecture**. We're setting the default dynamic linker based on the `-m` flag, which is labelled as "emulation" in the other linkers. It's not a very well-defined concept. I suspect it evolved organically over time as a series of portability hacks. It means binary format + architecture + system-level conventions. For instance, `elf_x86_64_sol2` is distinct from `elf_x86_64`, even though the final binary produced is still an ELF file for x86_64. The naming convention on x86 is not amazing - surely you would expect `elf_x86_64` to be right on Solaris? (It is not). The fact that emulations include OS-level configuration is clearer in the ARM names, e.g. `aarch64linux`.

Indeed, if you run `ld --verbose -m $emulation` you can see the linker script for each "emulation". The linker script contains default search paths for the dynamic linker. Here's the one for `aarch64linux`:

```
OUTPUT_FORMAT("elf64-littleaarch64", "elf64-bigaarch64", "elf64-littleaarch64")
OUTPUT_ARCH(aarch64)
ENTRY(_start)
SEARCH_DIR("=/usr/aarch64-redhat-linux/lib64"); SEARCH_DIR("=/usr/lib64"); SEARCH_DIR("=/usr/local/lib64"); SEARCH_DIR("=/lib64"); SEARCH_DIR("=/usr/aarch64-redhat-linux/lib"); SEARCH_DIR("=/usr/local/lib"); SEARCH_DIR("=/lib"); SEARCH_DIR("=/usr/lib");
```

Unfortunately, it seems that the various open-source toolchains have taken a dependency on this convention, and they expect the 'emulation' flag to require the linker to find the right dynamic linker unless otherwise explicitly provided.

mati865:
My bad, I meant emulation rather than architecture. So, my point is on x86_64 Linux you cannot assume the linker will use the correct interpreter (because it will not), and the caller has to supply it anyway.
The only difference between providing it or not, is whether you get a nice error or segfault. In any case, the binary won't work, so both GCC and Clang provide the proper interpreter path anyway.

Issue #1114, opened 2025-09-26
Slint fails to build on RISC-V (`Missing High relocation connected with R_RISCV_PCREL_LO12`)


davidlattimore:
> I failed to build today's slint using today's nightly rust with the wild linker on the Starfive VisionFive2:
> 
> ```
>    Compiling figma_import v1.14.0 (/home/riscv/slint/tools/figma_import)
> error: linking with `clang` failed: exit status: 1
>   |
>   = note:  "clang" "<1 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "/tmp/rustcGVNk8q/libtikv_jemalloc_sys-243dde6ed652ff83.rlib" "<sysroot>/lib/rustlib/riscv64gc-unknown-linux-gnu/lib/libcompiler_builtins-*.rlib" "-Wl,-Bdynamic" "-latomic" "-ldl" "-ldl" "-lgcc_s" "-lutil" "-lrt" "-lpthread" "-lm" "-ldl" "-lc" "-L" "/tmp/rustcGVNk8q/raw-dylibs" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-L" "/home/riscv/slint/target/release/build/tikv-jemalloc-sys-c1ff78f2c4883b73/out/build/lib" "-L" "<sysroot>/lib/rustlib/riscv64gc-unknown-linux-gnu/lib" "-o" "/home/riscv/slint/target/release/deps/slint_viewer-1c18b3d71b80dbe1" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-Wl,-O1" "-Wl,--strip-debug" "-nodefaultlibs" "--ld-path=wild"
>   = note: some arguments are omitted. use `--verbose` to show all linker arguments
>   = note: wild: error: Failed copying from /home/riscv/slint/target/release/deps/slint_viewer-1c18b3d71b80dbe1.slint_viewer.8419fd355d688f2d-cgu.02.rcgu.o (259 (1/3)) to output file
>             Caused by:
>               Failed to apply relocations in section `.text._ZN5winit13platform_impl5linux3x1118EventLoop$LT$T$GT$13run_on_demand17hae08a271a019a2fdE` of /home/riscv/slint/target/release/deps/slint_viewer-1c18b3d71b80dbe1.slint_viewer.8419fd355d688f2d-cgu.02.rcgu.o
[...]

PR #1198, opened 2025-10-12
RISC-V: fix lookup of the riscv-fix-hi-part

marxin:
In some cases, the high-part can actually follow the low part as seen when linking the `slint-viewer` binary:
```
0000000000000240  0000eb2200000018 R_RISCV_PCREL_LO12_I   00000000000002a0 .Lpcrel_hi26116 + 0
...
00000000000002a0  0000eb1100000017 R_RISCV_PCREL_HI20     0000000000002620 .LCPI3435_5 + 0
```

I've just tested it fixes the `slint-viewer` linking with the Nightly toolchain.

Fixes: #1114

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -1989,10 +1989,12 @@ fn apply_relocation<'data, A: Arch>(
             );
             let mut relocations_to_search = relocation_sequence
                 .crel_iter()
+                // In most cases, the high-part is just before the low-part.
                 .take(relocation_index)
-                // TODO
-                //.rev()
-                .chain(relocation_sequence.crel_iter().take(relocation_index + 1));
+                // TODO: iterate the relocations in reverse order for faster lookup!
+                // .rev()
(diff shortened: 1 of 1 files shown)


libwild/src/elf_writer.rs
```diff
@@ -1989,10 +1989,12 @@ fn apply_relocation<'data, A: Arch>(
             );
             let mut relocations_to_search = relocation_sequence
                 .crel_iter()
+                // In most cases, the high-part is just before the low-part.
                 .take(relocation_index)
-                // TODO
-                //.rev()
-                .chain(relocation_sequence.crel_iter().take(relocation_index + 1));
+                // TODO: iterate the relocations in reverse order for faster lookup!
+                // .rev()
```
mati865, line 1995: Maybe it's the time to test this todo and fix it or remove 😉

marxin: Sure, good point. That will need some care: #1207.

PR #1199, opened 2025-10-12
properly indent macro invocation

karolzwolak:


marxin:
Just a side question: is there any possibility in `cargo fmt` to cover also macros? From my experience, in most cases, these invocations are skipped and one has to do the indentation on its own.

davidlattimore:
Not that I know of. There have been various discussions about it:
https://www.reddit.com/r/rust/comments/1eh7g80/why_does_rustfmt_ignore_macros_when_formatting/
https://internals.rust-lang.org/t/formatting-of-macros/8128

Issue #1119, opened 2025-09-27
USDT probes not properly included in linked binary


VorpalBlade:
Given the following code:

```rust
fn main() {
    probe::probe!(testgroup, testprobe, 42);
    probe::probe_lazy!(lazygroup, lazyprobe, 84);
}
```

(And ` cargo add probe` to get the dependency.)

I would expect these probe points to be included in the program. However, the entire `.note.stapsdt` section is missing.

In fact diffing the section list of the binary built with ldd (default linker as of Rust 1.90) and wild:

```diff
❯ diff -Naur <(eu-readelf -S target/debug/wild-repro.with-lld | grep -Eo '\.[^ ]+' | sort) <(eu-readelf -S target/debug/wild-repro.with-wild | grep -Eo '\.[^ ]+' | sort)     
--- /proc/self/fd/11    2025-09-27 11:12:41.775245326 +0200
+++ /proc/self/fd/14    2025-09-27 11:12:41.776245337 +0200
@@ -17,23 +17,19 @@
 .fini
 .fini_array
 .gcc_except_table
-.gnu_debuglink
 .gnu.hash
 .gnu.version
 .gnu.version_r
 .got
-.got.plt
 .init
 .init_array
 .interp
 .note.ABI-tag
 .note.gnu.build-id
-.note.stapsdt
-.plt
+.note.gnu.property
+.plt.got
 .probes
 .rela.dyn
-.rela.plt
-.relro_padding
 .rodata
 .shstrtab
 .strtab
```

This means that tools like bpftrace can't find the probe points:

```console
❯ sudo bpftrace -lv 'usdt:target/debug/wild-repro.with-lld:*'
usdt:target/debug/wild-repro.with-lld:lazygroup:lazyprobe
usdt:target/debug/wild-repro.with-lld:testgroup:testprobe
❯ sudo bpftrace -lv 'usdt:target/debug/wild-repro.with-wild:*'
❯
```

To test actually using the probes (to make sure they work, not just are included):

```console
❯ sudo bpftrace -e 'usdt:target/debug/wild-repro:testgroup:testprobe { printf("1: %d\n", arg0); } usdt:target/debug/wild-repro:lazygroup:lazyprobe { printf("2: %d\n", arg0); }'
Attached 2 probes
1: 42
2: 84
^C
```

(Ensure the path to the binary is correct in both the probe statements. You will need to run the program in question from another terminal while bpftrace is actively running.)

[...]

PR #1202, opened 2025-10-13
Add support for .note.stapsdt sections

marxin:
Sample binary output with Wild:
```
Note section [39] '.note.stapsdt' of 144 bytes at offset 0x3a47bc:
  Owner          Data size  Type
  stapsdt               52  Version: 3
    PC: 0x14b36, Base: 0x13916, Semaphore: 0
    Provider: testgroup, Name: testprobe, Args: '-8@%rax'
  stapsdt               52  Version: 3
    PC: 0x14b8c, Base: 0x13916, Semaphore: 0x582ce
    Provider: lazygroup, Name: lazyprobe, Args: '-8@%rax'
```

readelf with ld built binary:
```
❯ readelf -SW bin-ld | rg 'prob|base|stap'
  [17] .stapsdt.base     PROGBITS        000000000004c152 04c152 000001 00   A  0   0  1
  [30] .probes           PROGBITS        00000000000579c0 0579c0 000002 00  WA  0   0  2
  [33] .note.stapsdt     NOTE            0000000000000000 057a14 000090 00      0   0  4
```

the same with Wild:
```
  [16] .stapsdt.base     PROGBITS        0000000000013916 013916 000001 00   A  0   0  1
  [30] .probes           PROGBITS        00000000000582ce 0562ce 000002 00  WA  0   0  2
  [39] .note.stapsdt     NOTE            0000000000000000 3a47bc 000090 00      0   0  4
```

Fixes: #1119

Plus, the tracing output prints filename for a loaded section (helps with navigation to the source object file).

### DIFF
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -150,8 +150,6 @@ impl Config {
                 "section.phdr",
                 // We don't yet support these sections.
                 "section.data.rel.ro",
-                "section.note.gnu.build-id",
-                "section.note.gnu.property",
                 "section.hash",
                 "section.sframe",
                 // We set this to 8. GNU ld sometimes does too, but sometimes to 0.
@@ -199,8 +197,6 @@ impl Config {
                 // GNU ld sometimes makes this writable sometimes not. Presumably this depends on
                 // whether there are relocations or some flags.
                 "section.eh_frame.flags",
-                // A package note section used by Ubuntu: https://systemd.io/ELF_PACKAGE_METADATA/
-                "section.note.package",
                 // TLSDESC relaxations aren't yet implemented.
                 "rel.match_failed.R_X86_64_GOTPC32_TLSDESC",
                 "rel.missing-opt.R_X86_64_TLSDESC_CALL.SkipTlsDescCall.*",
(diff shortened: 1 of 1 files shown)


libwild/src/layout.rs
```diff
@@ -1897,8 +1897,14 @@ fn compute_segment_layout(
                         output_sections.section_debug(section_id)
                     );
                 } else {
-                    // RISCV_ATTRIBUTES segment is kind of special as it maps a section that is non-ALLOC.
-                    if section_id == output_section_id::RISCV_ATTRIBUTES {
+                    // RISCV_ATTRIBUTES and // Similarly, NOTE_STAPSDT segments are kind of special
```
mati865: Badly formatted?

marxin: Yep, thanks for noticing that.

davidlattimore:
Is this section something that linkers need to deal with explicitly, or is there perhaps a more general rule that we're missing? e.g. perhaps we need to, by default, keep all sections with `sh_type=NOTE`. GNU ld's built-in linker script (`ld --verbose`) doesn't seem to mention stapsdt.

marxin:
No, you are right, there's nothing special. Let me rework the PR to cover all notes except the ones we synthetize by the linker.

mati865:
Maybe other notes sections should also be unignored from linker-diff?

marxin:
> Maybe other notes sections should also be unignored from linker-diff?

Heh, what a good idea. Let's see where can we go.

PR #1209, opened 2025-10-13
handle signedness and optimize x86_64 relocation bounds checks

### LINKED ISSUE #1201: x86_64 relocation bounds checks aren't right for unsigned relocations
Follow on from #1149.

While reviewing https://github.com/davidlattimore/wild/pull/1175, I suggested that signed vs unsigned didn't matter. Turns out I was wrong. When linking clang with debug info (which produces a 4.1GiB binary), it now fails with the following error:

```
wild: error: Failed copying from /home/david/tmp/clang-link-debug/0/libLLVMAsmPrinter.a @ DwarfExpression.cpp.o (19979 (78/11)) to output file
  Caused by:
    Failed to apply relocations in section `.debug_loclists` of /home/david/tmp/clang-link-debug/0/libLLVMAsmPrinter.a @ DwarfExpression.cpp.o
    Failed to apply relocation of type R_X86_64_32 to section `.debug_info` (596958 local=50) in file #19979 (78/11) (/home/david/tmp/clang-link-debug/0/libLLVMAsmPrinter.a @ DwarfExpression.cpp.o) (NON_INTERPOSABLE) at offset 0x3e
    Relocation 2148213222 outside of bounds [-2147483648, 2147483647)
```

So I think what we're doing is correct for signed relocations, but not for unsigned. For `R_X86_64_32` the range should be 0 to 4GiB, while for `R_X86_64_32S` the range should be what is is now (-2GiB to +2GiB).

So I think each row in `relocation_from_raw` will need a third value in the tuple. That third value could either be a `bool` to say whether the value is signed or not, or you could just put the `AllowedRange` as the third value.

Figuring out which relocation types are signed may require a little judgement, but if we get some wrong, we can always fix them later. If you look at the following page from the AMD64 ABI:

Image

The relocations that have `- P` in their formula are definitely signed. I guess `R_X86_64_32S` is also signed, although I'm not sure how it gets negative values unless either `S` or `A` is signed.

@karolzwolak, let me know if you'd like to do this, otherwise, I'm happy to. Don't feel that you need to try building clang/llvm in order to reproduce this. Building llvm takes ages. I'm happy to verify that it links successfully with a proposed change.

karolzwolak:
Follow up to #1175, fixes #1201.

Note that 64-bit values are still always bound-checked as signed because `AllowedRange::{min, max}` are i64.

Additionally, this mitigates performance regressions from the preceding PR by:
* precomputing the relocation ranges at compile time
* inlining the `relocation_from_raw` function

### DIFF
--- a/linker-utils/src/x86_64.rs
+++ b/linker-utils/src/x86_64.rs
@@ -267,6 +267,7 @@ const RELOC_NONE: RelocSizeAndRange = (RelocationSize::ByteSize(0), AllowedRange
 
 /// Returns the supplied x86-64 relocation as RelocationKindType. Returns `None` if the r_type isn't recognised.
 #[must_use]
+#[inline(always)]
 pub const fn relocation_from_raw(r_type: u32) -> Option<RelocationKindInfo> {
     let (kind, (size, range)) = match r_type {
         object::elf::R_X86_64_64 => (RelocationKind::Absolute, RELOC_8_BYTE_UNSIGNED),


karolzwolak:
@davidlattimore please test linking clang

linker-utils/src/x86_64.rs
```diff
@@ -239,48 +240,49 @@ impl RelaxationKind {
...
-        object::elf::R_X86_64_PC32 => (RelocationKind::Relative, 4),
-        object::elf::R_X86_64_PC64 => (RelocationKind::Relative, 8),
-        object::elf::R_X86_64_GOT32 => (RelocationKind::GotRelGotBase, 4),
-        object::elf::R_X86_64_GOT64 => (RelocationKind::GotRelGotBase, 8),
-        object::elf::R_X86_64_GOTOFF64 => (RelocationKind::SymRelGotBase, 8),
-        object::elf::R_X86_64_PLT32 => (RelocationKind::PltRelative, 4),
-        object::elf::R_X86_64_PLTOFF64 => (RelocationKind::PltRelGotBase, 8),
-        object::elf::R_X86_64_GOTPCREL => (RelocationKind::GotRelative, 4),
+    let (kind, size, sign) = match r_type {
+        object::elf::R_X86_64_64 => (RelocationKind::Absolute, 8, Sign::Unsigned),
```
karolzwolak: I think it should be unsigned, but as I indicated it doesn't matter for 64bit values

linker-utils/src/x86_64.rs
```diff
@@ -239,48 +240,49 @@ impl RelaxationKind {
...
+        object::elf::R_X86_64_PC32 => (RelocationKind::Relative, 4, Sign::Signed),
+        object::elf::R_X86_64_PC64 => (RelocationKind::Relative, 8, Sign::Signed),
+        object::elf::R_X86_64_GOT32 => (RelocationKind::GotRelGotBase, 4, Sign::Unsigned),
+        object::elf::R_X86_64_GOT64 => (RelocationKind::GotRelGotBase, 8, Sign::Unsigned),
```
karolzwolak: Same here

linker-utils/src/x86_64.rs
```diff
@@ -239,48 +240,49 @@ impl RelaxationKind {
...
-        object::elf::R_X86_64_GOT64 => (RelocationKind::GotRelGotBase, 8),
-        object::elf::R_X86_64_GOTOFF64 => (RelocationKind::SymRelGotBase, 8),
-        object::elf::R_X86_64_PLT32 => (RelocationKind::PltRelative, 4),
-        object::elf::R_X86_64_PLTOFF64 => (RelocationKind::PltRelGotBase, 8),
-        object::elf::R_X86_64_GOTPCREL => (RelocationKind::GotRelative, 4),
+    let (kind, size, sign) = match r_type {
+        object::elf::R_X86_64_64 => (RelocationKind::Absolute, 8, Sign::Unsigned),
+        object::elf::R_X86_64_PC32 => (RelocationKind::Relative, 4, Sign::Signed),
+        object::elf::R_X86_64_PC64 => (RelocationKind::Relative, 8, Sign::Signed),
+        object::elf::R_X86_64_GOT32 => (RelocationKind::GotRelGotBase, 4, Sign::Unsigned),
```
karolzwolak: it should be unsigned right?

linker-utils/src/x86_64.rs
```diff
@@ -239,48 +240,49 @@ impl RelaxationKind {
...
+            (RelocationKind::GotRelative, 4, Sign::Signed)
         }
-        object::elf::R_X86_64_TPOFF32 => (RelocationKind::TpOff, 4),
-        object::elf::R_X86_64_GOTPC32_TLSDESC => (RelocationKind::TlsDesc, 4),
-        object::elf::R_X86_64_TLSDESC_CALL => (RelocationKind::TlsDescCall, 0),
-        object::elf::R_X86_64_NONE => (RelocationKind::None, 0),
+        object::elf::R_X86_64_TPOFF32 => (RelocationKind::TpOff, 4, Sign::Signed),
+        object::elf::R_X86_64_GOTPC32_TLSDESC => (RelocationKind::TlsDesc, 4, Sign::Signed),
+        object::elf::R_X86_64_TLSDESC_CALL => (RelocationKind::TlsDescCall, 0, Sign::Signed),
+        object::elf::R_X86_64_NONE => (RelocationKind::None, 0, Sign::Signed),
```
karolzwolak: Doesn't matter here as well but I didn't want to add Sign::None nor wrap it into Option<>

linker-utils/src/x86_64.rs
```diff
@@ -239,48 +240,49 @@ impl RelaxationKind {
...
         _ => return None,
     };
 
     Some(RelocationKindInfo {
         kind,
         size: RelocationSize::ByteSize(size),
         mask: None,
-        range: AllowedRange::from_byte_size(size),
+        range: AllowedRange::from_byte_size(size, sign),
         alignment: 1,
```
karolzwolak, line 315: btw, is this alignment being always 1 correct?

linker-utils/src/elf.rs
```diff
@@ -1199,14 +1205,17 @@ impl AllowedRange {
     }
 
     #[must_use]
-    pub const fn from_byte_size(n_bytes: usize) -> Self {
+    /// Note: for the 8-byte size, we actually do signed checks regardless of the `sign` argument because the `min` and `max` are `i64` type
```
karolzwolak: I noticed that the fomatter doesn't care about extra long lines -- do we want to enforce max line length like 100?

linker-utils/src/x86_64.rs
davidlattimore: Instructions on x86_64 aren't aligned, which means that the relocations that go into those instructions don't have any particular alignment. I can't be sure that no relocations have alignment requirements, but at least most don't.

linker-utils/src/elf.rs
davidlattimore: I usually wrap at 100. My editor is configured to do this when I press alt-q. I'd be open to something that automatically applies to everyone's changes if you've got ideas.

davidlattimore:
> @davidlattimore please test linking clang

Yep, seems to work. Thanks :)

linker-utils/src/x86_64.rs
davidlattimore: I guess if you did make it an `Option`, then when it's `None`, that could just map to `AllowedRange::no_check()`

davidlattimore:
I just noticed that the previous PR had quite an effect on performance. It increased instruction count by about 10% when linking rustc, 2.9% when linking wild and 5% when linking clang. So it might be good to experiment with including the range as the 3rd value in the tuple rather than whether it's signed and see if that restores the performance.

linker-utils/src/elf.rs
karolzwolak: Not really -- I thought for some reason that `wrap_comments` got finally stabilized in rustfmt but that's not the case.

karolzwolak:
Ouch, 10% for rustc is baad. I pushed commit that precomputes the ranges.

mati865:
You can also try with const blocks:  https://doc.rust-lang.org/reference/expressions/block-expr.html#const-blocks

karolzwolak:
Oh interesting, so we could wrap every match arm inside const block like so:

```rust
object::elf::R_X86_64_64 => const { (RelocationKind::Absolute, 8, Sign::Unsigned) },
```

instead of having a dedicated constant:
```rust
object::elf::R_X86_64_64 => (RelocationKind::Absolute, RELOC_8_BYTE_UNSIGNED),
```
It's not immediately clear to me which style I like better. What do you think?
I think slightly prefer the const blocks, but I've included both of them in separate commits.

Let's verify precomputing the ranges actually mitigates the perf regression though.

davidlattimore:
I think it's likely that you need to ensure that the call to `AllowedRange::from_byte_size` is inside the const blocks, otherwise it might happen at runtime.

I did an experiment where the entire `RelocationKindInfo` was computed inside the const blocks. That did improve performance a bit, but still not to where it was on October 9. I think a large part of the performance change is because the function is now more complex and has been pushed over some threshold where it's not getting inlined anymore. Once I marked the function as `#[inline(always)]` almost all the performance loss was restored. That wasn't sufficient though. I tried adding `#[inline(always)]` without any of the changes in this PR and still saw the instruction count as up by 3.6% compared to an October 9 build. Much better than the 10%, but still some way to go. With both the inline directive and computing `RelocationKindInfo`, the instruction count is only up by 0.4% from the baseline. Some increase is to be expected, since we're doing range checks now where we weren't before. So I think 0.4% seems reasonable to me.

mati865:
Those const block won't do anything because the function call is outside them. Unfortunately, this code is too complex for const evaluation. I hope it could be done in a nicer way, but here is a version that optimises well: https://github.com/mati865/wild/commit/4ca0065854e69f4cf622eb32b5596b86b4378e65

I cannot tell from the assembly whether "precompute reloc ranges" commit or my change would be faster. You can see the assembly with `cargo asm relocation_from_raw -p linker-utils --rust` if you have `cargo-show-asm` installed.

On my machine, the benchmarks show nothing but noise in wall time for both linking Clang. CPU cycles and instruction counts do increase, but branch misses are lower, so I do not see the regression at all. When linking rustc both fixes still regress on instruction count, but all other metrics (including wall time) are improved compared to repository state from before the initial PR (AFAIK there were no other perf impacting changes):

<details>
<summary>Details</summary>

```
~/tmp/rustc took 20s
❯ OUT=/tmp/out.tmpfs powerprofilesctl launch -p performance poop './run-with ~/Projects/wild/target/release/wild-qtn' './run-with ~/Projects/wild/target/release/wild-xmy' './run-with ~/Pro
Benchmark 1 (170 runs): ./run-with ~/Projects/wild/target/release/wild-qtn
  wall_time          29.2ms ± 2.01ms    25.7ms … 36.4ms          2 ( 1%)        0%
  peak_rss           7.60MB ± 66.8KB    7.31MB … 7.73MB         14 ( 8%)        0%
  cpu_cycles         1.73G  ± 52.5M     1.57G  … 1.91G           5 ( 3%)        0%
  instructions        557M  ± 4.24M      547M  …  573M           2 ( 1%)        0%
  cache_references   17.3M  ±  203K     16.7M  … 17.9M           5 ( 3%)        0%
  cache_misses       4.13M  ± 54.3K     3.99M  … 4.35M           6 ( 4%)        0%
  branch_misses      1.77M  ± 15.9K     1.74M  … 1.81M           1 ( 1%)        0%
Benchmark 2 (169 runs): ./run-with ~/Projects/wild/target/release/wild-xmy
  wall_time          29.3ms ± 1.93ms    26.0ms … 35.5ms          7 ( 4%)          +  0.3% ±  1.4%
  peak_rss           7.60MB ± 73.2KB    7.33MB … 7.73MB         18 (11%)          -  0.0% ±  0.2%
  cpu_cycles         1.73G  ± 48.0M     1.58G  … 1.83G           3 ( 2%)          +  0.2% ±  0.6%
  instructions        577M  ± 4.42M      565M  …  589M           0 ( 0%)        💩+  3.6% ±  0.2%
  cache_references   17.3M  ±  191K     16.7M  … 17.8M           3 ( 2%)          +  0.1% ±  0.2%
  cache_misses       4.13M  ± 53.2K     3.96M  … 4.27M           3 ( 2%)          -  0.1% ±  0.3%
  branch_misses      1.72M  ± 14.0K     1.67M  … 1.75M           2 ( 1%)        ⚡-  2.7% ±  0.2%
Benchmark 3 (174 runs): ./run-with ~/Projects/wild/target/release/wild-constants
  wall_time          28.5ms ± 2.00ms    24.5ms … 35.5ms          5 ( 3%)          -  2.2% ±  1.5%
  peak_rss           7.60MB ± 63.5KB    7.32MB … 7.72MB         16 ( 9%)          +  0.0% ±  0.2%
  cpu_cycles         1.58G  ±  128M     1.35G  … 1.81G           0 ( 0%)        ⚡-  8.4% ±  1.2%
  instructions        566M  ± 9.29M      550M  …  585M           0 ( 0%)        💩+  1.6% ±  0.3%
  cache_references   17.0M  ±  242K     16.3M  … 17.5M           0 ( 0%)        ⚡-  1.8% ±  0.3%
  cache_misses       4.01M  ±  109K     3.84M  … 4.26M           0 ( 0%)        ⚡-  3.0% ±  0.4%
  branch_misses      1.73M  ± 23.9K     1.68M  … 1.79M           0 ( 0%)        ⚡-  2.2% ±  0.2%
Benchmark 4 (174 runs): ./run-with ~/Projects/wild/target/release/wild-proposed
  wall_time          28.5ms ± 2.17ms    24.1ms … 36.5ms          2 ( 1%)          -  2.3% ±  1.5%
  peak_rss           7.60MB ± 77.3KB    7.31MB … 7.73MB         19 (11%)          -  0.0% ±  0.2%
  cpu_cycles         1.57G  ±  127M     1.32G  … 1.93G           0 ( 0%)        ⚡-  9.1% ±  1.2%
  instructions        566M  ± 9.45M      549M  …  593M           0 ( 0%)        💩+  1.6% ±  0.3%
  cache_references   17.0M  ±  242K     16.3M  … 17.7M           1 ( 1%)        ⚡-  1.7% ±  0.3%
  cache_misses       4.00M  ±  107K     3.81M  … 4.27M           0 ( 0%)        ⚡-  3.2% ±  0.4%
  branch_misses      1.73M  ± 25.3K     1.68M  … 1.82M           1 ( 1%)        ⚡-  1.9% ±  0.3%
```
qtn is right before the previous bounds PR, xmy is the problematic change, constants is "precompute reloc ranges", finally proposed is my change

</details>

karolzwolak:
Whoops, yeah obviously the const blocks in my commit aren't doing anything. 
Dedicated constants (from [precompute reloc ranges](https://github.com/davidlattimore/wild/commit/9f392abdd66c6091957cb77870e6b93d7f351c8a)) for each combination should have the same perf outcome as const blocks proposed by @mati865 (since the ranges are gonna be computed once at compile time), and are cleaner in my opinion.
I also added `#[inline(always]` to `relocation_from_raw()`. Is there anything left to add here?

marxin:
Good work, thanks for the quick fix.

PR #1210, opened 2025-10-14
Replace all uses of bytemuck with zerocopy

### LINKED ISSUE #1200: Try to drop dependency on bytemuck
We currently use the `bytemuck` crate for a few things, mostly related to writing, but also in linker-diff. I recently added a dependency on `zerocopy` because I couldn't do what I needed with `bytemuck`. Specifically, I wanted to transmute between `&mut SomeNonAtomicType` and `&mut Some AtomicType`, but bytemuck requires types implement `Copy` and types containing atomics cannot implement `Copy`.

Anyway, since the two crates fill very similar roles, we probably shouldn't depend of both.

TechnoPorg:
Closes #1200. Unwraps have been used where bytemuck would panic on a size mismatch, or where the size is explicitly checked before performing the conversion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,26 +190,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bytemuck"
-version = "1.24.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
-dependencies = [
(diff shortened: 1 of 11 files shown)


davidlattimore:
Thanks! One of the failures looks unrelated to your change and likely to be a flake. Hopefully once you run cargo fmt and reupload, the flake won't happen. Not sure how the flake is possible since we're supposed to have file locking to prevent that kind of thing

PR #1211, opened 2025-10-14
Add a test for handling of note sections

davidlattimore:
Also added a TODO for some stuff that we're not currently getting right with non-alloc NOTE sections.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -1901,7 +1901,13 @@ fn compute_segment_layout(
                         output_sections.section_debug(section_id)
                     );
                 } else {
-                    // The note and RISCV_ATTRIBUTES sections  do not have ALLOC flag set.
+                    // TODO: Remove the NOTE exception. Non-alloc sections should be placed outside
+                    // of program segments. NOTE sections are sometimes alloc and sometimes not.
+                    // Alloc NOTE sections should be placed within a LOAD segment and within a NOTE
+                    // segment. Non-alloc NOTE sections shouldn't be in any segment.
+
+                    // The .riscv.attributes section is non-alloc but is expected to be put into a
+                    // RISCV_ATTRIBUTES segment.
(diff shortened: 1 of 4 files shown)


libwild/src/layout.rs
```diff
@@ -1901,7 +1900,10 @@ fn compute_segment_layout(
                         output_sections.section_debug(section_id)
                     );
                 } else {
-                    // The note and RISCV_ATTRIBUTES sections  do not have ALLOC flag set.
+                    // TODO: Remove these two exceptions. Non-alloc sections should be placed
+                    // outside of program segments. NOTE sections are sometimes alloc and sometimes
+                    // not. Alloc NOTE sections should be placed within a LOAD segment and within a
+                    // NOTE segment. Non-alloc NOTE sections shouldn't be in any segment.
                     if [NOTE, RISCV_ATTRIBUTES].contains(&section_info.ty) {
```
marxin, line 1911: If I'm not mistaken, the `RISCV_ATTRIBUTES` section is actually mapped by other linkers to a program segment (with name `RISCV_ATTRIBUTES`):

```
Program Headers:
  RISCV_ATTRIBUTES 0x00102a 0x0000000000000000 0x0000000000000000 0x000074 0x000000 R   0x1
...
Section Headers:
[25] .riscv.attributes    RISCV_ATTRIBUTES 0000000000000000 0000102a 00000074  0        0   0  1
```

davidlattimore: Thanks you're right. I've updated the TODO to just be about the NOTE section. It's possible that there's more we could do  with the RISCV_ATTRIBUTES segment, since it looks like GNU ld sets the size of the segment to 0, which makes sense given that it contains only non-alloc sections. We currently set it to the file-size. It likely has no effect on runtime though.

marxin (approved):

PR #1219, opened 2025-10-16
CI: Limit the conditions for executing `update-nix-lockfile`

lapla-cogito:
Some forks seem to periodically run the `update-nix-lockfile` action via cron. (My own fork doesn’t do this, since its `main` branch is outdated, but I noticed it while looking through other forks.)
We can prevent this by limiting the workflow’s trigger conditions to the repository owner.


davidlattimore (approved):
Thanks for noticing this!

PR #1220, opened 2025-10-16
Version script: Replace glob with custom matcher to support negation

lapla-cogito:
part of #990 

Let me start by noting that this change introduces a certain level of complexity, so I think there’s room for discussion about whether it should be merged.

As I mentioned in the issue above, using the negation character `^` in version scripts is not part of the GNU standard, but since both lld and mold support it, it would be ideal for Wild to support it as well.
The main reason this hasn’t been possible so far is that the `glob` crate hardcodes `!` as the negation character. Because of this, there are two possible solutions:

- Preprocess the version script, parsing it once to replace `^` with `!` before using the `glob` crate.
- Implement custom matching logic without relying on the `glob` crate.

The first approach would effectively require parsing twice, which is inefficient (simply replacing characters wouldn’t handle cases like [`version-script15.sh`](https://github.com/rui314/mold/blob/main/test/version-script15.sh)). Therefore, I implemented the second approach.

In practice, benchmarking with `bevy_dylib` showed no noticeable performance degradation (however the results may vary in other environments):

```
（▰╹◡╹）❯  hyperfine --warmup 2 '/tmp/wild/bevy/60/run-with ./wild.before' '/tmp/wild/bevy/60/run-with ./wild.after'
Benchmark 1: /tmp/wild/bevy/60/run-with ./wild.before
  Time (mean ± σ):      2.528 s ±  0.799 s    [User: 1.004 s, System: 0.346 s]
  Range (min … max):    2.008 s …  4.706 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: /tmp/wild/bevy/60/run-with ./wild.after
  Time (mean ± σ):      2.388 s ±  0.290 s    [User: 1.010 s, System: 0.341 s]
  Range (min … max):    1.957 s …  2.955 s    10 runs

Summary
  /tmp/wild/bevy/60/run-with ./wild.after ran
    1.06 ± 0.36 times faster than /tmp/wild/bevy/60/run-with ./wild.before

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,9 +215,9 @@ dependencies = [
 
 [[package]]
 name = "cfg-if"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fd1289c04a9ea8cb22300a459a72a385d7c73d3259e2ed7dcb2af674838cfa9"
+checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "clap"
@@ -527,14 +527,14 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.3.3"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26145e563e54f2cadc477553f1ec5ee650b00862f0a58bcd12cbdc5f0ea2d2f4"
+checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
  "libc",
  "r-efi",
- "wasi",
+ "wasip2",
 ]
 
 [[package]]
@@ -728,7 +728,6 @@ dependencies = [
  "foldhash",
  "gimli",
  "git-version",
- "glob",
  "hashbrown",
  "hex",
  "indexmap",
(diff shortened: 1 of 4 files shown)


marxin:
> Let me start by noting that this change introduces a certain level of complexity, so I think there’s room for discussion about whether it should be merged.

Thank you for your work, but I am strongly against adding an another glob evaluator the Wild linker. I would prefer staying with the `glob` crate.

> As I mentioned in the issue above, using the negation character ^ in version scripts is not part of the GNU standard, but since both lld and mold support it, it would be ideal for Wild to support it as well.

Fair enough. We should keep in mind it's super-esoteric feature (very rare).

> The main reason this hasn’t been possible so far is that the glob crate hardcodes ! as the negation character. Because of this, there are two possible solutions:
> Preprocess the version script, parsing it once to replace ^ with ! before using the glob crate.

That's correct, but we're doing the parsing anyway (in order to identify '*' chars) and so we can transform the input if `^` is present. Again, in 99.9% of the inputs, that's not going to be the case.

> In practice, benchmarking with bevy_dylib showed no noticeable performance degradation (however the results may vary in other environments):

Note this version script only contains exact matches, that's why can't measure any difference.

marxin (changes requested):
I don't like the idea of a new separate glob matcher.

lapla-cogito:
Alright. Thank you for the review. Considering this is indeed a rare use case, we should consider alternative approaches.

marxin:
Thank you @lapla-cogito.

davidlattimore:
We can always consider resurrecting this PR if we discover real-world projects that are using that syntax. Thanks for your work on it :)

PR #1222, opened 2025-10-16
Handle missing GNU_PROPERTY .note.gnu.property values

marxin:
Noticed while building `qtwayland` package where coverage for the following property was missing:
```
  GNU                  0x00000040	NT_GNU_PROPERTY_TYPE_0
      Properties: 1_needed: indirect external access
```

### DIFF
--- a/wild/tests/sources/trivial-main.c
+++ b/wild/tests/sources/trivial-main.c
@@ -40,6 +40,6 @@
 
 //#Config:gcc-indirect-external:default
 //#CompArgs:-fPIE -mno-direct-extern-access
-//#Arch:x86_64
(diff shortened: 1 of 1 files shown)


davidlattimore:
The failure message from gcc was a bit hard to spot, so I did https://github.com/davidlattimore/wild/pull/1224 to make it a bit easier.

Not sure if you'd already started looking at a solution to the unsupported flag (sorry if you had), but I thought I'd make the checking of compiler flags mechanism more generic - https://github.com/davidlattimore/wild/pull/1225

davidlattimore (approved):
Looks good to me once CI passes :)

marxin:
Thank you @davidlattimore for the infrastructure improving changes!

PR #1230, opened 2025-10-19
Add basic testing for update-in-place flag

el-yawd:
The new test mode actually breaks, but I'm not sure if I'm testing it right. The test pass when random bytes overwrite section is commented out.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -464,6 +464,7 @@ struct Config {
     auto_add_objects: bool,
     rustc_channel: RustcChannel,
     requires_rust_musl: bool,
+    test_update_in_place: bool,
     test_config: TestConfig,
     tracked_files: Vec<PathBuf>,
 }
@@ -713,6 +714,7 @@ impl Config {
             auto_add_objects: true,
             rustc_channel: RustcChannel::Default,
             requires_rust_musl: false,
+            test_update_in_place: false,
             test_config: test_config.clone(),
             tracked_files: Default::default(),
         }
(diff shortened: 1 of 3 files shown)


wild/tests/integration_tests.rs
```diff
@@ -1003,6 +1012,94 @@ impl ProgramInputs {
...
+                    " (first difference at offset 0x{:x}: original=0x{:02x}, final=0x{:02x})",
+                    offset, original_byte, final_byte
+                )
+            } else if original_content.len() != final_content.len() {
+                "(length difference)".to_string()
+            } else {
+                "(no differences found - this shouldn't happen)".to_string()
+            };
+
+            bail!(
```
davidlattimore, line 1106: Adding the filename of the binary would be useful, since the next thing someone would want to do is look at the binary to see what's at the specified offset in the file - e.g. what section it's in.

wild/tests/integration_tests.rs
```diff
@@ -1003,6 +1012,94 @@ impl ProgramInputs {
...
+                "Update-in-place test failed for {}: content differs after update-in-place linking. \
+                Original size: {}, Final size: {}{}",
+                self.source_file,
+                original_content.len(),
+                final_content.len(),
+                diff_info
+            );
+        }
+
+        println!("Update-in-place test passed for {}", self.source_file);
```
davidlattimore: Probably no need to print output when it succeeds

wild/tests/integration_tests.rs
el-yawd: Yeah, I think in case of error it's better to have both files to debug and print the respective path

davidlattimore:
I've fixed missing zero-fill bugs - #1237. If you rebase your change, then I think the tests should now pass.

el-yawd:
> I've fixed missing zero-fill bugs - #1237. If you rebase your change, then I think the tests should now pass.

Cool!

PR #1231, opened 2025-10-20
ci: fix potential security issues

mati865:
After reading https://ptrpa.ws/nixpkgs-actions-abuse, I decided to lint our CI scripts. I doubt the issue with `release.yml` would be exploitable because it's set to trigger only on tag push, which itself is difficult to exploit.

I didn't act on Zizmor's `unpinned-uses` and `dependabot-cooldown`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,6 +36,8 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -36,6 +36,8 @@ jobs:
 
     - name: Checkout code
       uses: actions/checkout@v4
+      with:
+        persist-credentials: false
```
davidlattimore, line 40: Have you tried running the release action on your fork with these changes? I have a vague recollection of trying to add this `persist-credentials` line once myself

mati865: Ah, forgot to write [wrkflw](https://github.com/bahdotsh/wrkflw) doesn't seem to work for me, so I didn't test release build. Perhaps `softprops/action-gh-release@v2` requires the credentials?
I'll try tomorrow.

mati865: Appears to work fine, there is a draft release at https://github.com/mati865/wild/releases/tag/untagged-721530ea245c31cf02d2 and the job passed: https://github.com/mati865/wild/actions/runs/18687086619

davidlattimore (approved):

PR #1238, opened 2025-10-21
Make --update-in-place default and fallback to unlinking on ETXTBSY

### LINKED ISSUE #1223: Make --update-in-place default and fallback to unlinking on ETXTBSY
Currently we unlink (delete) the old output file then create a new file in its place. We have a flag `--update-in-place` that overrides this. Generally updating the file in-place is slightly faster, however it can fail if the file is being executed. If the file is a shared object, then update-in-place won't fail, but would likely cause problems for any running program that has that shared object loaded.

So the idea would be that we default to update-in-place for executables, but not for shared objects. Then, if update-in-place fails due to ETXTBSY, we fall back to unlinking the file and creating a new one. If `--update-in-place` is actually specified on the command-line, then we shouldn't fall back if we get ETXTBSY and should instead report the error. `update_in_place` could probably be changed to an `Option<bool>`. We may want to add `--no-update-in-place` to force it to always unlink.

There's also kernel version 6.11, which removed the ETXTBSY error (it was restored in 6.12). We could detect that and always unlink, but it doesn't seem necessary, since 6.11 is not widely used anymore and is likely to get even less used as time goes by.

Because `--update-in-place` isn't the default, it hasn't had a lot of testing. There's a non-zero chance that we have bits of our writing code that don't write to some bytes of the file. This would be problematic if the file isn't initialised with all zeros. So I think a first step for this issue is to write a test that runs wild twice. Run once with the output file not existing or with `--no-update-in-place`. Read the contents of the file for later use. Overwrite the file with some random, non-zero byte. Possibly extend the file a bit larger than it was while we're at it. Run again with `--update-in-place`. Read the file again and check that it's identical to the bytes we read the first time. The simplest place to add this test is by making it an optional mode in `integration_test.rs`. i.e. add a test directive, that when turned on, performs this testing when linking with wild. We can then do a one-off test with it on for all tests, but then probably just leave it on for one or a few.

el-yawd:
Closes #1223

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -175,8 +175,8 @@ pub(crate) enum FileWriteMode {
 
     /// The existing output file, if any, will be edited in-place. Any hard links to the file will
     /// update accordingly. If the file is locked due to currently being executed, then our write
-    /// will fail.
-    UpdateInPlace,
+    /// will fail. Specified indicates whether the mode was explicitly specified by the user or not.
+    UpdateInPlace { specified: bool },
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
@@ -2191,7 +2191,7 @@ fn setup_argument_parser() -> ArgumentParser {
         .long("update-in-place")
         .help("Update file in place")
         .execute(|args, _modifier_stack| {
-            args.file_write_mode = Some(FileWriteMode::UpdateInPlace);
+            args.file_write_mode = Some(FileWriteMode::UpdateInPlace { specified: true });
             Ok(())
         });
 
--- a/libwild/src/file_writer.rs
+++ b/libwild/src/file_writer.rs
@@ -1,5 +1,6 @@
 use crate::args::Args;
 use crate::args::FileWriteMode;
+use crate::args::OutputKind;
 use crate::args::WRITE_VERIFY_ALLOCATIONS_ENV;
 use crate::error;
 use crate::error::Context as _;
@@ -11,6 +12,7 @@ use crate::output_section_map::OutputSectionMap;
 use crate::output_section_part_map::OutputSectionPartMap;
 use crate::output_trace::TraceOutput;
 use memmap2::MmapOptions;
+use std::io::ErrorKind;
 use std::io::Write;
 use std::ops::Deref;
 use std::ops::DerefMut;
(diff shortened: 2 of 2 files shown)


libwild/src/args.rs
```diff
@@ -175,8 +175,8 @@ pub(crate) enum FileWriteMode {
 
     /// The existing output file, if any, will be edited in-place. Any hard links to the file will
     /// update accordingly. If the file is locked due to currently being executed, then our write
-    /// will fail.
-    UpdateInPlace,
+    /// will fail. Specified indicates whether the mode was explicitly specified by the user or not.
+    UpdateInPlace { specified: bool },
```
davidlattimore, line 179: Looking at the code, I wonder if adding a third variant `Auto` might look better. But entirely up to you. I'm happy with either.

davidlattimore (commented):
Thanks. The change looks good. I did some additional testing and it looks like there's more places where we're missing writes. I'm happy to keep fixing them, but you're also welcome to have a go.

I did #1239, which makes it a bit easier to see multiple parts of the file that aren't being written. If I add `//#TestUpdateInPlace:true` to the `shared` variant of `libc-integration.c`, then I get the following sections with diffs:

```
[
    .note.gnu.build-id (0xab8..0xae8) 32 bytes differ between 0xac8 and 0xae7,
    .dynsym (0xb38..0xe80) [b3c],
    .gnu.version_r (0x1198..0x11d8) [11ac, 11ad, 11bc, 11bd, 11cc, 11cd],
    .dynamic (0x1cb8..0x1ed8) 112 bytes differ between 0x1e68 and 0x1ed7,
    .got (0x1ed8..0x1fc8) 239 bytes differ between 0x1ed8 and 0x1fc7,
]
```

The build-id depends on the contents of the rest of the file, so it won't match until all the others match, so that one should be dealt with last.

The rough idea would be to find the code responsible for each of these in `elf_writer.rs` and locate the spot where those bytes are being skipped without being written to, then zero them. If any prove too tricky, feel free to leave them for me. Or, if you'd rather I do them all, that's also fine. We should probably wait until they're all fixed before we merge this PR though.

davidlattimore:
Do let me know if you need help with writing to those uninitialised bytes - or would like me to do it. It's entirely possible that some of them could be tricky to figure out.

el-yawd:
I got quite busy in the last two weeks and I think I couldn't manage to do it. Feel free to take it, I'll follow the PRs and try to learn from you

davidlattimore:
Cool. I think I got them all with https://github.com/davidlattimore/wild/pull/1276. Now, if I change the default for `test_in_place` to `true`, the tests still pass.

Issue #1097, opened 2025-09-22
linker-scripts: Add support for PROVIDE


davidlattimore:
https://sourceware.org/binutils/docs/ld/PROVIDE.html

Note, this is not quite the same as a weak symbol, since presumably defining a weak symbol would cause a PROVIDE symbol to not be defined, so they're lower priority than weak.

The first step would be to implement parsing of `PROVIDE` in `linker_script.rs`.

The function `resolve_alternative_symbol_definitions` is where symbols that have multiple definitions are processed. That's one possible place where it might make sense to add PROVIDE support. I do worry slightly that this might affect performance. PROVIDE isn't commonly used, and we don't want to slow down linking when it isn't used.

Another possible place where support might be able to be added is in `canonicalise_undefined_symbols`. i.e. when we encounter an undefined symbol, if there's a PROVIDE definition, use that.

I'll tag this as a good-first-issue because I think it might be a good introduction for someone. It's not trivial though, so is best suited for someone who wants to spend a bit of time understanding how some bits of the linker work.

mdmoss:
I'd like to work on this. It'll be something of a learning experience so it might be slow going (happy for others to take over if it's blocking, of course).

I've [implemented parsing for PROVIDE](https://github.com/mdmoss/wild/tree/implement-provide) just to get going but I suspect that's the easy part. 

I'll come back at some point with a PR or questions if I get stuck.

mdmoss:
Saw the mention on #1257. Progress is pretty glacial but I'm still at it.

Always happy to hand it over if it's blocking, of course.

davidlattimore:
All good. If you're still working on it, that's fine. Do let us know if you have any questions

lapla-cogito:
@mdmoss Hi! Are you still working on this?

lapla-cogito:
Since there has been no response, I'll take a look at this, possibly handling not only `PROVIDE` but also `PROVIDE_HIDDEN`.

PR #1257, opened 2025-11-01
feat: add `__executable_start`

mati865:
cc https://github.com/davidlattimore/wild/issues/1096

### DIFF
--- a/libwild/src/output_section_id.rs
+++ b/libwild/src/output_section_id.rs
@@ -375,6 +375,7 @@ pub(crate) struct BuiltInSectionDetails {
     pub(crate) link: &'static [OutputSectionId],
     pub(crate) start_symbol_name: Option<&'static str>,
     pub(crate) end_symbol_name: Option<&'static str>,
+    pub(crate) synthetic_symbol_names: Option<&'static [&'static str]>,
     pub(crate) min_alignment: Alignment,
     info_fn: Option<fn(&InfoInputs) -> u32>,
     pub(crate) keep_if_empty: bool,
@@ -390,6 +391,7 @@ const DEFAULT_DEFS: BuiltInSectionDetails = BuiltInSectionDetails {
     link: &[],
     start_symbol_name: None,
     end_symbol_name: None,
+    synthetic_symbol_names: None,
     min_alignment: alignment::MIN,
     info_fn: None,
     keep_if_empty: false,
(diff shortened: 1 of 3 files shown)


libwild/src/output_section_id.rs
```diff
@@ -636,6 +637,7 @@ const SECTION_DEFINITIONS: [BuiltInSectionDetails; NUM_BUILT_IN_SECTIONS] = [
         kind: SectionKind::Primary(SectionName(TEXT_SECTION_NAME)),
         ty: sht::PROGBITS,
         section_flags: shf::ALLOC.with(shf::EXECINSTR),
+        start_symbol_name: Some(EXECUTABLE_START),
         ..DEFAULT_DEFS
```
mati865, line 641: This is wrong because that symbols should be emitted only for (some kinds of?) executables. Probably should be created inside `Prelude::new`.

Similarly, Wild outputs `__ehdr_start` even when other linkers don't, and it also misses `__dso_handle` symbol.

davidlattimore: I haven't looked into this too much, but a couple of things that you were probably aware of, but just in case... some of these symbols are `PROVIDE` in the GNU ld linker script, which AFAIK means that the linker script should only define the symbol if it's referenced and not defined by anything input objects. We don't yet support PROVIDE semantics, although it might be being worked on - #1097.

Some flags can affect the built-in linker script that GNU ld uses. You can see what changes for some particular flag with a command like `meld <(ld --verbose) <(ld --pie --verbose)`.

mati865:
Quick update:
Ah, I think now I understand how `__executable_start` and `__ehdr_start` are supposed to work. Looks like the linker should provide them if they are undefined in the input files, so I'm looking into creating something similar as `allocate_start_stop_symbol_id` does.

With the latest change, [eyra](https://github.com/sunfishcode/eyra) now passes many more tests and `example_crate_no_std` failure doesn't seem to be related to `__executable_start`. I'm not sure if merging such stopgap workaround is feasible.

libwild/src/output_section_id.rs
mati865: Thanks, I didn't see your comment when I was writing the other one (GitHub didn't refresh automatically). Well, I was unaware but figured that out eventually.

libwild/src/output_section_id.rs
```diff
@@ -375,6 +375,7 @@ pub(crate) struct BuiltInSectionDetails {
     pub(crate) link: &'static [OutputSectionId],
     pub(crate) start_symbol_name: Option<&'static str>,
     pub(crate) end_symbol_name: Option<&'static str>,
+    pub(crate) synthetic_symbol_names: Option<&'static [&'static str]>,
```
davidlattimore, line 378: Is there intended to be a difference between `None` and `Some(&[])` here? If not, then we could get rid of the `Option`.

Is this intended to be different to `start_symbol_name`? i.e. is this supposed to be start-symbols that have provide-semantics? I'm unsure about the naming, since these symbols seem just as synthetic as the start/end symbols. I guess until such time as we support provide-semantics, one option would be to change `start_symbol_name: `Option<&'static str>` to `start_symbol_names: &'static [&'static str]`. Then, once we support provide-semantics, split it out into two separate fields. e.g. `start_symbol_names` and `provide_start_symbol_names` or something like that.

mati865: That was just a proof of concept  because I'm not sure whether it's worth doing it.

The only difference from start symbols is the fact they shouldn't be provided by default, but it's not fulfilled yet.

The naming is based on LLD. I tried to find any docs regarding what those symbols are and LLD comments are probably the best (at least they exist). There I noticed they are referred as synthetic symbols. I'm not happy with that name though.

The next release is probably far away, so we can wait for provide. I don't think this is something of high priority.

mati865:
Let's wait until we can do it in a more reasonable way.

PR #1264, opened 2025-11-03
release: Release 0.7.0

davidlattimore:


davidlattimore:
Here's a draft of some release notes for our next release. I took the output of git-cliff and then edited extensively. Hopefully as we following conventional commits more going forward there will be less edited needed. There's probably more work needed on the git-cliff config and the changes to CONTRIBUTING.md. I may merge those separately.

Also, somewhat relatedly, I've disabled rebasing on the repo, so now the only option is to squash. I documented the motivation for this in CONTRIBUTING.md, but I'm open to discussion on this if anyone feels strongly. I've also set the default commit message for squashes to be the PR description. We had quite a few commits where the commit message was a list of the commits that got squashed, which in some cases is a bit noisy, especially when there's lot of fixups, so I thought we could have a try at using the PR description and see how that goes. In either case, it can be edited when squashing.

Anyway, what do you think of the groupings in the release notes? Any extra headings we should add (or remove)? I was unsure about porting. RISC-V work is kind of porting, but I guess I classified early-stage-porting as porting and later stage as just bug fixing. I did wonder if we should have separate headings for platform-specific work. e.g. a RISC-V heading, but then I realised that there's a lot of testing changes that were RISC-V specific and I wasn't sure if they would then go under RISC-V or under testing.

CONTRIBUTING.md
```diff
@@ -153,24 +153,36 @@ WILD_IGNORE_SKIP=mold cargo test --features mold_tests
...
+## Commit messages
+
+The title of your commit message should ideally be written with a view to it being included in the
+release notes. i.e. write what would be useful for a user of the linker to read. Implementation
+details can go in the extended description.
+
+When we write the release notes, we try to categorise commits into groups. You can help by prefixing
+the commit title with a word that gives a clue as to which section it should go in. If you think it
+shouldn't go in the release notes, you can prefix the title with "chore:" or "typo:", neither of
+which get included in the notes.
```
mati865: We could also filter by footer `changelog: skip` like in https://github.com/davidlattimore/wild/commit/921fa4b79a93311d52c6f7bc553126f903d993cc but that's up to you. Example of git-cliff config: https://github.com/davidlattimore/wild/pull/1256/files#diff-e1372c8b03c40942b5d828a90975054cb8aaed3b38189f434396f922ec41a584R90

mati865:
> Anyway, what do you think of the groupings in the release notes? Any extra headings we should add (or remove)? I was unsure about porting. RISC-V work is kind of porting, but I guess I classified early-stage-porting as porting and later stage as just bug fixing.

I view it the same way. Porting for me is creating the foundation, once something mostly works, it's no longer a porting work.

PR #1268, opened 2025-11-04
style: Wrap comments at 100 characters

daniel-levin:
This is a small stylistic change. I normally rely on Rustfmt's comment wrapping feature. Even though it's not a stable feature, it has worked well and stably for the last few years. Also, I added a second commit so that the changes affecting these 28 files can be excluded from `git blame`. See also: https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/

GitHub respects this file by default:
image

### DIFF
--- a/linker-diff/src/x86_64.rs
+++ b/linker-diff/src/x86_64.rs
@@ -284,9 +284,9 @@ impl Arch for X86_64 {
                     0xf2, 0xe9, 0, 0, 0, 0,    // bnd jmp {plt[0]}(%rip)
                     0x90, // nop
                 ];
-                // Note: Some variants use jmp instead of bnd jmp, then a different padding instruction.
-                // Because we use the index that gets pushed, we ignore the bytes of the later
-                // instructions, so that we support these variants.
+                // Note: Some variants use jmp instead of bnd jmp, then a different padding
+                // instruction. Because we use the index that gets pushed, we ignore the bytes of
+                // the later instructions, so that we support these variants.
                 if plt_entry[..5] == PLT_ENTRY_TEMPLATE[..5] {
                     let index = u32_from_slice(&plt_entry[5..]);
                     return Some(PltEntry::JumpSlot(index));
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -2,3 +2,5 @@ edition = "2024"
 use_field_init_shorthand = true
 imports_granularity = "Item"
 group_imports = "One"
+wrap_comments = true
+comment_width = 100
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -2009,12 +2009,12 @@ impl LinkCommand {
                         Compiler::Gcc(_) => {
                             match linker {
                                 Linker::Wild => {
-                                    // GCC unfortunately doesn't provide any way to use a custom linker.
-                                    // Their flag for switching linkers only accepts a hard-coded list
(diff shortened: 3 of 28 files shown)


linker-diff/src/x86_64.rs
```diff
@@ -284,9 +284,10 @@ impl Arch for X86_64 {
                     0xf2, 0xe9, 0, 0, 0, 0,    // bnd jmp {plt[0]}(%rip)
                     0x90, // nop
                 ];
-                // Note: Some variants use jmp instead of bnd jmp, then a different padding instruction.
-                // Because we use the index that gets pushed, we ignore the bytes of the later
-                // instructions, so that we support these variants.
+                // Note: Some variants use jmp instead of bnd jmp, then a different padding
+                // instruction. Because we use the index that gets pushed, we ignore
```
davidlattimore: Some of the wrapping decisions that it makes are a bit weird. "the bytes of" could have stayed on this line. On the up-side, if I reflow this paragraph with my editor, rustfmt doesn't undo my change. Perhaps you could look for any that stand out as being weird and reflow them?

wild/tests/integration_tests.rs
```diff
@@ -2009,11 +2009,13 @@ impl LinkCommand {
                         Compiler::Gcc(_) => {
                             match linker {
                                 Linker::Wild => {
-                                    // GCC unfortunately doesn't provide any way to use a custom linker.
-                                    // Their flag for switching linkers only accepts a hard-coded list
+                                    // GCC unfortunately doesn't provide any way to use a custom
+                                    // linker. Their flag for
```
davidlattimore: Another weird one here

rustfmt.toml
```diff
@@ -2,3 +2,5 @@ edition = "2024"
 use_field_init_shorthand = true
 imports_granularity = "Item"
 group_imports = "One"
+wrap_comments = true
+comment_width = 120
```
davidlattimore: I've used 100 as the wrapping width for comments since the start of the project. I've mostly used the rewrap extension in vscode to do this. So I think it'd be good to stick to what has been (mostly) used to date. Also, the default code width in rustfmt is 100, so it's good to be consistent with that.

Interestingly, when I just tried different settings for this, 120 and 100 produced exactly the same result. As far as I can tell, it's capping `comment_width` at the value of `max_width` (100).

davidlattimore (commented):
Thanks. I think this is a good idea.

I recently (as in yesterday) disabled rebases on the repo. My motivation was that if we always squash, then it's easier to track commits back to their PRs, since github adds the PR number to the commit message. So perhaps we could add `.git-blame-ignore-revs` in a separate PR once this one is merged?

daniel-levin:
> So perhaps we could add .git-blame-ignore-revs in a separate PR once this one is merged?

Sure, we can do that.

> I've mostly used the rewrap extension in vscode to do this. So I think it'd be good to stick to what has been (mostly) used to date.

Ah, I normally use Rustfmt to reflow my comments, because it's very good at it, and understands markdown and code embedded in markdown.

daniel-levin:
To fix the weirdly flowed cases, I put all the text onto the same line and then re-ran Rustfmt. It then produced sensible wrapping ([diff](https://github.com/davidlattimore/wild/compare/f6a9c5bace1a5ae792358071a93168afc3cd1f8c..c2aa466524802f42f29885c7a21ee8734a9284ed)) - including putting "the bytes of" in the right place. I don't know why this is.

PR #1272, opened 2025-11-04
AArch64: Support `variant_pcs`

lapla-cogito:
For AArch64, when using SVE or SIMD instructions, the types of floating-point registers saved differ from normal operation. To detect this, the `.variant_pcs` directive exists. While this feature appears to be primarily used by mathematical libraries like glibc, I think implementing it now still makes sense because failure to do so may cause the linker to silently produce invalid binaries.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -1081,7 +1081,7 @@ impl<'data> SymbolRequestHandler<'data> for DynamicLayoutState<'data> {
         }
 
         // Check for VARIANT_PCS flag in AArch64 symbols
-        if matches!(A::elf_header_arch_magic(), object::elf::EM_AARCH64)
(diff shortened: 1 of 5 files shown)


libwild/src/layout.rs
```diff
@@ -1086,6 +1079,17 @@ impl<'data> SymbolRequestHandler<'data> for DynamicLayoutState<'data> {
                     true;
             }
         }
+
+        // Check for VARIANT_PCS flag in AArch64 symbols
+        if matches!(A::elf_header_arch_magic(), object::elf::EM_AARCH64)
```
davidlattimore: I feel like this would be a little bit neater if we added an extra method or associated const to the `Arch` trait.

```rust
fn kind() -> Architecture;
```

or:

```rust
const KIND: Architecture;
```

PR #1281, opened 2025-11-08
feat: Support `--hash-style={sysv, both}`

lapla-cogito:
`--hash-style=sysv` is still widely used in many places. I found it being used in [gvisor builds](
https://github.com/google/gvisor/blob/cd0901c24964b819586498bd6d593e15b187d9e9/vdso/BUILD#L46).

Creating a `.hash` section requires determining the total number of entries in the `.dynsym` table before calculating the chain array length for the SysV hash table. The `update_existing_dynsym_entries` function performs this calculation.

Additionally, mold appears to also support `--hash-style=none`, but since no other linkers implement this feature and I couldn't find any practical use cases, I haven't implemented it this time (though adding it would be relatively straightforward).

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3827,12 +3827,11 @@ impl<'data> EpilogueLayoutState<'data> {
                 common.allocate(part_id::DYNAMIC, dynamic_entry_size as u64);
             }
 
-            self.update_existing_dynsym_entries(symbol_db, per_symbol_flags)?;
-
             if symbol_db.args.hash_style.includes_gnu() {
                 self.allocate_gnu_hash(common);
             }
             if symbol_db.args.hash_style.includes_sysv() {
+                self.update_existing_dynsym_entries(symbol_db, per_symbol_flags)?;
                 self.allocate_sysv_hash(common)?;
             }
 


libwild/src/layout.rs
```diff
@@ -2361,19 +2372,36 @@ impl<'data> GroupState<'data> {
...
+
             file_state.finalise_sizes(
                 &mut self.common,
                 symbol_db,
                 output_sections,
                 per_symbol_flags,
             )?;
         }
+
+        if let Some(epilogue) = self
```
davidlattimore, line 2390: Can you add a comment describing why this needs to be done separately? From just reading the code, I can't see why. The epilogue is always last, so from what I can see with both the old code and the new code, all the same calls to `finalise_sizes` will be made and in the same order.

libwild/src/layout.rs
```diff
@@ -3638,6 +3678,35 @@ impl<'data> EpilogueLayoutState<'data> {
         };
     }
 
+    fn update_existing_dynsym_entries(
```
davidlattimore, line 3687: It looks to me like this function is counting the number of non-definition dynamic symbols. i.e. the number of undefined dynamic symbols. I think this is the same as `dynsym_start_index` in `EpilogueLayoutState::finalise_layout`, so perhaps this function isn't needed?

libwild/src/layout.rs
```diff
@@ -2384,6 +2412,18 @@ impl<'data> GroupState<'data> {
...
+
+        if let Some(epilogue_layout) = files.iter().find_map(|file| {
+            if let FileLayout::Epilogue(epilogue) = file {
+                Some(epilogue)
+            } else {
+                None
+            }
+        }) {
+            let size = epilogue_layout.sysv_hash_size.unwrap_or(0);
+            *self.common.mem_sizes.get_mut(part_id::SYSV_HASH) = size;
```
davidlattimore, line 2427: Could this and the computation of sysv_hash_size move to `EpilogueLayoutState::finalise_sizes`?

davidlattimore (commented):
Thanks! That's great to fill this gap in our supported features :-)

libwild/src/layout.rs
lapla-cogito: `dynsym_start_index` is a value determined within `finalize_layout`, and while it ultimately becomes the same value, to determine the size of the SysV hash, we need to know the number of existing entries before `finalize_layout`. To handle this more efficiently, I believe redesigning the surrounding area would be necessary. I've noted this as a comment.

libwild/src/layout.rs
lapla-cogito: Added.

libwild/src/layout.rs
lapla-cogito: There're doubts about if this approach can correctly compute the hash table. This is because `finalise_sizes` doesn't have the final starting index for the `.dynsym` section yet, which could lead to incorrect calculation results. `finalize_layout` computes the number of buckets and chains after this index is finalized, then uses the `sysv_hash_size` function to calculate the actual byte size.

libwild/src/layout.rs
lapla-cogito: Sorry, I realized I didn't push correctly so I'm doing it again.

libwild/src/layout.rs
```diff
@@ -3750,7 +3827,14 @@ impl<'data> EpilogueLayoutState<'data> {
                 common.allocate(part_id::DYNAMIC, dynamic_entry_size as u64);
             }
 
-            self.allocate_gnu_hash(common);
+            self.update_existing_dynsym_entries(symbol_db, per_symbol_flags)?;
```
davidlattimore: I might do some refactoring after you merge this to see if I can remove the need for this method. In the meantime, could the call to this method be made conditional on `includes_sysv()`? Perhaps it can be moved into the if-block below?

lapla-cogito: Thanks! Done.

davidlattimore (approved):

PR #1284, opened 2025-11-10
feat: support negation ('^') in version script globs

marxin:
Apparently, there are various glob patterns in the `libstdc++` version script file and so we need to support it properly:

```
      std::co[^dln]*;
      std::cu[^r]*;
      std::error[^_]*;
      std::h[^a]*;
```

Fixes: #990

CC: @lapla-cogito


marxin:
Yeah, good suggestions, let's apply it.

PR #1287, opened 2025-11-10
feat: Support SFrame

### LINKED ISSUE #1025: Support SFrame
Running the tests on Arch Linux I'm getting failures in two of them after recent updates:
```
     Summary [   4.934s] 119 tests run: 117 passed, 2 failed, 0 skipped
        FAIL [   2.062s] wild-linker::integration_tests integration_test::program_name_02___trivial_main_c__
  stdout ───

    running 1 test
    test integration_test::program_name_02___trivial_main_c__ ... FAILED

    failures:

    failures:
        integration_test::program_name_02___trivial_main_c__

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 71 filtered out; finished in 2.06s

  stderr ───
    wild: /home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.wild
    ld: /home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.ld
    section.sframe
      wild Section missing
      ld OK

    segment.UNKNOWN_P_TYPE_1685382484.alignment
      wild
      ld 0x8

    segment.UNKNOWN_P_TYPE_1685382484.flags
      wild
      ld 0x4


    Error: Validation failed.
    Binary `/home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.wild`. Relink with:
    WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.wild /home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.save/run-with cargo run --bin wild --
     To revalidate:
    cargo run --bin linker-diff -- --wild-defaults --ignore 'section.rodata,section.data,section.data.alignment,section.rela.plt.link,section.sdata' --ref /home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.ld /home/mateusz/Projects/wild/wild/tests/build/trivial-main.c-clang-static-host.wild

        FAIL [   0.824s] wild-linker::integration_tests integration_test::program_name_48___cpp_integration_cc__
  stdout ───

    running 1 test
    test integration_test::program_name_48___cpp_integration_cc__ ... FAILED

    failures:

    failures:
        integration_test::program_name_48___cpp_integration_cc__

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 71 filtered out; finished in 0.82s

  stderr ───
    wild: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.wild
    ld: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.ld
    lld: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.lld
    section.sframe
      wild Section missing
      ld OK
      lld OK


    Error: Validation failed.
    Binary `/home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.wild`. Relink with:
    WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.wild /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.save/run-with cargo run --bin wild --
     To revalidate:
    cargo run --bin linker-diff -- --wild-defaults --ignore 'section.rodata,section.data,section.sdata,section.rodata.alignment,section.bss.alignment,dynsym.__stack_chk_guard.section' --ref /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.ld --ref /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.lld /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc-no-pie-host.wild

error: test run failed
```
Depends on: https://github.com/gimli-rs/object/pull/799

lapla-cogito:
[SFrame](https://sourceware.org/binutils/docs/sframe-spec.html) is a relatively new mechanism for storing frame information that occasionally appears in environments such as Arch Linux. I confirmed that all tests are green on Arch.

close #1025

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -81,6 +81,9 @@
 //!
 //! RequiresGlibcVersion:{version} Specifies the minimum version of glibc required for this test.
 //!
+//! RequiresSFrameBacktrace:{bool} Defaults to false. Set to true to disable this test if we're
+//! running on a system without sframe backtrace support.
+//!
 //! RequiresNightlyRustc:{bool} Defaults to false. Set to true to disable this test if we detect
 //! that the version of rustc available to us is not nightly.
 //!


lapla-cogito:
Results from testing with the `gcc-static` config in `trivial-main.c`:

```
（▰╹◡╹）❯ objdump --sframe wild/tests/build/trivial-main.c-gcc-static-host.wild

wild/tests/build/trivial-main.c-gcc-static-host.wild:     file format elf64-x86-64

Contents of the SFrame section .sframe:
  Header :

    Version: SFRAME_VERSION_2
    Flags: SFRAME_F_FDE_SORTED,
           SFRAME_F_FDE_FUNC_START_PCREL
    CFA fixed RA offset: -8
    Num FDEs: 1
    Num FREs: 1

  Function Index :

    func idx [0]: pc = 0x48b1f0, size = 5 bytes
    STARTPC         CFA       FP        RA
    000000000048b1f0  sp+8      u         f
```

```
（▰╹◡╹）❯ readelf -s wild/tests/build/trivial-main.c-gcc-static-host.wild | grep 48b1f0
   638: 000000000048b1f0     5 FUNC    GLOBAL HIDDEN    12 _dl_relocate_sta[...]
```

lapla-cogito:
Since current CI environments can't verify this feature, adding tests using Arch Docker might be worthwhile. Given its relative simplicity, I'd like to consider implementing this.

davidlattimore:
From looking at maskray's blog posts on the topic - [Stack walking: space and time trade-offs](https://maskray.me/blog/2025-10-26-stack-walking-space-and-time-trade-offs) and  [Remarks on SFrame](https://maskray.me/blog/2025-09-28-remarks-on-sframe) - I'd have expected that we'd need to do a few other things to support SFrames. In particular, it sounds like the entries need to be sorted by the addresses of the functions to which they refer.

It'd also be interesting to verify if `--gc-sections` works correctly when there are sframes. i.e. that the relocations in the sframe sections aren't preventing otherwise unused sections from being GCed.

linker-diff/src/lib.rs
```diff
@@ -249,12 +248,14 @@ impl Config {
                 "segment.GNU_RELRO.*",
                 "segment.GNU_STACK.*",
                 "segment.GNU_PROPERTY.*",
-                "segment.GNU_SFRAME.*",
                 // Wild currently generates PT_NOTE even for non-alloc note sections, while the
                 // other linkers don't.
                 "segment.NOTE.*",
                 // TODO: RISC-V
                 "segment.LOAD.RW.alignment",
+                // GNU ld and lld sometimes don’t generate .sframe sections in cases where we do.
```
mati865, line 259: It would be good to find out when and why it happens.
Maybe it's related to GC?

lapla-cogito:
OK, this should now allow us to add functionality for sorting `.sframe` sections. The results of running `objdump --sframe` on the binaries created with each config in `trivial-main.c` within Arch Docker are as follows:

<details><summary>output</summary>

```
[root@0fcee70c27b6 wild]# objdump --sframe wild/tests/build/trivial-main.c*.wild

wild/tests/build/trivial-main.c-clang-host.wild:     file format elf64-x86-64

No .sframe section present


wild/tests/build/trivial-main.c-clang-static-host.wild:     file format elf64-x86-64

Contents of the SFrame section .sframe:
  Header :

    Version: SFRAME_VERSION_2
    Flags: SFRAME_F_FDE_SORTED,
           SFRAME_F_FDE_FUNC_START_PCREL
    CFA fixed RA offset: -8
    Num FDEs: 1
    Num FREs: 1

  Function Index :

    func idx [0]: pc = 0x48b230, size = 5 bytes
    STARTPC         CFA       FP        RA
    000000000048b230  sp+8      u         f

wild/tests/build/trivial-main.c-clang-static-pie-no-relax-host.wild:     file format elf64-x86-64

Contents of the SFrame section .sframe:
  Header :

    Version: SFRAME_VERSION_2
    Flags: SFRAME_F_FDE_FUNC_START_PCREL
    CFA fixed RA offset: -8
    Num FDEs: 2
    Num FREs: 9

  Function Index :

    func idx [0]: pc = 0x95bd0, size = 176 bytes
    STARTPC         CFA       FP        RA
    0000000000095bd0  sp+8      u         f
    0000000000095bd1  sp+16     c-16      f
    0000000000095bd6  fp+16     c-16      f
    0000000000095bff  sp+8      c-16      f
    0000000000095c00  fp+16     c-16      f

    func idx [1]: pc = 0x95c80, size = 51 bytes
    STARTPC         CFA       FP        RA
    0000000000095c80  sp+8      u         f
    0000000000095c85  sp+16     c-16      f
    0000000000095c8f  fp+16     c-16      f
    0000000000095cae  sp+8      c-16      f

wild/tests/build/trivial-main.c-gcc-host.wild:     file format elf64-x86-64

No .sframe section present


wild/tests/build/trivial-main.c-gcc-indirect-external-host.wild:     file format elf64-x86-64

No .sframe section present


wild/tests/build/trivial-main.c-gcc-static-host.wild:     file format elf64-x86-64

Contents of the SFrame section .sframe:
  Header :

    Version: SFRAME_VERSION_2
    Flags: SFRAME_F_FDE_SORTED,
           SFRAME_F_FDE_FUNC_START_PCREL
    CFA fixed RA offset: -8
    Num FDEs: 1
    Num FREs: 1

  Function Index :

    func idx [0]: pc = 0x48b230, size = 5 bytes
    STARTPC         CFA       FP        RA
    000000000048b230  sp+8      u         f

wild/tests/build/trivial-main.c-gcc-static-pie-no-relax-host.wild:     file format elf64-x86-64

Contents of the SFrame section .sframe:
  Header :

    Version: SFRAME_VERSION_2
    Flags: SFRAME_F_FDE_FUNC_START_PCREL
    CFA fixed RA offset: -8
    Num FDEs: 2
    Num FREs: 9

  Function Index :

    func idx [0]: pc = 0x95bf0, size = 176 bytes
    STARTPC         CFA       FP        RA
    0000000000095bf0  sp+8      u         f
    0000000000095bf1  sp+16     c-16      f
    0000000000095bf6  fp+16     c-16      f
    0000000000095c1f  sp+8      c-16      f
    0000000000095c20  fp+16     c-16      f

    func idx [1]: pc = 0x95ca0, size = 51 bytes
    STARTPC         CFA       FP        RA
    0000000000095ca0  sp+8      u         f
    0000000000095ca5  sp+16     c-16      f
    0000000000095caf  fp+16     c-16      f
    0000000000095cce  sp+8      c-16      
```
</details>

However, there are several points of concern. All these issues have been documented as TODO comments in the code:
- To ensure proper sorting for all `.sframe` files, it appears we would need to make redundant modifications to `elf_writer::write_file_contents`. This would affect linking processes that don't require SFrames, which we wish to avoid. From my understanding, SFrames aren't strictly required to be sorted unless they carry the `SFRAME_F_FDE_SORTED` flag, so even in the current state, we consider this to meet the minimum requirements for SFrame support.
- As noted [here](https://github.com/davidlattimore/wild/pull/1287#discussion_r2512370502), we should clarify the specific causes of cases where Wild creates `.sframe` sections while GNU ld or lld do not. This doesn't seem to be due to section garbage collection, and I don't currently have any leads on the potential cause...

libwild/src/elf_writer.rs
```diff
@@ -221,6 +222,21 @@ fn write_file_contents<A: Arch>(sized_output: &mut SizedOutput, layout: &Layout)
             Ok(())
         })?;
 
+    let sframe_buffer = section_buffers.get_mut(output_section_id::SFRAME);
+    if !sframe_buffer.is_empty() {
+        // TODO: To reliably ensure `sort_sframe_section` is called for all `.sframe` sections, you
```
davidlattimore: This sounds a bit like what we currently do for `.eh_frame_hdr`. https://github.com/davidlattimore/wild/blob/1b15e1daacf414d5501b22c5154cbf1a88871849/libwild/src/elf_writer.rs#L138
Perhaps we can call the function to sort the sframe section at the same point?

libwild/src/sframe.rs
```diff
@@ -0,0 +1,147 @@
...
+            bytes,
+            func_addr,
+            original_index: index,
+        });
+    }
+
+    entries.sort_by(|a, b| {
+        a.func_addr
+            .cmp(&b.func_addr)
+            .then_with(|| a.original_index.cmp(&b.original_index))
```
davidlattimore: `sort_by` is a stable sort, so I don't think it should be necessary to sort by `original_index`, since the sort algorithm will already preserve the order for you where `func_addr` is equal.

libwild/src/sframe.rs
```diff
@@ -0,0 +1,147 @@
...
+const FDE_SIZE: usize = 20;
+const FDE_START_OFFSET_FIELD: usize = 0x14;
+const NUM_FDES_FIELD: usize = 0x08;
+const AUX_LENGTH_FIELD: usize = 0x07;
+const FLAGS_FIELD: usize = 0x03;
+const VERSION_FIELD: usize = 0x02;
+
+struct Entry {
+    bytes: [u8; FDE_SIZE],
+    func_addr: i128,
```
davidlattimore, line 36: Why i128? On a 64 bit platform, I'd expect that a u64 would be sufficient to represent all addresses

libwild/src/sframe.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    for (index, entry) in entries.into_iter().enumerate() {
+        let offset_in_section = fde_start + index * FDE_SIZE;
+        let new_value = if pc_rel {
+            entry.func_addr - (section_base + offset_in_section as i128)
+        } else {
+            entry.func_addr - section_base
+        };
+        let new_value_i32 = i32::try_from(new_value).with_context(|| {
+            "Function start address out of 32-bit range for SFrame entry".to_string()
```
davidlattimore: Looks like you can probably use `.context` instead of `.with_context`, then you won't need the closure or the `.to_string()`

davidlattimore (commented):
If there are two (or more) input objects containing sframe sections, do we emit an sframe section with multiple headers? Is it valid to do that, or do the sections need to be merged and have a single header?

Is there a consumer of sframes that we can test this with? I'm not sure what uses SFrames, perhaps gdb or libunwind? It'd be good confirm that it works with some consumer - ideally with eh_frames disabled so that we can be sure it's using the SFrames.

libwild/src/elf_writer.rs
lapla-cogito: Ah, I didn't notice that such processing existed there! Indeed, it would be good to sort the SFrame entries at the same location too, thanks.

libwild/src/sframe.rs
lapla-cogito: When the function appears before the FDE, `start_value` could be negative, so I took a conservative approach. While we could make `func_addr` a `u64` and use types like `i128` for intermediate calculations, that would require additional `u64::try_from()` conversions.

lapla-cogito:
To verify SFrame's behavior beyond what `objdump --sframe` provides, I conducted the following experiment on native Arch Linux:
1. Used `wild/tests/build/libc-integration.c-gcc-static-pie-host.wild` as the target binary (I copied this and named it `tmp.wild`)
2. Used `objdump --remove-section` to remove the `.eh_frame` and `.eh_frame_hdr` sections
3. Ran `perf record -e cycles -g ./tmp.wild` followed by `perf report` to verify that stack information was being captured correctly
(You can see `perf`'s support for SFrame with user events from links like this: https://lists.openwall.net/linux-kernel/2023/11/09/22)

First, here's the result of `perf report` without removing the eh_frame sections:
```
  Children      Self  Command          Shared Object     Symbol
+  100.00%   100.00%  tmp_before.wild  tmp_before.wild   [.] 0x00000000000a6af7
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] _start
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] __libc_start_main_impl
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] __libc_start_call_main
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] 0x00007f0990f10e0e
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] __run_exit_handlers
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] call_fini
+  100.00%     0.00%  tmp_before.wild  tmp_before.wild   [.] 0x00007f0990f0eaf7
```
And here's the result after removal:

```
  Children      Self  Command          Shared Object     Symbol
+  100.00%   100.00%  tmp.wild  tmp.wild          [.] 0x00000000000a6b18
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] _start
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] __libc_start_main_impl
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] __libc_start_call_main
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] 0x00007f3e19290e0e
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] __run_exit_handlers
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] call_fini
+  100.00%     0.00%  tmp.wild  tmp.wild          [.] 0x00007f3e1928eb18
```

thesamesam:
Note that the perf support for SFrames is still not yet upstream (but has been making recent progress). The only support is in glibc's `backtrace()` right now.

davidlattimore:
Thanks, that's good to know that glibc's backtrace supports it. I wrote a test - https://github.com/davidlattimore/wild/pull/1353 - the commented out part of the test removes eh_frames and relies on sframes instead. It fails at head and passes with this PR.

davidlattimore:
I just realised that my test wasn't testing multiple object files. Unfortunately, once I extended the test to have a second object file, it failed, presumably because the sframe sections need to be merged.

lapla-cogito:
While still in debugging, I pushed because I wanted to observe CI results across various distributions. The main changes include:
- Enhanced functionality to merge SFrame sections when multiple object files contain them. Since FRE address sizes and offset sizes may differ between input files, this implementation uses the maximum size among them to enable merging.
- Support for obtaining backtraces using SFrame has been available since [glibc 2.42](https://sourceware.org/binutils/wiki/sframe#Support_SFrame_based_stack_tracing_in_glibc_backtrace:~:text=SFrame%20support%20for%20backtrace%20()%20is%20available%20in%20glibc%202.42). To ensure compatibility, I added a `RequiresGlibcVersion` directive that checks the installed glibc version and skips the operation if it's lower than the specified version.

These changes allows the `backtrace.c` tests to pass on my Arch Linux ~~, but further investigation is needed with several distributions...~~ edit: Since failing tests other than those involving Clippy appear to be from GNU ld, it seems appropriate to set `SkipLinker:ld` for SFrame-related tests.

lapla-cogito:
To get CI green, I needed to take the following points into account, so I implemented the corresponding changes:

- Simply detecting in the test harness whether the glibc version is newer than 2.42 (added in https://github.com/davidlattimore/wild/pull/1287/commits/af3e4e235d3fc09080f5d619b120732945448ac4), where SFrame support was introduced, was not sufficient. This is because the glibc --enable-sframe option is false by default (see https://sourceware.org/git/?p=glibc.git;a=blob;f=configure.ac;hb=refs/heads/release/2.42/master#l445
). To address this, I added a `RequiresSFrameBacktrace` directive and now check, per architecture, whether the `.sframe` section is present in `libc.so.6`: https://github.com/davidlattimore/wild/pull/1287/commits/beeadbbabb843cb4c097d9aab7ec78226267c7f8
- However, this alone turned out to be insufficient (for example, see this run: https://github.com/davidlattimore/wild/actions/runs/20312855178/job/58348562000
). As a result, when the RequiresSFrameBacktrace directive is set, I additionally run a small program that uses SFrame internally to verify that SFrame-based unwinding actually works: https://github.com/davidlattimore/wild/pull/1287/commits/1b10d1be86be108332db86293d67162033d8bb01

There may be a better approach, but at least this one is reliable I think.

wild/tests/integration_tests.rs
```diff
@@ -79,6 +79,8 @@
 //! RequiresGlibc:{bool} Defaults to false. Set to true to disable this test if we're running on a
 //! system without glibc.
 //!
+//! RequiresGlibcVersion:{version} Specifies the minimum version of glibc required for this test.
+//!
```
davidlattimore, line 83: Can you document "RequiresSFrameBacktrace" too?

libwild/src/elf_writer.rs
```diff
@@ -134,11 +134,39 @@ pub(crate) fn write<A: Arch>(sized_output: &mut SizedOutput, layout: &Layout) ->
...
     }
 
+    let mut section_buffers = split_output_into_sections(layout, &mut sized_output.out);
+
     if layout.args().should_write_eh_frame_hdr {
-        let mut section_buffers = split_output_into_sections(layout, &mut sized_output.out);
         sort_eh_frame_hdr_entries(section_buffers.get_mut(output_section_id::EH_FRAME_HDR));
     }
 
+    let sframe_buffer = section_buffers.get_mut(output_section_id::SFRAME);
```
davidlattimore: Could you extract this code out to a separate function and add a timing phase for it?

libwild/src/sframe.rs
davidlattimore: Fair enough. i128 is more expensive to work with, but I'd be surprised if it made a measurable difference, so fine to leave it like that for now.

libwild/src/sframe.rs
```diff
@@ -0,0 +1,331 @@
...
+            })?;
+
+            let curr_fre_len = (next_fre_offset - curr_fre_offset) as usize;
+            let curr_fre_abs_start = fre_start + curr_fre_offset as usize;
+
+            if curr_fre_abs_start + curr_fre_len > offset + len {
+                bail!("SFrame FRE data truncated");
+            }
+
+            let mut fre_bytes = vec![0u8; curr_fre_len];
```
davidlattimore: Using `slice::to_owned` to create the Vec would avoid the need to zero-fill the vec.

libwild/src/elf_writer.rs
```diff
@@ -134,11 +134,39 @@ pub(crate) fn write<A: Arch>(sized_output: &mut SizedOutput, layout: &Layout) ->
...
+    if !sframe_buffer.is_empty() {
+        let sframe_start_address = layout.mem_address_of_built_in(output_section_id::SFRAME);
+        let mut sframe_ranges = Vec::new();
+
+        for group in &layout.group_layouts {
```
davidlattimore: I tried benchmarking linking the zed editor with and without this PR. With this PR, it links 3% slower. The output binary has an sframe section, but it's only 0x38 bytes in size. I'm guessing that some C dependency got compiled with sframes, but the majority of the code being Rust, didn't. It looks like the additional time is from this loop. Perhaps we could store a Vec of sframe sections on the group so that we don't need to scan all sections of all objects here?

davidlattimore (commented):
Nice work! This looks like it's really close to being ready to merge :)

wild/tests/integration_tests.rs
lapla-cogito: Ah, I totally forgot about it :) https://github.com/davidlattimore/wild/pull/1287/commits/73a5b753133ec7c8861c9b46930c8cd8fde83e3d

libwild/src/elf_writer.rs
lapla-cogito: Done: https://github.com/davidlattimore/wild/pull/1287/commits/08f1065d60b598e191473e492bbcb9f939bd22b0

libwild/src/sframe.rs
lapla-cogito: Done: https://github.com/davidlattimore/wild/pull/1287/commits/34bfe72ad0b146f5b195f168e7f8bab4a24b1afc

libwild/src/elf_writer.rs
lapla-cogito: Understood, but simply having GroupLayout store SFrame sections would mean scanning all files in a group even when only one object has a small SFrame section. Therefore, I moved this to ObjectLayout, collecting ranges during the existing section resolution loop: https://github.com/davidlattimore/wild/pull/1287/commits/4fe8ebcc7833c43f8042dddc41c87cff74aac2d4 and https://github.com/davidlattimore/wild/pull/1287/commits/9ff2e43f99c178bd5a33919eb708818a4175a0fa

davidlattimore (approved):

linker-diff/src/lib.rs
mati865: Yeah, did some testing and this is indeed related to GC.

GNU ld and LLD seem to always output `.sframe`, unless `--gc-sections` is added. Since Wild does that by default, you had to mark sframe sections as must keep.

Sure enough, all tests pass with this patch:
```diff
diff --git a/libwild/src/args.rs b/libwild/src/args.rs
index 0ff9701e7c..74b5c6a766 100644
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -379,7 +379,7 @@
             // but do end up writing more, so --no-gc-sections will almost always be as
             // slow or slower than --gc-sections. For that reason, the latter is
             // probably a good default.
-            gc_sections: true,
+            gc_sections: false,
             prepopulate_maps: false,
             sym_info: None,
             merge_sections: true,
diff --git a/libwild/src/layout_rules.rs b/libwild/src/layout_rules.rs
index e37d17d41f..44992fb295 100644
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -359,7 +359,7 @@
     SectionRule::exact(secnames::EH_FRAME_SECTION_NAME, SectionRuleOutcome::EhFrame),
     SectionRule::exact(
         secnames::SFRAME_SECTION_NAME,
-        SectionRuleOutcome::Section(SectionOutputInfo::keep(output_section_id::SFRAME)),
+        SectionRuleOutcome::Section(SectionOutputInfo::regular(output_section_id::SFRAME)),
     ),
     SectionRule::exact(
         secnames::NOTE_GNU_PROPERTY_SECTION_NAME,
diff --git a/linker-diff/src/lib.rs b/linker-diff/src/lib.rs
index b08bad44de..40e652d820 100644
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -257,9 +257,6 @@
                 "segment.GNU_PROPERTY.alignment",
                 "segment.GNU_PROPERTY.flags",
                 // GNU ld and lld sometimes don’t generate .sframe sections in cases where we do.
-                // TODO: Figure out why this is happening.
-                "segment.GNU_SFRAME.alignment",
-                "segment.GNU_SFRAME.flags",
             ]
             .into_iter()
             .map(ToOwned::to_owned),
```
I think, the current behaviour of always emitting sframe is fine. It doesn't slow us down much.

PR #1291, opened 2025-11-11
doc: Add some info about our monthly dev team meetings

davidlattimore:


CONTRIBUTING.md
```diff
@@ -15,6 +15,12 @@ You're also welcome to reach out directly to the following people:
   maintainer. I love talking about this stuff, so feel free to set up a video call to discuss.
   You're also welcome to PM me on the [rust-lang zulip](https://rust-lang.zulipchat.com/).
 
+## Meetings
+
+We try to have about one video call per month. To add this to your calendar, join the
+[wild-dev-meetings](https://groups.google.com/g/wild-dev-meetings) google group. If this doesn't
```
lapla-cogito, line 21: Does this group exist? I get a "Content unavailable" error with this link, and it doesn’t show up in my "My Groups" section.

mati865: I think it might be private or unlisted (I also cannot open the URL).

davidlattimore: Thanks. I had a permission wrong. Should be fixed now.

lapla-cogito: Thanks, I've confirmed that the link is working now.

PR #1293, opened 2025-11-11
refactor: decouple OutputKind from Args

mati865:


libwild/src/args.rs
```diff
@@ -2587,7 +2484,7 @@ mod tests {
     }
 
     fn input1_assertions(args: &Args) {
-        assert!(args.is_relocatable());
+        // assert!(args.is_relocatable());
```
mati865: I have no idea why those asserts were there.

libwild/src/symbol_db.rs
```diff
@@ -84,6 +85,9 @@ pub struct SymbolDb<'data> {
 
     /// The name of the entry symbol if overridden by a linker script.
     entry: Option<&'data [u8]>,
+
+    // FIXME
+    pub(crate) output_kind: OutputKind,
```
mati865, line 89: Putting it here seems suboptimal, but simplifies the refactor.

libwild/src/args.rs
```diff
@@ -1434,6 +1330,7 @@ fn setup_argument_parser() -> ArgumentParser {
             "nocopyreloc",
             "Disable copy relocations",
             |args, _modifier_stack, _value| {
+                // At this point we already know whether we output executable or a shared object.
```
davidlattimore: I'm not sure what relation this comment has to the copy relocations flag

libwild/src/args.rs
davidlattimore: I'm sure they made sense at the time for the test when it was originally written, but I'd say it's fine to just remove them if they don't make sense today.

libwild/src/symbol_db.rs
davidlattimore: I'm fine with having it here. We've already got `Args` on here. True, it does make the name `SymbolDb` not quite right, but the name still describes mostly what it does and seems better than some super-generic name like `LinkerContext`. Can you remove the FIXME?

davidlattimore (approved):
Thanks!

libwild/src/args.rs
mati865: Good catch. It got misplaced at some point, but I'll remove it altogether.

libwild/src/args.rs
```diff
@@ -119,9 +117,8 @@ pub struct Args {
     pub(crate) allow_multiple_definitions: bool,
     pub(crate) z_interpose: bool,
 
-    output_kind: Option<OutputKind>,
-    pub(crate) is_dynamic_executable: AtomicBool,
-    relocation_model: RelocationModel,
+    pub(crate) relocation_model: RelocationModel,
+    pub(crate) outputting_executable: bool,
```
mati865: This field could use a better name, but I haven't got anything fitting.

davidlattimore: Perhaps `is_output_executable` or `should_output_executable`. Another option would be to make this an enum. The only trouble is that the obvious name for such an enum is... `OutputKind`, which would mean we'd need a new name for the other `OutputKind`. Or maybe this one could be `OutputClass`.

PR #1297, opened 2025-11-13
feat: Support `PT_GNU_PROPERTY`

lapla-cogito:


libwild/src/output_section_id.rs
```diff
@@ -340,6 +339,10 @@ impl<'data> OutputSections<'data> {
 
         match segment_def.segment_type {
             pt::NOTE => info.ty == sht::NOTE,
+            pt::GNU_PROPERTY => {
+                info.ty == sht::NOTE
+                    && matches!(info.kind, SectionKind::Primary(name) if name.0 == NOTE_GNU_PROPERTY_SECTION_NAME)
+            }
```
davidlattimore: Why is this bit of code needed? It looks to me like the `other` branch below should handle this since the output section has `target_segment_type` set to `pt::GNU_PROPERTY`

lapla-cogito: Ah, sorry it looks like some experimental code I wrote while I was testing my implementation got left in.

davidlattimore (approved):

PR #1300, opened 2025-11-13
feat: Support `--defsym=symbol=expression`

lapla-cogito:
`--defsym` appears to be used in [many places](https://github.com/search?q=%22--defsym%3D%22&type=code). Moreover, once this feature is merged, we can begin extending linker scripts further by implementing support for symbol aliases within the scripts themselves.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       fail-fast: false
 
     runs-on: ${{ matrix.runs-on }}
+    timeout-minutes: 10
 
     container:
       image: ${{ matrix.container }}
@@ -64,7 +65,7 @@ jobs:
         if: ${{ contains(matrix.container, 'ubuntu') }}
       - run: apt-get update && apt-get -y install qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-riscv64-linux-gnu g++-riscv64-linux-gnu
         if: ${{ matrix.test-qemu }}
-      - run: zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap
+      - run: zypper refresh && zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup bubblewrap
         if: ${{ contains(matrix.container, 'opensuse') }}
       - run: apk add build-base lld clang clang-extra-tools bash curl
         if: ${{ contains(matrix.container, 'alpine') }}
@@ -93,8 +94,8 @@ jobs:
       - run: WILD_TEST_CROSS=$WILD_TEST_CROSS cargo test --profile ci --workspace
 
   minimal-versions:
-    runs-on:
-      - ubuntu-24.04
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
       - run: sudo apt-get update && sudo apt-get -y install gcc g++ clang clang-format lld curl
       - uses: actions/checkout@v5
@@ -118,6 +119,7 @@ jobs:
   clippy:
     name: Clippy
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v5
       with:
@@ -143,6 +145,7 @@ jobs:
   rustfmt:
     name: Check formatting
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v5
       with:
@@ -155,6 +158,7 @@ jobs:
   riscv-build:
     name: RISC-V build
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v5
       with:
@@ -169,6 +173,7 @@ jobs:
   spelling:
     name: Spell Check with Typos
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v5
         with:
@@ -179,6 +184,7 @@ jobs:
   calc-matrix:
     name: Find Nix checks
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     outputs:
       matrix: ${{ steps.calc-matrix.outputs.matrix }}
 
@@ -210,6 +216,7 @@ jobs:
       matrix:
         check: ${{ fromJson(needs.calc-matrix.outputs.matrix) }}
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     name: Nix/${{ matrix.check }}
 
     steps:
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -11,6 +11,7 @@ jobs:
   markdown-link-check:
     if: github.repository_owner == 'davidlattimore'
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v5
         with:
--- a/.github/workflows/external_tests.yml
+++ b/.github/workflows/external_tests.yml
@@ -18,6 +18,7 @@ jobs:
   external-tests:
     name: Run mold tests
     runs-on: ubuntu-24.04
+    timeout-minutes: 10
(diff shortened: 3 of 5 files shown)


libwild/src/layout.rs
```diff
@@ -375,16 +424,68 @@ fn get_epilogue_mut<'a, 'data>(
...
+                    .version_script
+                    .version_for_symbol(&UnversionedSymbolName::prehashed(name), version_name)?
+            {
+                version = v;
+                if !is_default {
+                    version |= object::elf::VERSYM_HIDDEN;
+                }
+            }
+
+            dynamic_symbol_definitions.push(DynamicSymbolDefinition::new(symbol_id, name, version));
```
lapla-cogito, line 497: Make symbols specified by defsym be exported to `.dynsym`.

libwild/src/elf_writer.rs
```diff
@@ -3113,6 +3124,44 @@ fn write_linker_script_dynsym(
...
+    ));
+
+    let resolution = layout
+        .local_symbol_resolution(symbol_id)
+        .with_context(|| format!("Missing resolution for {}", layout.symbol_debug(symbol_id)))?;
+    let address = resolution.raw_value;
+    let name = layout.symbol_db.symbol_name(symbol_id)?;
+
+    let entry = dynsym_writer
+        .define_symbol(false, object::elf::SHN_ABS, address, 0, name.bytes())
```
davidlattimore: Should this always be SHN_ABS? I'd have thought that for a symbol alias that it would refer to the section for the symbol it was aliasing?

libwild/src/layout.rs
```diff
@@ -298,6 +300,53 @@ pub fn compute<'data, A: Arch>(
     })
 }
 
+/// Update resolutions for defsym symbols that reference other symbols.
+fn update_defsym_symbol_resolutions(symbol_db: &SymbolDb, resolutions: &mut [Option<Resolution>]) {
+    let Some(Group::Prelude(prelude)) = symbol_db.groups.first() else {
+        return;
```
davidlattimore: If somehow the first group isn't the prelude, then that's a bug. Should we panic here? Perhaps use `unreachable!()`

libwild/src/layout.rs
```diff
@@ -3204,6 +3305,35 @@ impl<'data> PreludeLayoutState<'data> {
             );
         }
 
+        let needs_dynsym = resources.symbol_db.args.needs_dynsym();
```
davidlattimore, line 3327: Can this bit of code be extracted out into a separate function so as to keep this function from getting too large?

libwild/src/layout.rs
davidlattimore: Would it be possible to make the prelude populate `group.common.dynamic_symbol_definitions` like the other objects do so that this code isn't needed?

If not, could we extract this code into a separate function?

libwild/src/layout.rs
lapla-cogito: Thanks. Using `unreachable!()` seems reasonable.

libwild/src/layout.rs
lapla-cogito: Done.

libwild/src/layout.rs
lapla-cogito: Since `DefsymSymbol` doesn't have a corresponding section in the input object file, transferring this processing to the prelude would require passing this information back to `CommonGroupState`. In this context, I thought that performing this processing at this merge stage was acceptable. However, I also agree with splitting the code would improve readability so I did so. Thanks.

libwild/src/elf_writer.rs
lapla-cogito: Other linkers like GNU ld appear to treat any aliases defined for symbols via `--defsym` as simple absolute values. This implementation aligns with this behavior.
From the GNU ld man:
```
       --defsym=symbol=expression
           Create  a  global symbol in the output file, containing the absolute address given by expression.
```

davidlattimore: I just tried both GNU ld and lld and both of them put symbol aliases in the section for the symbol that was aliased. e.g.:

```
# gcc -fuse-ld=lld -Wl,--defsym=foobar=foo,--defsym=aaa1=42 -shared a.c -o liba.so

# readelf -W --dyn-sym liba.so

Symbol table '.dynsym' contains 8 entries:
     1: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
     2: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_deregisterTMCloneTable
     3: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND _ITM_registerTMCloneTable
     4: 0000000000000000     0 FUNC    WEAK   DEFAULT  UND __cxa_finalize
     5: 0000000000001649    15 FUNC    GLOBAL DEFAULT   10 foo
     6: 0000000000001649     0 FUNC    GLOBAL DEFAULT   10 foobar
     7: 000000000000002a     0 NOTYPE  GLOBAL DEFAULT  ABS aaa1
```

`aaa1` is in section `ABS`, however `foobar` which aliases `foo` is in section 10, the same as `foo`.

It could be worthwhile to add a test to check this behaviour. e.g. `//#ExpectSym:foobar section=".text"

lapla-cogito: Thanks. I've modified the implementation to inherit attributes from the referenced symbol when creating symbol aliases. Additionally, I've added test cases to the integration tests. Since I wanted to verify cases like `--defsym=${addr}`, I also added test directives such as `//#ExpectSym:bar address=0x12345678`.

libwild/src/parsing.rs
```diff
@@ -127,6 +128,13 @@ pub(crate) enum SymbolPlacement {
 
     /// An undefined symbol supplied by the user, e.g. via `--undefined=symbol-name`.
     ForceUndefined,
+
+    /// A symbol defined via --defsym with an absolute address.
+    DefsymAbsolute(u64),
+
+    /// A symbol defined via --defsym that references another symbol.
+    /// The SymbolId is initially None and will be filled in during layout.
+    DefsymSymbol(Option<SymbolId>),
```
davidlattimore: I can't see anything that populates the SymbolId, nor anything that uses it. If it's too tricky / messy to get the SymbolId - and i suspect it might be - then perhaps just store the target name here? 
The code in `update_defsym_symbol_resolutions` and `write_prelude_dynsym` both do linear scans over all the defsyms in order to find the one with the right name. Those could both be simplified if they already had the name.

libwild/src/elf_writer.rs
```diff
@@ -3207,24 +3331,66 @@ fn write_internal_symbols(
...
-            .transpose()?
-            .unwrap_or(0);
+                    } else {
+                        None
+                    }
+                });
+
+            if let Some(target_id) = target_symbol_id {
+                get_symbol_attributes(layout, target_id)?
+            } else {
```
davidlattimore, line 3342: I assume this block is for handling the case where the target symbol doesn't exist? It's not clear to me what the expected behaviour is though. I would have expected an error. Perhaps add a test for that error case? e.g. `//#LinkArg:--defsym=foo=doesnotexist`. Assuming we do want to raise an error for this case, consider if it can be done earlier than here - e.g. perhaps when the target symbol is requested, since the write phase is quite late to be raising such errors.

libwild/src/resolution.rs
```diff
@@ -661,7 +661,7 @@ fn load_prelude<'definitions>(
     // symbol record to the definition.
     for (def_info, definition_out) in prelude.symbol_definitions.iter().zip(definitions_out) {
         match def_info.placement {
-            SymbolPlacement::ForceUndefined => {
+            SymbolPlacement::ForceUndefined | SymbolPlacement::DefsymSymbol(_) => {
                 load_symbol_named(resources, definition_out, def_info.name, work_sender);
```
davidlattimore, line 665: Is `def_info.name` the right name to be loading here? I thought that was the alias and that you'd actually want to load the target symbol?

PR #1301, opened 2025-11-13
ci: Set `timeout-minutes` for CI jobs

lapla-cogito:
The exact cause remains unclear, but checks for `minimal-versions` sometimes appear to hang (e.g. https://github.com/davidlattimore/wild/actions/runs/19277294367/job/55120115423). GitHub Actions' default timeout is 6 hours - while retries can eventually pass these checks, catching such issues early is clearly beneficial. Therefore, we'll set `timeout-minutes` for each CI job. For now, we've configured all jobs with a 10-minute timeout, but based on actual execution times, a slightly shorter duration might be appropriate (though we've included some buffer to account for potential future CI time increases).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       fail-fast: false
 
     runs-on: ${{ matrix.runs-on }}
+    timeout-minutes: 10
 
     container:
       image: ${{ matrix.container }}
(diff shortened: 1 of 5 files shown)


karolzwolak:
I'm not sure exactly which test is the culprit — but testing on my machine also sometimes hangs (or takes **significantly** longer). Maybe we should investigate what's happening.

Issue #838, opened 2025-06-05
Add support for non-string merge sections


davidlattimore:
When we encounter an input section with the merge flag set, but without the strings flag set, we currently just ignore the merge flag. We should actually deduplicate that input section with other input sections with the same name.

Some links for additional context (thanks to Trevor Gross):

https://inbox.sourceware.org/binutils/CALNs47sfhiiCPi4o=otZ4k3nEt=byB=hv3yEowLO5rKU8CKt+Q@mail.gmail.com/T/#u

https://github.com/rust-lang/libs-team/issues/466

davidlattimore:
@karolzwolak 

To add some additional context, the existing string merging is done in `string_merging.rs`. I expect that the simplest option is likely to be to reuse that and have some code that skips splitting the section and instead uses the entire section as one string.

The string merging algorithm is kind of complex for performance reasons, but hopefully you won't need to touch most of it.

The function `part_id::should_merge_strings` will need to change. Mostly likely it'll need to ignore the `shf::STRINGS` flag. The strings flag would then need to be checked somewhere else, perhaps somewhere like `group_merge_string_sections_by_output`.

karolzwolak:
I"ve got hacky MVP working I think. Any tips on setting up the tests for this feature? Oh and please assign me to this. @davidlattimore

davidlattimore:
Great! For testing, I think either add to `string_merging.c` or add a new test that does something similar but with non-string sections. In `string_merging1.s`, where the section is declared `.section .rodata.strings, "aSM", @progbits, 1` you can do the same, but without the "S" bit. When declaring the strings - e.g. `.ascii "Hello\0"` you can remove the `\0` at the end.

karolzwolak:
As discussed with @davidlattimore there's no need to follow up the implementations with additional renaming so this should be closed as completed.

davidlattimore:
Thanks for implementing this!

PR #1306, opened 2025-11-15
feat: support merging non-string sections

karolzwolak:
Implements #838.

### DIFF
--- a/libwild/src/string_merging.rs
+++ b/libwild/src/string_merging.rs
@@ -1,6 +1,9 @@
 //! Input sections that are marked as string-merge sections need special processing. Our algorithm
 //! is somewhat complicated in an attempt to get good performance. A rough outline of our algorithm
-//! is here with more details throughout the code.
+//! is here with more details throughout the code. Contrary to what the name might suggest, this
+//! algorithm also supports merging non-string sections. The only difference between handling string
+//! and non-string sections is we split the former into multiple slices at the null terminators, and
+//! treat the latter as a single slice.
 //!
 //! We group input sections by the output section into which they are to be placed. We then process
 //! each output section one at a time.
@@ -10,11 +13,11 @@
 //!
 //! With multiple threads, we alternate between two phases:
 //!
-//! Phase 1: We split input sections by looking for null terminators, then we hash the resulting
-//! string and store it in a bucket based on its hash.
+//! Phase 1: We take the whole input sections or split string sections by looking for null
+//! terminators, then we hash the resulting slices and store it in a bucket based on its hash.
 //!
-//! Phase 2: We take the outputs of phase 1 and insert the strings into a hashmap for the bucket
-//! that the string is in. As we do this, we compute bucket-relative offsets for each string and
+//! Phase 2: We take the outputs of phase 1 and insert the slices into a hashmap for the bucket
+//! that the slice is in. As we do this, we compute bucket-relative offsets for each string and
 //! store these into entries in a map that we set up in phase 1.
 //!
 //! Threads can switch between phases multiple times until all work for the section is complete. At


karolzwolak:
This is WIP but I wanted to get some feedback.
Things left to do:
* update all the types and docs when necessary (this is only partially done atm)
* add clearer separation between string and non-string sections in code to be less confusing and more efficient

libwild/src/string_merging.rs
```diff
@@ -319,7 +326,11 @@ fn process_input_section<'data, 'offsets>(
     let mut input_offset = input_section.start_input_offset;
     let mut remaining = input_section.section_data;
     while !remaining.is_empty() {
-        let string = MergeString::take_hashed(&mut remaining)?;
+        let string = if input_section.is_string {
```
karolzwolak: The non-string merge sections shouldn't even reach this path — there's no preprocessing necessary.

libwild/src/string_merging.rs
```diff
@@ -122,6 +125,9 @@ struct StringMergeInputSection<'data> {
 
     /// The sum of the sizes of the input sections prior to this one with the same `part_id`.
     start_input_offset: LinearInputOffset,
+
+    // TODO: remove this flag and store non string sections elsewhere
+    is_string: bool,
```
karolzwolak, line 132: This probably isn't needed. We should just separate the non-string sections to avoid preprocessing. Not sure yet which type to change to achieve that though.

wild/tests/sources/non_string_merging1.s
```diff
@@ -0,0 +1,43 @@
+.section .rodata.foo, "aM", @progbits, 1
+.align 1
+
+.globl s1h
+s1h: .ascii "Hello"
```
davidlattimore, line 5: It could be good to put some null bytes within one of the strings, just to verify that this works correctly.

libwild/src/string_merging.rs
davidlattimore: I think it's probably fine to leave this here. I'm not noticing any change in memory consumption with this change.

libwild/src/string_merging.rs
davidlattimore: They don't need splitting, but they do still need to be hashed.

I would have been concerned about the performance impact on string-merging of doing this check inside this loop, but when I benchmarked, I couldn't see any change, so perhaps the compiler was smart enough to move this check outside the loop, with two separate versions of the loop in each branch. Or maybe the CPU's branch prediction means that the check becomes insignificant. Anyway, I'm fine with leaving this as-is if it's convenient.

karolzwolak: Yeah I did that because it was convenient. I refactored it slightly and moved the check outside of the loop.

libwild/src/string_merging.rs
karolzwolak: Okay. We can always refactor it later if it turns out it's an issue.

karolzwolak:
Glad to hear that. I added a few commits that addressed you comments and did minor refactor and improved docs.

karolzwolak:
The ubuntu 24.04 on aarch64 check failure looks like some CI issue.

```
Get:12 http://ports.ubuntu.com/ubuntu-ports noble-updates/universe arm64 Packages [1865 kB]
Err:12 http://ports.ubuntu.com/ubuntu-ports noble-updates/universe arm64 Packages
  File has unexpected size (1864981 != 1864881). Mirror sync in progress? [IP: 91.189.91.103 80]
  Hashes of expected file:
   - Filesize:1864881 [weak]
   - SHA256:35183757127012ad30f38ba2ca56973aadbbb45f0905438203b93cfce75e4630
   - SHA1:f435e67f90a4dfa54287c3e31b173e55f6653ca9 [weak]
   - MD5Sum:637252ef5d5162dd76d0f6117f8a673a [weak]
  Release file created at: Sun, 16 Nov 2025 10:56:23 +0000
Fetched 34.6 MB in 4s (8310 kB/s)
Reading package lists...
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/dists/noble-updates/universe/binary-arm64/Packages.gz  File has unexpected size (1864981 != 1864881). Mirror sync in progress? [IP: 91.189.91.103 80]
   Hashes of expected file:
    - Filesize:1864881 [weak]
    - SHA256:35183757127012ad30f38ba2ca56973aadbbb45f0905438203b93cfce75e4630
    - SHA1:f435e67f90a4dfa54287c3e31b173e55f6653ca9 [weak]
    - MD5Sum:637252ef5d5162dd76d0f6117f8a673a [weak]
   Release file created at: Sun, 16 Nov 2025 10:56:23 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
Error: Process completed with exit code 100.
```

mati865:
Yeah, AArch64 GHA runners do that sometimes.

davidlattimore (approved):

PR #1310, opened 2025-11-16
fix: put common TLS symbols into correct section

### LINKED ISSUE #1303: GCC testcase failure: emutls-3.c crashes
https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/gcc.dg/tls/emutls-3.c

```
gcc /home/marxin/BigProgramming/gcc/gcc/testsuite/gcc.dg/tls/emutls-3.c -B ~/Programming/wild -no-pie && ./a.out
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/15.2.1/liblto_plugin.so is not yet supported
Segmentation fault         (core dumped) ./a.out
```

Apparently, LLD rejects the test-case with the following error message:
```
ld.lld: error: a-emutls-3.o has an STT_TLS symbol but doesn't have a PT_TLS segment
```

mati865:
Code like this is probably a niche scenario and the issue manifests only with dynamic executables, causing them to crash.

Fixes #1303


mati865:
Thanks.

PR #1317, opened 2025-11-17
fix: Track RISC-V HI20/LO12 relaxations in symbol sizes

lapla-cogito:
In RISC-V, certain 32-bit or larger immediate values can be represented using HI20/LO12 pairs in instructions like `auipc` or `addi`, where the linker can apply relaxations optimizations. When relaxations reduces an instruction by 4 bytes or 2 bytes, the actual number of bytes occupied by the symbol also decreases. The linker should reflect this size reduction in writing symbols to `.symtab` and `.dynsym`. This PR implements the following changes:

- Introduced a RISC-V-specific `RelaxRecorder` in `layout.rs` to track the number of bytes saved for each HI20/LO12 pair
- Added new hooks in `riscv64.rs` to detect HI20 variants and determine whether compressed instructions are applicable
- Modified `elf_writer.rs` to overwrite the sizes written to `.symtab`/`.dynsym` with the relaxed values

### DIFF
--- a/wild/tests/external_tests/mold_skip_tests.toml
+++ b/wild/tests/external_tests/mold_skip_tests.toml
@@ -162,9 +162,7 @@ tests = [
   "arch-riscv64-obj-compatible.sh",
   "arch-riscv64-relax-align.sh",
   "arch-riscv64-relax-got.sh",
-  "arch-riscv64-relax-hi20.sh",
   "arch-riscv64-reloc-overflow.sh",
-  "arch-riscv64-symbol-size.sh",
 ]
 
 [skipped_groups.tls]


davidlattimore:
Nice! It'd be good to get an opinion from @marxin 

My only worry is the extent to which this might slow down linking on risc-v, which is already not super-fast as it is. I wonder if we need to store info about each relaxation into a Vec or if we could work stuff out as we process each relaxation. Or maybe store less in the Vec and avoid scanning the Vec multiple times. e.g. each time a relaxation is applied, write just the offset and number of bytes saved into a Vec, then sort the Vec by offset and do binary searches to find the range that applies to the current function. Or maybe instead store the function offsets and sizes in a Vec, then as each relaxation is applied, find the function that owns those bytes and decrement its size.

That said, I'm OK with going with a slow approach and optimising later if that's the consensus.

lapla-cogito:
> avoid scanning the Vec multiple times

`ObjectLayout::riscv_adjusted_symbol_size` is called at most once per symbol, and this call only processes relaxed HI20/LO12 pairs, so it will never be called more than this maximum number of times.

> e.g. each time a relaxation is applied, write just the offset and number of bytes saved into a Vec, then sort the Vec by offset and do binary searches to find the range that applies to the current function

Doesn't this approach eventually require information like `start` and `end` like in `ObjectLayout::riscv_adjusted_symbol_size` from `symbol_id`? I think this could introduce unnecessary complexity.

> store the function offsets and sizes in a Vec, then as each relaxation is applied, find the function that owns those bytes and decrement its size

This is purely speculative, but this method likely involves preparing a shared vector to store symbol information during layout processing. Accessing this vector simultaneously from multiple threads might actually negatively impact parallelism.

marxin:
What an interesting relaxation :) Well, there are 2 aspects to the relaxations:

1. we can potentially reduce the symbol table - am curious how much can we save on an average binary?
1. we can simplify the instruction stream, however, here we can't actually shrink the code size (unless we implemented a full support of RISC-V relaxation: https://maskray.me/blog/2021-03-14-the-dark-side-of-riscv-linker-relaxation)

> riscv_adjusted_symbol_size
> // Sum up all the bytes saved by RISC-V relaxations in this symbol.

Can you elaborate on why you sum `riscv_entry_bytes_saved` for a particular symbol? My expectation here would be that you'll find the smallest possible saving (for a symbol). Or do i miss something?

As @davidlattimore said, I am also worried about the run-time aspect of the relaxation and would like to see some numbers.

lapla-cogito:
> 1. we can potentially reduce the symbol table - am curious how much can we save on an average binary?
> 2. we can simplify the instruction stream, however, here we can't actually shrink the code size (unless we implemented a full support of RISC-V relaxation: https://maskray.me/blog/2021-03-14-the-dark-side-of-riscv-linker-relaxation)

This PR is more focused on accurately recording the symbol size reduction caused by relaxation into the symbol table's metadata than on implementing relaxation itself. Therefore, it's best to observe the actual size reduction achieved by implementing additional relaxation methods. Moreover, for this reason, incorporating this fix before implementing full relaxation proves valuable.

> Can you elaborate on why you sum riscv_entry_bytes_saved for a particular symbol? My expectation here would be that you'll find the smallest possible saving (for a symbol). Or do i miss something?

Based on my understandingAccording to my understanding, the size for each HI20/LO12 pair within a specific function can be reduced. To properly reflect this in the symbol table, I believe we need to obtain the total of the reduced sizes.

Additionally, I'll try measuring some performance metrics over the weekend. Since I have a Milk-V Mars board, I'll be using that for this.

lapla-cogito:
Here are the performance measurement results. I conducted measurements for Wild with debug builds. `wild.before` is `7ef64ed` and `wild.after` is `fc61d0a`:
```
user@milkv:~/repos/wild$ hyperfine --warmup 2 '/tmp/wild/wild/30/run-with ~/wild.before' '/tmp/wild/wild/30/run-with ~/wild.after'
Benchmark 1: /tmp/wild/wild/30/run-with ~/wild.before
  Time (mean ± σ):      1.832 s ±  0.067 s    [User: 0.016 s, System: 0.003 s]
  Range (min … max):    1.750 s …  1.983 s    10 runs

Benchmark 2: /tmp/wild/wild/30/run-with ~/wild.after
  Time (mean ± σ):      1.813 s ±  0.047 s    [User: 0.009 s, System: 0.010 s]
  Range (min … max):    1.732 s …  1.898 s    10 runs

Summary
  '/tmp/wild/wild/30/run-with ~/wild.after' ran
    1.01 ± 0.05 times faster than '/tmp/wild/wild/30/run-with ~/wild.before'
```

In addition to this test, I tried the same command several times afterward, and while the patched version occasionally produces outliers, it's not clear whether these are directly caused by this specific change...

```
user@milkv:~/repos/wild$ hyperfine --warmup 2 '/tmp/wild/wild/30/run-with ~/wild.before' '/tmp/wild/wild/30/run-with ~/wild.after'
Benchmark 1: /tmp/wild/wild/30/run-with ~/wild.before
  Time (mean ± σ):      2.091 s ±  0.736 s    [User: 0.016 s, System: 0.004 s]
  Range (min … max):    1.797 s …  4.182 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: /tmp/wild/wild/30/run-with ~/wild.after
  Time (mean ± σ):      2.104 s ±  0.762 s    [User: 0.009 s, System: 0.010 s]
  Range (min … max):    1.723 s …  4.205 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  '/tmp/wild/wild/30/run-with ~/wild.before' ran
    1.01 ± 0.51 times faster than '/tmp/wild/wild/30/run-with ~/wild.after'
```

davidlattimore:
That's great that it doesn't slow down linking. I'll defer to Martin, but I did have one question... From what I understand, this PR adjusts the sizes of the functions in the symtab, but do these size adjustments reflect changes that we currently make to the functions? i.e. are the functions reduced in size by this much? Because if they're not, then that'd mean that if you're stepping through the function in gdb, I think it'd say that the function ended earlier than it did and then there's be some extra instructions that gdb didn't know what they were.

davidlattimore:
To clarify my comment about gdb, I'm thinking of what happens if you're stepping through the assembly of a function without debug info.

I guess what I'm wondering is, are we changing the symtab function sizes to a size that the function would have been had we applied relaxations, but we didn't actually apply those relaxations, so the true size of the function is actually larger? If so, then I'm not sure that we should be doing that. i.e. we should only reduce the size of the function for relaxations that we apply. Ideally the code that applies the relaxations would aid in computing the function size reduction - e.g. by recording the offsets and sizes of all reductions.

marxin:
> Moreover, for this reason, incorporating this fix before implementing full relaxation proves valuable.

Sure, so then please collect some numbers about how much space can be potentially saved? Btw. do we know if other linkers support the described relaxation?

> Based on my understandingAccording to my understanding, the size for each HI20/LO12 pair within a specific function can be reduced. To properly reflect this in the symbol table, I believe we need to obtain the total of the reduced sizes.

Yes, we can theoretically shrink functions (as opposed to the other supported platforms), but we would have to support e.g. `R_RISCV_ALIGN` (and probably other things as well - described in the MaskRay's blog post). That's a non-trivial effort. What I've taken from the post is that it's probably better to sacrifice a bit of disk space and rather use `-fno-relax` - that way one ends up with a much smaller number of relocations. The ability to shrink a function goes against the linking speed as a codegen has to emit much more local labels (for relocation purposes) from e.g. debug info sections.

marxin:
@lapla-cogito Can we close this? Or please reply to my previous comment so that we can make an agreement about what's the final goal of the PR. Thanks.

lapla-cogito:
As I briefly mentioned during our recent call, RISC-V’s relaxation differs from other architectures in that it can actually reduce byte sizes. The current Wild architecture resolves symbol addresses during layout, so performing this reduction during symbol emission would create discrepancies between the addresses in the actual code and the addresses calculated during layout, resulting in invalid binaries.

While we could address this by running layout twice, it would slow down linking. So ideally, we should handle byte-size reduction during layout, though it’s not yet clear how best to implement that.

That said, since this has already caused conflicts, I don’t have any strong objections to closing this PR.

PR #1318, opened 2025-11-19
Nix: Update for inclusion in Nixpkgs

RossSmyth:
[Wild is now in Nixpkgs](https://search.nixos.org/packages?channel=unstable&show=wild&query=wild), which is great. So in this PR I move some stuff around because of that.

1. Bump the lock file
2. Have the overlay set `wild-unwrapped` as Wild needs to be wrapped to find libraries correctly. The wrapper is in Nixpkgs.
3. Run the tests that are in Nixpkgs (these are the same tests as were in the Wild repo before)
4. Remove the stdenv adapter from the Wild repo
5. Remove the two seperate overlay files and centralize to just one
6. Move the GCC wrappers to their own file

Then there is a commit that does several things all at once since there's not a great way to split most of them up:
1. Ensure all accesses of Wild are done through the overlay
2. Move the check inputs into the derivation
3. Remove some of the hacks for running unit tests by using `stdenvNoCC` when running unit tests

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,39 +11,39 @@
       crane,
     }:
     let
+      # Generate an output for each flake-exposed system. Flakes suck.
       forAllSystems = nixpkgs.lib.genAttrs nixpkgs.lib.systems.flakeExposed;
 
-      common = forAllSystems (system: rec {
+      # Make an attribute-set that instances Nixpkgs with our overlay for each
+      # system
+      common = forAllSystems (system: {
         pkgs = import nixpkgs {
           inherit system;
           overlays = [
             (import self)
           ];
         };
-
-        craneLib = crane.mkLib pkgs;
       });
     in
     {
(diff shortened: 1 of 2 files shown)


davidlattimore:
Thank! Looks good. There's a typo in one of the files - see CI failure.

I've recently changed the repo to only allow squash merges, since it makes it easier to write the release notes because all commits link back to the relevant PR. Is this PR OK to squash, or should we figure out an alternative?

RossSmyth:
> Is this PR OK to squash, or should we figure out an alternative?

Yeah this is fine to squash. 

@dawnofmidnight @Noratrieb if yall have any input :ferrisclueless:

whispersofthedawn:
I have no particular comments. This generally seems reasonable to me, but this interacts with a lot of machinery I don't really understand, so I can't offer anything particularly meaningful.

PR #1320, opened 2025-11-20
feat: handle `-z x86-64-*` ISA needed

mati865:


libwild/src/args.rs
```diff
@@ -1380,6 +1386,38 @@ fn setup_argument_parser() -> ArgumentParser {
...
+            },
+        )
+        .sub_option(
+            "x86-64-v4",
+            "Mark x86-64-v4 ISA as needed",
+            |args, _modifier_stack, _value| {
+                args.z_isa = GNU_PROPERTY_X86_ISA_1_V4;
+                Ok(())
+            },
+        )
```
mati865, line 1422: Should we list them when on non x86_64?

libwild/src/args.rs
```diff
@@ -116,6 +120,7 @@ pub struct Args {
     pub(crate) error_unresolved_symbols: bool,
     pub(crate) allow_multiple_definitions: bool,
     pub(crate) z_interpose: bool,
+    pub(crate) z_isa: u32,
```
mati865: Maybe `Option<NonZeroU32>` would be better?

libwild/src/args.rs
davidlattimore: We support cross-linking, so I think we should list them

libwild/src/args.rs
davidlattimore: `Option<NonZeroU32>` looks like it would fit nicely here, but either is fine by me.

davidlattimore (approved):

PR #1322, opened 2025-11-21
Update/Fix: nix documentation

beckend:
The old approach is broken.
This one works.

### DIFF
--- a/nix/nix.md
+++ b/nix/nix.md
@@ -7,38 +7,34 @@ your flake inputs. A devShell example is also shown with the flake.
 ```nix
 {
   inputs = {
-    nixpkgs.url = "github:NixOS/nix/nixos-unstable";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     wild = {
       url = "github:davidlattimore/wild";
       inputs.nixpkgs.follows = "nixpkgs";
     };
   };
 
-  outputs =
-    {
-      self,
-      nixpkgs,
-      wild,
-    }:
+  outputs = { self, nixpkgs, wild }:
     let
-      pkgs = import nixpkgs {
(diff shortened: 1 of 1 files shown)


beckend:
Applied `forAllSystems`

davidlattimore:
@RossSmyth - any thoughts?

RossSmyth:
There was a typo in the old docs, but overall this change does not do what it should. This pulls down a stable Wild linker built from Nixpkgs. The documentation in this repo should show how to use an unstable Wild built from this repo.

`forAllSystems` isn't usually required in downstream leaf packages because users generally only use one, maybe two systems in reality. The examples here should be as clear and as simple as possible to help users do what they need without much confusion.

RossSmyth:
Wait I don't even know what this change is supposed to do? 

The old docs showed how to use Wild with an arbitrary user-defined Nix package. All this update does is create a flake that outputs Wild from Nixpkgs.

davidlattimore:
Thanks for taking a look. I guess there are different use-cases for what people want to do with wild on nix. Perhaps it would help to update the docs with a section for each use-case

RossSmyth:
Yeah, I can add a section for using released version from Nixpkgs (probably in the top-level README), and using the unstable Wild built in-repo.

beckend:
I will leave this one to you.

beckend:
Just to clarify, if copied the current flake docs. It doesn't even work.

RossSmyth:
Yes, thanks for letting me know. I will update them today.

davidlattimore:
Thanks @beckend for raising this. If you have any comments on https://github.com/davidlattimore/wild/pull/1326 feel free to add them, either before or after submission

PR #1324, opened 2025-11-22
Update rustc building instructions & format all the rustflags consistently

karolzwolak:
I've updated instructions on building rustc with wild linker because rust-lang/rust#148795 got merged.
This makes you can now put rustflags in the `bootstrap.toml` just like in `.cargo/config.toml`. This should means building rustc with wild should finally be both easy & reliable.
This is a follow up to #1283 which turned out to not work for everyone as on some systems took `clang` took `lld` not from PATH but from other some specific path.

Additionally it was bothering that every instance of rustflags looks different so I made them all consistent.

### DIFF
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -236,10 +236,10 @@ PATH="$WILD_REPO_PATH/target/release:$PATH" WILD_SAVE_BASE=/tmp/rustc-link ./x b
 ```
 
 For more information about building rustc see [building instructions on the rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html).
-You should now have a few subdirectories under `$HOME/tmp/rustc-link`. You can identify which one is
+You should now have a few subdirectories under `/tmp/rustc-link`. You can identify which one is
 `rustc_driver` by looking at the last line of the `run-with` script in each directory.
 
-If the directory `$HOME/tmp/rustc-link` didn't get created, then most likely wild wasn't used to
+If the directory `/tmp/rustc-link` didn't get created, then most likely wild wasn't used to
 link.
 
 ### Other tools


BENCHMARKING.md
```diff
@@ -223,25 +220,22 @@ interesting benchmark because it's a shared object rather than an executable.
...
+To build rustc just cd into the rust repo root and run:
 
-Cd into the rust repo root and run:
 ```sh
-touch compiler/rustc_driver/src/lib.rs
-PATH=$WILD_REPO_PATH/fakes:$PATH echo "Using $(ld.lld --version) to link rustc"
-PATH=$WILD_REPO_PATH/fakes:$PATH WILD_SAVE_BASE=$HOME/tmp/rustc-link ./x build --keep-stage 1
-# verify that wild was used to link rustc
-readelf -p .comment build/x86_64-unknown-linux-gnu/stage1/bin/rustc | grep -i 'linker:'
+PATH="$WILD_REPO_PATH/target/release/wild:$PATH" WILD_SAVE_BASE=/tmp/rustc-link ./x build rustc
```
davidlattimore: `target/release/wild` is a file. I think PATH only supports directories

karolzwolak: oh yeah right

karolzwolak: I fixed it

BENCHMARKING.md
```diff
@@ -223,25 +220,22 @@ interesting benchmark because it's a shared object rather than an executable.
...
-touch compiler/rustc_driver/src/lib.rs
-PATH=$WILD_REPO_PATH/fakes:$PATH echo "Using $(ld.lld --version) to link rustc"
-PATH=$WILD_REPO_PATH/fakes:$PATH WILD_SAVE_BASE=$HOME/tmp/rustc-link ./x build --keep-stage 1
-# verify that wild was used to link rustc
-readelf -p .comment build/x86_64-unknown-linux-gnu/stage1/bin/rustc | grep -i 'linker:'
+PATH="$WILD_REPO_PATH/target/release:$PATH" WILD_SAVE_BASE=/tmp/rustc-link ./x build rustc
 ```
 
+For more information about building rustc see [building instructions on the rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/building/how-to-build-and-run.html).
 You should now have a few subdirectories under `$HOME/tmp/rustc-link`. You can identify which one is
```
davidlattimore: The command above now uses `/tmp/rustc-link`. I pushed a commit to your branch to fix it here and a few lines below.

PR #1326, opened 2025-11-22
Nix: docs

RossSmyth:
1. Add a short blurb in the top-level readme on how to use released Wild versions from Nixpkgs
2. Add much more detail to the Nix docs on how to use Wild with Nix in different scenarios

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ in
 pkgs.callPackage ./package { stdenv = wildStdenv; }  
 ```
 
-to use the latest unstable git revision of wild, see [./nix/nix.md](the nix documentation)
+to use the latest unstable git revision of wild, see [the nix documentation](./nix/nix.md)
 
 ## Using as your default linker
 


README.md
```diff
@@ -37,7 +37,16 @@ cargo install --locked --bin wild --git https://github.com/davidlattimore/wild.g
...
+To use a stable Wild from Nixpkgs:
+
+```nix
+let
+ wildStdenv = pkgs.useWildLinker pkgs.stdenv;
+in
+pkgs.callPackage ./package { stdenv = wildStdenv; }  
+```
+
+to use the latest unstable git revision of wild, see [./nix/nix.md][the nix documentation]
```
lapla-cogito:
```suggestion
to use the latest unstable git revision of wild, see [./nix/nix.md](the nix documentation)
```

nix/nix.md
```diff
@@ -42,7 +63,7 @@ your flake inputs. A devShell example is also shown with the flake.
     };
 }
 ```
-Without flakes (npins shown):
+Without flakes (npins shown, but any solution can be used):
 
 1. `$ npins add github davidlattimore wild -b main`
```
lapla-cogito: This isn't directly related to the current changes, but having steps numbered only as "1" when using npins seems somewhat unusual. Might it be better to remove this numbering altogether?
```suggestion
Pull the dependencies with npins: `$ npins add github davidlattimore wild -b main`

Then, import the pinned sources in your Nix expression:
```

nix/nix.md
```diff
@@ -1,16 +1,26 @@
...
-    nixpkgs.url = "github:NixOS/nix/nixos-unstable";
+    # Have Nixpkgs
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+
+    # Include Wild
     wild = {
       url = "github:davidlattimore/wild";
-      inputs.nixpkgs.follows = "nixpkgs";
+      # If using the Wild Flake (not required)
+      # inputs.nixpkgs.follows = "nixpkgs";
```
lapla-cogito, line 20: When using Wild as a flake, simply uncommenting this line and commenting out `flake = false;` won't work properly. Here's a `flake.nix` file that minimizes differences from the original —wouldn't it be more user-friendly if the differences were clearly marked as comments? (Since I don't use Nix myself, if the differences are obvious to Nix users, this comment might be unnecessary.)

```nix
{
  inputs = {
    # Have Nixpkgs
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";

    # Include Wild
    wild = {
      url = "github:davidlattimore/wild";
      # If using the Wild Flake (not required)
      inputs.nixpkgs.follows = "nixpkgs";
      #
      # If not using the Wild flake, and just using the overlay
      flake = false;
    };
  };

  outputs =
    {
      self,
      nixpkgs,
      wild,
    }:
    let
      # Create an instance of Nixpkgs targeting x64 Linux with the
      # Wild overlay applied
      pkgs = import nixpkgs {
        system = "x86_64-linux";
        overlays = [
          wild.overlays.default
        ];
      };

      # Create a stdenv that uses the Wild linker
      wildStdenv = pkgs.useWildLinker pkgs.stdenv;
    in
    {
      # Add an output of some very cool package that is linked with the Wild linker
      #
      # Note that if a Rust package is being linked with `buildRustPackage`, you will
      # need to create a `rustPlatform` using `makeRustPlatform` with this stdenv. See
      # below how to do that.
      overlays.default = wild.overlays.default;

      # A devShell that demonstrates Wild linker usage
      devShells.x86_64-linux.default = pkgs.mkShell.override { stdenv = wildStdenv; } {
        packages = [
          pkgs.rust-analyzer
          pkgs.rustc
          pkgs.cargo
        ];
      };
    };
}
```

lapla-cogito (commented):
Thanks for working on this!

nix/nix.md
RossSmyth: Oh yeah, that is an artifact of when I was writing it initially.

README.md
RossSmyth: Good catch.

nix/nix.md
RossSmyth: The flake you provided is incorrect, and also does not really show what is useful.

1. Marking `inputs.nixpkgs.follows = "nixpkgs";` and `flake = false;` may work but will produce an error, as the inputs key is for flake
2. It just attempts reexports the wild overlay for some reason, though if `flake = false;` is set, `wild.overlays.default` will just result in an error as that attribute set does not exist
3. It does not have a package as an output, which is an important use-case
4. The devShell is fine, though for actual packages one should always use `inputsFrom`.

RossSmyth: I hope the what to uncomment and when is clear to Nix users.

RossSmyth: > When using Wild as a flake, simply uncommenting this line and commenting out flake = false; won't work properly.

Why do you think so? It imports the overlay which is all that is needed.

lapla-cogito: > I hope the what to uncomment and when is clear to Nix users.

Understood. In that case, I think this arrangement is fine as it is. Thanks!

README.md
lapla-cogito: Sorry—I was mistaken too, but the correct fix requires both correcting the parentheses and modifying it to `[the nix documentation](./nix/nix.md)`.

davidlattimore (approved):
I pushed a fix to the markdown link

RossSmyth:
Thanks! Sorry I forgot about it :sweat_smile:

PR #1329, opened 2025-11-23
refactor: Preparations to run string merging in parallel with GC

davidlattimore:
* Split string merging to reduce mutable borrow duration
* Move string merging to layout phase
* Move allocation of merge sections to finalise_sizes


davidlattimore:
Note, I'm not necessarily going to run string merging in parallel with GC. Whether I do will depend on whether it improves runtime. At the moment it doesn't, but it's possible that with more refactorings it might. I'll submit this anyway, since I think the changes are probably reasonable changes to make and they don't appear to affect performance.

PR #1332, opened 2025-11-24
chore: Update opensuse docker

davidlattimore:


davidlattimore:
It seems like `riscv64gc-unknown-linux-musl` isn't available for the last week. For now, I've updated the rustup command so that it'll select the latest nightly that has all required components.

mati865:
I also noticed `riscv64gc-unknown-linux-musl` absence on Saturday or Sunday. Since nobody fixed it still, I posted a fix at https://github.com/rust-lang/rust/pull/149312

PR #1334, opened 2025-11-24
docs: add alternative Clang invocation

mati865:
Clang accepts `-fuse-ld=wild` if `ld.wild` is present in PATH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,13 @@ linker = "clang"
 rustflags = ["-Clink-arg=--ld-path=wild"]
 ```
 
+Alternatively, you can create symlink `ld.wild` pointing to `wild` and use:
+```toml
+[target.x86_64-unknown-linux-gnu]
+linker = "clang"
+rustflags = ["-Clink-arg=-fuse-ld=wild"]
+```
+
 On Illumos:
 ```
 [target.x86_64-unknown-illumos]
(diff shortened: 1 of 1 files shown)


mati865:
Maybe we should also put `ld.wild` symlink in prebuilt binaries and somehow recommend this symlink to distro packagers?

mati865:
You can include symlink in the tarball, IIRC hard-links as well, so there would be no additional disk usage.
Obviously, it has to be created manually by calling `ln` in release job and won't work with `cargo install`.



> Recommending that distros include a symlink seems sensible.

I don't think we've got packager recommendations anywhere. Where should it go?

davidlattimore:
> You can include symlink in the tarball, IIRC hard-links as well, so there would be no additional disk usage. Obviously, it has to be created manually by calling `ln` in release job and won't work with `cargo install`.

That's true, I was mostly thinking about `cargo install`. I wonder if `cargo binstall` would pick up the symlink from the tarball.

> > Recommending that distros include a symlink seems sensible.
> 
> I don't think we've got packager recommendations anywhere. Where should it go?

I suspect we should perhaps have a new markdown file, since it's a pretty different audience. Perhaps `PACKAGING.md`?

mati865:
From the docs it seems `cargo binstall` would not install the symlink:
> Then it will concat the dir with "{ bin }{ binary-ext }" and use that as the final bin-dir.

https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md#defaults

> I suspect we should perhaps have a new markdown file, since it's a pretty different audience. Perhaps `PACKAGING.md`?

Sure, should I add it here or as a separate PR?

davidlattimore:
Probably a separate PR. I'd say that it deserves its own entry in the changelog

PR #1337, opened 2025-11-24
fix: `save_dir::make_relative_path()` now works as expected

### LINKED ISSUE #1274: Save dir can contain linker scripts with absolute paths
When `WILD_SAVE_DIR` or `WILD_SAVE_BASE` is set, we produce a save-dir into which we copy more-or-less everything that's needed to repeat the link. However, some .so files are actually linker scripts which instruct the linker to load additional files. If the linker script contains absolute paths, then the save-dir is copied to a different OS which doesn't have those same files, the link won't work due to the files not being found. When copying the linker script to the save-dir, we should update it so that any absolute paths are converted to relative paths. That way the linker script will reference files in the save-dir.

karolzwolak:
Fixes #1274.

### DIFF
--- a/libwild/src/save_dir.rs
+++ b/libwild/src/save_dir.rs
@@ -226,7 +226,13 @@ impl SaveDirState {
 
             if target.is_absolute() {
                 self.copy_file(&target)?;
-                target = make_relative_path(&target, directory);
+                target = make_relative_path(&target, directory).with_context(|| {
+                    format!(
+                        "Failed to make path `{}` relative to `{}` while copying symlink",
+                        target.display(),
+                        directory.display()
+                    )
+                })?;
             } else {
                 let absolute_target = directory.join(&target);
                 self.copy_file(&absolute_target)?;
@@ -331,6 +337,10 @@ fn make_linker_script_relative(bytes: &[u8], source_path: &Path) -> Result<Vec<u
 
     for path in absolute_paths {
         let relative_path = make_relative_path(&path, script_dir);
+        // This shouldn't happen, but we don't want to fail in case it does.
+        let Some(relative_path) = relative_path else {
+            continue;
+        };
         let relative_str = relative_path.to_str().context("Path isn't valid UTF-8")?;
         let path_str = path.to_str().context("Path isn't valid UTF-8")?;
         text = text.replace(path_str, relative_str);
@@ -341,28 +351,31 @@ fn make_linker_script_relative(bytes: &[u8], source_path: &Path) -> Result<Vec<u
 
 /// Removes any backtracking components (`..`) and the next component from `path`
 /// For example, `/a/b/../c` becomes `/a/c`.
-/// The path must be absolute, otherwise it might start with `..` which we can't handle.
-fn remove_backtracking_components(path: &Path) -> PathBuf {
+/// The path must be absolute. Will return `None` if backtracking past the root is attempted.
+fn normalize_abs_path(path: &Path) -> Option<PathBuf> {
     assert!(path.is_absolute());
     let mut out = PathBuf::new();
     for comp in path.components() {
         if comp.as_os_str() == ".." {
-            out.pop();
+            if !out.pop() {
+                return None;
+            }
         } else {
             out.push(comp);
         }
     }
-    out
+    Some(out)
 }
 
-/// Returns a relative path to reach `target` from `directory`. Both should be absolute paths.
-fn make_relative_path(target: &Path, directory: &Path) -> PathBuf {
+/// Returns a relative path to reach `target` from `directory`. Both must be absolute paths.
+/// Returns `None` if the paths are invalid (e.g. contain backtracking past the root).
+fn make_relative_path(target: &Path, directory: &Path) -> Option<PathBuf> {
     assert!(target.is_absolute());
     assert!(directory.is_absolute());
     let mut out = PathBuf::new();
 
-    let target = remove_backtracking_components(target);
-    let directory = remove_backtracking_components(directory);
+    let target = normalize_abs_path(target)?;
+    let directory = normalize_abs_path(directory)?;
 
     let mut target_comps = target.components().peekable();
     let mut dir_comps = directory.components().peekable();
@@ -371,7 +384,7 @@ fn make_relative_path(target: &Path, directory: &Path) -> PathBuf {
     loop {
         match (target_comps.peek(), dir_comps.peek()) {
             // identical paths
-            (None, None) => return PathBuf::from("."),
+            (None, None) => return Some(PathBuf::from(".")),
             (Some(t), Some(d)) if t == d => {
                 target_comps.next();
                 dir_comps.next();
@@ -387,7 +400,7 @@ fn make_relative_path(target: &Path, directory: &Path) -> PathBuf {
 
     out.extend(target_comps);
 
-    out
+    Some(out)
 }
 
 fn write_copied_file_arg(out: &mut BufWriter<&mut std::fs::File>, path: &Path) -> Result {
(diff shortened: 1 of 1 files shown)


karolzwolak:
So it turns out that we had this functionality for a while but the function that was supposed to make relative paths wasn't working correctly. I tested this in a linker script locally but since that requires putting an existing absolute path I have no idea how to integration test this. Instead I made unit tests for this function.

karolzwolak:
Interestingly now some linker tests fail under QEMU but pass everywhere else (modulo the aarch64 u24.04 failing on `apt install`...):
```
---- integration_test::program_name_52___rust_integration_dynamic_rs__ stdout ----
Error: Test failed: `rust-integration-dynamic.rs` with linker `ld` config `default`
  Caused by:
    Linker failed:
/usr/aarch64-linux-gnu/bin/ld: cannot find ../../../../../../../../../../aarch64-linux-gnu/lib/libm.so.6: No such file or directory
/usr/aarch64-linux-gnu/bin/ld: cannot find ../../../../../../../../../../aarch64-linux-gnu/lib/libmvec.so.1: No such file or directory
```

karolzwolak:
I looked into that and I can reproduce this. 
The diff of the linker scripts looks roughly like this:
```
diff --git a/before-libm.so b/after-libm.so
index f809537..1cafc6a 100644
--- a/before-libm.so
+++ b/after-libm.so
@@ -1,4 +1,4 @@
 /* GNU ld script
 */
 OUTPUT_FORMAT(elf64-littleaarch64)
-GROUP ( /lib/libm.so.6  AS_NEEDED ( /lib/libmvec.so.1 ) )
+GROUP ( ../../../../../../../../../../../../../../../lib/libm.so.6  AS_NEEDED ( ../../../../../../../../../../../../../../../lib/libmvec.so.1 ) )
```
I think it might to be because the function doesn't handle `..` properly. Gonna fix that.
```
Swapped path `/usr/lib/libm.so.6` to path `../../../../../../../../../../../lib/libm.so.6` in linker script `/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/15.2.1/../../../../lib64/libm.so
```

karolzwolak:
These tests still fail but now the function handles `..` properly which results in sane paths and I improved the tests.J
EDIT: Okay nvm it doesn't pass locally but does in CI.

libwild/src/save_dir.rs
```diff
@@ -339,21 +339,53 @@ fn make_linker_script_relative(bytes: &[u8], source_path: &Path) -> Result<Vec<u
     Ok(text.into_bytes())
 }
 
+/// Removes any backtracking components (`..`) and the next component from `path`
+/// For example, `/a/b/../c` becomes `/a/c`.
+/// The path must be absolute, otherwise it might start with `..` which we can't handle.
+fn remove_backtracking_components(path: &Path) -> PathBuf {
```
karolzwolak: there's also this: https://doc.rust-lang.org/std/fs/fn.canonicalize.html
However that function will also resolve links. Do we want that or not?

libwild/src/save_dir.rs
```diff
@@ -339,21 +339,53 @@ fn make_linker_script_relative(bytes: &[u8], source_path: &Path) -> Result<Vec<u
...
+    assert!(path.is_absolute());
+    let mut out = PathBuf::new();
+    for comp in path.components() {
+        if comp.as_os_str() == ".." {
+            out.pop();
```
karolzwolak: Hm I now that I think about it, the path can be wrong at this point but I'm not sure. Therefore it should return an Option if this is an illegal path.

karolzwolak: I tested it and couldn't produce a crash naively.

davidlattimore: Note that `pop` won't crash if there is nothing to pop. I guess you could return an error if there's nothing there. e.g. `ensure!(out.pop(), "Invalid path")`

libwild/src/save_dir.rs
davidlattimore: There's a bunch of subtleties in the save-dir mechanism as to when we do and don't resolve symlinks. I'd say for now, if we can get away without calling canonicalize, that we should do without it

libwild/src/save_dir.rs
karolzwolak: > Note that `pop` won't crash if there is nothing to pop.

Oh yeah right

> I guess you could return an error if there's nothing there. e.g. `ensure!(out.pop(), "Invalid path")`

We could leave the path absolute if that happened I guess. Not sure if that can happen in practice. Maybe that's worth doing it though because in a case of a bug we're gonna know there's something wrong when the path in the linker script is still absolute.

davidlattimore: Sure, leaving it absolute seems reasonable to me

karolzwolak: Done. I think this is ready to merge if you're fine with the current revision. Unless we could test that the paths are made relative when copying the linker script with WILD_SAVE_DIR.

mati865:
I'm getting similar errors as mentioned in this thread in rust integration tests since this commit on my Arch Linux based box:
```
Linker failed:
    /usr/aarch64-linux-gnu/bin/ld: cannot find ../../../lib/libm.so.6: No such file or directory
    /usr/aarch64-linux-gnu/bin/ld: cannot find ../../../lib/libmvec.so.1: No such file or directory
```

Dunno if related, but lib64 is just a symlink to lib in the official packages and `libc.so` is located under `$SYSROOT/lib`, not `$SYSROOT/usr/lib` (this one doesn't exist):
```
❯ ls /usr/aarch64-linux-gnu/
drwxr-xr-x - root  1 paź 15:41  bin
drwxr-xr-x - root  1 paź 15:41  include
drwxr-xr-x - root  1 paź 15:41  lib
lrwxrwxrwx - root 17 lip 12:32  lib64 -> ./lib
drwxr-xr-x - root  1 paź 15:41  usr
```
Contents of the linker script:
```
❯ bat /home/mateusz/Projects/wild/wild/tests/build/rust-integration.llvm-dynamic-aarch64-5e0a1ab956944e49.d/usr/aarch64-linux-gnu/lib/libc.so
─────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
     │ File: /home/mateusz/Projects/wild/wild/tests/build/rust-integration.llvm-dynamic-aarch64-5e0a1ab956944e49.d/usr/aarch64-linux-gnu/lib/libc.so
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ /* GNU ld script
   2 │    Use the shared library, but some functions are only in
   3 │    the static library, so try that secondarily.  */
   4 │ OUTPUT_FORMAT(elf64-littleaarch64)
   5 │ GROUP ( ../../../lib/libc.so.6 ../../../lib/libc_nonshared.a  AS_NEEDED ( ../../../lib/ld-linux-aarch64.so.1 ) )
─────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

davidlattimore:
I merged a change yesterday - https://github.com/davidlattimore/wild/pull/1393 - that skips making the linker script have relative paths if it's in the sysroot. If you're still hitting problems even with that, then perhaps we need to canonicalise both paths before doing the check.

mati865:
I'm 90% certain I was testing with and without #1393 but I'll make sure tomorrow.

mati865:
Found some time, it still fails with latest main for me. Keep in mind that you have to `rm -r wild/tests/build/*` to reproduce it.

PR #1338, opened 2025-11-24
fix: do not add duplicated rpath entries

mati865:


mati865:
I'm not sure whether `fix` is the proper label here. I think the duplicates shouldn't have caused problems.

davidlattimore:
I don't have any strong thoughts on the label. `fix` seems a reasonable choice

PR #1339, opened 2025-11-25
perf: Run string merging and GC in parallel

davidlattimore:


davidlattimore:
If I merge this, I'll split the first two commits off into separate PRs first.

davidlattimore:
Note, the other two commits are now separate PRs - #1343 and #1344. This PR without the other two doesn't help at all because with the two stages both using separate channels to control jobs, they end up blocking on their respective channels waiting for work. Merging the two stages to use the same channel ends up being pretty impractical.

davidlattimore:
I wanted to see if this PR was still a win after https://github.com/davidlattimore/wild/pull/1345. Seems like it still is, so I'll probably go ahead with it. The following are the changes I see in benchmarks for this change relative to #1345.

clang-debug: no change
clang-release: -1.1%
librustc-driver: -3.9%
wild: -3.1%

PR #1340, opened 2025-11-25
chore: Add RISC-V to Arch Linux container

mati865:


mati865:
We will see if that really does the trick, but I don't see why it would not.
With some luck, it might land before next nightly is built (couple minutes after 0:00 UTC).

PR #1344, opened 2025-11-26
refactor: Change GC phase to use rayon scopes

davidlattimore:


Cargo.toml
```diff
@@ -106,7 +106,6 @@ debug = true
 # The profile that 'dist' will build with
 [profile.dist]
 inherits = "release"
-lto = "thin"
```
mati865, line 109: Accidental change?

davidlattimore: Oops! I clearly need to get better at always running `jj new` when I'm about to make a change that I intend to throw away. Thanks for noticing!

PR #1348, opened 2025-11-28
refactoring: Rewrite input opening work control

davidlattimore:


davidlattimore:
Mostly this is just more straightforward code than what was there before. Although, for very small links, it's actually a little bit faster. e.g. it speeds up wild linking itself by about 2.3% and a trivial (non-libc) link is sped up by about 15%. Although that's 8.47ms down to 7.17ms, so the absolute saving is pretty negligible.

PR #1349, opened 2025-11-28
perf: Preallocate a Vec to avoid resizes

davidlattimore:


libwild/src/resolution.rs
```diff
@@ -854,100 +856,123 @@ fn resolve_sections_for_object<'data>(
...
-        .object
-        .sections
-        .enumerate()
-        .map(|(input_section_index, input_section)| {
-            let section_name = obj.object.section_name(input_section).unwrap_or_default();
+    // Note, we build up the collection with push rather than collect because at the time of
+    // writing, object's `SectionTable::enumerate` isn't an exact-size iterator, so using collect
+    // would result in resizing.
+    let mut sections = Vec::with_capacity(obj.object.sections.len());
+    for (input_section_index, input_section) in obj.object.sections.enumerate() {
```
mati865, line 863: FYI you can also use:
```rust
    let mut sections = Vec::with_capacity(obj.object.sections.len());
    sections.extend(obj.object.sections.enumerate());
```
I cannot test right now if it will build.

davidlattimore: Thanks. I did attempt to use extend, but the iterator returns a result, so I'm not sure it'd work

mati865: Ah, didn't see that question mark in-between the parentheses.

PR #1353, opened 2025-12-02
test: Add a test for backtraces

davidlattimore:


mati865:
Do we have any CI job that supports sframe? Probably Tumbleweed will be the most likely.

davidlattimore:
My dev box is running Ubuntu 25.10. AFAICT, the sframe version of the test appeared to work as expected. When enabled, it passed with GNU ld and failed with wild. I guess we'll probably need to do some check. Perhaps `RequiresCompilerFlags:-Wa,--gsframe` will be sufficient. Hopefully none of our test systems have support for sframes in the assembler, but not in glibc's `backtrace`.

Unrelatedly, the non-sframe version of the test that this PR introduces was failing on Nix. I tried in the Nix docker image and the failure didn't reproduce, so I added some print statements to help debug the issue in CI and now it passes. If anyone has any idea why, I'd be interested. My best guess is inlining, but we're compiling without any optimisation flags, so I wouldn't expect it to be doing any inlining. Also, I tried with `-O0` before and it still failed. I guess I could restructure the test to make sure that each function call crosses into a different file. Or I could leave the print statements there, since they seem to make it pass - but that just seems wrong.

davidlattimore:
Although, I'm not sure that inlining makes sense. The exit code when it fails is 21, which indicates that `foo` was the first function that wasn't found in the backtrace. `foo` is in a separate file all by itself, so the compiler couldn't have inlined it.

davidlattimore:
I ended up changing the test to not rely on symbol names and that seems to have worked.

PR #1355, opened 2025-12-03
perf: optimize version script from rustc

### LINKED ISSUE #1014: Avoid building a hashmap during version script parsing
When the version script is very large, building a hashmap takes substantial time. See parent issue.

Even if the version script isn't very large, the approach of building a HashSet then later looking up symbol names in it seems suboptimal. The kind of version script that rustc produces has `global:` then a possibly long list of symbols to be exported, then `local: *`. We should perhaps optimise for this usage.

One possibility would be to store the list of `global` symbol names in a `Vec` instead of a `HashSet`, then during the GC phase, `par_iter` that `Vec` and look up each name in the `SymbolDb` before requesting export of the identified `SymbolId`.

More thought is probably needed as to how to optimise for the use-case of version scripts produced rustc, while still being generic enough to support more complex version scripts that have glob matching and other features.

There is work currently ongoing to improve version script support, so this optimisation should likely wait until after that's finished and submitted.

karolzwolak:
Fixes #1014.

### DIFF
--- a/libwild/src/symbol_db.rs
+++ b/libwild/src/symbol_db.rs
@@ -37,6 +37,7 @@ use crate::value_flags::PerSymbolFlags;
 use crate::value_flags::RawFlags;
 use crate::value_flags::ValueFlags;
 use crate::verbose_timing_phase;
+use crate::version_script::RustVersionScript;
 use crate::version_script::VersionScript;
 use crossbeam_queue::SegQueue;
 use hashbrown::HashMap;
@@ -281,6 +282,39 @@ struct PendingSymbolHashBucket<'data> {
 }
 
 impl<'data> SymbolDb<'data> {
+    /// If the version script is optimized fur rust, we downgraded all symbols to local visibility.
+    /// This promotes symbols marked for global visibility in a Rust version script back to global.
+    /// Also adds the non-interposable flag to all local symbols.
+    fn handle_rust_version_script(
+        &self,
+        rust_vscript: &RustVersionScript<'data>,
+        per_symbol_flags: &mut PerSymbolFlags,
+    ) {
+        verbose_timing_phase!("Upgrade locals for export");
+        let atomic_per_symbol_flags = per_symbol_flags.borrow_atomic();
+
+        rust_vscript.global.par_iter().for_each(|symbol| {
+            let prehashed = UnversionedSymbolName::prehashed(symbol);
+            if let Some(symbol_id) = self.get_unversioned(&prehashed) {
+                let symbol_atomic_flags = atomic_per_symbol_flags.get_atomic(symbol_id);
+                symbol_atomic_flags.remove(ValueFlags::DOWNGRADE_TO_LOCAL);
+            }
+        });
+
+        // Don't forget to add the non-interposable flag the local symbols.
+        // We coudn't do this earlier as we didn't know which symbols would remain
+        // local.
+        per_symbol_flags
+            .flags_mut()
+            .par_iter_mut()
+            .for_each(|flags| {
+                let flags_val = flags.get();
+                if flags_val.is_downgraded_to_local() {
+                    *flags = (flags_val | ValueFlags::NON_INTERPOSABLE).raw();
+                }
+            });
+    }
+
     pub fn build(
         groups: Vec<Group<'data>>,
         version_script_data: Option<ScriptData<'data>>,
@@ -380,39 +414,15 @@ impl<'data> SymbolDb<'data> {
                 // If it's a rust version script, apply the global symbol visibility now.
                 // We previously downgraded all symbols to local visibility.
                 if let VersionScript::Rust(rust_vscript) = &index.version_script {
-                    verbose_timing_phase!("Upgrade locals for export");
-                    let atomic_per_symbol_flags = per_symbol_flags.borrow_atomic();
-
-                    rust_vscript.global.par_iter().for_each(|symbol| {
-                        let prehashed = UnversionedSymbolName::prehashed(symbol);
-                        if let Some(symbol_id) = index.get_unversioned(&prehashed) {
-                            let symbol_atomic_flags = atomic_per_symbol_flags.get_atomic(symbol_id);
-                            symbol_atomic_flags.remove(ValueFlags::DOWNGRADE_TO_LOCAL);
-                        }
-                    });
-
-                    // Don't forget to add the non-interposable flag the local symbols.
-                    // We coudn't do this earlier as we didn't know which symbols would remain
-                    // local.
-                    per_symbol_flags
-                        .flags_mut()
-                        .par_iter_mut()
-                        .for_each(|flags| {
-                            let flags_val = flags.get();
-                            if flags_val.is_downgraded_to_local() {
-                                *flags = (flags_val | ValueFlags::NON_INTERPOSABLE).raw();
-                            }
-                        });
+                    index.handle_rust_version_script(rust_vscript, &mut per_symbol_flags);
                 }
 
-                {
-                    index.apply_wrapped_symbol_overrides(args, herd);
+                index.apply_wrapped_symbol_overrides(args, herd);
 
-                    verbose_timing_phase!("Apply linker scripts");
+                verbose_timing_phase!("Apply linker scripts");
 
-                    for script in linker_scripts {
-                        index.apply_linker_script(script);
-                    }
+                for script in linker_scripts {
+                    index.apply_linker_script(script);
                 }
             },
         );


karolzwolak:
Actually this doesn't speed the things up that much.
Benching libbevy_dylib (560k line version script):
```
Benchmark 1: ./run-with /home/karolz/dev/contrib/wild/target/release/wild --no-fork
  Time (mean ± σ):      1.068 s ±  0.017 s    [User: 7.228 s, System: 1.607 s]
  Range (min … max):    1.042 s …  1.092 s    10 runs
 
Benchmark 2: ./run-with /home/karolz/dev/contrib/wild/target/release/wild-before --no-fork
  Time (mean ± σ):      1.088 s ±  0.032 s    [User: 7.129 s, System: 1.636 s]
  Range (min … max):    1.054 s …  1.148 s    10 runs
 
Summary
  ./run-with /home/karolz/dev/contrib/wild/target/release/wild --no-fork ran
    1.02 ± 0.03 times faster than ./run-with /home/karolz/dev/contrib/wild/target/release/wild-before --no-fork
```

Did I miss something performance wise or just this doesn't provide any substantial speedup?

EDIT: this actually is slower for compiling smaller libs like librustc_driver (although 300ms is probably too low to be a representative test)
```
Benchmark 1: ./run-with /home/karolz/dev/contrib/wild/target/release/wild --no-fork
  Time (mean ± σ):     292.5 ms ±   7.9 ms    [User: 1992.1 ms, System: 536.3 ms]
  Range (min … max):   280.4 ms … 306.7 ms    10 runs
 
Benchmark 2: ./run-with /home/karolz/dev/contrib/wild/target/release/wild-before --no-fork
  Time (mean ± σ):     286.2 ms ±   8.9 ms    [User: 1968.3 ms, System: 536.9 ms]
  Range (min … max):   277.6 ms … 305.7 ms    10 runs
 
Summary
  ./run-with /home/karolz/dev/contrib/wild/target/release/wild-before --no-fork ran
    1.02 ± 0.04 times faster than ./run-with /home/karolz/dev/contrib/wild/target/release/wild --no-fork
```

libwild/src/value_flags.rs
```diff
@@ -234,6 +234,14 @@ impl AtomicValueFlags {
     pub(crate) fn or_assign(&self, flags: ValueFlags) {
         self.0.fetch_or(flags.bits(), Ordering::Relaxed);
     }
+
+    pub(crate) fn remove(&self, flags_to_remove: ValueFlags) {
+        let mut res = self.get();
+        res.remove(flags_to_remove);
+        if res != flags_to_remove {
+            self.0.store(res.bits(), Ordering::Relaxed);
```
davidlattimore: At the moment, `remove` is only used to clear that one bit and there aren't other threads doing anything other than clearing that one bit. But if other atomic operations were occurring concurrently, this could cause lost updates. e.g. this function reads the value, another thread makes some different change to the value then this thread writes the updated value and the change made by the other thread is lost.

Probably better to implement this function like `or_assign` above. e.g. `self.0.fetch_and(!flags.bits(), ...);`. Note the bitwise not (`!`)

libwild/src/version_script.rs
```diff
@@ -386,6 +404,49 @@ impl<'data> VersionScript<'data> {
...
+        match self {
+            VersionScript::Regular(script) => script.version_count(),
+            VersionScript::Rust(_) => 0,
+        }
+    }
+
+    pub(crate) fn version_iter(&self) -> impl Iterator<Item = &Version<'data>> {
+        match self {
+            VersionScript::Regular(script) => script.version_iter(),
+            VersionScript::Rust(_) => todo!(),
```
davidlattimore: Perhaps the caller of `version_iter` should do the match instead?

libwild/src/symbol_db.rs
```diff
@@ -374,6 +375,21 @@ impl<'data> SymbolDb<'data> {
             output_kind,
         };
 
+        let mut per_symbol_flags = PerSymbolFlags::new(per_symbol_flags);
+        let atomic_per_system_flags = per_symbol_flags.borrow_atomic();
+
+        // If it's a rust version script, apply the global symbol visibility now.
+        // We previously downgraded all symbols to local visibility.
+        if let VersionScript::Rust(rust_vscript) = &index.version_script {
+            rust_vscript.global.par_iter().for_each(|symbol| {
```
davidlattimore: It'd be good to track how long this block takes to run. Can you add `verbose_timing_phase!("Upgrade globals for export");`?

libwild/src/symbol_db.rs
```diff
@@ -1474,7 +1490,12 @@ impl<'data> SymbolLoader<'data> for RegularObjectSymbolLoader<'_, 'data> {
     }
 
     fn should_downgrade_to_local(&self, name: &PreHashed<UnversionedSymbolName>) -> bool {
```
davidlattimore, line 1528: Looking at the the caller of this method (in `layout.rs`), we've got the following code:

```rust
            if self.should_downgrade_to_local(&name) {
                flags |= ValueFlags::DOWNGRADE_TO_LOCAL;
                // If we're downgrading to a local, then we're writing a shared object. Shared
                // objects should never bypass the GOT for TLS variables.
                if symbol.st_type() != object::elf::STT_TLS {
                    flags |= ValueFlags::NON_INTERPOSABLE;
                }
            }
```

That could be problematic, since we're marking symbols as non-interposable then leaving them that way even after we remove the downgrade-to-local flag. I suspect we'd need to move the code that sets non-interposable to after we've removed the downgrade-to-local flags.

libwild/src/symbol_db.rs
```diff
@@ -374,6 +375,21 @@ impl<'data> SymbolDb<'data> {
             output_kind,
         };
 
+        let mut per_symbol_flags = PerSymbolFlags::new(per_symbol_flags);
+        let atomic_per_system_flags = per_symbol_flags.borrow_atomic();
```
davidlattimore: This line could move into the if-block below

libwild/src/symbol_db.rs
```diff
@@ -374,6 +375,21 @@ impl<'data> SymbolDb<'data> {
...
+
+        // If it's a rust version script, apply the global symbol visibility now.
+        // We previously downgraded all symbols to local visibility.
+        if let VersionScript::Rust(rust_vscript) = &index.version_script {
```
davidlattimore: This stage could run in parallel with `drop(per_group_outputs)` above. When I try that, this PR causes no slowdown for librustc-driver.

libwild/src/symbol_db.rs
davidlattimore: Not related to your change, but this scope doesn't seem to be necessary. Feel free to remove it if you're moving the `drop(per_group_outputs)` as suggested below.

davidlattimore (commented):
That's unfortunate that there wasn't a speedup for librustc-driver. Although, I see a 9.9% speedup for the bevy dylib, which is still definitely something! I just looked at some profiles and it seems that `analyze_glob_pattern` is taking quite a bit of time.

I wonder if the changes you've made here would be a step towards moving the call to `analyze_glob_pattern` to run in parallel. e.g. we parse the file from a single thread, but don't identify the kinds of patterns, then from multiple threads, we classify matchers. I wouldn't suggest doing that in this PR, but we could explore doing it afterwards. If you don't want to, then I might.

mati865:
If you don't want to regress performance in the short term, we could merge it into a new branch, do the optimizations, and then rebase main onto that branch (without squashing).
Although I'm not sure if it's worth risking conflicts.

davidlattimore:
The performance regression for librustc-driver was pretty small, about 1% and went away completely (at least for me) once the extra phase was run in parallel with dropping a bunch of Vecs.

libwild/src/version_script.rs
karolzwolak: Yeah, I forgot to remove it.

libwild/src/symbol_db.rs
karolzwolak: I did that but I had to resort to using `rayon::scope` which turned out a bit ugly. Is there a better way of doing that? Directly dropping in the `rayon::spawn` doesn't compile as it requires `'static`.

davidlattimore: `rayon::join` is perfectly suited to that kind of thing. It runs two closures in parallel. It even returns the values produced by the two closures, although you don't need that in this case.

libwild/src/symbol_db.rs
karolzwolak: Good catch! How about also removing `NON_INTERPOSABLE` when we remove `DOWNGRADE_TO_LOCAL` instead of moving this code?

EDIT: I Don't think it's possible as we won't have access to the `symbol.st_type()` where we remove the local downgrade flag.

karolzwolak:
> I wonder if the changes you've made here would be a step towards moving the call to `analyze_glob_pattern` to run in parallel. e.g. we parse the file from a single thread, but don't identify the kinds of patterns, then from multiple threads, we classify matchers. I wouldn't suggest doing that in this PR, but we could explore doing it afterwards. If you don't want to, then I might.

Sure, I will look into that when this is merged.

libwild/src/symbol_db.rs
davidlattimore: I don't think that would be equivalent since other things might have added `NON_INTERPOSABLE`

libwild/src/symbol_db.rs
karolzwolak: Okay, I've changed to use join then.

libwild/src/symbol_db.rs
```diff
@@ -374,15 +368,43 @@ impl<'data> SymbolDb<'data> {
             output_kind,
         };
 
-        index.apply_wrapped_symbol_overrides(args, herd);
+        let mut per_symbol_flags = PerSymbolFlags::new(per_symbol_flags);
 
-        verbose_timing_phase!("Apply linker scripts");
+        rayon::join(
```
karolzwolak, line 407: We should probably benchmark that this doesn't slow things down when the version script isn't coming from rustc. I can do that tomorrow.

libwild/src/symbol_db.rs
```diff
@@ -374,15 +368,43 @@ impl<'data> SymbolDb<'data> {
...
+                            let symbol_atomic_flags = atomic_per_system_flags.get_atomic(symbol_id);
+                            symbol_atomic_flags.remove(ValueFlags::DOWNGRADE_TO_LOCAL);
+                        }
+                    });
+                }
 
-        for script in linker_scripts {
-            index.apply_linker_script(script);
-        }
+                {
```
karolzwolak: Maybe we should move this scope out of the rayon task and only do the join if the version script is coming from rustc

karolzwolak: It's not clear to me which is better, this way we are always dropping in parallel but not sure that's actually worth it when the script isn't from rustc.

davidlattimore: `rayon::join` is pretty lightweight. It doesn't even heap-allocate the closures and we're not in a loop here, so this is one extra join during the entire linker execution.

libwild/src/symbol_db.rs
karolzwolak: I've looked a bit for a good place to move this code but couldn't find anything. Where would you move it?

davidlattimore: The simplest option might be right after the code that removes the `DOWNGRADE_TO_LOCAL` flags. You could `par_iter_mut` over all the flags from within the same join.

karolzwolak: Oh yeah right, I thought for some reason that we needed to check the `symbol.st_type()`. I've done fixed it.

karolzwolak:
> That's unfortunate that there wasn't a speedup for librustc-driver. Although, I see a 9.9% speedup for the bevy dylib, which is still definitely something! I just looked at some profiles and it seems that `analyze_glob_pattern` is taking quite a bit of time.

This is great! The speedup is negligible on my system.
```
./run-with /home/karolz/dev/contrib/wild/target/release/wild --no-fork ran
    1.01 ± 0.02 times faster than ./run-with /home/karolz/dev/contrib/wild/target/release/wild-before --no-fork
```
Could you do the benchmark again on the latest revision?

davidlattimore:
Sure! Still looks good to me:

```
B=bevy F="--threads=32 --strip-debug" R=/home/david/save/$B/run-with OUT=/run/user/1000/ttt bench poop -d 30000  "$R $HOME/wild-builds/2025-12-06 $F" "$R target/release/wild $F"
  Temperature: 54.8 C
Benchmark 1 (87 runs): /home/david/save/bevy/run-with /home/d/wild-builds/2025-12-06 --threads=32 --strip-debug
  wall_time           345ms ± 7.04ms     332ms …  366ms          2 ( 2%)        0%
  peak_rss           7.29MB ± 24.2KB    7.24MB … 7.30MB         16 (18%)        0%
  cpu_cycles         10.5G  ±  136M     10.1G  … 10.7G           0 ( 0%)        0%
  instructions       10.2G  ± 17.2M     10.2G  … 10.2G           0 ( 0%)        0%
  cache_references    237M  ± 1.56M      233M  …  243M           2 ( 2%)        0%
  cache_misses       61.2M  ±  350K     60.3M  … 62.5M           3 ( 3%)        0%
  branch_misses      17.2M  ± 61.0K     17.1M  … 17.4M           0 ( 0%)        0%
Benchmark 2 (99 runs): /home/david/save/bevy/run-with target/release/wild --threads=32 --strip-debug
  wall_time           305ms ± 6.35ms     296ms …  326ms          5 ( 5%)        ⚡- 11.4% ±  0.6%
  peak_rss           7.29MB ± 21.4KB    7.20MB … 7.34MB         19 (19%)          +  0.1% ±  0.1%
  cpu_cycles         10.6G  ±  111M     10.2G  … 10.7G           4 ( 4%)          +  0.5% ±  0.3%
  instructions       10.1G  ± 16.8M     10.1G  … 10.2G           1 ( 1%)          -  0.4% ±  0.0%
  cache_references    239M  ± 1.78M      236M  …  246M           4 ( 4%)          +  0.9% ±  0.2%
  cache_misses       61.8M  ±  342K     61.3M  … 63.2M           1 ( 1%)          +  1.0% ±  0.2%
  branch_misses      17.5M  ± 58.8K     17.4M  … 17.7M           5 ( 5%)        💩+  1.6% ±  0.1%
  Temperature: 61.4 C
```

No significant change on librustc-driver.

libwild/src/symbol_db.rs
```diff
@@ -374,15 +368,56 @@ impl<'data> SymbolDb<'data> {
...
+            || {
+                // This can take a while, so do it in parallel with other work.
+                verbose_timing_phase!("Drop per-group outputs");
+                drop(per_group_outputs);
+            },
+            || {
+                // If it's a rust version script, apply the global symbol visibility now.
+                // We previously downgraded all symbols to local visibility.
+                if let VersionScript::Rust(rust_vscript) = &index.version_script {
+                    verbose_timing_phase!("Upgrade locals for export");
```
davidlattimore: There's possibly enough in this if-block that it'd be worthwhile moving it to a separate function.

karolzwolak: Done.

PR #1373, opened 2025-12-11
docker: Add Dockerfile for Gentoo

csfore:
Tests passed for me within the image.

Binary packages get installed to reduce friction, as Clang can have a long build time depending on the hardware in use.

One thing I'm not 100% sure on is the method for ensuring `PATH` is correct by updating `CMD`. It was the simplest solution I could think of though, open to better ideas!

### DIFF
--- a/docker/gentoo.Dockerfile
+++ b/docker/gentoo.Dockerfile
@@ -1,0 +1,37 @@
+# Run on Gentoo
+#
+# docker build --progress=plain -t wild-dev-gentoo . -f docker/gentoo.Dockerfile
+# docker run -it wild-dev-gentoo
+
+FROM gentoo/stage3:20251208 AS chef
+
+RUN echo "dev-util/rustup" >> /etc/portage/package.accept_keywords/wild
+RUN emerge-webrsync
+# Required for installing binary packages.
+RUN getuto
+RUN emerge -g dev-util/rustup llvm-core/clang llvm-core/lld
+# Ensures the PATH is correct upon running the image.
+RUN echo 'export PATH="/root/.cargo/bin:${PATH}"' >> /etc/profile
+
+RUN rustup-init --default-toolchain stable -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup toolchain install nightly \
+        --allow-downgrade \
+        --target x86_64-unknown-linux-musl,aarch64-unknown-linux-gnu,aarch64-unknown-linux-musl,riscv64gc-unknown-linux-gnu,riscv64gc-unknown-linux-musl \
+        --component rustc-codegen-cranelift-preview
+RUN cargo install --locked cargo-chef
+WORKDIR /wild
+
+# This ensures LLVM's PATH update from the emerge is recognized, while also ensuring
+# rustup's environment in /root is added to PATH, as Docker does not source
+# /etc/profile by default.
+CMD ["/bin/bash", "-l"]
+
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+COPY --from=planner /wild/recipe.json recipe.json
+RUN cargo chef cook --all-targets --recipe-path recipe.json
+COPY . .


docker/gentoo.Dockerfile
```diff
@@ -0,0 +1,42 @@
+# Run on Gentoo
+#
+# docker build --progress=plain -t wild-dev-gentoo . -f docker/gentoo.Dockerfile
+# docker run -it wild-dev-gentoo
+
+FROM gentoo/stage3:desktop-20251208 AS chef
```
davidlattimore: "desktop" sound like it'd be pretty big and contain lots of stuff that wouldn't be needed. Is there a lighter alternative? From a quick look at dockerhub, it looks like the systemd images are about 370MiB compared to about 1GiB for the desktop image.

docker/gentoo.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN emerge-webrsync
+# Required for installing binary packages.
+RUN getuto
+RUN emerge -g dev-util/rustup llvm-core/clang llvm-core/lld
+# Ensures the PATH is correct upon running the image.
+RUN echo 'export PATH="/root/.cargo/bin:${PATH}"' >> /etc/profile
+
+RUN rustup-init --default-toolchain stable -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup toolchain install nightly && \
```
davidlattimore: This is how most of our docker files install the toolchain, but recently we've started using the following:

```docker
RUN rustup toolchain install nightly \
        --allow-downgrade \
        --target x86_64-unknown-linux-musl,aarch64-unknown-linux-gnu,aarch64-unknown-linux-musl,riscv64gc-unknown-linux-gnu,riscv64gc-unknown-linux-musl \
        --component rustc-codegen-cranelift-preview
```

This allows us to use a slightly older nightly if not all of the components are available rather than failing outright. I think we're possibly only using it in `arch.Dockerfile` and maybe one other so far, but it could be good to use here.

docker/gentoo.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+        riscv64gc-unknown-linux-musl \
+        && \
+    rustup component add rustc-codegen-cranelift-preview --toolchain nightly
+RUN cargo install --locked cargo-chef
+WORKDIR /wild
+
+# This ensures LLVM's PATH update from the emerge is recognized, while also ensuring
+# rustup's environment in /root is added to PATH, as Docker does not source
+# /etc/profile by default.
+CMD ["/bin/bash", "-l"]
```
davidlattimore, line 28: At least one of our other dockerfiles does `ENV PATH="/root/.cargo/bin:$PATH"`. I don't have any strong preference though.

davidlattimore (commented):
Thanks! That will be useful if we get bug reports that are gentoo specific.

docker/gentoo.Dockerfile
csfore: Definitely! I think I set it to desktop originally because they come with LLVM pre-installed and did the binary package portion after the fact.

docker/gentoo.Dockerfile
csfore: When I tried this originally it dropped the `/usr/lib/llvm/<LLVM Version>/bin` part of the PATH that got updated within the image (it gets set in `/etc/profile.env`) when using just ENV, resulting in `ld.lld` not being found.

docker/gentoo.Dockerfile
mati865: FYI the other one would be openSUSE IIRC.

docker/gentoo.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# docker build --progress=plain -t wild-dev-gentoo . -f docker/gentoo.Dockerfile
+# docker run -it wild-dev-gentoo
+
+FROM gentoo/stage3:20251208 AS chef
+
+RUN echo "dev-util/rustup" >> /etc/portage/package.accept_keywords/wild
+RUN emerge-webrsync
+# Required for installing binary packages.
+RUN getuto
+RUN emerge -g dev-util/rustup llvm-core/clang llvm-core/lld
```
zyxhere, line 12: Wonder if `emerge -G` would be better? otherwise if for some reason the binary package isn't available then cmake/llvm/clang/lld would be compiled from source. Would it be better for the CI to stop in that case?

https://wiki.gentoo.org/wiki/Binary_package_guide#Installing_binary_packages

davidlattimore: Probably that would be better. Note that we don't use these dockerfiles in CI. They're just for when we want to try to reproduce a bug on a particular platform.

csfore: Definitely doable, we'd just need to split the emerge line between rustup and the other binary packages since rustup doesn't have a binary package.

PR #1374, opened 2025-12-12
refactor: Allow SymbolDb to be built in stages

davidlattimore:


davidlattimore:
The bump to the minimal version of `bumpalo-herd` was because I added the herd to a struct that had a derived `Debug` impl and herd doesn't implement `Debug` until the latest version.

PR #1382, opened 2025-12-16
refactor: Make InputRef implement Copy

davidlattimore:


mati865:
Oh, that's annoying. This PR is probably the best solution until https://github.com/rust-lang/rust/issues/125687 is complete.

Issue #1325, opened 2025-11-22
Wild doesn't check libs soname with `--no-allow-shlib-undefined`


mati865:
Noticed by @davidlattimore in https://github.com/davidlattimore/wild/pull/1321#discussion_r2550977513

When checking whether we loaded a needed DSO, we don't check soname and instead rely entirely on the filename, which is not quite right.

Test based on https://github.com/rui314/mold/blob/5b710298fcae04bfcb00d3e3dd324c5638f02696/test/no-allow-shlib-undefined2.sh:
```bash
❯ bat -p undefined-in-dso.sh
#!/bin/bash

if [ $# -eq 0 ]; then
  echo "Supply linker name/path as an argument!"
fi

LD="$1"

mkdir -p out

cat <<EOF | clang --ld-path="$LD" -shared -fPIC -o out/libfoo.so -Wl,-soname=libfoo.0 -xc -
void foo() {}
EOF

cat <<EOF | clang --ld-path="$LD" -shared -fPIC -o out/libbar.so -xc - -Lout -lfoo
void foo();
void bar() { foo(); }
EOF

cat <<EOF | clang -o out/a.o -c -xc -
int bar();
int main() { bar(); }
EOF

# First case: libfoo lost soname, only GNU ld prints error
echo | clang --ld-path="$LD" -shared -fPIC -o out/libfoo.so -xc -
not clang --ld-path="$LD" -o out/exe3 out/a.o -Wl,--no-allow-shlib-undefined -Lout -lfoo -lbar > /dev/null 2>&1
if [ $? -ne 0 ]; then
  echo "First case failed - linker didn't report the error"
fi

# Second case: libfoo has soname different from filename, GNU ld, LLD, Mold print error, Wild doesn't
echo | clang --ld-path="$LD" -shared -fPIC -o out/libfoo.so -Wl,-soname=libfoo.0 -xc -
not clang --ld-path="$LD" -o out/exe3 out/a.o -Wl,--no-allow-shlib-undefined -Lout -lfoo -lbar > /dev/null 2>&1
if [ $? -ne 0 ]; then
  echo "Second case failed - linker didn't report the error"
fi
```

Testing different linkers:
```
❯ ./undefined-in-dso.sh ld.bfd

❯ ./undefined-in-dso.sh ld.lld
First case failed - linker didn't report the error

❯ ./undefined-in-dso.sh target/debug/wild
First case failed - linker didn't report the error
Second case failed - linker didn't report the error

[...]

PR #1389, opened 2025-12-19
refactor: Split ResolvedObject into dynamic and non-dynamic variants

davidlattimore:
The main motivation for doing this is so that we can store state that is specific to dynamic objects.

Issue #1325


davidlattimore:
I made this change originally to help with #1390. However, I ended up changing that PR a bit, so the dynamic tags are stored on the parsed object rather than the resolved object, which means this PR isn't strictly needed. That said, I think the change here makes the code slightly nicer and it's certainly easier if we do decide that we want to add state to ResolvedDynamic in future, so I think I'll merge it anyway.

PR #1392, opened 2025-12-19
chore: Ignore a mold test that fails on Ubuntu 25.05

davidlattimore:


wild/tests/external_tests/mold_skip_tests.toml
```diff
@@ -199,6 +199,7 @@ tests = [
   "stdout.sh",
   "undefined.sh",                   # Passes when `--no-gc-sections` is passed.
   "undefined2.sh",                  # Passes when `--no-gc-sections` is passed.
+  "unkown-section-type.sh",         # Wild doesn't yet error on unknown section types
```
marxin, line 202: There's a typo in the filename:
```
❯ fd unknown
test/unknown-section-type.sh
```

mati865:
Make sure to do `git submodule update` after https://github.com/davidlattimore/wild/pull/1386. That should fix the problem.

davidlattimore:
For some reason I had assumed that this was a different typo, but you're right, I just hadn't updated the submodules. Thanks :)

PR #1400, opened 2025-12-22
feat: Support `-z max-page-size`

lapla-cogito:
This option is used when building [gVisor](https://github.com/google/gvisor/blob/89151a08e48f15c73ae24d5d460f1d6400c86675/vdso/BUILD#L49).

This is a prefix-style option, but the existing argument parser did not support prefix options that take a value. To address this, the second commit refactors the option handling logic to distinguish between sub-options that take a value and those that do not.

In addition, for testing, this also introduces a new test directive, `ExpectLoadAlignment`, which checks the alignment of the first `PT_LOAD` segment in the test binary.

### DIFF
--- a/wild/tests/sources/max-page-size.c
+++ b/wild/tests/sources/max-page-size.c
@@ -10,10 +10,14 @@
 //#Config:1m:default
 //#LinkArgs:-z max-page-size=0x100000 -z now
 //#ExpectLoadAlignment:0x100000
+// It seems that large page sizes are not permitted in RISC-V QEMU
+//#Arch: x86_64,aarch64
 
 //#Config:2m:default
 //#LinkArgs:-z max-page-size=0x200000 -z now
 //#ExpectLoadAlignment:0x200000
+// It seems that large page sizes are not permitted in RISC-V QEMU
+//#Arch: x86_64,aarch64
 
 #include "runtime.h"
 


libwild/src/args.rs
```diff
@@ -928,7 +942,19 @@ impl ArgumentParser {
 
                 // Check if this value corresponds to a registered sub-option
                 if let Some(sub) = handler.sub_options.get(value.as_str()) {
-                    (sub.handler)(args, modifier_stack, &value)?;
+                    match sub.handler {
+                        SubOptionHandler::NoValue(f) => f(args, modifier_stack)?,
+                        SubOptionHandler::WithValue(f) => f(args, modifier_stack, &value)?,
```
davidlattimore: Seems odd that value is passed here given that it's the name of the suboption. I'd have thought that this case would be an error.

libwild/src/args.rs
```diff
@@ -928,7 +942,19 @@ impl ArgumentParser {
...
+                    }
+                } else if let Some((sub_name, sub)) = handler
+                    .sub_options
+                    .iter()
+                    .find(|(name, sub)| sub.with_value() && value.starts_with(*name))
```
davidlattimore: If we split on '=' then used the left hand side as the key when looking up `sub_options`, then we could just use the hashmap lookup above rather than doing a find over all the options.

wild/tests/sources/max-page-size.c
```diff
@@ -0,0 +1,25 @@
+//#AbstractConfig:default
+//#LinkerDriver:gcc
```
davidlattimore: Does this test need to link against libc? It looks like it just define `_start` instead. Tests that link against libc are a lot slower than non-libc tests.

libwild/src/args.rs
lapla-cogito: Initially, I was concerned that options like `-l :filename` wouldn't be properly handled in this context, but fixing [this](https://github.com/davidlattimore/wild/pull/1400#discussion_r2638434411) actually resolved the issue neatly. Thanks!

wild/tests/sources/max-page-size.c
lapla-cogito: Fixed.

wild/tests/sources/max-page-size.c
```diff
@@ -0,0 +1,25 @@
...
+
+//#Config:1m:default
+//#LinkArgs:-z max-page-size=0x100000 -z now
+//#ExpectLoadAlignment:0x100000
+
+//#Config:2m:default
+//#LinkArgs:-z max-page-size=0x200000 -z now
+//#ExpectLoadAlignment:0x200000
+
+#include "init.h"
```
davidlattimore: `init.h` shouldn't be needed. It's for testing of things like `.init_array` etc.

wild/tests/sources/max-page-size.c
```diff
@@ -0,0 +1,25 @@
...
+#include "runtime.h"
+
+int main() {
```
davidlattimore: Probably best to call this `_start`. `main` won't be used and would get GCed.

wild/tests/sources/max-page-size.c
```diff
@@ -0,0 +1,25 @@
+//#AbstractConfig:default
+//#RunEnabled:false
```
davidlattimore: I think it should be fine to enable running of this. At least I don't see any reason why it should be able to be run provided `main` is renamed to `_start`.

wild/tests/sources/max-page-size.c
lapla-cogito: Ah, I completely forgot about that. Now it makes sense why the test failed without `RunEnabled: false`.

wild/tests/sources/max-page-size.c
davidlattimore: Sorry, I didn't mean for you to remove the call to `runtime_init`, just the include of "init.h". `runtime_init` is declared in `runtime.h` and is needed for risc-v - although I'm unsure if that would matter or not if we're just calling the exit syscall.

lapla-cogito: I had assumed `runtime_init` would be included in `runtime.h`, my apologies for the confusion.

PR #1406, opened 2025-12-23
fix(save-dir): Normalize source paths before checking if it's in sysroot

karolzwolak:
Follow up to #1393, fixes https://github.com/davidlattimore/wild/pull/1337#issuecomment-3679330435


mati865:
Thanks, it works with this change.

PR #1407, opened 2025-12-24
Enable Windows build

8Keep:
I am maybe/sorta going to try to get wild working on windows. The first (easy step to that is to get it buildable on windows. This PR gets it building on windows, disables some tests that don't yet work, and sets up github actions CI to also build+test on windows.

### DIFF
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -5,10 +5,8 @@
 use crate::bail;
 use crate::error::Context as _;
 use crate::error::Result;
-use std::ffi::OsStr;
 use std::ops::Range;
-use std::os::unix::ffi::OsStrExt as _;
-use std::path::Path;
+use std::path::PathBuf;
 use zerocopy::FromBytes;
 use zerocopy::Immutable;
 use zerocopy::KnownLayout;
@@ -18,6 +16,20 @@ pub(crate) enum ArchiveEntry<'data> {
     Regular(ArchiveContent<'data>),
     Filenames(ExtendedFilenames<'data>),
     Thin(ThinEntry<'data>),
+}
+
+fn path_from_bytes(bytes: &[u8]) -> PathBuf {
+    #[cfg(unix)]
+    {
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt as _;
+        std::path::Path::new(OsStr::from_bytes(bytes)).to_path_buf()
+    }
+    #[cfg(not(unix))]
+    {
+        let path = std::str::from_utf8(bytes).expect("Invalid UTF-8 in archive path name");
+        PathBuf::from(path)
+    }
 }
 
 #[derive(Clone, Copy)]
@@ -324,8 +336,8 @@ impl<'data> Identifier<'data> {
         }
     }
 
-    pub(crate) fn as_path(&self) -> &'data std::path::Path {
-        Path::new(OsStr::from_bytes(self.as_slice()))
+    pub(crate) fn as_path(&self) -> PathBuf {
+        path_from_bytes(self.as_slice())
     }
 }
 
--- a/libwild/src/file_writer.rs
+++ b/libwild/src/file_writer.rs
@@ -263,7 +263,10 @@ impl SizedOutput {
         // assuming it eventually calls exec, this at least means that it inherits the file
         // descriptor for less time. i.e. this doesn't really fix anything, but makes problems less
         // bad.
-        std::os::unix::fs::OpenOptionsExt::custom_flags(&mut open_options, libc::O_CLOEXEC);
+        #[cfg(unix)]
+        {
+            std::os::unix::fs::OpenOptionsExt::custom_flags(&mut open_options, libc::O_CLOEXEC);
+        }
 
         match output_config.file_write_mode {
             FileWriteMode::UnlinkAndReplace => {
@@ -316,7 +319,10 @@ impl SizedOutput {
 
         // Making the file executable is best-effort only. For example if we're writing to a pipe or
         // something, it isn't going to work and that's OK.
-        let _ = crate::fs::make_executable(&self.file);
+        #[cfg(unix)]
+        {
+            let _ = crate::fs::make_executable(&self.file);
+        }
 
         Ok(())
     }
--- a/libwild/src/fs.rs
+++ b/libwild/src/fs.rs
@@ -1,6 +1,9 @@
+#[cfg(unix)]
 use crate::error::Result;
+#[cfg(unix)]
 use std::fs::File;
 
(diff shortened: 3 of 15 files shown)


libwild/src/timing.rs
```diff
@@ -4,18 +4,21 @@ use crate::args::CounterKind;
...
+use anyhow::Context;
+#[cfg(feature = "perfetto")]
 use std::sync::Mutex;
 use std::time::Duration;
 use std::time::Instant;
 use tracing::field::Visit;
 
 const PERFETTO_ENV_VAR: &str = "WILD_PERFETTO_OUT";
 
+#[cfg(feature = "perfetto")]
```
davidlattimore: I just published version 0.3.0 of perfetto-recorder, which adds support for Windows. At least it builds and the tests pass on Windows.

libwild/src/archive.rs
```diff
@@ -346,18 +358,24 @@ impl<'data> EntryMeta<'data> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[cfg(unix)]
```
davidlattimore: Too much use of cfg will make it hard to maintain. Can you try to reduce the number of cfg directives? Ideally get rid of all cfg directives on use statements.

libwild/src/archive.rs
```diff
@@ -20,6 +22,16 @@ pub(crate) enum ArchiveEntry<'data> {
...
 }
 
+#[cfg(unix)]
+fn path_from_bytes(bytes: &[u8]) -> PathBuf {
+    std::path::Path::new(OsStr::from_bytes(bytes)).to_path_buf()
+}
+
+#[cfg(not(unix))]
+fn path_from_bytes(bytes: &[u8]) -> PathBuf {
+    PathBuf::from(String::from_utf8_lossy(bytes).into_owned())
```
davidlattimore: If a path is supplied that isn't a valid Windows path, then from_utf8_lossy won't do what we want. Most likely we'd end up reporting that the file couldn't be found which would be confusing. Reporting an error or even panicking would likely be better.

libwild/src/archive.rs
```diff
@@ -450,6 +469,7 @@ mod tests {
         inner(path, limit).with_context(|| format!("Failed to process {}", path.display()))
     }
 
+    #[cfg(unix)]
```
davidlattimore: Rather than using `cfg` to disable tests, how about using `#[cfg_attr(not(unix), ignore)]`. That way the test will still be built, it just won't be executed. That will then mean that you won't get dead code warnings for stuff that was only used by that test.

libwild/src/fs.rs
```diff
@@ -2,13 +2,21 @@ use crate::error::Result;
 use std::fs::File;
 
 pub(crate) fn make_executable(file: &File) -> Result {
+    #[cfg(unix)]
     use std::os::unix::prelude::PermissionsExt;
```
davidlattimore, line 8: Can this use-statement just be moved inside the cfg block below?

8Keep:
Thanks for the review. I've pushed a new version.

Some notes:
1) All unit tests now work, none disabled on windows.
2) I had to disable the integration tests, too much broken. Will need to put together windows specific integration tests in the future.
3) I have made some changes to args.rs for char escaping for @argfiles, otherwise it clobbers backslashes in windows paths. I am making an assumption there. It is also not wrapped in a platform specific cfg. Maybe this should be redone.

Cargo build+test now pass

davidlattimore:
Thanks. Looks good. A few CI failures. Looks like you need to run `cargo +nightly fmt`, there's some clippy warnings that need fixing and a test that's failing. Not sure why the test would fail in CI, but not when you run locally, but if you need help to figure it out, let me know.

8Keep:
Ok, I've fixed some lints/errors/format.

I did use cfg a bit more to make the executable bit code only available on unix, let me know if you'd prefer something else there.

The test error was due to clrf endings on windows, and git was thinking that .a file is a text file. I've added a .gitattributes so git doesn't treat it like a text file. I have that config set locally, so I didn't see that error locally. I think this would be right... Let's see what ci says now

davidlattimore:
Looks like there's just one clippy warning left. The clippy command that gets run in CI is `cargo clippy --workspace --target x86_64-unknown-linux-gnu`. Hopefully if you run that you can see the warning.

libwild/src/file_writer.rs
```diff
@@ -316,7 +319,10 @@ impl SizedOutput {
 
         // Making the file executable is best-effort only. For example if we're writing to a pipe or
         // something, it isn't going to work and that's OK.
-        let _ = crate::fs::make_executable(&self.file);
+        #[cfg(unix)]
+        {
+            let _ = crate::fs::make_executable(&self.file);
+        }
```
bjorn3, line 325: Maybe pull the `#[cfg(unix)]` into the `make_executable` body?

Edit: Didn't notice your comment above.

libwild/src/save_dir.rs
```diff
@@ -252,31 +255,44 @@ impl SaveDirState {
...
-                        "Failed to make path `{}` relative to `{}` while copying symlink",
-                        target.display(),
-                        directory.display()
+                        "Failed to symlink {} to {}",
+                        dest_path.display(),
+                        target.display()
+                    )
+                })?;
+            }
+            #[cfg(not(unix))]
```
bjorn3, line 287: I think it would be cleaner to write this as:

```rust
if meta.is_dir() {
    // ...
    return Ok(());
}

#[cfg(unix)]
if meta.is_symlink() {
    // ...
    return Ok(());
}

if let Ok(data) = FileData::new(source_path, false) {
    // ...
}

// hardlink or copy

**Comment by davidlattimore** [comment]
@8Keep - it'd be good to get this merged so that we can progress on https://github.com/davidlattimore/wild/issues/1462. Let us know if you don't think you'll get back to this and we can finish it off.

**Comment by davidlattimore** [comment]
Nothing heard. @bjorn3 would you like to pick this up or should I?

**Comment by Blazearth** [comment]
I ran into the tidy::check_sources_format failure locally.

It turns out the failure is purely due to clang-format enforcing brace spacing in
tests/sources/export-dynamic.c:
```bash
    void foo(void) {};
→  void foo(void){};
```

After running `clang-format -i tests/sources/*.{c,cc,h}`, the tidy test passes.
No functional or linker-related changes, just formatting.

Mentioning this in case anyone else hits the same failure while testing this PR.

8Keep:
So sorry everyone, I've just been very busy and out of town most of the time since. If you need it soon - feel free to take it over

davidlattimore:
Thanks for your work on this. Since #1629 has been merged, I'll close this PR now.

Issue #1381, opened 2025-12-16
LoongArch support


marxin:
I am planning to work on the port in the upcoming weeks.

marxin:
Currently developed under the following git branch: https://github.com/marxin/wild/tree/loongarch64-support.

PR #1409, opened 2025-12-29
feat: LoongArch64 support

marxin:
The LoongArch64 port is looking good based on the integration tests (and the capability to bootstrap Wild itself!). I would like to list a couple of significant changes/needs for the port:

1. The port needs a couple of High/Low part relocations where the former needs a bias (typically `0x800`), so I decided to make it a first class citizen in the `RelocationKindInfo` type (plus I noticed a couple of relocations need to first apply addend before masking -> that's why I touched so many relocation expression calculations).
2. LoongArch64 does not use a specific TLSGD Low part relocations for GOT and so we need to special-case it a bit.
3. The target also uses the page masking, so I decided to make it an explicit parameter, rather than the currently used AArch64 constants.
4. There's one super-complex relocation calculation that I decided to move to a separate function `highest_relocation_with_bias`.
5. Just recently, there was a new addition of relocation in the latest spec release (`v2.50`) - for now I haven't implemented them.
6. `PairSubtractionULEB128` has been factored out as it's something used by both AArch64 and the newly added LoongArch64 port.

Some early benchmarks for linking the Mold project on Loongson-3A6000 8 core CPU:
```
| Linker | With debug info (s) | Without debug info (s) |
|--------|----------------------|-------------------------|
| bfd    | 65.0                 | 24.0                    |
| mold   | 4.4                  | 1.82                    |
| lld    | 10.4                 | 8.0                     |
| wild   | 11.8                 | 0.953                   |
```

I would like to thank @Gelbpunkt for the SSH access to the Alpine LoongArch64 machine! Really helped me.

Resolves: #1381

### DIFF
--- a/linker-utils/src/elf.rs
+++ b/linker-utils/src/elf.rs
@@ -1208,8 +1208,7 @@ impl DynamicRelocationKind {
             DynamicRelocationKind::Relative => object::elf::R_LARCH_RELATIVE,
             DynamicRelocationKind::Absolute => object::elf::R_LARCH_64,
             DynamicRelocationKind::GotEntry => object::elf::R_LARCH_64,
-            // TODO: add new object constant: https://github.com/gimli-rs/object/pull/827
-            DynamicRelocationKind::TlsDesc => 14,
+            DynamicRelocationKind::TlsDesc => object::elf::R_LARCH_TLS_DESC64,
             DynamicRelocationKind::JumpSlot => object::elf::R_LARCH_JUMP_SLOT,
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,7 +784,7 @@ dependencies = [
  "linker-utils",
  "memchr",
  "memmap2",
- "object 0.38.0",
+ "object 0.38.1",
  "perf-event",
  "perfetto-recorder",
  "rayon",
@@ -829,7 +829,7 @@ dependencies = [
  "linker-utils",
  "memchr",
  "memmap2",
- "object 0.38.0",
+ "object 0.38.1",
  "symbolic-demangle",
  "tabled",
  "tempfile",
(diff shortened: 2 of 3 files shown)


wild/tests/sources/section-start.c
```diff
@@ -6,6 +6,10 @@
 //#DiffIgnore:section.rodata.alignment
 //#DiffIgnore:section.data
 //#DiffIgnore:section.sdata
+//#Arch:x86_64,aarch64,riscv64
```
davidlattimore, line 9: I wonder if we should really have a SkipArch directive. If we add another arch, then this test will be skipped for that arch unless we list it here.

linker-utils/src/elf.rs
```diff
@@ -1043,6 +1196,23 @@ impl DynamicRelocationKind {
...
+        match self {
+            DynamicRelocationKind::Copy => object::elf::R_LARCH_COPY,
+            DynamicRelocationKind::Irelative => object::elf::R_LARCH_IRELATIVE,
+            DynamicRelocationKind::DtpMod => object::elf::R_LARCH_TLS_DTPMOD64,
+            DynamicRelocationKind::DtpOff => object::elf::R_LARCH_TLS_DTPREL64,
+            DynamicRelocationKind::TpOff => object::elf::R_LARCH_TLS_TPREL64,
+            DynamicRelocationKind::Relative => object::elf::R_LARCH_RELATIVE,
+            DynamicRelocationKind::Absolute => object::elf::R_LARCH_64,
+            DynamicRelocationKind::GotEntry => object::elf::R_LARCH_64,
+            // TODO: add new object constant: https://github.com/gimli-rs/object/pull/827
```
davidlattimore: PR has been merged. Given that Cargo.toml currently pulls object from github, I assume this could be changed now?

wild/tests/sources/section-start.c
marxin: Good idea, hope it's fine doing that in a separate PR.

PR #1416, opened 2026-01-04
CI: include next upcoming LTS Ubuntu release

marxin:


marxin:
@davidlattimore Can you please take a look at the new linker-diff failures?

davidlattimore:
Looks like GNU ld now uses `VER_NDX_LOCAL` for undefined symbols without a version rather than `VER_NDX_GLOBAL`. This appears to be from [this commit](https://github.com/bminor/binutils-gdb/commit/f685e3953f9a38a41bbd0a597f9882870cee13d5). This [later change](https://github.com/bminor/binutils-gdb/commit/78894bcb483d45c4d7be5ef9499ec017a9727079) is also related.

LLD still emits `VER_NDX_GLOBAL` for undefined symbols without a version.

[Maskray's blog post](https://maskray.me/blog/2020-11-26-all-about-symbol-versioning) on symbol versioning has this to say:

> * Index 0 is called VER_NDX_LOCAL (misleading name, should have been VER_NDX_NONE). For defined symbols, the binding of the symbol will be changed to STB_LOCAL.
> * Index 1 is called VER_NDX_GLOBAL. It has no special effect and is used for unversioned symbols.
> * Index 2 to 0xffef are used for user defined versions.

[The spec](https://sourceware.org/gnu-gabi/program-loading-and-dynamic-linking.txt#:~:text=versioning) doesn't seem to really say which is correct for undefined symbols. It just says:

>   VER_NDX_LOCAL  0 - The symbol is private, and is not available outside this object.
>   VER_NDX_GLOBAL 1 - The symbol is globally available (ie the base or default version).

I guess maybe it doesn't matter which is used, so perhaps we should change linker-diff to treat them as the same, at least for undefined symbols.

The other option would be enable LLD for all affected tests, although there are quite a few.

davidlattimore:
I've created https://github.com/davidlattimore/wild/pull/1424 to make linker-diff ignore the difference. Feel free to merge that PR if you like. With that PR, the tests now pass for me on a 26.04 docker image.

marxin:
> I'm no longer developing on 22.04, so I'd be happy to remove that unless folks think we still need it for some reason.
Good idea - I welcome that.

Anyway, thanks for the investigation related to the symbol versioning.

LeoniePhiline:
Debian Bookworm only provides glibc 2.36, while new `wild-linker` builds now require glibc 2.39.

Would you recommend users of Debian oldstable use the musl builds, or is there any chance you could consider providing backward compatibility?

davidlattimore:
Thanks for bringing that to our attention @LeoniePhiline. I've raised #1490. We'll ensure that the next release has lower glibc requirements for our release builds. In the meantime, if you have a rust toolchain installed, I'd probably suggest `cargo install wild-linker`. The musl build would also be fine.

PR #1418, opened 2026-01-04
feat: Support expressions in `--defsym` and linker script symbol assignment

lapla-cogito:
Previously, symbol assignments in the `--defsym` option or  in linker scripts could only contain absolute addresses or symbol names on the right-hand side of the assignment. However, other existing linkers allow well-formed expressions on the right-hand side.

The following statements are from the description of the `--defsym` option in the GNU ld man page:

> A limited form of arithmetic is supported for the expression in this context: you may specify a hexadecimal constant or the name of an existing symbol, or use "+" and "-" to add or subtract hexadecimal constants or symbols. For more complex expressions, consider using the linker command language from a script. Note: there should be no whitespace between the symbol, the equals sign ("="), and the expression.

Additionally, documentation on symbol assignments in linker scripts can be found here:
https://sourceware.org/binutils/docs-2.17/ld/Simple-Assignments.html#Simple-Assignments

While this documentation mentions expressions such as `symbol += expression`, it is unclear whether these are widely used in practice. Therefore, I plan to implement this support in a separate PR.

### DIFF
--- a/libwild/src/parsing.rs
+++ b/libwild/src/parsing.rs
@@ -131,7 +131,7 @@ pub fn parse_symbol_expression(s: &str) -> ParsedSymbolExpression<'_> {
 
     for (i, ch) in s.bytes().enumerate().skip(token_start) {
         if ch == b'+' || ch == b'-' {
-            let token = &s[token_start..i];
+            let token = s[token_start..i].trim();
             if let Ok(val) = parse_number(token) {
                 offset = offset.wrapping_add(current_sign * val as i64);
             } else if symbol.is_none() && !token.is_empty() {
@@ -143,7 +143,7 @@ pub fn parse_symbol_expression(s: &str) -> ParsedSymbolExpression<'_> {
     }
 
     // Process the last token
-    let token = &s[token_start..];
+    let token = s[token_start..].trim();
     if let Ok(val) = parse_number(token) {
         offset = offset.wrapping_add(current_sign * val as i64);
     } else if symbol.is_none() && !token.is_empty() {
--- a/wild/tests/sources/linker-script.ld
+++ b/wild/tests/sources/linker-script.ld
@@ -10,5 +10,5 @@ SECTIONS {
 defsym_start_aaa = start_aaa;
 defsym_addr = 0x1234;
 
-defsym_decimal = defsym_addr+10;
-defsym_hex = defsym_addr+0x10;
+defsym_decimal = defsym_addr + 10;
+defsym_hex = defsym_addr + 0x10;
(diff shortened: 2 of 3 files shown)


libwild/src/parsing.rs
```diff
@@ -91,8 +91,78 @@ pub(crate) enum SymbolPlacement<'data> {
...
+        match self {
+            ParsedSymbolExpression::Absolute(value) => SymbolPlacement::DefsymAbsolute(value),
+            ParsedSymbolExpression::SymbolWithOffset(sym, offset) => {
+                SymbolPlacement::DefsymSymbol(sym, offset)
+            }
+        }
+    }
+}
+
+pub fn parse_symbol_expression(s: &str) -> ParsedSymbolExpression<'_> {
```
davidlattimore, line 118: Are there any parse errors that we should report?

wild/tests/sources/linker-script.ld
```diff
@@ -9,3 +9,6 @@ SECTIONS {
 }
 defsym_start_aaa = start_aaa;
 defsym_addr = 0x1234;
+
+defsym_decimal = defsym_addr+10;
```
davidlattimore: Is whitespace permitted around the `+`?

libwild/src/parsing.rs
lapla-cogito: For example, having multiple symbols on the right side of an expression is problematic. While we could consider implementing an error check for this, at least GNU ld appears to simply ignore such cases:
```
（▰╹◡╹）❯  cat main.c

int test_func(void) {
    return 1;
}

int main(void) {
}

（▰╹◡╹）❯  gcc main.c -Wl,--defsym=foo=test_func+test_func

（▰╹◡╹）❯  readelf -s a.out
...
Symbol table '.symtab' contains 37 entries:
...
    23: 0000000000001129    15 FUNC    GLOBAL DEFAULT   12 test_func
...
    28: 00000000000001d2     0 NOTYPE  GLOBAL DEFAULT  ABS foo
```

wild/tests/sources/linker-script.ld
lapla-cogito: Thanks, the parser now accepts whitespace.

libwild/src/parsing.rs
davidlattimore: I'd probably be inclined to report errors for such cases, but up to you. It seems unlikely that anyone would make such a mistake. I guess other errors could be numbers that don't parse correctly. e.g. if someone were to write a hex number without the "0x" prefix, although if the first digit is a letter then it'd be ambiguous anyway.

PR #1421, opened 2026-01-05
CI: Run loongarch64 external tests

lapla-cogito:


marxin:
Thank you @lapla-cogito!

Issue #1396, opened 2025-12-20
CREL relocations consume too much memory


davidlattimore:
Chromium has switched to using CREL relocations by default when building if it thinks the linker is LLD. Wild supports CREL relocations, but apparently uses heaps more RAM when doing so.

I can sometimes link Chromium on my mini PC, which has 92GB of RAM. Apparently Wild hits peak RSS of 94GB, so I guess a little bit of swap space is being used. This is in contrast to linking Chromium without CREL relocations which hits peak RSS of less than 8GB.

You can switch whether Chromium is built with CREL relocations by editing `build/config/compiler/BUILD.gn` and commenting out the line that sets the flags `-Wa,--crel,--allow-experimental-crel` .

We currently copy all the CREL relocations for a section into a Vec so that we can access relocations by index. I'm not sure if a bidirectional CREL iterator is possible, but if it is, then we might not need the Vec. It'd probably also help if the CREL iterator was smaller. It's currently 80 bytes, so if we made lots of copies of different iterators, it might take up a bit of space. I'm not sure how practical it would be to reduce its size. I think the main spot where we would need lots of copied iterators is when storing the relocations for eh_frame for each section. Perhaps there's a different approach we could take there that might avoid the need for that, but I haven't thought too much about it yet.

marxin:
Let me take a look at the issue - I think we can simplify the logic where we currently seek in the relocation stream (when finding a particular HI-LO part for the RISC-V target).

davidlattimore:
Turned out that I was wrong about the cause of the high memory consumption. It wasn't related to the Vecs that we were storing. Rather, it was because we weren't discarding the '.crel' sections (like we do for '.rela' sections). That meant that every '.crel' section was being made into a custom section, since they all have different names. We don't handle having lots of custom sections well, so were blowing up memory usage. This was fixed by #1438.

Thanks @marxin for the work to optimise processing of CREL relocations and sorry for sending you down the wrong path! That said, the changes you made were still beneficial. I just benchmarked the effect of #1426 on Linking chrome CREL. The baseline is with #1426 reverted.

```
Benchmark 1 (114 runs): /home/david/save/chrome-crel/run-with target/release/wild --strip-debug --no-fork --threads=32
  measurement          mean ± σ            min … max           outliers         delta
  wall_time          1.06s  ± 9.08ms    1.04s  … 1.09s           1 ( 1%)        0%
[...]

PR #1426, opened 2026-01-06
fix(CREL): reduce allocation for non-EH relocations

marxin:
This is a first part of the fix for #1396 where we currently allocate a vector for `CRelIterator` when we apply relocations.
My solution uses a RelocationCache instead, which is likely leading to a slower link time, but just on targets that use HI/LO relocations (RISC-V).

@davidlattimore am interested if it helps with the peak memory on Chromium?

One merged, I'm planning to drop `DynamicRelocationSequence` type as well and use again an iterator ;)


davidlattimore (approved):
Unfortunately it looks like peak RSS when linking chromium with CREL didn't change much. It was about 92GB both with and without this change. I guess most of the memory is related to eh_frame relocations.

I benchmarked linking wild for riscv64, although I ran the benchmark on x86_64 (i.e. cross linking). It looks like this change actually sped it up by about 7%.

marxin:
> It looks like this change actually sped it up by about 7%.

Even better ;)

PR #1428, opened 2026-01-06
linker-diff: LoongArch64 support

lapla-cogito:
ABI spec: https://github.com/loongson/la-abi-specs/blob/release/laelf.adoc


davidlattimore (approved):
Thanks for implementing this!

marxin:
Thank you @lapla-cogito.

PR #1441, opened 2026-01-11
fix: ifunc address equality for data section references

lapla-cogito:
When an ifunc is referenced from a data section (e.g., `void *ptr = my_ifunc`), the pointer should have the same value as a direct function pointer reference to ensure address equality comparisons work correctly.
For dynamic executables (both PIE and non-PIE), we now emit IRELATIVE relocations in `.rela.dyn` for such references, allowing the dynamic linker to resolve them to the same target address.
For static executables without a dynamic linker, the PLT address is used directly, which is handled automatically by `value_with_addend()`.


davidlattimore (approved):
Nice improvement!

PR #1443, opened 2026-01-11
feat: include metrics connected to '.eh_frame'

marxin:


libwild/src/layout.rs
```diff
@@ -2570,6 +2575,18 @@ fn find_required_sections<'data, A: Arch>(
     let mut group_states = unwrap_worker_states(&resources.worker_slots);
     let sections_with_content = resources.sections_with_content.into_map(|v| v.into_inner());
 
+    let total: usize = group_states
```
davidlattimore: I think generally it's best to do the computation inside the `tracing::debug!` invocation. That way the computation will be skipped when the trace message isn't being emitted.

marxin: Sure - makes sense!

Issue #1442, opened 2026-01-11
error: Failed copying from <epilogue> to output file after sort capability of `.init_array/.fini_array`


marxin:
I noticed after the revision a152d28a15bace4e705977fcf238835acf0d82f8 we can't link `rustc`:

```
❯ ./run-with ~/Programming/wild/ld
WARNING: wild: --plugin /usr/lib/gcc/x86_64-pc-linux-gnu/15.1.1/liblto_plugin.so is not yet supported
wild: error: Failed copying from <epilogue> to output file
  Caused by:
    Failed to copy dynamic symbol `std::sys::args::unix::imp::ARGV_INIT_ARRAY` (845180 local=4481) in file #41249 (161/33) (./home/marxin/BigProgramming/rust/build/x86_64-unknown-linux-gnu/stage0-sysroot/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-cd399c36e87ede14.rlib @ std-cd399c36e87ede14.std.c965b0aa6c3056df-cgu.0.rcgu.o) (DIRECT | EXPORT_DYNAMIC)
    internal error: tried to copy symbol `_ZN3std3sys4args4unix3imp15ARGV_INIT_ARRAY17haf171eaa51dd6a2cE` that's in section 54 which is not being output
```

I uploaded the reproducer here: https://splichal.eu/tmp/rustc.tar.zst

CC: @YamasouA

YamasouA:
Is this the reason?
https://github.com/davidlattimore/wild/pull/1445

PR #1445, opened 2026-01-11
bug fix init_array order

YamasouA:
#1442


marxin:
I can confirm it fixes the underlying issue!

PR #1446, opened 2026-01-12
fix: ifunc address equality for GOT-relative relocations

lapla-cogito:
When code takes the address of an ifunc symbol via a GOT-relative relocation (e.g., `R_X86_64_REX_GOTPCRELX`), it should return the same address as direct references to the ifunc, which is the PLT stub address. Previously, wild used a single GOT entry for both the PLT stub (with `R_IRELATIVE`) and GOT-relative references. This caused address inequality: direct references returned the PLT stub address, while GOT-relative references returned the resolver result (the actual function address). For non-PIE executables, we now allocate a separate GOT entry for address equality that contains the PLT stub address statically. This ensures all references to an ifunc return the same address.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -900,7 +900,6 @@ pub(crate) struct Resolution {
     /// offset within module).
     pub(crate) got_address: Option<NonZeroU64>,
     pub(crate) plt_address: Option<NonZeroU64>,
-    pub(crate) ifunc_got_for_address: Option<NonZeroU64>,
     pub(crate) flags: ValueFlags,
 }
 
@@ -934,7 +933,6 @@ impl SectionResolution {
             dynamic_symbol_index: None,
             got_address: None,
             plt_address: None,
-            ifunc_got_for_address: None,
             flags: ValueFlags::empty(),
         })
     }
@@ -5720,7 +5718,6 @@ fn create_resolution(
         dynamic_symbol_index,
         got_address: None,
         plt_address: None,
-        ifunc_got_for_address: None,
         flags,
     };
     if flags.needs_plt() {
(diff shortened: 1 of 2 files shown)


libwild/src/layout.rs
```diff
@@ -899,6 +899,7 @@ pub(crate) struct Resolution {
     /// offset within module).
     pub(crate) got_address: Option<NonZeroU64>,
     pub(crate) plt_address: Option<NonZeroU64>,
+    pub(crate) ifunc_got_for_address: Option<NonZeroU64>,
```
davidlattimore: We create a lot of resolutions, so increasing the size of the struct is likely to have a performance impact. Could we do the same thing that we do for TLS GOT entries and have this GOT entry be relative to `got_address` so that we don't need a separate field to store it?

I just ran a benchmark to sure sure. Here are the results for zed:

```
Benchmark 1 (220 runs): /home/david/save/zed/run-with /home/d/wild-builds/2026-01-13 --strip-debug --no-fork
  wall_time           546ms ± 12.8ms     512ms …  573ms          0 ( 0%)        0%
  peak_rss           3.49GB ± 1.11MB    3.48GB … 3.49GB          2 ( 1%)        0%
  cpu_cycles         22.6G  ±  218M     22.0G  … 23.1G           0 ( 0%)        0%
  instructions       20.3G  ± 65.3M     20.2G  … 20.6G           7 ( 3%)        0%
  cache_references    489M  ± 3.52M      482M  …  501M          10 ( 5%)        0%
  cache_misses        127M  ±  772K      126M  …  130M           1 ( 0%)        0%
  branch_misses      38.1M  ±  171K     37.2M  … 38.7M           7 ( 3%)        0%
Benchmark 2 (216 runs): /home/david/save/zed/run-with target/dist/wild --strip-debug --no-fork
  wall_time           556ms ± 14.4ms     521ms …  586ms          0 ( 0%)        💩+  1.7% ±  0.5%
  peak_rss           3.54GB ± 1.07MB    3.54GB … 3.55GB          4 ( 2%)        💩+  1.6% ±  0.0%
  cpu_cycles         22.8G  ±  217M     22.2G  … 23.4G           0 ( 0%)          +  0.9% ±  0.2%
  instructions       20.5G  ± 62.2M     20.4G  … 20.7G           7 ( 3%)          +  0.8% ±  0.1%
  cache_references    492M  ± 3.38M      486M  …  505M          11 ( 5%)          +  0.6% ±  0.1%
  cache_misses        127M  ±  809K      126M  …  129M           0 ( 0%)          -  0.2% ±  0.1%
  branch_misses      39.0M  ±  233K     37.9M  … 39.5M           6 ( 3%)        💩+  2.5% ±  0.1%
```

lapla-cogito: Thanks, it makes sense.

PR #1448, opened 2026-01-12
feat: Support `-z stack-size=SIZE`

lapla-cogito:
This option is commonly used in environments with small default stack sizes like musl, or in scenarios where the stack size is typically specified explicitly, such as with WebAssembly: https://github.com/search?q=%22-z+stack-size%22&type=code
(Of course, we still don't support linking to WebAssembly yet)

### DIFF
--- a/wild/tests/external_tests/mold_skip_tests.toml
+++ b/wild/tests/external_tests/mold_skip_tests.toml
@@ -109,7 +109,6 @@ tests = [
   "z-pack-relative-relocs.sh",
   "z-rodynamic.sh",
   "z-sectionheader.sh",
-  "z-stack-size.sh",
   "z-start-stop-visibility.sh",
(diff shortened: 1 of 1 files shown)


wild/tests/sources/stack-size.c
```diff
@@ -0,0 +1,13 @@
+//#AbstractConfig:default
+//#Object:runtime.c
+
+//#Config:stack-size:default
+//#LinkArgs:-z stack-size=0x1000 -z now
+//#DiffIgnore:section.note.gnu.property
```
lapla-cogito, line 6: Because of this: https://github.com/davidlattimore/wild/blob/ca8a334062db67c0b5c8d90f2c234b0584541aa2/libwild/src/layout.rs#L5517

PR #1452, opened 2026-01-13
fix: Address equality for exported ifunc symbols

lapla-cogito:
This is the final patch set in a series of changes addressing address equivalence for ifunc functions. When running the tests in `ifunc-export.c` with the current `main` implementation, the dynamic linker reports a circular dependency at runtime:
```
---- integration_test::program_name_97___ifunc_export_c__ stdout ----
Error: Failed to run program. Binary `/home/lapla/repos/wild/wild/tests/build/ifunc-export.c/ifunc-export.c-default-host.wild`. Relink with:
WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/home/lapla/repos/wild/wild/tests/build/ifunc-export.c/ifunc-export.c-default-host.wild /home/lapla/repos/wild/wild/tests/build/ifunc-export.c/ifunc-export.c-default-host.save/run-with cargo run --bin wild --
  Caused by:
    Binary exited with unexpected exit code 127: /home/lapla/repos/wild/wild/tests/build/ifunc-export.c/ifunc-export.c-default-host.wild: IFUNC symbol 'foo' referenced in '/home/lapla/repos/wild/wild/tests/build/ifunc-export.c/ifunc-export-1.default-host-5051386a5e303a08.wild.so' is defined in the executable and creates an unsatisfiable circular dependency.
```

Even when an ifunc symbol is exported from a binary and referenced by a dynamic library, Wild has historically resolved this symbol with the `STT_GNU_IFUNC` type and kept the resolution address. In such cases, the actual processing required should be as follows:
1. Change the symbol value to the PLT stub address
2. Change the symbol type from `STT_GNU_IFUNC` to `STT_FUNC`
3. Change the symbol's section to the PLT section

### DIFF
--- a/wild/tests/sources/ifunc-export.c
+++ b/wild/tests/sources/ifunc-export.c
@@ -1,5 +1,6 @@
-//#AbstractConfig:default
+//#Config:ifunc-export
 //#LinkerDriver:gcc
+//#CompArgs:-fno-pic -fno-pie
 //#LinkArgs:-no-pie -Wl,-z,now
 //#Shared:ifunc-export-1.c:-fPIC
 //#RequiresGlibc:true
@@ -8,11 +9,6 @@
 //#DiffIgnore:.dynamic.DT_NEEDED
 //#DiffIgnore:section.data
 //#DiffIgnore:dynsym.foo.section
-
-//#Config:ifunc-export-base:default
-
-//#Config:nopic:default
-//#CompArgs:-fno-pic -fno-pie
 
 typedef void (*Func)(void);
 static int foo_called = 0;


davidlattimore (commented):
Nice series of PRs!

Looks like CI failed on openSUSE. When you figure out what the problem is, if there's a way to repro the same failure on other platforms, can you extend the test case? Might just be an additional flag. I think openSUSE has some different defaults, although I forget the details.

lapla-cogito:
The test failure on OpenSUSE was caused by the creation of `R_X86_64_32S` relocations, which set the `DIRECT` flag. To make the test OpenSUSE-independent, I added a test that adds `-fno-pic -fno-pie` compilation flags:
```
（▰╹◡╹）❯  readelf -r wild/tests/build/ifunc-export.c/ifunc-export.nopic-host-*.o | grep "R_X86_64_32S"
000000000034  00070000000b R_X86_64_32S      foo()            foo + 0
```

However, the test failure in Nix seems to be due to another issue caused by GNU ld, so I'll continue investigating...

lapla-cogito:
During CI testing on NixOS, I investigated why test binaries linked with GNU ld were failing with exit code 1. As a result, GCC 14 (on NixOS) creates `R_X86_64_REX_GOTPCRELX` relocations for address resolution, whereas GCC 15 creates `R_X86_64_PC32` relocations. This discrepancy was causing the CI failures. As noted above, specifying `-fno-pic -fno-pie` fixes the generated relocations, so I updated the existing tests to apply these flags and standardize the relocation behavior.

mati865:
This PR is the cause for https://github.com/davidlattimore/wild/issues/1485.
What should we do to fix it?

PR #1458, opened 2026-01-16
Use parent directory of args.tmp for filesystem checking

TechnoPorg:
As per the discussion in #1454.


benchmarks/runner/src/benchmarking.rs
```diff
@@ -59,7 +60,8 @@ pub(crate) fn run_bench(args: &BenchArgs, config: &Config) -> Result {
 }
 
 fn check_tmpfs(args: &BenchArgs) -> Result {
-    let tmpdir = &args.tmp;
+    let tmpfile = current_dir()?.join(&args.tmp); // Allow relative path
```
davidlattimore: I think std::path::absolute would be clearer - `https://doc.rust-lang.org/stable/std/path/fn.absolute.html`

PR #1460, opened 2026-01-16
Use the object crate for parsing archive files

bjorn3:
This saves 257 lines of code. The object crate also supports a couple more archive formats, but they seem to mostly differ by how the symbol table is encoded, which wild doesn't use. The main exception would be aix big, which should now be supported. Rustc also uses the object crate for reading archives, so using it in wild keeps archive format support in sync between wild and rustc.

### DIFF
--- a/libwild/src/archive.rs
+++ b/libwild/src/archive.rs
@@ -2,33 +2,19 @@
 //! Read trait and we want to borrow the data of each entry. We do however use the ar crate as a dev
 //! dependency in our tests so that we can verify consistency.
 
-use crate::bail;
-use crate::error::Context as _;
 use crate::error::Result;
 use std::ffi::OsStr;
 use std::ops::Range;
 use std::os::unix::ffi::OsStrExt as _;
 use std::path::Path;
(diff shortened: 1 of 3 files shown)


bjorn3:
Fixed clippy and ran nightly rustfmt.

mati865:
Nice! Have you checked if performance has changed in any direction?

bjorn3:
I haven't. If it changes at all, I would personally expect the difference to be within a couple of percent, which is too little for me to reliably detect on my laptop. Thermal throttling and turbo boost easily cause much more significant perf changes.

davidlattimore:
Thanks for doing this! I hadn't realised the object crate supported this.

I've just benchmarked linking of Zed and Wild and haven't see any measurable change.

What should go in the release notes? I can put the current title as-is in the refactoring section, unless there's something more appropriate?

bjorn3:
This is effectively just a refactoring. While this adds support for reading AIX archives, AIX uses XCOFF, so it still wouldn't be supported. Using the PR title as changelog entry seems fine.

PR #1466, opened 2026-01-17
fix: delete the file instead of truncating it when updating busy executable

karolzwolak:
When we try to update a running executable, and we fail because the file is busy, we retried with truncate which fails too. Instead, we have to remove the file, and create it again.

I encountered this bug sporadically when building stuff in the LLVM repo.

### DIFF
--- a/libwild/src/file_writer.rs
+++ b/libwild/src/file_writer.rs
@@ -277,7 +277,9 @@ impl SizedOutput {
                         FileWriteMode::UpdateInPlaceWithFallback
                     )
                 {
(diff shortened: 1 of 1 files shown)


karolzwolak:
I don't think our tests infra can tests this scenario. We'd have to have to try to modify some binary while it was running. Maybe we can rely on mold test suite to cover this.

karolzwolak:
I see no new passes on mold test suite.
But weirdly enough, `--gbd-index` tests that should be failing, are passing (not that this changed with this PR).

Issue #1472, opened 2026-01-19
`integration_test::program_name_91___section_start_c__` fails on riscv64 musl


Kladki:
With this latest release, `0.8.0`, the aforementioned test fails on riscv64 on musl:
```
failures:
---- integration_test::program_name_91___section_start_c__ stdout ----
Error: Failed to run program. Binary `/builds/Matthias/aports/community/wild/src/wild-0.8.0/wild/tests/build/section-start.c/section-start.c-section-start-host.wild`. Relink with:
WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/builds/Matthias/aports/community/wild/src/wild-0.8.0/wild/tests/build/section-start.c/section-start.c-section-start-host.wild /builds/Matthias/aports/community/wild/src/wild-0.8.0/wild/tests/build/section-start.c/section-start.c-section-start-host.save/run-with cargo run --bin wild --
  Caused by:
    Binary exited (core dumped)  with signal 11: 
failures:
    integration_test::program_name_91___section_start_c__
test result: FAILED. 97 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 18.50s
```
https://gitlab.alpinelinux.org/Matthias/aports/-/jobs/2179603#L821

lapla-cogito:
I haven't yet tested this issue locally, but I suspect the problem might be due to the large starting address for `foo` being hardcoded in `section-start.c`, which is interfering with the base address of the PIE.

davidlattimore:
Looks like the segfault occurs if the binary is linked with `-no-pie`. The `--section-start` flag used by that test isn't needed for it to reproduce. The segfault occurs during startup. One of the init_array functions - `frame_dummy` calls the PLT entry for `__register_frame_info`. That PLT entry then tries to call address 0.

`frame_dummy` comes from `crtbegin.o`. The reference to `__register_frame_info` is weak. `__register_frame_info` is defined in `libgcc_s.so.1`, however we didn't link against that shared object because it was `--as-needed` and nothing needed it.

Before calling the `frame_dummy` PLT, `__register_frame_info` checks if the address of `__register_frame_info` is null. It appears that check is incorrectly checking if the PLT entry address is null rather than checking if the function address is null.

I've been trying to reproduce this in a more convenient environment, but haven't managed as yet.

lapla-cogito:
Ah, I missed that `-no-pie` was specified.
[...]

PR #1474, opened 2026-01-20
fix: Resolve weak undefined symbols to zero in read-only sections

lapla-cogito:
#1472 

When a weak undefined symbol is referenced from a read-only section via an absolute relocation, the resolved value should be 0, not the PLT entry address.

### DIFF
--- a/wild/tests/sources/undefined-weak-sym.c
+++ b/wild/tests/sources/undefined-weak-sym.c
@@ -1,0 +1,25 @@
+// A test for #1472.
+
+//#Object:runtime.c
+//#CompArgs:-fno-PIC
+//#Mode:dynamic
+//#Shared:force-dynamic-linking.c
+//#DiffIgnore:section.got
+//#DiffIgnore:.dynamic.DT_NEEDED
+//#DiffIgnore:.dynamic.DT_RELA
+//#DiffIgnore:.dynamic.DT_RELAENT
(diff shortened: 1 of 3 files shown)


wild/tests/sources/undefined-weak-sym.c
```diff
@@ -0,0 +1,21 @@
+// A test for #1472.
+
+//#Object:runtime.c
+//#CompArgs:-fno-PIC
+//#Mode:dynamic
+//#Shared:force-dynamic-linking.c
+//#DiffEnabled:false
```
davidlattimore: Did you intend to leave diffing disabled?

lapla-cogito: Yes, the purpose of this test is to verify that no segfaults occur, so I deliberately disabled diffing. If this test needs more appropriate settings, please let me know.

davidlattimore: Having diffing enabled would still test that, it's just that it would diff first, so wouldn't get to the segfault. A test with diffing is a stronger test than a test without diffing. Also, if the bug were to be reintroduced, the diff failure is easier to interpret than the segfault. I think the main reason for disabling diffing is if we don't have a reference linker to diff against - e.g. tests where we disable linking with GNU ld. But that doesn't apply here.

lapla-cogito: Understood. Diffing has been enabled.

PR #1475, opened 2026-01-21
chore: Install additional dependencies in external tests

lapla-cogito:
context: https://github.com/davidlattimore/wild/pull/1466#pullrequestreview-3675223830

Some mold tests check for specific tool availability and skip them if the tool is missing (e.g., `command -v gdb >& /dev/null || skip`), which can lead to tests passing accidentally if additional packages aren't installed.

### DIFF
--- a/wild/tests/external_tests/mold_skip_tests.toml
+++ b/wild/tests/external_tests/mold_skip_tests.toml
@@ -114,6 +114,10 @@ tests = [
 reason = "x86_64 specific tests"
 tests = [
   "arch-x86_64-address-equality.sh",
+  "arch-x86_64-apx-gotpcrelx.sh",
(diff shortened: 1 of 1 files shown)


wild/tests/external_tests/mold_skip_tests.toml
```diff
@@ -260,7 +264,6 @@ tests = [
   "versioned-undef.sh",
   "warn-symbol-type.sh",
   "weak-export-exe.sh",
-  "weak-undef.sh",
```
lapla-cogito, line 263: This can be removed from [this commit](https://github.com/davidlattimore/wild/commit/5805151cffccc087942b9dd8d7ac10ca5f4f80a8).

PR #1477, opened 2026-01-22
feat: Allow specifying library search paths in version scripts and export lists

lapla-cogito:
When the `-L` option specifies a library search path, Wild now also searches that path for version scripts and export lists if none are found in the current directory.

### DIFF
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -192,16 +192,26 @@ pub(crate) struct AuxiliaryFiles<'data> {
 
 impl<'data> AuxiliaryFiles<'data> {
     pub(crate) fn new(args: &'data Args, inputs_arena: &'data Arena<InputFile>) -> Result<Self> {
+        let resolve_script_path = |path: &Path| -> PathBuf {
+            if path.exists() {
+                path.to_owned()
+            } else if let Some(found) = search_for_file(&args.lib_search_path, None, path) {
+                found
+            } else {
+                path.to_owned()
+            }
+        };
+
         Ok(Self {
             version_script_data: args
                 .version_script_path
                 .as_ref()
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):

libwild/src/input_data.rs
```diff
@@ -192,16 +192,26 @@ pub(crate) struct AuxiliaryFiles<'data> {
...
                 .version_script_path
                 .as_ref()
-                .map(|path| read_script_data(path, inputs_arena))
+                .map(|path| read_script_data(&resolve_script_path(path), inputs_arena))
                 .transpose()?,
             export_list_data: args
                 .export_list_path
                 .as_ref()
-                .map(|path| read_script_data(path, inputs_arena))
+                .map(|path| read_script_data(&resolve_script_path(path), inputs_arena))
```
lapla-cogito, line 214: At first, I thought there was no need to apply the same change to `export_list`, based on how mold implements the corresponding part. However, I found that GNU ld searches paths specified via `-L` when options such as `--dynamic-list` are provided. Therefore, to match GNU ld’s behavior, this change also applies to `export_list`.
```
（▰╹◡╹）❯  tree .
.
├── searchdir
│   └── test.list
└── test.c

2 directories, 2 files

（▰╹◡╹）❯  gcc -shared -o test.so test.c -fPIC -Wl,-Lsearchdir -Wl,--dynamic-list,test.list -fuse-ld=mold
mold: fatal: cannot open test.list: No such file or directory
collect2: error: ld returned 1 exit status

（▰╹^╹）❯  gcc -shared -o test.so test.c -fPIC -Wl,-Lsearchdir -Wl,--dynamic-list,test.list -fuse-ld=bfd

（▰╹◡╹）❯  ls test.so
test.so
```

PR #1478, opened 2026-01-22
chore: Add configuration for Zed

lapla-cogito:


davidlattimore (approved):

.zed/settings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "lsp": {
+    "rust-analyzer": {
+      "initialization_options": {
+        "linkedProjects": ["Cargo.toml"],
```
mati865: This should be a noop.

lapla-cogito: Initially I tried to align with VSCode's settings, but as you mentioned, whether this is present or not won't affect the behavior.

PR #1484, opened 2026-01-25
build(deps): bump to gimli 0.33

marxin:


Cargo.toml
```diff
@@ -43,7 +43,7 @@ dhat = { version = "0.3.3" }
 fallible-iterator = "0.3.0"
```
lapla-cogito: Can't we peel off our dependency on `fallible_iterator`?

marxin: Sure - good point.

PR #1494, opened 2026-01-28
feat: Allow excluding symbols from specific libraries via `--exclude-libs`

### LINKED ISSUE #812: Add support for -exclude-libs
This is currently required when building MLIR with wild set as the linker via `--ld-path` unless the cmake build files are patched.  See https://github.com/llvm/llvm-project/blob/main/mlir/lib/ExecutionEngine/CMakeLists.txt#L212

Here's the cmake configure command that I used:
```
cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXE_LINKER_FLAGS="--ld-path=wild" -DCMAKE_SHARED_LINKER_FLAGS="--ld-path=wild" -DLLVM_ENABLE_ASSERTIONS=ON -DCLANG_ENABLE_CIR=ON -DCLANG_DEFAULT_LINKER=lld -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra;mlir" -DLLVM_CCACHE_BUILD=ON -DLLVM_ENABLE_RUNTIMES="compiler-rt;libcxx;libcxxabi;libunwind" -S llvm -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DLLVM_INCLUDE_TESTS=OFF
```

Also, `-r` (creating a relocatable object file) seems to be required for compiler-rt but I figured getting compiler-rt to link is less of a priority.

lapla-cogito:
close #812

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -39,6 +39,8 @@
 //! NoSym:symbol-name Checks that the specified symbol name is not defined in either .symtab or
 //! .dynsym.
 //!
+//! NoDynSym:symbol-name Checks that the specified symbol name is not defined in .dynsym.
+//!
 //! ExpectComment: Checks that the comment in the .comment section is equal to the supplied
 //! argument. If no ExpectComment directives are given then .comment isn't checked. The argument may
 //! end with '*' which matches anything.
@@ -836,6 +838,7 @@ struct Assertions {
     expected_dynsym_entries: Vec<ExpectedSymtabEntry>,
     expected_comments: Vec<String>,
     no_sym: HashSet<String>,
+    no_dynsym: HashSet<String>,
     does_not_contain: Vec<String>,
     contains_strings: Vec<String>,
     expect_dynamic: bool,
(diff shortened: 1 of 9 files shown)


wild/tests/integration_tests.rs
```diff
@@ -1681,17 +1687,26 @@ fn build_linker_input(
...
+    let first_source_filename = dep
+        .files
+        .first()
+        .context("At least one file is required")?
+        .filename
+        .as_str();
+    let archive_basename = Path::new(first_source_filename)
+        .file_stem()
+        .context("Invalid source filename")?;
+    let archive_path = config.build_dir.join(archive_basename).with_extension("a");
```
lapla-cogito, line 1713: In previous implementations, archive files created by directives like `Archive` included a hash in their filenames, which prevented specifying library names using `--exclude-libs`. Therefore, this rename library files with a `.a` extension instead (e.g., `foo.c` becomes `foo.a`).

davidlattimore: Since I recently changed each test to have its own subdirectory, I had been wondering if we should go further and put each test config into a subdirectory. So it'd be `build/test-name/config-name/*`. If we do that, then I don't think we need the hashes anymore. I guess as you've got this at the moment, if a test has two configurations, the archives for each config will end up with the same name, so the file will get overwritten? Possibly that doesn't matter for the moment, especially if the tests that use this don't have multiple configurations.

lapla-cogito: > if we should go further and put each test config into a subdirectory. So it'd be `build/test-name/config-name/*`.

Ah, nice idea. I'll implement that first and then rebase this branch. That way, we should be able to revert these changes.

davidlattimore (approved):

PR #1495, opened 2026-01-29
test: Use a separate subdirectory for each test config

lapla-cogito:
context: https://github.com/davidlattimore/wild/pull/1494#discussion_r2739750556

Previously, integration tests created a separate subdirectory for each test file. However, no further subdirectories were created for each config within those directories. This caused problems when specifying filenames created by directives like `Archive` in other directives.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1695,17 +1695,26 @@ fn build_linker_input(
         .context("At least one object is required")?
         .path;
 
+    let first_source_filename = dep
+        .files
+        .first()
+        .context("At least one file is required")?
+        .filename
+        .as_str();
+    let archive_basename = Path::new(first_source_filename)
+        .file_stem()
+        .context("Invalid source filename")?;
+    let archive_path = config.build_dir.join(archive_basename).with_extension("a");
+
     let mut linker_input = match dep.input_type {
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -465,6 +465,9 @@ fn is_musl_used() -> bool {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 struct Config {
+    /// The base build directory for the test (without config name).
+    base_build_dir: PathBuf,
+    /// The build directory for this config (base_build_dir + config name).
     build_dir: PathBuf,
```
lapla-cogito, line 471: Unifying these two members would mean that, in many places within `integration_tests.rs`, we would need to perform operations equivalent to `.join(&config.name)` whenever `build_dir` is used. For this reason, they are intentionally kept separate as `base_build_dir` and `build_dir`.

davidlattimore: I guess the other option would be to add a method `base_dir` that computes the `PathBuf`. But I don't mind either way.

lapla-cogito:
> Should we get rid of the hashing in this PR or leave that until later?

Thanks, from a separation of responsibilities perspective, I originally planned to remove hashing for archive files in this PR, but I completely forgot about it...

PR #1496, opened 2026-01-29
fix: Align RELRO segment end to page boundary

lapla-cogito:
Since `mprotect(2)` sets memory protection attributes on a page-by-page basis, if the RELRO region exists, its end should be aligned to page boundaries.

This approach, similar to lld/mold, creates a NOBITS section named `.relro_padding` and terminates the RW LOAD segment after it. Since it's a NOBITS section, creating this section does not increase the file size.

### DIFF
--- a/libwild/src/output_section_id.rs
+++ b/libwild/src/output_section_id.rs
@@ -247,23 +247,16 @@ impl<'scope, 'data> OutputOrderBuilder<'scope, 'data> {
 
     /// Returns true if processing the given section will cause the RELRO segment to end.
     fn relro_segment_will_end(&self, section_id: OutputSectionId) -> bool {
-        let relro_def_index = PROGRAM_SEGMENT_DEFS
+        self.active_segment_kinds
             .iter()
-            .position(|def| def.segment_type == pt::GNU_RELRO);
-
-        let Some(def_index) = relro_def_index else {
-            return false;
-        };
-
-        if self.active_segment_kinds[def_index].is_none() {
-            return false;
-        }
-
-        let should_be_in_relro = self
-            .output_sections
-            .should_include_in_segment(section_id, PROGRAM_SEGMENT_DEFS[def_index]);
-
-        !should_be_in_relro
+            .zip(PROGRAM_SEGMENT_DEFS)
+            .any(|(id, def)| {
+                id.is_some()
+                    && def.segment_type == pt::GNU_RELRO
+                    && !self
+                        .output_sections
+                        .should_include_in_segment(section_id, *def)
+            })
     }
 
     /// Ends the currently active RW LOAD segment, if any. This is used when the RELRO segment


davidlattimore (commented):
I had a quick look at the output files produced by GNU ld and LLD.

It looks like LLD emits two adjance RW LOAD segments, one of which is covered entirely by the RELRO segment.

```
Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x001750 0x0000000000003750 0x0000000000003750 0x0002f0 0x0008b0 RW  0x1000
  LOAD           0x001a40 0x0000000000004a40 0x0000000000004a40 0x000048 0x000054 RW  0x1000
  GNU_RELRO      0x001750 0x0000000000003750 0x0000000000003750 0x0002f0 0x0008b0 R   0x1
```

GNU ld emits a single RW LOAD segment, but selects the start address of that load segment such that the RELRO segment ends on a page boundary.

```
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x002cd8 0x0000000000003cd8 0x0000000000003cd8 0x000370 0x000378 RW  0x1000
  GNU_RELRO      0x002cd8 0x0000000000003cd8 0x0000000000003cd8 0x000328 0x000328 R   0x1
```

Mold does the same as LLD.

Wild, with this PR has a single RW LOAD segment, but in order to achieve page boundary alignment of the end of RELRO, padding is inserted.

```
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x001e1c 0x0000000000003e1c 0x0000000000003e1c 0x00123c 0x001245 RW  0x1000
  GNU_RELRO      0x001e1c 0x0000000000003e1c 0x0000000000003e1c 0x000344 0x0011e4 R   0x8
```

This means that the total file-size of Wild's RW segments is 0x123c, compared with  0x338 for LLD, 0x370 for GNU ld and 0x39c for Wild without this fix.

File / segment sizes are from `gcc-dynamic-pie-host/libc-integration.c`.

I suspect LLD's approach here is likely to result in a smaller file. I haven't checked, but, I would assume that GNU ld's approach of selecting the start address of the RW segment would still require padding, it's just that the padding would be before the RW segment instead of in the middle of it.

I'd obviously rather have this fix as-is than not at all, since this seems like an important fix, but if we can do it without significantly increasing the file size for small binaries, that would be even better :)

lapla-cogito:
Both lld and mold appear to align the end of RELRO with page boundaries without increasing file size by inserting a NOBITS section named `.relro_padding`. While I'm not sure whether this approach would work well with Wild’s current architecture, I'd like to consider it further.

```
（▰╹◡╹）❯  readelf --sections -W wild/tests/build/libc-integration.c/gcc-dynamic-pie-host/libc-integration.c.lld | grep relro
  [27] .relro_padding    NOBITS          00000000000039b0 0019b0 000650 00  WA  0   0  1
```

davidlattimore:
Is the NOBITS section actually needed? It feels to me like it might not be, although having it there might make it easier to implement. In fact it's possible that if we added such a section in between RELRO and the regular RW segment, that might automatically cause the additional RW LOAD segment to be created.

lapla-cogito:
> In fact it's possible that if we added such a section in between RELRO and the regular RW segment, that might automatically cause the additional RW LOAD segment to be created.

Yes, this is the recognition I had IIUC, so creating a NOBITS section seemed appropriate.

lapla-cogito:
I thought I had successfully implemented this without using a NOBITS section for padding in [`2425d4b`](https://github.com/davidlattimore/wild/pull/1496/commits/2425d4b1ffcabe7cb7e99d8e5ab9ad365a53187c), but one mold's test uncovered an issue: https://github.com/davidlattimore/wild/actions/runs/21514703547/job/61990069760?pr=1496

The approach in this commit modifies the `p_memsz` value of the LOAD segment corresponding to the RELRO region to match the `p_memsz` value held by RELRO itself. This was necessary because failing to do so would cause the loader to potentially access invalid memory regions when calling `mprotect(2)` at runtime. However, when using `strip(1)` on the generated binary, it apparently recalculates the new LOAD segment `p_memsz` based on the section headers, resulting in the loss of the RELRO region.

As a result, I think creating a NOBITS section remains the most logical approach, so I plan to retain this commit and add a follow-up commit to create the `.relro_padding` section.

libwild/src/output_section_id.rs
```diff
@@ -238,6 +245,41 @@ impl<'scope, 'data> OutputOrderBuilder<'scope, 'data> {
         }
     }
 
+    /// Returns true if processing the given section will cause the RELRO segment to end.
+    fn relro_segment_will_end(&self, section_id: OutputSectionId) -> bool {
+        let relro_def_index = PROGRAM_SEGMENT_DEFS
```
davidlattimore: So this function is checking if there's an active RELRO segment, but the new section ID shouldn't be included in it. I'd probably write that as:

```rust
        self.active_segment_kinds
            .iter()
            .zip(PROGRAM_SEGMENT_DEFS)
            .any(|(id, def)| {
                id.is_some()
                    && def.segment_type == pt::GNU_RELRO
                    && !self
                        .output_sections
                        .should_include_in_segment(section_id, *def)
            })
```

But what you've got here is fine too.

edit: Adding missing not

davidlattimore (approved):
Thanks for fixing this!

davidlattimore (approved):

PR #1497, opened 2026-01-29
chore: Separate test build directories by architecture

lapla-cogito:
follow-up of #1495 

For example, after running `cargo test`, if I then execute `WILD_TEST_CROSS=aarch64 cargo test` without clearing the `wild/tests/build` directory, the integration tests attempt to interpret x86_64 archive files as AArch64 ones, resulting in numerous test failures.
To prevent this, I modified integration tests so that the subdirectory names created during test runs include architecture information (e.g., `wild/tests/build/trivial.c/default-host/trivial.c-default-host.wild`).

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1224,16 +1224,6 @@ fn parse_configs(src_filename: &Path, default_config: &Config) -> Result<Vec<Con
         bail!("Missing non-abstract Config");
     }
 
-    for config in &mut configs {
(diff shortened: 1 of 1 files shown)


lapla-cogito:
Removing "default-host" equivalent information from the filename makes sense, but eliminating the hashing process seems risky. I roughly searched and found that, for example, `preinit-array.c` references `preinit-array.s`—in cases like this, identically named files could be created within the same configuration. While I only encountered one such case, there's certainly potential for more of these tests to emerge in the future.

davidlattimore (approved):

PR #1501, opened 2026-02-01
fix: Use actual segment alignment in `layout_section_parts`

lapla-cogito:
When computing section layouts, `align_modulo` was using only the page alignment instead of the actual maximum alignment of sections within each segment. This caused incorrect `file_offset`/`mem_offset` relationships when segments contained sections with large alignments, resulting in `file_size` > `mem_size` which Linux kernel rejects.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -5979,6 +5979,9 @@ fn layout_section_parts(
     output_order: &OutputOrder,
     args: &Args,
 ) -> OutputSectionPartMap<OutputRecordLayout> {
+    let segment_alignments =
+        compute_segment_alignments(sizes, program_segments, output_order, args);
+
     let mut file_offset = 0;
     let mut mem_offset = output_sections.base_address;
     let mut nonalloc_mem_offsets: OutputSectionMap<u64> =
@@ -5995,7 +5998,10 @@ fn layout_section_parts(
             }
             OrderEvent::SegmentStart(segment_id) => {
                 if program_segments.is_load_segment(segment_id) {
-                    let segment_alignment = program_segments.segment_alignment(segment_id, args);
+                    let segment_alignment = segment_alignments
+                        .get(&segment_id)
+                        .copied()
+                        .unwrap_or_else(|| args.loadable_segment_alignment());
                     if let Some(location) = pending_location.take() {
                         mem_offset = location.address;
                         file_offset =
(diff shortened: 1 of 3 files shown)


libwild/src/layout.rs
```diff
@@ -6003,6 +6003,12 @@ fn layout_section_parts(
                     } else {
                         mem_offset = segment_alignment.align_modulo(file_offset as u64, mem_offset);
                     }
+
+                    // Ensure file_offset >= mem_offset to prevent FileSiz > MemSiz in the segment.
+                    if (file_offset as u64) < mem_offset {
+                        file_offset =
```
davidlattimore: Bumping the file offset will cause padding to be inserted in the file. I don't think file_offset being less than `mem_offset` is inherently a problem. In fact, when we emit a non PIE executable, that's almost always the case. Adding the padding increases the size of the output file for `trivial.c` from 1.7K to 14K.

I suspect the problem is that we used 4K as the `segment_alignment` above.

In `compute_segment_layout`, we compute the alignment of each segment by taking into account the maximum alignment of any of the sections in that segment. But we need to know the actual segment alignment here and `compute_segment_layout` hasn't been called yet. Possibly we need to compute the segment alignments earlier, or maybe some other refactoring is possible. We unfortunately can't just compute the segment layout before  the section layout since the later (at least currently) depends on the former.

libwild/src/layout.rs
```diff
@@ -6003,6 +6003,12 @@ fn layout_section_parts(
                     } else {
                         mem_offset = segment_alignment.align_modulo(file_offset as u64, mem_offset);
                     }
+
+                    // Ensure file_offset >= mem_offset to prevent FileSiz > MemSiz in the segment.
```
davidlattimore: I don't think `FileSiz` and `MemSiz` are standard terms. `readelf` shows those as the column names just because it's abbreviating `file_size` and `mem_size` so that the columns aren't too wide.

davidlattimore:
I just tried running the test with GNU ld. Interestingly, it appears to create a separate LOAD segment for each of the large-alignment sections:

```
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x0006c0 0x0006c0 R   0x1000
  LOAD           0x001000 0x0000000000001000 0x0000000000001000 0x000162 0x000162 R E 0x1000
  LOAD           0x008000 0x0000000000008000 0x0000000000008000 0x00001f 0x00001f R E 0x8000
  LOAD           0x010000 0x0000000000010000 0x0000000000010000 0x00002d 0x00002d R E 0x8000
  LOAD           0x011000 0x0000000000011000 0x0000000000011000 0x0001d4 0x0001d4 R   0x1000
  LOAD           0x011db8 0x0000000000012db8 0x0000000000012db8 0x000258 0x000260 RW  0x1000
```

Which kind of makes sense from the perspective of trying to keep the file size in check. Although segments with such large alignment are presumably very rare, so it's probably not worthwhile trying to optimise that use-case.

davidlattimore:
Someone coincidentally (I think) `large-alignment.sh` has started passing on my system, which means that the negative test is failing. I added support for ignoring specific external tests https://github.com/davidlattimore/wild/pull/1502 to work around this for now. I think the reason that it passes is that the test isn't checking much. When it fails it segfaults, however slightly different linker inputs can result in the segment being page-aligned, which seems to be enough for the test to pass. It might be worthwhile adding our own test for this scenario, since the mold test doesn't seem to reliably detect the problem.

lapla-cogito:
I started addressing this issue because `large-alignment.sh` was incorrectly passing from [`588c7f6`](https://github.com/davidlattimore/wild/commit/588c7f686d570bcf7aa67cbabce1f7aae5300038) on CI (but fails on my machine). I'll be revisiting this problem within the next few days.

lapla-cogito:
I've implemented pre-calculation of maximum alignment requirements for sections within each LOAD segment and incorporated them into `layout_section_parts`. This differs from GNU ld's approach as noted by David above, but since extremely large alignments are rare, this shouldn't be a significant issue.

libwild/src/layout.rs
```diff
@@ -6087,6 +6088,49 @@ fn layout_section_parts(
     records_out
 }
 
+/// Computes the maximum alignment for each LOAD segment by examining the alignments of all sections
+/// that will be placed in that segment.
+fn compute_segment_alignments(
```
davidlattimore, line 6093: Could you add a timing phase to the start of this function so that we know how long it takes?

davidlattimore (approved):

libwild/src/layout.rs
lapla-cogito: Oh sure, done.

PR #1506, opened 2026-02-04
chore: Add some missing deps in a couple of our CI actions

davidlattimore:


davidlattimore:
For additional context, #1505 failed CI ([logs](https://github.com/davidlattimore/wild/actions/runs/21654469746/job/62429066636)) in the loongarch build. I reran the loongarch build check for another PR for which it previously passed and it still passed. I'm guessing perhaps something changed in the base image that meant that we couldn't build host (x86_64) anymore, but we didn't notice because caching meant that we weren't building that stuff. My pending PR, by virtue of updating Cargo.lock required additional stuff to build. That's just a guess. In any case, I was able to reproduce the build failure in a local docker image and determined which things were missing. With this PR submitted, #1505 now passed CI.

PR #1525, opened 2026-02-10
feat: Support `PROVIDE` and `PROVIDE_HIDDEN` linker script directives

### LINKED ISSUE #1097: linker-scripts: Add support for PROVIDE

lapla-cogito:
The `PROVIDE` directive defines a symbol with lower priority than even weak definitions. The symbol is only used if no other definition (strong or weak) exists. `PROVIDE_HIDDEN` additionally prevents the symbol from being exported.

close #1097

### DIFF
--- a/wild/tests/sources/linker-script-provide.c
+++ b/wild/tests/sources/linker-script-provide.c
@@ -22,6 +22,11 @@
 //#DiffIgnore:rel.R_AARCH64_ADR_GOT_PAGE.R_AARCH64_ADR_GOT_PAGE
 //#DiffIgnore:section.riscv.attributes
 //#DiffIgnore:segment.RISCV_ATTRIBUTES.*
+// GNU ld behaves strangely when a symbol referenced in a linker script is
+// empty. See this:
+// https://github.com/davidlattimore/wild/pull/1525#discussion_r2785478582
+//#DiffIgnore:dynsym.__data_start.section
+//#DiffIgnore:dynsym.__data_end.section
 
 extern char provided_absolute __attribute__((weak));
 extern char provided_hidden_absolute __attribute__((weak));
@@ -29,8 +34,6 @@ extern char __text_start __attribute__((weak));
 extern char __text_end __attribute__((weak));
 extern char __data_start __attribute__((weak));
 extern char __data_end __attribute__((weak));
-
-int test_data = 42;
 
 void* get_provided(void) {
   return &provided_absolute + (long)&provided_hidden_absolute;
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3896,14 +3896,20 @@ impl<'data> PreludeLayoutState<'data> {
                 *keep_sections.get_mut(primary_id) |= keep_secondary;
             }
 
-            // Remove any sections without a type except for section 0 (the file header). This
-            // should just be the .phdr and .shdr sections which contain the program headers and
-            // section headers. We need these sections in order to allocate space for those
+            // Remove any built-in sections without a type except for section 0 (the file header).
+            // This should just be the .phdr and .shdr sections which contain the program headers
+            // and section headers. We need these sections in order to allocate space for those
             // structures, but other linkers don't emit section headers for them, so neither should
-            // we.
+            // we. Custom sections (e.g. from linker scripts) that still have NULL type get
+            // PROGBITS assigned instead, since an empty but explicitly defined section should still
+            // be emitted if something references it.
             let section_info = output_sections.section_infos.get(section_id);
             if section_info.ty == sht::NULL && section_id != output_section_id::FILE_HEADER {
-                *keep_sections.get_mut(section_id) = false;
+                if section_id.as_usize() >= output_section_id::NUM_BUILT_IN_SECTIONS {
+                    output_sections.section_infos.get_mut(section_id).ty = sht::PROGBITS;
+                } else {
+                    *keep_sections.get_mut(section_id) = false;
+                }
             }
         }
 
@@ -6636,6 +6642,15 @@ impl<'data> LinkerScriptLayoutState<'data> {
             let symbol_id = self.symbol_id_range.offset_to_id(offset);
             if !resources.symbol_db.is_canonical(symbol_id) {
                 continue;
+            }
+
+            // Mark the section referenced by this symbol so that empty sections
+            // defined by the linker script are still emitted.
+            if let Some(section_id) = def_info.section_id() {
+                resources
+                    .must_keep_sections
+                    .get(section_id)
+                    .fetch_or(true, atomic::Ordering::Relaxed);
             }
 
             // PROVIDE_HIDDEN symbols should not be exported to dynsym.


wild/tests/sources/linker-script-provide.c
```diff
@@ -0,0 +1,45 @@
...
+//#DiffIgnore:segment.RISCV_ATTRIBUTES.*
+
+extern char provided_absolute __attribute__((weak));
+extern char provided_hidden_absolute __attribute__((weak));
+extern char __text_start __attribute__((weak));
+extern char __text_end __attribute__((weak));
+extern char __data_start __attribute__((weak));
+extern char __data_end __attribute__((weak));
+
+int test_data = 42;
```
lapla-cogito: This is the reason I'm keeping this PR in draft. If you comment out this line, the `.data` section becomes empty, and wild will fail to link. This is a clear issue that needs to be addressed.

```
---- integration_test::program_name_055___linker_script_provide_c__ stdout ----
Error: Test failed: `linker-script-provide.c` with linker `wild` config `default`
  Caused by:
    libwild reported error. Rerun command(s):
        cargo run --bin wild -- -dynamic-linker /lib64/ld-linux-x86-64.so.2 --gc-sections -shared -T /home/lapla/repos/wild/wild/tests/sources/linker-script-provide.ld -o /home/lapla/repos/wild/wild/tests/build/linker-script-provide.c/default-host/linker-script-provide.c.wild /home/lapla/repos/wild/wild/tests/build/linker-script-provide.c/default-host/linker-script-provide.aeb0bf64dfeb9435.o --validate-output --write-layout --write-trace --dependency-file=/home/lapla/repos/wild/wild/tests/build/linker-script-provide.c/default-host/linker-script-provide.c.wild.deps
    Failed copying from /home/lapla/repos/wild/wild/tests/sources/linker-script-provide.ld to output file
    symbol '__data_start' in section '`.data`' that we're not going to output Resolution { raw_value: 0, dynamic_symbol_index: Some(5), got_address: None, plt_address: None, flags: ValueFlags(NON_INTERPOSABLE | DIRECT | EXPORT_DYNAMIC | PROVIDE) }
```

After checking how other linkers handle this situation, I found that GNU ld appears to resolve symbols referenced by `PROVIDE` to point to the next section (=`.dynamic` in here) when the `.data` section is empty:
```
（▰╹◡╹）❯  gcc -c wild/tests/sources/linker-script-provide.c -o /tmp/ld-no-data.o && ld -shared -T wild/tests/sources/linker-script-provide.ld /tmp/no-data-test.o -o /tmp/ld-no-data.so && readelf -sSW /tmp/ld-no-data.so|grep -E "data_start|data_end|.dynamic"
/usr/bin/x86_64-linux-gnu-ld.bfd: warning: /tmp/ld-no-data.so has a LOAD segment with RWX permissions
  [ 8] .dynamic          DYNAMIC         0000000000000168 001168 0000e0 10  WA  4   0  8
     2: 0000000000000168     0 NOTYPE  GLOBAL DEFAULT    8 __data_start
     3: 0000000000000168     0 NOTYPE  GLOBAL DEFAULT    8 __data_end
     6: 0000000000000168     0 NOTYPE  GLOBAL DEFAULT    8 __data_end
     7: 0000000000000168     0 NOTYPE  GLOBAL DEFAULT    8 __data_start
```

However, this seems very strange to me - I don't see any positive motivation for doing it this way.

In fact, lld appears to output sections referenced via linker scripts even when they are empty. This approach seems natural to me.
```
（▰╹◡╹）❯  gcc -c wild/tests/sources/linker-script-provide.c -o /tmp/lld-no-data.o && ld.lld -shared -T /tmp/test-location.ld /tmp/lld-no-data.o -o /tmp/no-data-test-lld.so && readelf -sSW /tmp/no-data-test-lld.so | grep -E "data_start|data_end|.data"
  [11] .data             PROGBITS        00000000000004d0 0014d0 000000 00  WA  0   0  1
     7: 00000000000003d6    36 FUNC    GLOBAL DEFAULT    8 get_data_size
     8: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __data_end
     9: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __data_start
    10: 00000000000003fa     0 NOTYPE  GLOBAL DEFAULT    8 __before_data
    11: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __after_data
    10: 00000000000003d6    36 FUNC    GLOBAL DEFAULT    8 get_data_size
    11: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __data_end
    12: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __data_start
    13: 00000000000003fa     0 NOTYPE  GLOBAL DEFAULT    8 __before_data
    14: 00000000000004d0     0 NOTYPE  GLOBAL DEFAULT   11 __after_data
```

libwild/src/value_flags.rs
```diff
@@ -98,6 +98,10 @@ bitflags! {
         /// contains the PLT stub address (for address equality), rather than the IRELATIVE GOT
         /// entry which will be resolved to the actual function address at runtime.
         const IFUNC_GOT_FOR_ADDRESS = 1 << 14;
+
+        /// This symbol is from a PROVIDE directive in a linker script. It should only be used if
+        /// the symbol is referenced but not defined elsewhere.
+        const PROVIDE = 1 << 15;
```
davidlattimore: Our last available bit. I guess if we need another bit we'll need to either expand to 32 bits and incur the associated performance penalty (about a 1.3% slowdown on the zed benchmark) or refactor to free a bit. Probably PROVIDE could be done without using a bit, since it looks like the only place the bit is used is when selecting alternatives.

wild/tests/sources/linker-script-provide.c
davidlattimore: I agree that keeping sections associated with a `PROVIDE` symbol that is referenced makes sense. We have code that does that for regular symbols - https://github.com/davidlattimore/wild/blob/6a7660ba30b24b6214ab7fa301bde38ce52ac137/libwild/src/layout.rs#L3876

davidlattimore (commented):
Great to have this implemented!

libwild/src/value_flags.rs
lapla-cogito: One way to achieve this without adding new bits to `ValueFlags` would be to add a method to `SymbolDb` to determine whether a symbol is a PROVIDE symbol. However, this would mean the method would be called every time `select_symbol()` is invoked, which raises concerns about potential performance impact. However indeed the current implementation consumes the last bit of `ValueFlags`, let me think a bit more.

davidlattimore: Another way would be the collect up a list of all PROVIDE symbols, then before alternative symbols are selected, look up each of the PROVIDEd symbols to see if there is a definition. If there is, do nothing, if there isn't, then add the symbol.

lapla-cogito: I realized that `symbol_strength()` returns `SymbolStrength::Undefined` for linker script symbols, making it possible to implement this much more simply. https://github.com/davidlattimore/wild/pull/1525/commits/e0894b7acc07b1093f9d3344ccfbcb0d1a55b08b

wild/tests/sources/linker-script-provide.c
lapla-cogito: All right, thanks for double-checking: https://github.com/davidlattimore/wild/pull/1525/changes/ada108fc7343adaae7bae71f00ac7ce421f14041

davidlattimore (approved):
Nice!

PR #1526, opened 2026-02-11
feat: Reject conflicting RISC-V ISA extensions

lapla-cogito:
part of #916 

While base ISA conflicts are already caught by `merge_eflags`, there were no checks to detect conflicting combinations of ISA extensions.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -202,7 +202,7 @@ pub fn compute<'data, A: Arch>(
         merge_dynamic_symbol_definitions(&group_states, &symbol_db)?;
     let gnu_property_notes = merge_gnu_property_notes::<A>(&group_states, symbol_db.args.z_isa)?;
     let eflags = merge_eflags::<A>(&group_states)?;
-    let riscv_attributes = merge_riscv_attributes::<A>(&group_states);
+    let riscv_attributes = merge_riscv_attributes::<A>(&group_states)?;
 
     let finalise_sizes_resources = FinaliseSizesResources {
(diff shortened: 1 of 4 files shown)


libwild/src/layout.rs
```diff
@@ -816,9 +817,40 @@ fn merge_riscv_attributes<A: Arch>(group_states: &[GroupState]) -> RiscVAttribut
...
+}
+
+/// Conflicting pairs of RISC-V ISA extensions.
+const RISCV_CONFLICTING_EXT_PAIRS: &[(&str, &str)] = &[
+    ("f", "zfinx"),
+    ("d", "zdinx"),
+    ("q", "zqinx"),
+    ("zfh", "zhinx"),
+    ("zfhmin", "zhinxmin"),
+];
```
lapla-cogito, line 833: For record-keeping, let me explain why I'm listing these pairs: [Zfinx](https://github.com/riscvarchive/riscv-zfinx/blob/main/Zfinx_spec.adoc) specifies the following:
> Every current and future extension that uses `F` registers will also have an `ZFinx` version that is incompatible with the extension.

According to this, enabling any of these pairs simultaneously in the same implementation would violate the intent of the specification.

lapla-cogito: Additionally, this list is somewhat excessive. If we were strictly limited to rv64gc, the minimum necessary would be just `(f, zfinx)` and `(d, zdinx)`. However, rather than properly updating this list in the future, I considered it more meaningful to proactively exclude other pairs here.

Issue #874, opened 2025-06-11
RISC-V: support basic set of relaxations


marxin:
Similarly to other platforms, we should implement the basic set of relaxations, mostly related to the TLS variables.

PR #1527, opened 2026-02-11
feat: Add base to track byte reduction due to relaxation

lapla-cogito:
part of #874 

Some relaxations in architectures such as RISC-V and LoongArch actually reduce the generated code size. Although similar relaxations exist in x86_64 and AArch64, those implementations pad the shortened instructions with NOPs, so the overall symbol sizes remain unchanged. In contrast, the relaxations mentioned above truly shrink the symbols themselves, which means we must track how much each symbol’s size is reduced during the layout phase. A naive implementation would require running the layout process twice (the second pass accounting for size reductions in other symbols) which would clearly hurt performance.

Therefore, while this PR does not introduce any specific relaxation, it lays the groundwork for tracking size reductions caused by relaxations without requiring a second layout pass.

`SectionRelaxDeltas` maintains a list of offsets from which each section should remove a specific number of bytes. This information is used during the write phase to calculate offsets individually. The `relaxation_deleted` field within `Section` records the amount of byte reduction achieved for each section. From this data, we can calculate the actual size of each section, which can be used for the layout phase.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -102,6 +102,8 @@ use linker_utils::elf::sht;
 use linker_utils::elf::stt;
 use linker_utils::loongarch64::highest_relocation_with_bias;
 use linker_utils::relaxation::RelocationModifier;
+use linker_utils::relaxation::SectionRelaxDeltas;
+use linker_utils::relaxation::opt_input_to_output;
 use object::LittleEndian;
 use object::SymbolIndex;
 use object::elf::NT_GNU_BUILD_ID;
@@ -1103,7 +1105,7 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
         name: &[u8],
         output_section_id: OutputSectionId,
         value: u64,
-    ) -> Result {
+    ) -> Result<&mut SymtabEntry> {
         let shndx = self
             .output_sections
             .output_index_of_section(output_section_id)
@@ -1125,14 +1127,14 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
         name: &[u8],
         shndx: u16,
         value: u64,
-    ) -> Result {
+    ) -> Result<&mut SymtabEntry> {
         let e = LittleEndian;
         let is_local = sym.is_local();
         let size = sym.st_size(e);
         let entry = self.define_symbol(is_local, shndx, value, size, name)?;
         entry.st_info = sym.st_info();
         entry.st_other = sym.st_other();
-        Ok(())
+        Ok(entry)
     }
 
     fn copy_absolute_symbol(&mut self, sym: &crate::elf::Symbol, name: &[u8]) -> Result {
@@ -1458,14 +1460,49 @@ fn write_section_raw<'out>(
             );
         }
         let out = section_buffer.split_off_mut(..allocation_size).unwrap();
-        // Cut off any padding so that our output buffer is the size of our input buffer.
         let object_section = object.object.section(sec.index)?;
-        let section_size = object.object.section_size(object_section)?;
-        let (out, padding) = out.split_at_mut(section_size as usize);
-        object.object.copy_section_data(object_section, out)?;
-        // Fill padding. This is especially important if we're writing the file in place.
-        padding.fill(0);
-        Ok(out)
+        let relax_deltas = object.section_relax_deltas.get(sec.index.0);
+
+        match relax_deltas {
+            None => {
+                let section_size = object.object.section_size(object_section)?;
+                let (out, padding) = out.split_at_mut(section_size as usize);
+                object.object.copy_section_data(object_section, out)?;
+                padding.fill(0);
+                Ok(out)
+            }
+            Some(deltas) => {
+                let input_data = object.object.raw_section_data(object_section)?;
+                let effective_size = sec.size as usize;
+
+                let mut input_pos: usize = 0;
+                let mut output_pos: usize = 0;
+
+                for delta in deltas.deltas() {
+                    let skip_start = delta.input_offset as usize;
+                    // Copy everything from input_pos up to the deletion point.
+                    let copy_len = skip_start - input_pos;
+                    if copy_len > 0 {
+                        out[output_pos..output_pos + copy_len]
+                            .copy_from_slice(&input_data[input_pos..skip_start]);
+                        output_pos += copy_len;
+                    }
+                    // Skip over the deleted bytes in the input.
+                    input_pos = skip_start + delta.bytes_deleted as usize;
+                }
+
+                // Copy the remainder after the last deletion.
+                let remaining = input_data.len() - input_pos;
+                if remaining > 0 {
+                    out[output_pos..output_pos + remaining]
+                        .copy_from_slice(&input_data[input_pos..]);
+                    output_pos += remaining;
+                }
+                out[output_pos..].fill(0);
+
+                Ok(&mut out[..effective_size])
+            }
+        }
     } else {
         Ok(&mut [])
     }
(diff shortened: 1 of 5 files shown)


libwild/src/layout.rs
```diff
@@ -1765,10 +1768,12 @@ pub(crate) struct DynamicSymbolDefinition<'data> {
 pub(crate) struct Section {
     pub(crate) index: object::SectionIndex,
     pub(crate) part_id: PartId,
-    /// Size in memory.
+    /// Size in memory (original input section size, before relaxation).
     pub(crate) size: u64,
     pub(crate) flags: ValueFlags,
     pub(crate) is_writable: bool,
+    /// Total bytes deleted by relaxation. 0 when no relaxation applies.
+    pub(crate) relaxation_deleted: u64,
```
lapla-cogito: One might consider merging this and `ObjectLayout::section_relax_deltas` by creating a `Section` member with type `Option<Box<linker_utils::relaxation::SectionRelaxDeltas>>`. However, [the requirement to implement `Copy` for `Section`](https://github.com/davidlattimore/wild/blob/326bb2a7a6172c1e291cc56804d0634f2fd58d2a/libwild/src/resolution.rs#L1362-L1370) forces us to maintain this separation. `Section::relaxation_deleted` tracks the total bytes reduced, while `ObjectLayout::section_relax_deltas` holds per-offset delta values.

lapla-cogito:
I don't think the CI failure is caused by this change: #1528

lapla-cogito:
Note: I tried something similar before in #1317, but it didn’t work out. This PR is another attempt (and it took some time to implement, as you can see from the gap). Based on this, I'm currently working on implementing call relaxation for RISC-V locally, and it seems to be working quite well so far.

libwild/src/layout.rs
davidlattimore: I benchmarked this change and unfortunately it causes a slowdown of about 2% (linking zed). Most of the slowdown seems to be due to the addition of this field.

Here's a flamegraph showing where the performance differences show up:

Image

Looks like the extra hashmap lookups are also contributing. At least I'm guessing that's what the blake3 stuff is.

One possibility is to just store a boolean here to indicate that deletions have occurred, then store the amount in the hashmap. You can then skip the hashmap lookup in the boolean is set to false. That would mean that getting the section size when there are deletions would require a hashmap lookup where currently it doesn't, but would optimise for the case where there aren't any deletions, which I think is, at least currently, the more important case.

Another option would be to reduce the size of this field to be less than a u64. That by itself wouldn't be sufficient to reduce the struct size and so we'd likely need other changes to have an effect. e.g. the `object::SectionIndex` is a usize and probably doesn't need to be. With this option, the hashmap lookup in the write phase can probably be avoided by just checking if this field is zero.

Yet another option would be to make `Section` be generic over `Arch`, allowing the field to have size zero on architectures that don't support this kind of relaxation.

lapla-cogito: Thanks. By storing `section_relax_deltas` as a `Vec<Option<SectionRelaxDeltas>>` instead of a `HashMap`, we can use the `Option` type to indicate whether deletions occurred during relaxation, allowing us to remove the `relaxation_deleted` field.

davidlattimore (commented):
Unfortunately the performance loss is larger:

```
Benchmark 1 (671 runs): /home/david/save/zed/run-with env-rand /home/d/wild-builds/2026-02-13.cg1 --no-fork
  wall_time           537ms ± 13.2ms     494ms …  565ms          1 ( 0%)        0%
  peak_rss           3.47GB ± 1.21MB    3.47GB … 3.47GB          4 ( 1%)        0%
  cpu_cycles         21.7G  ±  182M     21.0G  … 22.4G           8 ( 1%)        0%
  instructions       19.4G  ± 51.6M     19.3G  … 19.6G          18 ( 3%)        0%
  cache_references    473M  ± 2.89M      466M  …  485M          14 ( 2%)        0%
  cache_misses        124M  ±  627K      123M  …  126M           2 ( 0%)        0%
  branch_misses      38.3M  ±  183K     37.1M  … 38.8M          23 ( 3%)        0%
Benchmark 2 (639 runs): /home/david/save/zed/run-with env-rand /home/d/wild-builds/2026-02-13.cg1.relaxation-tracking --no-fork
  wall_time           564ms ± 14.6ms     516ms …  592ms          1 ( 0%)        💩+  5.0% ±  0.3%
  peak_rss           3.59GB ± 1.06MB    3.59GB … 3.59GB          8 ( 1%)        💩+  3.5% ±  0.0%
  cpu_cycles         22.2G  ±  172M     21.5G  … 22.8G          11 ( 2%)        💩+  2.1% ±  0.1%
  instructions       19.6G  ± 48.9M     19.5G  … 19.9G           7 ( 1%)        💩+  1.3% ±  0.0%
  cache_references    496M  ± 2.83M      489M  …  507M           8 ( 1%)        💩+  4.8% ±  0.1%
  cache_misses        128M  ±  629K      126M  …  130M           4 ( 1%)        💩+  2.7% ±  0.1%
  branch_misses      39.5M  ±  172K     38.6M  … 40.1M          16 ( 3%)        💩+  3.2% ±  0.1%
```

Memory consumption is also up. I assume this is because we're now storing an `Option<Vec<..>>` for every input section.

lapla-cogito:
Implemented sparse maps for each object file to track byte reduction. While this is clearly more memory-efficient than using `Vec` per section, it also means each object file will incur an additional 24 bytes of struct size, even for architectures where byte reduction tracking isn't necessary. But I think this trade-off is acceptable.

libwild/src/elf_writer.rs
```diff
@@ -1770,7 +1832,13 @@ fn write_eh_frame_relocations<A: Arch>(
                                 })?;
 
                             if let Some(hdr_out) = table_writer.take_eh_frame_hdr_entry() {
-                                let frame_ptr = (section_address + offset_in_section) as i64
+                                // When relaxation has deleted bytes fromq the target section, the
```
davidlattimore: typo - 'fromq'

linker-utils/src/relaxation.rs
```diff
@@ -3,3 +3,201 @@ pub enum RelocationModifier {
...
+    #[must_use]
+    pub fn output_to_input_offset(&self, output_offset: u64) -> u64 {
+        if self.deltas.is_empty() {
+            return output_offset;
+        }
+
+        let n = self.deltas.len();
+        let mut lo = 0usize;
+        let mut hi = n;
+        while lo < hi {
```
davidlattimore: Would it work to use `binary_search_by_key` here? If not, could you add a comment to the code saying why.

davidlattimore:
Performance looks good now. Thanks :)

linker-utils/src/relaxation.rs
lapla-cogito: Since `output_pos` is strictly monotonically increasing because the input offsets are strictly ascending and the deletion ranges don't overlap, we didn't actually need to use a binary search here in the first place (the remnants of the first approach's implementation were still present).

marxin:
> Therefore, while this PR does not introduce any specific relaxation, it lays the groundwork for tracking size reductions caused by relaxations without requiring a second layout pass.

Just wanted to mention we can reduce the binary size, but we must obey the align relocations, e.g. `R_RISCV_ALIGN`.
@lapla-cogito can you please create a PR that will respect the alignment in the tracking?

lapla-cogito:
Since I've merged #1552, I'll begin addressing `R_RISCV_ALIGN`-related things soon. However, the current code still outputs alignment NOPs specified by `R_RISCV_ALIGN` at their original size, which should not cause alignment issues.

Issue #1529, opened 2026-02-11
Use PROVIDE semantics for built-in symbols as appropriate


davidlattimore:
https://github.com/davidlattimore/wild/pull/1525 added support for `PROVIDE` in linker scripts. Some linker-defined symbols should be using PROVIDE semantics, but aren't. e.g "_end". There may be some code changes required to allow linker-defined symbols to use PROVIDE semantics rather than just linker scripts. Which symbols use PROVIDE semantics can be checked by looking at GNU ld's default linker script (`ld --verbose`).

PritamP20:
Hi @davidlattimore  I would like to work on this issue

From what I understand, Wild currently defines symbols like _end unconditionally, but they should follow PROVIDE semantics (only define them if the user hasn’t already). My plan is to check GNU ld’s default script to see which symbols require PROVIDE behavior and then update Wild to make those linker-defined symbols conditional

Does this approach sound correct or is there anything specific you’d recommend I look at before starting?

davidlattimore:
That's right, although the main work is to make the changes to support having the linker-provided symbols use PROVIDE semantics.

PR #1534, opened 2026-02-13
Implement PROVIDE semantics for linker-defined symbols

PritamP20:
This PR fixes **#1529**, which reported that some linker-defined symbols (like `_end`, `__init_array_start`, etc.) were always being defined by Wild  even when the user had already defined them.

`#1525` previously added support for `PROVIDE` in linker scripts, but the linker’s built-in symbols weren’t using it. This update ensures those symbols now follow `PROVIDE` semantics  meaning they’re only defined if the user hasn’t already defined them. This brings Wild’s behavior in line with **GNU ld**.

---

#### **Changes**

1. **Metadata Infrastructure (`libwild/src/output_section_id.rs`)**

   * Added flags to mark which built-in symbols should use `PROVIDE` semantics.
   * Updated `DEFAULT_DEFS` to mark:

     * `_end` (BSS section)
     * `__init_array_start` / `__init_array_end`
     * `__fini_array_start` / `__fini_array_end`
     * `__preinit_array_start` / `__preinit_array_end`

2. **Symbol Definition Tracking (`libwild/src/parsing.rs`)**

   * Added `use_provide: bool` to `InternalSymDefInfo` for better control over how symbols are defined.
   * Updated constructors and the `Prelude` initialization to respect the section metadata.

3. **Linker Script Handling (`libwild/src/layout_rules.rs`)**

   * Updated `PROVIDE` command parsing to set `use_provide: true`.
   * Regular symbol assignments continue using `use_provide: false`.

4. **Implementation Logic (`libwild/src/layout.rs`)**

   * Documented how the existing `is_canonical()` check naturally enforces `PROVIDE` behavior:

     * User-defined symbols take precedence.
     * Linker-defined symbols are skipped when a user definition exists.

---

#### **Testing**

*  **Unit Tests:** All 48 tests in `libwild` passed.
* **Integration Tests:** All 108 tests in `integration_tests` passed.
*  **New Test:** Added `wild/tests/sources/linker-defined-provide.c`

  * Verifies user-defined `_end` overrides linker’s version.
  * Confirms linker-defined symbols still exist when not overridden.
* **Manual Check:** Linked a C program with a custom `_end`  no duplicate symbol errors, correct symbol used.

---

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -3725,6 +3725,7 @@ fn write_internal_dynsym(
     Ok(())
 }
 
+/// Writes a dynsym entry for a symbol defined via --defsym or linker script symbol assignment.
 fn write_defsym_dynsym(
     dynsym_writer: &mut SymbolTableWriter,
     layout: &ElfLayout,
--- a/libwild/src/parsing.rs
+++ b/libwild/src/parsing.rs
@@ -230,6 +230,8 @@ impl<'data> Prelude<'data> {
                 continue;
             }
             // .rela.plt start/stop symbols are only emitted for non-relocatable executables.
+            // Emitting them for relocatable binaries causes glibc to try to call the resolver
+            // functions without taking into account that the binary has been relocated.
             if output_kind == OutputKind::StaticExecutable(RelocationModel::Relocatable)
                 && section_id == output_section_id::RELA_PLT
             {


wild/tests/sources/linker-defined-provide.c
```diff
@@ -0,0 +1,11 @@
+// User defines _end - linker should not create a conflicting symbol
+char _end = 42;
+
+// These are linker-defined symbols that should still be created
+extern char __init_array_start __attribute__((weak));
+extern char __init_array_end __attribute__((weak));
+
+int main(void) {
```
davidlattimore: Tests that use `main` need to link against libc, which makes linking a lot slower, so usually we only use main (and libc) if the test actually needs it, which I don't think this test does. Could you change this test to use `_start`. See `trivial.c` for an example.

wild/tests/sources/linker-defined-provide.c
davidlattimore, line 1: New tests need to be added to the list in `integration_tests.rs`, otherwise they won't get run.

wild/tests/sources/linker-defined-provide.c
```diff
@@ -0,0 +1,11 @@
+// User defines _end - linker should not create a conflicting symbol
+char _end = 42;
```
davidlattimore: I'd be tempted to make this symbol weak, since even a weak symbol should take priority over a PROVIDE symbol.

wild/tests/sources/linker-defined-provide.c
PritamP20: Sure! I will make the change

libwild/src/output_section_id.rs
lapla-cogito, line 565: Shouldn't `__rela_iplt_start` and `__rela_iplt_end` also follow `PROVIDE` semantics (more precisely `PROVIDE_HIDDEN` - see my other review)? Actually, on my system, I get the following results with GNU ld's default linker script:
```
（▰╹◡╹）❯  ld --verbose|rg PROVIDE
  PROVIDE (__executable_start = SEGMENT_START("text-segment", 0x400000));
      PROVIDE_HIDDEN (__rela_iplt_start = .);
      PROVIDE_HIDDEN (__rela_iplt_end = .);
  PROVIDE (__etext = .);
  PROVIDE (_etext = .);
  PROVIDE (etext = .);
     PROVIDE_HIDDEN (__tdata_start = .);
    PROVIDE_HIDDEN (__preinit_array_start = .);
    PROVIDE_HIDDEN (__preinit_array_end = .);
    PROVIDE_HIDDEN (__init_array_start = .);
    PROVIDE_HIDDEN (__init_array_end = .);
    PROVIDE_HIDDEN (__fini_array_start = .);
    PROVIDE_HIDDEN (__fini_array_end = .);
  PROVIDE (edata = .);
  PROVIDE (end = .);
```

libwild/src/parsing.rs
```diff
@@ -305,12 +324,13 @@ impl<'data> Prelude<'data> {
...
                 }
             };
-            InternalSymDefInfo::notype(placement, name.as_bytes())
+            InternalSymDefInfo::notype(placement, name.as_bytes(), false)
         }));
 
         symbol_definitions.push(InternalSymDefInfo::notype(
             SymbolPlacement::LoadBaseAddress,
             b"__executable_start",
+            false,
```
lapla-cogito: Same as above. `__executable_start` should also follow `PROVIDE` semantics.

libwild/src/output_section_id.rs
```diff
@@ -464,6 +464,9 @@ pub(crate) struct BuiltInSectionDetails {
     pub(crate) start_symbol_name: Option<&'static str>,
     pub(crate) end_symbol_name: Option<&'static str>,
     pub(crate) group_end_symbol_name: Option<&'static str>,
+    pub(crate) start_symbol_use_provide: bool,
+    pub(crate) end_symbol_use_provide: bool,
+    pub(crate) group_end_symbol_use_provide: bool,
```
lapla-cogito: This appears to only support the `PROVIDE` semantics. The default linker script in GNU ld contains several symbols defined with `PROVIDE_HIDDEN`. For details on the differences between them, see the description in #1525.

wild/tests/sources/linker-defined-provide.c
PritamP20: I will add a week symbol for the test

PritamP20: Yup the test is passing with the the week symbol also

Screenshot 2026-02-15 at 1 57 49 AM

libwild/src/output_section_id.rs
PritamP20: I’m running into an issue where __rela_iplt_end still gets exported to .dynsym even after switching it to ProvideHidden. From the logs, PreludeLayoutState detects is_hidden = true and skips the explicit export, but define_symbol is still called later during ELF generation, bypassing export_dynamic checks. Do you know if there’s any implicit path in elf_writer.rs or layout.rs that could be promoting __rela_iplt_* symbols to dynamic despite the hidden flag?

davidlattimore: It's probably easiest for us to check by looking at what's happening with your PR. Could you upload a new version of it with your latest changes? If you're not ready for it to be submitted, just mark it as a draft in the Github UI.

lapla-cogito: Since `ProvideHidden` doesn't currently exist in this branch, I think we should discuss [this commit](https://github.com/PritamP20/wild/commit/65a86654808ea07b0c143871550306bd0ad3c03d).

- This commit includes changes to make `write_prelude_defsym` handle non-defsym symbols as well, but it seems the `is_hidden` check isn't being performed here.
- While this isn't the direct cause of this issue, this commit appears to simultaneously skip checking for `RELA_PLT` in the existing `parsing.rs`. This implies that the change was previously missing the definition of `__rela.iplt_*` in dynamic executables, and now it is being added. When any object file references this symbol, it could potentially set the `EXPORT_DYNAMIC` flag, add it to `dynamic_symbol_definitions`, and write it to `.dynsym`.

BTW, looking at the committed files, are you working on macOS? I was under the impression that developing Wild on macOS has some challenges - is that not the case? 🤔

PritamP20: Yeah, I’m on macOS,  it’s been fine so far since I run everything in Docker for Linux builds and test them. I’ll push the updated draft PR so we can check where the hidden flag’s getting overridden.

lapla-cogito: Ah, I see - since `.DS_Store` is being committed (which shouldn’t be), I assumed you were working entirely on macOS :)

lapla-cogito:
I added some special symbols in #1563. Therefore, it would be good if this PR could include changes to apply PROVIDE semantics to these symbols as well.

PritamP20:
I’ve updated the PR to implement PROVIDE and PROVIDE_HIDDEN for all built-in symbols, correctly fixing the .dynsym leakage issue. I also refactored the tests to use _start and weak symbols as requested. Verified passing in a Linux Docker environment.


Screenshot 2026-02-23 at 10 38 20 PM

libwild/src/parsing.rs
```diff
@@ -248,34 +265,61 @@ impl<'data> Prelude<'data> {
...
-            // Emitting them for relocatable binaries causes glibc to try to call the resolver
-            // functions without taking into account that the binary has been relocated.
-            if output_kind != OutputKind::StaticExecutable(RelocationModel::NonRelocatable)
-                && section_id == output_section_id::RELA_PLT
-            {
-                continue;
+
+            // ... (imports)
+
+            // ... inside Prelude::new ...
```
mati865: When pasting code from LLM you should at least review it first.

benchmarks/images/.DS_Store
lapla-cogito, line 1: No `.DS_Store` files should be committed.

libwild/src/elf_writer.rs
```diff
@@ -1155,6 +1155,7 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
         size: u64,
         name: &[u8],
     ) -> Result<&mut SymtabEntry> {
+        let entry_size = size_of::<SymtabEntry>();
```
lapla-cogito: This does not seem to be used anywhere.

libwild/src/parsing.rs
```diff
@@ -72,6 +73,9 @@ pub(crate) struct InternalSymDefInfo<'data> {
     pub(crate) elf_symbol_type: SymbolType,
     /// If true, this symbol should have hidden visibility (from PROVIDE_HIDDEN).
     pub(crate) is_hidden: bool,
+    /// If true, this symbol should only be defined if not already defined (PROVIDE semantics).
+    #[allow(dead_code)]
```
lapla-cogito: It would seem sufficient to simply remove `use_provide`.

wild/rust-integration-dynamic.8zwq9oiug1vyzdlccs3ch0h6m.rcgu.o
lapla-cogito, line 1: No binary files produced by tests should be committed.

lapla-cogito (changes requested):
Some preliminary comments. Also, please resolve the conflicts.

benchmarks/images/.DS_Store
davidlattimore: It is however fine to add to the .gitignore if that helps.

.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
     ],
     "search.exclude": {
         "/external_test_suites/": true
-    }
+    },
+    "kiroAgent.configureMCP": "Disabled"
```
davidlattimore: I'm not sure what this does. Perhaps revert it? Possibly in a separate PR I should rename "settings.json" to "settings.json.example" so that people can copy it then make their own customisations.

wild/tests/integration_tests.rs
```diff
@@ -3329,22 +3295,23 @@ impl PartialEq for ErrorMatcher {
...
-        Linker::ThirdParty(ThirdPartyLinker {
+    let mut linkers = vec![Linker::ThirdParty(ThirdPartyLinker {
+        name: "ld",
+        gcc_name: "bfd",
+        path: find_bin(&["ld.bfd", "ld"])?,
+        cross_paths: find_cross_paths("ld"),
+        enabled_by_default: true,
+    })];
+
+    if let Ok(path) = find_bin(&["ld.lld"]) {
```
davidlattimore, line 3355: Note that if lld is disabled, then many of the tests will fail if diffing is enabled.

libwild/src/parsing.rs
PritamP20: Sure, noted. I did go through and verify everything before pushing, but I’ll be extra careful next time. Thanks for pointing it out

benchmarks/images/.DS_Store
PritamP20: Yeah, that slipped in by mistake
I’ll remove it and add .DS_Store to .gitignore so it doesn’t happen again

libwild/src/elf_writer.rs
PritamP20: Yup, that was left over from an earlier version. I’ll clean it up in the next commit

wild/rust-integration-dynamic.8zwq9oiug1vyzdlccs3ch0h6m.rcgu.o
PritamP20: Got it,  I’ll remove the binary and will make sure those outputs are excluded

.vscode/settings.json
PritamP20: Yeah, I’ll revert it

wild/tests/integration_tests.rs
PritamP20: @davidlattimore I will tested it, in both the ways and update you

libwild/src/parsing.rs
PritamP20: tnx for pointing it out, i just checked is_canonical is handling the checks, so i will remove it and update the changes

PritamP20:
I have updated and cleaned things up and fixed the symtab issue
- Dropped use_provide since is_canonical() already handles the logic.
- Fixed the null symtab entry mismatch by marking it hidden.
- Removed .DS_Store, *.rcgu.o, the .vscode/setting.json
- Ran clang-format on the test files.

Added DiffIgnore for .data/TLS and turned off diffing for the shared object variants verified the test passes with lld diffing

wild/tests/sources/rela_iplt_def.c
davidlattimore, line 1: What are these two files? They don't seem to be used

libwild/src/elf_writer.rs
```diff
@@ -3509,7 +3509,35 @@ fn write_prelude_dynsym(
         .get(offset)
         .with_context(|| format!("Invalid prelude symbol {}", layout.symbol_debug(symbol_id)))?;
 
-    write_defsym_dynsym(dynsym_writer, layout, symbol_id, def_info)
+    if matches!(
```
davidlattimore, line 3696: It looks like you've copied and pasted a substantial chunk of code from one of the other functions in this file. Could you refactor so that they instead share code?

.vscode/settings.json
```diff
@@ -4,5 +4,5 @@
     ],
     "search.exclude": {
         "/external_test_suites/": true
-    }
+    },
```
lapla-cogito: This change seems unnecessary.

wild/.DS_Store
lapla-cogito, line 1: still

.gitignore
```diff
@@ -22,3 +22,5 @@ fakes-debug/sde-cet-checker-out.txt
 ##
 *.pftrace
 *.bench-results
+.DS_Store
+*.rcgu.o
```
lapla-cogito: Could you add a line break at the end? That would prevent unnecessary diffs when adding new entries to `.gitignore` in the future.

libwild/src/layout.rs
```diff
@@ -3578,6 +3624,37 @@ impl<'data> PreludeLayoutState<'data> {
 
         self.load_entry_point::<A>(resources, queue, scope);
 
+        for (offset, def_info) in self.internal_symbols.symbol_definitions.iter().enumerate() {
```
davidlattimore: This whole loop looks to be copied from elsewhere too. Can you refactor this one as well?

.gitignore
mati865: Also `.rcgu.o` is a temporary rustc object. It shouldn't be there.

libwild/src/elf_writer.rs
PritamP20: The reason for that if matches! block was that write_defsym_dynsym had a debug_assert! that only allowed defsym placements.

When I added PROVIDE semantics for symbols like __init_array_start and __init_array_end, those became internal symbols with SectionStart or SectionEnd placements. Since they weren’t regular defsyms, calling the original function would’ve hit the assert, so I just handled them manually at that point

wild/tests/sources/rela_iplt_def.c
PritamP20: Those were just scratchpad files I used to manually verify the __rela_iplt symbol behavior during development. Once that logic got integrated into the linker-defined-provide.c test variants, I’ll remove them

.gitignore
PritamP20: Got it , I’ll add the newline at the end

The .rcgu.o binaries were added because they were generated locally during testing and accidentally got included in the previous commit

libwild/src/elf_writer.rs
davidlattimore: I'm not sure I understand. The code between this line and the end of the function is more or less identical to code in `write_linker_script_dynsym`. Is there a reason it can't be deduplicated?

libwild/src/layout.rs
PritamP20: Yeah i will refactor it

davidlattimore: Probably best to let us know after you've made and pushed the change. Of course if you disagree with something or have further questions, then replying without first pushing is fine

PritamP20:
Made all the requested changes:

Removed unused entry_size variable
- Extracted write_internal_dynsym helper to deduplicate write_prelude_dynsym and write_linker_script_dynsym
- Extracted InternalSymbols::activate_symbols to remove the duplicated loop in layout.rs
- Deleted the scratchpad test files and .DS_Store files

Also fixed two bugs I caught while testing on Ubuntu aarch64:
- The loop in PreludeLayoutState::activate was incorrectly bulk-exporting all non-hidden prelude symbols (_end, etext, etc.) into dynsym  removed it since GNU ld only adds these when explicitly referenced
- The RELA_PLT skip for static PIE was missing, which caused __rela_iplt_start/end to be non-zero  glibc uses zero as a signal that there are no IPLT entries to process, so this broke libc-integration

Tests pass on aarch64 Ubuntu (92 passed, 17 pre-existing failures unrelated to this PR)

libwild/src/layout.rs
```diff
@@ -5616,37 +5652,7 @@ impl<'data> LinkerScriptLayoutState<'data> {
         common: &mut CommonGroupState<'data, O>,
         resources: &GraphResources<'data, '_, O>,
     ) -> Result {
-        for (offset, def_info) in self.internal_symbols.symbol_definitions.iter().enumerate() {
-            let symbol_id = self.symbol_id_range.offset_to_id(offset);
-            if !resources.symbol_db.is_canonical(symbol_id) {
-                continue;
-            }
-
-            // Mark the section referenced by this symbol so that empty sections
```
davidlattimore, line 5625: This code was moved, but it looks like the comment got lost. Unless there's a reason for the comment being removed, can you restore it? Same for the comment below.

libwild/src/parsing.rs
```diff
@@ -228,36 +231,53 @@ impl<'data> Prelude<'data> {
             {
                 continue;
             }
-
-            let def = section_id.built_in_details();
-            // .rela.plt start/stop symbols are only emitted for non-relocatable executables.
```
davidlattimore, line 233: This comment seems to have been lost as well.

.vscode/settings.json
davidlattimore: Looks like you missed this

libwild/src/layout.rs
PritamP20: Sorry didn't notice it, will restore it

PritamP20:
Both of these bugs were actually introduced in this PR the dynsym loop was added to PreludeLayoutState::activate, and the RELA_PLT skip was removed as part of these changes. So addressing them here seems appropriate since they’re regressions from this PR itself

Regarding the 12 test failures, I’ll review `ci.yml` to check if i am missing any packages or versions of dependencies

davidlattimore:
Ah, nice. If they're bugs introduced in this PR, then fixing them in this PR is definitely the right thing to do :)

I haven't looked at the CI failures, but let me know if you need help to interpret what's going on.

.gitignore
mati865: Doing such tests in root project directory isn't a good idea. You should use some temporary directory instead.

davidlattimore:
Getting close. Looks like it's just clippy and rustfmt. You can fix the formatting by running `cargo +nightly fmt`. You should be able to see the clippy warnings by running `cargo clippy`.

PritamP20:
Yeah learning a lot, will fix the fmt's and commit after testing everything again once

PritamP20:
All the test cases are running, if there are any other changes required please let me know

libwild/src/parsing.rs
davidlattimore: It looks like you restored one line of the comment, but the other two lines are still removed

libwild/src/elf_writer.rs
```diff
@@ -3696,11 +3684,47 @@ fn write_prelude_dynsym(
...
+    let address = resolution.address()?;
+    let name = layout.symbol_db.symbol_name(symbol_id)?;
+
+    let entry = dynsym_writer.define_symbol(false, shndx, address, 0, name.bytes())?;
+    entry.set_st_info(object::elf::STB_GLOBAL, object::elf::STT_NOTYPE);
+
+    Ok(())
 }
 
-/// Writes a dynsym entry for a symbol defined via --defsym or linker script symbol assignment.
```
davidlattimore, line 3703: Can you restore this comment too unless there's a reason to remove it?

libwild/src/parsing.rs
```diff
@@ -295,11 +314,33 @@ impl<'data> Prelude<'data> {
...
+        symbol_definitions.push(InternalSymDefInfo::hidden(
             SymbolPlacement::LoadBaseAddress,
             b"__executable_start",
         ));
 
+        // etext aliases
+        symbol_definitions.push(InternalSymDefInfo::notype(
+            SymbolPlacement::SectionEnd(output_section_id::RODATA),
+            b"etext",
+        ));
```
lapla-cogito, line 328: `etext` shouldn't be placed at the end of RODATA, but rather at the end of TEXT, should it not?

libwild/src/parsing.rs
```diff
@@ -228,36 +229,54 @@ impl<'data> Prelude<'data> {
...
+                defs: &mut Vec<InternalSymDefInfo<'data>>,
+                mode: DefinitionMode,
+                placement: SymbolPlacement<'data>,
+                name: &'static str,
+            ) {
+                let is_hidden = matches!(mode, DefinitionMode::ProvideHidden);
+                if is_hidden {
+                    defs.push(InternalSymDefInfo::hidden(placement, name.as_bytes()));
+                } else {
+                    defs.push(InternalSymDefInfo::notype(placement, name.as_bytes()));
```
lapla-cogito, line 253: This seems to imply that `DefinitionMode::Provide` and `DefinitionMode::Strong` aren't distinguished.

lapla-cogito (commented):
Thanks! I've added some comments.

libwild/src/parsing.rs
PritamP20: I initially placed etext at SectionEnd(RODATA) thinking it should represent the end of the full read-only segment, similar to how GNU ld shows it after .rodata

But since _etext is already defined at SectionEnd(TEXT) having etext point elsewhere makes them inconsistent. I’ll move it to SectionEnd(TEXT) to keep both aligned

libwild/src/parsing.rs
PritamP20: `DefinitionMode::Provide` and `String` don't need to be distinguished at runtime because is_canonical() already enforces PROVIDE behavior for all prelude symbols , when a user defines the same symbol, the prelude's entry becomes non-canonical and gets skipped automatically. Only`ProvideHidden` needs a separate flag (is_hidden) since it has a real runtime effect preventing dynsym export

davidlattimore: Makes sense to me. @lapla-cogito, is this resolved?

lapla-cogito (approved):
Thanks! Looks good.

PR #1535, opened 2026-02-15
linker-diff: Support RISC-V attributes

lapla-cogito:
part of #916


wild/tests/sources/linker-script-provide.c
```diff
@@ -22,6 +22,8 @@
 //#DiffIgnore:rel.R_AARCH64_ADR_GOT_PAGE.R_AARCH64_ADR_GOT_PAGE
 //#DiffIgnore:section.riscv.attributes
 //#DiffIgnore:segment.RISCV_ATTRIBUTES.*
+// GNU ld emits `.riscv.attributes`, but Wild does not
+//#DiffIgnore:riscv_attributes.*
```
lapla-cogito, line 26: This discrepancy might warrant investigation, but in any case, it's outside the scope of this PR.

Issue #1538, opened 2026-02-16
Abstract ELF-specific code


davidlattimore:
I'm going to put as much ELF-specific code as possible behind traits. This should hopefully make porting to non-ELF platforms easier.

marxin:
Great you started intensively working on this @davidlattimore. That's something that will unblock #756 and #757.
Feel free to ping once you have a feeling a macOS integration steps are worth doing.

davidlattimore:
I'm hesitant to call this done because I know there will be plenty more changes needed as porting to other platforms occurs. That said, I think that the bulk of the abstraction is now done. The main stages like input reading, symbol DB building, resolution and layout at least now don't have obvious references to ELF types or constants.

Layout code still references `linker_utils::elf::RelocationKind`, but that enum is already an abstraction that is a superset of the relocations kinds supported by the different ELF architectures. So I'm unsure what we should do with it. Perhaps it is ELF-specific and should be made into an associated type on `Platform`. Or maybe it's generic enough that it can be shared with other platforms. If we do decide to share it, then that possibly suggests that some of the code in `elf_writer::apply_relocation` should be moved to a common location so that it can be shared.

When porting, if you find yourself copying significant chunks of logic from the ELF implementation and only making minor changes, consider if the logic should be moved to common location. I'm sure there is ELF-specific logic that will need to move still, but it's also likely that the reverse is true and there's logic that should still be shared, but currently isn't.

Anyway, thanks everyone for putting up with the noise of so many PRs as well as probably many merge conflicts. I'm calling this done for now.

PR #1549, opened 2026-02-16
refactor: Rename ElfArch to Platform

davidlattimore:
Also renamed existing ELF-specific implementations to indicate that they're ELF-specific.

Issue #1538


davidlattimore:
In https://github.com/davidlattimore/wild/pull/1536 I renamed this trait from Arch to ElfArch. Saying that the trait itself was ELF-specific turned out to be too painful to work with. The implementations of the trait are definitely ELF-specific. The trait does have a few ELFy things in it, but I think we can reduce those with time. The Platform trait now has an associated type that specifies which object file format it's for.

PR #1550, opened 2026-02-17
refactor: Move associated types from ObjectFile to Format

davidlattimore:
Issue #1538


davidlattimore:
The whole `<Self::Format as Format>::` is a bit annoying, but rustc complains that the associated type is ambiguous otherwise.

PR #1553, opened 2026-02-17
fix: Placed downgraded symbols from version script correctly in .symtab

Blazearth:
## Fix: Place version script downgraded symbols in .symtab local section

Fixes #1508

I have fixed the problem by first making changes in the layout where the space for local symbols is calculated. Made a function to check for both the flags local or downgraded_to_local. Since in the starting only when symbols are loaded from .o files symtab it then only checks it against the version script. In rust version scripts it works by downgrading all symbols to local then it declare global from them so that doesnt causes problem.
Secondly made changes in the elf_writer to check against the flag and change the binding.
This ensures the elf format that all locals comes first before global.

PR #1554, opened 2026-02-17
chore: Update mold skip tests list

davidlattimore:


davidlattimore:
Presumably fixed by #1553

PR #1555, opened 2026-02-17
Revert "refactor: Move associated types from ObjectFile to Format"

davidlattimore:
Reverts #1550

This reverts commit 9c18a80b278e8d2997f50b2f443c0109d73d44fc.


davidlattimore:
I introduced the `Format` trait so that we could refer to various types without needing a lifetime. That meant using associated types on the trait that were generic over their lifetime. e.g. `type File<'data>: ObjectFile<'data>`. Unfortunately this kind of structure results in invariant lifetimes rather than the usual covariant lifetimes. So it looks like we'll need to just make stuff generic over `ObjectFile<'data>` rather than generic over `Format`. I'll get rid of `Format` in a subsequent PR.

PR #1574, opened 2026-02-20
feat: Support RISC-V HI20 relaxation

lapla-cogito:
Compress 4-byte lui to 2-byte c.lui when the HI20 immediate fits in a 6-bit signed non-zero range.

part of #874

### DIFF
--- a/linker-utils/src/riscv64.rs
+++ b/linker-utils/src/riscv64.rs
@@ -28,6 +28,9 @@ pub enum RelaxationKind {
 
     /// Rewrite auipc+jalr to jal.
     CallToJal,
+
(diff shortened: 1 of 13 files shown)


linker-utils/src/riscv64.rs
```diff
@@ -28,6 +28,9 @@ pub enum RelaxationKind {
 
     /// Rewrite auipc+jalr to jal.
     CallToJal,
+
+    /// Rewrite 4-byte `lui rd, imm` to 2-byte `c.lui rd, nzimm`.
+    HI20ToCLui,
```
davidlattimore: For consitency with naming elsewhere in Wild, I think the "I" in "HI" should be lowercase

PR #1575, opened 2026-02-20
perf: Parallelize section and symbol address computation in relaxation scan

lapla-cogito:
Merge `compute_object_section_addresses()` and `build_symbol_output_infos()` into a single pass that runs in parallel across groups.


davidlattimore (approved):
I benchmarked this linking wild for riscv (benchmark was run on x86_64) and it got a 12% speedup. Nice!

Issue #1576, opened 2026-02-20
SFrame version 3 support


mati865:
Linking anything on up to date Arch Linux now fails with:
```
wild: error: Failed to parse SFrame header at offset 0: Unsupported SFrame version 3 (expected 2)
```
Apparently Binutils 2.46 now defaults to SFRAME version 3.

lapla-cogito:
It might have been better to skip rather than bail when trying to process an unsupported SFrame version. Sorry for that.

However, since the specification isn't documented at https://sourceware.org/binutils/wiki/sframe, I'm unsure how reliably we can implement SFrame V3 at this stage.

mati865:
Yeah, that's exactly what I'm planning to do for the short term.

Documentation is sort of all around the place. For the version 3 you can find the docs here: https://sourceware.org/binutils/docs-2.46/sframe-spec.html
Version 2 docs are best read at https://sourceware.org/binutils/docs-2.45/sframe-spec.html

lapla-cogito:
Ah, I didn't know that link. Thanks!

mati865:
This SFrame sh*t show is not funny. Apparently, Arch Linux had to disable SFrame support in glibc to ease version 2 to version 3 transition and reenabled it only recently. Except it doesn't work for me with `backtrace()` called in C. I had even rebuilt glibc and rebooted.

In test case from `run_sframe_backtrace_test`, `backtrace()` returns 1 with `.eh_frame*` removed and 6 without removing them.

There are a few bugs in our current version 2 implementation that I wanted to fix before working on version 3, but I cannot test anything at the moment. I'll try 

To make things more complicated, `as` supports only the latest version of SFrame:
```
❯ as --help | rg sframe-
  --gsframe-<N>           generate SFrame version <N> information. 3 == <N>
```
So, `--gsframe-2` is declined while `--gsframe-3` works with Binutils 2.46. This makes me wonder whether supporting multiple SFrame versions is even feasible.

marxin:
Just updated my Arch Linux machine and got to the very same error - do we have any plan what to do about it?

mati865:
[...]

PR #1577, opened 2026-02-20
fix: Do not bail when encountering unsupported SFrame version

mati865:
cc #1576

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,9 +93,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.101"
+version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


mati865:
I can remove `thiserror` if we don't plan to use it in more places. I'm just used to this kind of handling errors.

mati865:
In this case, being able to differentiate the error is handy. It can be done without thiserror as well, that was just more natural to me.

lapla-cogito (approved):
I'm in favor of using `thiserror` when it makes the implementation cleaner, but I'm not sure how common such cases are beyond SFrame.

marxin:
Don't have a strong preference here - I am slightly leaning towards the opaque error type (e.g. `bail!`), but don't have any strong arguments against the `thiserror`. But I would like to see an unified approach for the error handling.

mati865:
Removed thiserror. Now the implementation is more in line with `VersionScriptError` and `LinkerScriptError`, although they handle it that way because of winnow.

PR #1585, opened 2026-02-23
refactor: Get rid of ElfPlatform trait

davidlattimore:
Issue #1538


davidlattimore:
I thought this would make some types a little bit simpler, but ends up making some things not work since even though `ElfPlatform` only ever has `Platform<File = crate::elf::File>`, the compiler can't be sure that that will always be the case.

PR #1596, opened 2026-02-26
feat: support matching based on linker script input filenames

levietduc0712:
This pull request adds support for matching input sections based on input file names (including glob patterns) in linker scripts. Now, section mapping rules can specify not only section name patterns but also which input files (or file patterns) they apply to, enabling more precise control over section placement. The implementation includes parsing, rule matching, and comprehensive tests for this new feature.

**Enhancements to section rule matching:**

* Added an `input_file_pattern` field to `SectionRule` and `Matcher`, allowing rules to optionally specify a glob pattern for matching input filenames. If `None`, the rule matches all files. 
* Updated the `SectionRule::matches` method and section rule lookup logic to consider both section name and input file name, using a new `glob_match_bytes` function that supports `*` and `?` wildcards.

**Parsing and propagation of file patterns:**

* Modified the linker script parser to support file patterns in section commands, correctly populating the `input_file_pattern` field in `Matcher`.
* Updated the section rule builder and related code to propagate file patterns from parsed matchers to section rules.

**Integration and testing:**

* Updated the section resolution logic to pass the input file name to rule lookup, ensuring correct rule selection based on both section and file name.
* Added extensive tests for parsing file patterns, glob matching logic, and section command parsing with file-specific and glob-specific matchers.

### DIFF
--- a/wild/tests/sources/linker-script-filename-match.c
+++ b/wild/tests/sources/linker-script-filename-match.c
@@ -1,6 +1,6 @@
 //#LinkerScript:linker-script-filename-match.ld
 //#Object:linker-script-filename-match-startup.c
-//#Object:linker-script-filename-match-app.c
+//#Archive:linker-script-filename-match-app.c
 //#Object:runtime.c
 //#DiffIgnore:segment.LOAD.RW.alignment
 //#SkipArch:riscv64


lapla-cogito:
How about adding test linker scripts to integration tests?

levietduc0712:
@lapla-cogito 
Quick clarification: are you suggesting moving test_section_command_with_filename() to the integration tests, or adding a new integration test that includes a test linker script?

lapla-cogito:
I intend the latter. Place the test code (in this case, likely a linker script and a C source file) under `wild/tests/sources`. After that, add the necessary test directives to the source file (see the beginning of `wild/tests/integration_tests.rs` for details), and then add the name of the newly added source file to `integration_test()`. Once this is done, running `cargo test` will execute the newly added test.

levietduc0712:
Is it okay now? Since it’s my first time, there might still be some changes needed.

libwild/src/resolution.rs
```diff
@@ -1196,7 +1196,15 @@ fn resolve_section<'data, O: ObjectFile<'data>>(
...
-    match rules.lookup(section_name, section_flags, section_type) {
+    let file_name = obj
+        .common
+        .input
+        .file
+        .filename
+        .file_name()
+        .map(|n| n.as_encoded_bytes());
+
+    match rules.lookup(section_name, file_name, section_flags, section_type) {
```
lapla-cogito, line 1212: I think this currently never matches archive members 🤔  . You pass `obj.common.input.file.filename.file_name()` to `rules.lookup()`, so the pattern sees like `libfoo.a` instead of the member name inside the archive. If you tweak the new test to replace `//#Object:input-filename-app.c` with `//#Archive:input-filename-app.c`, `app_code` ends up in `.text` rather than `.text.app`.

wild/tests/sources/linker-script-filename-match.c
lapla-cogito, line 1: Since this is a test related to linker scripts, it would be good to give it a prefix like "linker-script" in line with other similar tests.

wild/tests/sources/linker-script-filename-match.c
```diff
@@ -0,0 +1,20 @@
+//#LinkerScript:input-filename.ld
+//#Object:input-filename-startup.c
+//#Object:input-filename-app.c
+//#Object:runtime.c
+//#DiffIgnore:segment.LOAD.RW.alignment
+//#SkipArch:riscv64
```
lapla-cogito, line 6: Do these 2 directives really need to exist?

lapla-cogito (changes requested):
Thanks for working on this! I've added some comments. Also, CI seems to be failing checks for test code formatting and `cargo-minimal-versions`.

libwild/src/layout_rules.rs
```diff
@@ -449,6 +481,38 @@ fn section_name_prefix_hash(name: &[u8]) -> Option<u64> {
     Some(hash_bytes(name.get(..4)?))
 }
 
+/// Performs glob-style matching of `pattern` against `text`. Supports `*` (matches any sequence of
+/// characters) and `?` (matches any single character).
+fn glob_match_bytes(pattern: &[u8], text: &[u8]) -> bool {
```
davidlattimore: Wild already has a dependency on the `glob` crate. It's used in `version_script.rs`. Can you look and see if that'd be suitable to use here? Probably call `Pattern::new` just once for each pattern, since I think that part of the process is slightly expensive, so it's good to not repeat it.

wild/tests/integration_tests.rs
```diff
@@ -3580,7 +3580,8 @@ fn integration_test(
         "riscv-call-relaxation.s",
         "riscv-cross-object-call-relaxation.s",
         "riscv-hi20-relaxation.s",
-        "segment-end-syms.c"
+        "segment-end-syms.c",
+        "input-filename.c"
```
davidlattimore: The name "input-filename" doesn't tell me that this is related to linker scripts. What about something like "linker-script-filename-matcher.c" or something similar?

wild/tests/sources/linker-script-filename-match.c
levietduc0712: For #DiffIgnore:segment.LOAD.RW.alignment: Yes, the RW LOAD segment alignment can differ between Wild and the reference linker (ld.bfd/lld) due to implementation differences in page size handling. Without this directive, the linker-diff comparison would flag a false mismatch. The existing linker-script-executable.c test uses the same directive for the same reason.

For #SkipArch:riscv64: Yes, this test uses a custom linker script, and RISC-V relies on __global_pointer$ which is defined in the default linker script. BFD complains when it's missing from custom scripts. The other linker-script test linker-script-executable.c skips riscv64 for exactly this reason (see its comment on line 4-5).

libwild/src/resolution.rs
lapla-cogito: It seems you've marked this as resolved @levietduc0712 , but I believe the issue is still unresolved.

levietduc0712: Good catch! You're right, the current code passes the archive filename (e.g., libfoo.a) instead of the member name (e.g., app.o). I've fixed this by checking obj.common.input.entry first: for archive members, we now use entry.identifier.as_slice() for matching, and only fall back to file.filename.file_name() for regular object files.

wild/tests/sources/linker-script-filename-match.c
lapla-cogito: All right, thanks for the explanation.

libwild/src/layout_rules.rs
levietduc0712: Good suggestion! I've replaced the custom glob_match_bytes with [glob::Pattern](vscode-file://vscode-app/c:/Users/duc.le-viet/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) from the existing dependency. The pattern is now compiled once in [SectionRule::new()](vscode-file://vscode-app/c:/Users/duc.le-viet/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and stored as [Option<Pattern>](vscode-file://vscode-app/c:/Users/duc.le-viet/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so the parsing cost isn't repeated on every match. The [matches()](vscode-file://vscode-app/c:/Users/duc.le-viet/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) method now delegates to [pattern.matches()](vscode-file://vscode-app/c:/Users/duc.le-viet/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with a UTF-8 conversion of the filename.

davidlattimore:
Might be worthwhile installing [typos](https://crates.io/crates/typos-cli) so that you can run it before you upload the change. I have a script that gets run whenever I'm uploading a change. It runs typos, runs the tests in a few different modes etc.

davidlattimore:
My script, in case you're wondering currently looks like the following.

```sh
#!/usr/bin/bash
set -e
typos
cargo +nightly fmt --check
cargo clippy --all-targets --features mold_tests -- -D warnings
cargo clippy --no-default-features --all-targets -- -D warnings
cargo clippy --features plugins
cargo test --features plugins
if [ -z "$SKIP_MOLD_TESTS" ]; then
    WILD_TEST_CROSS=aarch64 cargo nextest run -v --features mold_tests
fi
```

Or at least that's the wild-specific part of the script. I also have other more generic scripts that grep the diff of changes for things like "DO NOT SUBMIT", notify me of any added files, etc.

wild/tests/sources/linker-script-filename-match.c
```diff
@@ -0,0 +1,20 @@
+//#LinkerScript:linker-script-filename-match.ld
+//#Object:linker-script-filename-match-startup.c
+//#Object:linker-script-filename-match-app.c
```
davidlattimore: Probably worthwhile having two variants of the test, one that uses object files and one that uses archives. Or just change one of the objects to an archive.

davidlattimore:
Where possible, it's good if you can make the title of the PR what will go into the release notes. If you prefix it with "feat:" then it'll go into the features section. I'll usually edit it when I'm merging the change if it doesn't seem quite right, but I sometimes forget.

lapla-cogito (approved):
I've modified the PR title to clearly indicate it relates to linker scripts (since this is typically documented in release notes, I think the relevant section should be more obvious). Otherwise, LGTM.

PR #1601, opened 2026-02-27
fix: Use correct symbol priority when primary definition is dynamic

levietduc0712:
This pull request improves the symbol resolution logic to correctly prioritize symbol bindings (strong, common, weak, GNU unique) when multiple object files define the same symbol. It also adds new integration tests to verify the correct behavior.


libwild/src/grouping.rs
```diff
@@ -265,6 +267,22 @@ impl<'data, O: ObjectFile<'data>> SequencedInputObject<'data, O> {
     pub(crate) fn is_dynamic(&self) -> bool {
         self.parsed.object.is_dynamic()
     }
+
+    pub(crate) fn symbol_strength(&self, symbol_id: crate::symbol_db::SymbolId) -> SymbolStrength {
```
davidlattimore, line 270: It looks like this function is more or less a copy of `ObjectFile::symbol_strength`. Can you deduplicate?

libwild/src/symbol_db.rs
```diff
@@ -1093,22 +1093,52 @@ impl<'data> SymbolBucket<'data> {
...
+                }
+                SymbolStrength::Undefined => {}
+            }
+        }
+
+        if let Some((_, id)) = max_common {
+            return Some(id);
+        }
+
+        first_weak
```
davidlattimore: The logic here seems to be more or less the same as `select_symbol`. Can they share code?

wild/tests/sources/symbol-binding.c
davidlattimore, line 1: I suspect this test passes even without your change. Can you adapt the test so that tests what you changed? You'll need to look at where `get_non_dynamic` is called and figure out when that code path is exercised. That will hopefully also allow you to make the commit message more specific as to what's being fixed.

levietduc0712:
All 3 review comments have been addressed.

wild/tests/sources/symbol-binding.c
davidlattimore: That makes sense now. However, I just tried reverting your non test changes and the test still passes. Does it fail for you?

levietduc0712: It passes for me as well

davidlattimore: Any idea why it passes without your fix?

levietduc0712: The test passes without the fix because of how definition() works with two-level indirection. So the test can't distinguish between the old and new `get_non_dynamic` because `select_symbol` + `definition()` always resolves the correct winner regardless of which non-dynamic alternative `get_non_dynamic` picks.

Let me update it.

davidlattimore: Thanks. I can confirm that the test now fails for me without your fix.

wild/tests/sources/symbol-binding-2.c
davidlattimore, line 1: I think the names of the two secondary files now don't make so much sense since they each contain one weak and one strong symbol.

levietduc0712: Do you have a suggestion for better names?

davidlattimore: I'm not sure what you're asking

levietduc0712: Sorry, I got a bit lost.

levietduc0712: The two files have been renamed.

davidlattimore (approved):

PR #1604, opened 2026-02-27
fix: Differentiate between local and global version symbols in linker-diff

levietduc0712:
This pull request refines how symbol versions are reported in the `read_gnu_version` function. Previously, local and global symbols were not differentiated and were both labeled as "local or global". Now, the function distinctly labels version 0 as "local" and version 1 as "global", providing more accurate information.

### DIFF
--- a/linker-diff/src/version_diff.rs
+++ b/linker-diff/src/version_diff.rs
@@ -120,10 +120,11 @@ fn read_gnu_version(bin: &crate::Binary) -> Result<FieldValues> {
         let version_index = version_index_raw & VERSYM_VERSION;
         let hidden = version_index_raw & VERSYM_HIDDEN == VERSYM_HIDDEN;
 
-        let version_name = if version_index == 0 {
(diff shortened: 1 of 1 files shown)


davidlattimore:
Note that GNU ld recently switched which version index they use for undefined symbols, so you'll possibly see differences in behaviour between different GNU ld versions

levietduc0712:
Based on your note about GNU ld recently switching which version index it uses for undefined symbols, I've reverted the differentiation and instead replaced the old TODO comment with a clearer explanation of why local and global are treated as equivalent since GNU ld's behavior varies across versions, comparing them would produce false positives. What do you think?

lapla-cogito:
I believe the original TODO comment better captures the problem's fundamental nature. Additionally, while type annotations have been added, this doesn't actually introduce any meaningful changes.

levietduc0712:
I think I’ll proceed with closing this PR.

PR #1605, opened 2026-02-27
feat: Support GDB index

### LINKED ISSUE #811: Add support for --gdb-index
This flag originated from GNU gold (no longer maintained), but has been adopted by lld and mold. As far as I can tell, GNU ld doesn't support this flag. Users of GNU ld would need to instead use the separate program `gdb-add-index`. Supporting this flag would be useful, since it would help with users migrating from lld. It sounds like it's used in conjunction with split debug info. I've observed the flag being used when building chromium.

[Background information](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Index-Files.html)

[Format docs](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Index-Section-Format.html#Index-Section-Format)

levietduc0712:
This pull request adds support for generating a `.gdb_index` section in ELF binaries, which is used to speed up debugging with GDB. The implementation introduces a new command-line flag (`--gdb-index` and `--no-gdb-index`), updates the layout and writer logic to allocate and fill in the section, and integrates the section into the ELF output as an optional part. It also updates relevant tests and internal data structures to accommodate the new section.

Resolve #811

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -4008,15 +4008,6 @@ fn write_eh_frame_hdr(table_writer: &mut TableWriter, layout: &Layout) -> Result
 
 fn write_gdb_index(table_writer: &mut TableWriter, layout: &Layout) -> Result {
     timing_phase!("Write .gdb_index");
-    if !layout.args().gdb_index {
-        return Ok(());
-    }
-
-    let gdb_layout = layout.section_layouts.get(output_section_id::GDB_INDEX);
-    if gdb_layout.mem_size == 0 {
-        return Ok(());
-    }
-
     let allocated_len = table_writer.gdb_index.len();
     if allocated_len == 0 {
         return Ok(());
@@ -4097,45 +4088,50 @@ fn write_gdb_index(table_writer: &mut TableWriter, layout: &Layout) -> Result {
     let address_area_size = addr_count * size_of::<elf::GdbIndexAddressEntry>();
     let symbol_table_offset = address_area_offset + address_area_size;
 
-    // Compute the symbol table + constant pool from .debug_gnu_pubnames sections.
-    // We need a second pass over objects-with-CUs, this time with their assigned CU indices.
-    let pubnames_symbol_size: usize;
-    let pubnames_hash_table_bytes: usize;
-    {
+    // Single pass to collect pubnames/pubtypes for each object that has a resolved .debug_info CU.
+    // This data is used for both computing the symbol table size and writing it.
+    let objects_with_pubnames = if let Some(di_id) = debug_info_section_id {
         let mut cu_index: u32 = 0;
-        let sections_iter = layout.group_layouts.iter().flat_map(|g| g.files.iter()).filter_map(|f| {
-            if let FileLayout::Object(object) = f { Some(object) } else { None }
-        }).filter_map(|object| {
-            let has_di = debug_info_section_id.is_some_and(|di_id| {
-                object.sections.iter().zip(object.section_resolutions.iter()).any(|(slot, res)| {
-                    matches!(slot, SectionSlot::LoadedDebugInfo(s) if s.output_section_id() == di_id)
-                        && res.address().is_some()
-                })
-            });
-            if has_di {
-                let this_cu_index = cu_index;
-                // Count how many .debug_info sections this object has
-                let di_count = object.sections.iter().zip(object.section_resolutions.iter())
+        layout
+            .group_layouts
+            .iter()
+            .flat_map(|g| g.files.iter())
+            .filter_map(|f| match f {
+                FileLayout::Object(obj) => Some(obj),
+                _ => None,
+            })
+            .filter_map(|object| {
+                let di_count = object
+                    .sections
+                    .iter()
+                    .zip(object.section_resolutions.iter())
                     .filter(|(slot, res)| {
-                        debug_info_section_id.is_some_and(|di_id| {
-                            matches!(slot, SectionSlot::LoadedDebugInfo(s) if s.output_section_id() == di_id)
-                                && res.address().is_some()
-                        })
+                        matches!(slot, SectionSlot::LoadedDebugInfo(s) if s.output_section_id() == di_id)
+                            && res.address().is_some()
                     })
-                    .count() as u32;
-                cu_index += di_count;
-                let pubnames = object.object.section_by_name(".debug_gnu_pubnames")
+                    .count();
+                if di_count == 0 {
+                    return None;
+                }
+                let this_cu_index = cu_index;
+                cu_index += di_count as u32;
+                let pubnames = object
+                    .object
+                    .section_by_name(".debug_gnu_pubnames")
                     .and_then(|(_, sec)| object.object.raw_section_data(sec).ok());
-                let pubtypes = object.object.section_by_name(".debug_gnu_pubtypes")
+                let pubtypes = object
+                    .object
+                    .section_by_name(".debug_gnu_pubtypes")
                     .and_then(|(_, sec)| object.object.raw_section_data(sec).ok());
                 Some((pubnames, pubtypes, this_cu_index))
-            } else {
-                None
-            }
-        });
-        (pubnames_hash_table_bytes, pubnames_symbol_size) =
-            gdb_index::compute_symbol_table_size(sections_iter.map(|(p, t, _)| (p, t)));
-    }
+            })
+            .collect::<Vec<_>>()
+    } else {
+        Vec::new()
+    };
+
+    let (pubnames_hash_table_bytes, pubnames_symbol_size) =
+        gdb_index::compute_symbol_table_size(objects_with_pubnames.iter().map(|&(p, t, _)| (p, t)));
 
     let constant_pool_offset = symbol_table_offset + pubnames_hash_table_bytes;
     let needed = symbol_table_offset + pubnames_symbol_size;
@@ -4172,51 +4168,7 @@ fn write_gdb_index(table_writer: &mut TableWriter, layout: &Layout) -> Result {
     // Write symbol table + constant pool if there are any pubnames entries.
     if pubnames_symbol_size > 0 {
         let sym_table_buf = &mut rest[..pubnames_symbol_size];
-
-        // Third pass: collect pubnames data with CU indices for writing.
-        let mut cu_index: u32 = 0;
-        #[allow(clippy::type_complexity)]
-        let sections_for_write: Vec<(Option<&[u8]>, Option<&[u8]>, u32)> = layout
-            .group_layouts
-            .iter()
-            .flat_map(|g| g.files.iter())
-            .filter_map(|f| {
-                if let FileLayout::Object(object) = f {
-                    Some(object)
-                } else {
-                    None
-                }
-            })
-            .filter_map(|object| {
-                let di_slots: Vec<_> = object
-                    .sections
-                    .iter()
-                    .zip(object.section_resolutions.iter())
-                    .filter(|(slot, res)| {
-                        debug_info_section_id.is_some_and(|di_id| {
-                            matches!(slot, SectionSlot::LoadedDebugInfo(s) if s.output_section_id() == di_id)
-                                && res.address().is_some()
-                        })
-                    })
-                    .collect();
-                if di_slots.is_empty() {
-                    return None;
-                }
-                let this_cu_index = cu_index;
-                cu_index += di_slots.len() as u32;
-                let pubnames = object
-                    .object
-                    .section_by_name(".debug_gnu_pubnames")
-                    .and_then(|(_, sec)| object.object.raw_section_data(sec).ok());
-                let pubtypes = object
-                    .object
-                    .section_by_name(".debug_gnu_pubtypes")
-                    .and_then(|(_, sec)| object.object.raw_section_data(sec).ok());
-                Some((pubnames, pubtypes, this_cu_index))
-            })
-            .collect();
-
-        gdb_index::write_symbol_table(sections_for_write.into_iter(), sym_table_buf);
+        gdb_index::write_symbol_table(objects_with_pubnames.into_iter(), sym_table_buf);
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


libwild/src/elf_writer.rs
```diff
@@ -819,42 +822,40 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
...
+            let allocated = *mem_sizes.get(part_id::GOT);
+            bail!("Allocated too much space in .got. {remaining} of {allocated} bytes remain.");
         }
         if !self.rela_dyn_relative.is_empty() {
-            return Err(excessive_allocation(
-                ".rela.dyn (relative)",
-                self.rela_dyn_relative.len() as u64 * elf::RELA_ENTRY_SIZE,
-                *mem_sizes.get(part_id::RELA_DYN_RELATIVE),
-            ));
+            let remaining = self.rela_dyn_relative.len() as u64 * elf::RELA_ENTRY_SIZE;
```
davidlattimore: Is this related to this change? I'm also curious as to the motivation for this bit of the change. It looks like you've inlined the function `excessive_allocation` into several of its usage sites.

libwild/src/layout.rs
```diff
@@ -4191,6 +4218,15 @@ impl<'data> ObjectLayoutState<'data> {
 
         tracing::debug!(loaded_debug_section = %self.object.section_display_name(section_index),);
         common.section_loaded(part_id, header, section);
+
+        if self
```
davidlattimore: Currently we read the section names only during section resolution. Doing an extra read of the section name can cause a noticeable slowdown. So I think we likely need to find a different way to do this.

libwild/src/elf_writer.rs
```diff
@@ -3984,6 +3996,101 @@ fn write_eh_frame_hdr(table_writer: &mut TableWriter, layout: &Layout) -> Result
...
+    buf[0x00..0x04].copy_from_slice(&VERSION.to_le_bytes());
+    buf[0x04..0x08].copy_from_slice(&(cu_list_offset as u32).to_le_bytes());
+    buf[0x08..0x0c].copy_from_slice(&(tu_list_offset as u32).to_le_bytes());
+    buf[0x0c..0x10].copy_from_slice(&(address_area_offset as u32).to_le_bytes());
+    buf[0x10..0x14].copy_from_slice(&(symbol_table_offset as u32).to_le_bytes());
+    buf[0x14..0x18].copy_from_slice(&(constant_pool_offset as u32).to_le_bytes());
+
+    for (i, &(cu_offset, cu_length)) in cu_entries.iter().enumerate() {
+        let off = cu_list_offset + i * CU_ENTRY_SIZE;
+        buf[off..off + 8].copy_from_slice(&cu_offset.to_le_bytes());
```
davidlattimore: Would it work to simplify some of this code by using zerocopy?

davidlattimore (commented):
Nice! Have you tried loading a binary with the gdb index into gdb to confirm that it's happy with it?

Any ideas about any other testing - automated or otherwise we could do? Checking that the section exists is a start, but it leaves a lot of scope for things being not quite right.

libwild/src/elf_writer.rs
levietduc0712: It’s not required for the .gdb_index work it was just a small refactor to make the error paths explicit, which I forgot to remove before committing. I’ve restored the original code.

libwild/src/layout.rs
levietduc0712: Done

libwild/src/elf_writer.rs
levietduc0712: Done also

levietduc0712:
> Nice! Have you tried loading a binary with the gdb index into gdb to confirm that it's happy with it?

Yes. I tested that explicitly.

I built a small test binary with `-Wl,--gdb-index` and then loaded it into GDB in batch mode to check for any structural issues. GDB accepted the binary without reporting any corruption, errors, or warnings.

In addition, I compared the generated `.gdb_index` section against an LLD-linked reference build. The section exists, the size matches, and the hex dump is identical.

So from GDB’s perspective, it appears to be fully valid and functional.

levietduc0712:
> Any ideas about any other testing - automated or otherwise we could do? Checking that the section exists is a start, but it leaves a lot of scope for things being not quite right.

I’ve added a `ValidateGdbIndex` directive that parses the raw `.gdb_index` and verifies: version = 7, header offsets are ordered and in-bounds, CU entries are non-overlapping and fully cover .debug_info, and address area entries fall within `.text`. I also added `ExpectGdbIndexCuCount: 3` to assert the exact CU count. Additionally, I loaded the binary in GDB and confirmed `info` sources lists files from all three input objects. Would this satisfy your request?

davidlattimore (approved):
The extra testing is great. Thanks!

libwild/src/layout.rs
```diff
@@ -3008,6 +3015,22 @@ fn resolution_flags(rel_kind: RelocationKind) -> ValueFlags {
...
+    let has_addresses = cu_count > 0;
+
+    size_of::<elf::GdbIndexHeader>() as u64
+        + cu_count * size_of::<elf::GdbIndexCuEntry>() as u64
+        + if has_addresses {
+            size_of::<elf::GdbIndexAddressEntry>() as u64
+        } else {
+            0
+        }
+        + MIN_SYMBOL_TABLE_SLOTS * 8
```
lapla-cogito: Wouldn't this result in the creation of a zero-padded hash table? In other words, it would just create unnecessary space of 1024*8=8192 bytes, which seems inefficient.

libwild/src/elf_writer.rs
```diff
@@ -3984,6 +3998,100 @@ fn write_eh_frame_hdr(table_writer: &mut TableWriter, layout: &Layout) -> Result
...
+                        .iter()
+                        .zip(object.section_resolutions.iter())
+                    {
+                        if let SectionSlot::LoadedDebugInfo(section) = slot
+                            && section.output_section_id() == di_section_id
+                            && let Some(address) = resolution.address()
+                        {
+                            cu_entries.push(elf::GdbIndexCuEntry {
+                                cu_offset: address - debug_info_start,
+                                cu_length: section.capacity(),
```
lapla-cogito: This seems to include alignment padding, making it inappropriate as a proper CU length.

libwild/src/elf_writer.rs
```diff
@@ -3984,6 +3998,100 @@ fn write_eh_frame_hdr(table_writer: &mut TableWriter, layout: &Layout) -> Result
...
+
+    let (cu_list, rest) =
+        <[elf::GdbIndexCuEntry]>::mut_from_prefix_with_elems(rest, cu_count).unwrap();
+    for (out, entry) in cu_list.iter_mut().zip(cu_entries.iter()) {
+        *out = *entry;
+    }
+
+    if has_addresses {
+        let (addr_entry, _rest) = elf::GdbIndexAddressEntry::mut_from_prefix(rest).unwrap();
+        let text_layout = layout.section_layouts.get(output_section_id::TEXT);
```
lapla-cogito: Doesn't this effectively bind all CUs to the `.text`? For cases where multiple CUs exist and have address ranges other than `.text`, wouldn't this become problematic?

libwild/src/output_section_id.rs
```diff
@@ -88,6 +88,8 @@ pub(crate) const RELA_PLT: OutputSectionId = part_id::RELA_PLT.output_section_id
 pub(crate) const EH_FRAME: OutputSectionId = part_id::EH_FRAME.output_section_id();
 pub(crate) const EH_FRAME_HDR: OutputSectionId = part_id::EH_FRAME_HDR.output_section_id();
 pub(crate) const SFRAME: OutputSectionId = part_id::SFRAME.output_section_id();
+#[allow(dead_code)]
```
lapla-cogito: why this?

libwild/src/elf_writer.rs
lapla-cogito, line 1: Shouldn't `validate_empty()` also check for buffer residue?

libwild/src/output_section_id.rs
levietduc0712: The #[allow(dead_code)] annotation suppresses the compiler warning that GDB_INDEX as an OutputSectionId constant is never directly referenced. Currently, the code only uses part_id::GDB_INDEX (the PartId) and calls .output_section_id() on it when needed, so the pre-computed OutputSectionId constant is unused.

lapla-cogito: I think this is used [here](https://github.com/levietduc0712/wild/blob/2a20ff7264318e1f0a9bec820d4222fba04e66a9/libwild/src/output_section_id.rs#L1008).

levietduc0712: Let me correct that

levietduc0712:
> Through `2a20ff7` in this PR, all tests in the GDB index category appear to pass (see [this run](https://github.com/davidlattimore/wild/actions/runs/22495559590/job/65168996536)), but the latest commit seems to cause many of these tests to fail again.

I fixed it. Check this [one](https://github.com/davidlattimore/wild/actions/runs/22512957084/job/65225797678)

davidlattimore:
Can you delete the lines from `mold_skip_tests.toml` for the tests that now pass?

Also, I was thinking that it might be good to enable `TestUpdateInPlace` for your new test, just to make sure that all the bytes of the new section are getting written.

levietduc0712:
This feature appears to be more complex than I initially estimated. I need to add gdb_index.rs file to implements hashing, hash-table sizing, parsing of .debug_gnu_pubnames/.debug_gnu_pubtypes, and writing of the symbol table + constant-pool required by the GDB index format see this [one](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Index-Section-Format.html).

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,227 @@
+use hashbrown::HashMap;
+
+pub(crate) fn hash_symbol_name(name: &[u8]) -> u32 {
+    let mut r: u32 = 0;
+    for &c in name {
+        r = r
+            .wrapping_mul(67)
+            .wrapping_add(u32::from(c))
```
marxin, line 8: Looking at the formula: `The formula is r = r * 67 + tolower (c) - 113.`
Don't we miss the `tolower`?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+    r
+}
+
+pub(crate) fn compute_hash_table_slots(num_symbols: usize) -> usize {
+    if num_symbols == 0 {
+        return 0;
+    }
+    let min_slots = num_symbols * 4 / 3 + 1;
+    let mut slots = 4usize;
```
marxin, line 19: What about `slot = (num_symbols * 4 / 3 + 1).next_multiple_of(4)`?

libwild/src/layout.rs
```diff
@@ -2136,6 +2138,90 @@ fn find_required_sections<'data, P: Platform<'data, File = crate::elf::File<'dat
...
+                    debug_info_section_id.is_some_and(|di_id| {
+                        obj.sections.iter().any(|s| {
+                            matches!(s, SectionSlot::LoadedDebugInfo(sec) if sec.output_section_id() == di_id)
+                        })
+                    })
+                })
+                .map(|obj| {
+                    let pubnames = obj
+                        .object
+                        .section_by_name(".debug_gnu_pubnames")
```
marxin, line 2211: Please add new constants to https://github.com/davidlattimore/wild/blob/56f438d1e3c1ddbd9ff85cb0681f25aeec1ab249/linker-utils/src/elf.rs#L687

libwild/src/layout.rs
```diff
@@ -4191,6 +4306,15 @@ impl<'data> ObjectLayoutState<'data> {
 
         tracing::debug!(loaded_debug_section = %self.object.section_display_name(section_index),);
         common.section_loaded(part_id, header, section);
+
+        if resources
+            .output_sections
+            .custom_name_to_id(SectionName(b".debug_info"))
```
marxin, line 4312: Similarly here.

wild/tests/integration_tests.rs
```diff
@@ -2852,6 +2890,201 @@ impl Assertions {
...
+        // Validate minimum header size (6 x u32 = 24 bytes).
+        const HEADER_SIZE: usize = 24;
+        if data.len() < HEADER_SIZE {
+            bail!(
+                ".gdb_index section too small for header: {} bytes",
+                data.len()
+            );
+        }
+
+        let version = u32::from_le_bytes(data[0..4].try_into().unwrap());
```
marxin, line 2923: Can't we directly read the `GdbIndexHeader` here (using the `zerocopy` library)?

wild/tests/integration_tests.rs
```diff
@@ -2852,6 +2890,201 @@ impl Assertions {
...
+        let cu_list_offset = u32::from_le_bytes(data[4..8].try_into().unwrap()) as usize;
+        let tu_list_offset = u32::from_le_bytes(data[8..12].try_into().unwrap()) as usize;
+        let address_area_offset = u32::from_le_bytes(data[12..16].try_into().unwrap()) as usize;
+        let symbol_table_offset = u32::from_le_bytes(data[16..20].try_into().unwrap()) as usize;
+        let constant_pool_offset = u32::from_le_bytes(data[20..24].try_into().unwrap()) as usize;
+
+        // Check version.
+        if version != 7 {
```
marxin, line 2931: Maybe we should declare a constant somwhere for the currently supported version.

wild/tests/integration_tests.rs
```diff
@@ -2852,6 +2890,201 @@ impl Assertions {
...
+        let exec_sections: Vec<_> = obj
+            .sections()
+            .filter(|s| {
+                matches!(s.flags(), object::SectionFlags::Elf { sh_flags } if sh_flags & u64::from(object::elf::SHF_EXECINSTR) != 0)
+            })
+            .map(|s| (s.address(), s.address() + s.size()))
+            .collect();
+
+        let addr_entry_count = address_area_size / 20;
+        for i in 0..addr_entry_count {
```
marxin, line 3055: Maybe use [chunks_exact](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks_exact)?

levietduc0712:
> Really cool PR, good work. Am curious if we want to support the latest version`9` (could be a subsequent PR)? Thank you.

Thanks! I’ll look into supporting the latest version (9) as well.

PR #1606, opened 2026-02-27
feat: Add R_AARCH64_PLT32 and R_AARCH64_GOTPCREL32 relocations

levietduc0712:
This pull request adds support for two missing AArch64 relocation types, `R_AARCH64_PLT32` and `R_AARCH64_GOTPCREL32`, which are not present in the upstream `object` crate or ELF header. The changes define these constants, handle them in relocation kind mapping, and improve their string representation for diagnostics.


davidlattimore:
Do you think it'd be possible to make a test that validates that we're handling these relocations correctly? Unless we can get a C compiler to emit them, it might be necessary write the test in assembly.

levietduc0712:
Yes, I’ll work on adding a test for this.

PR #1607, opened 2026-02-27
feat: Add ASSERT command parsing support to linker scripts

Blazearth:
Adds parsing support for `ASSERT(expression, "message")` commands in linker scripts. Both top-level and within `SECTIONS` blocks are supported.

Assertions are parsed and stored in the AST but not yet evaluated. This is commonly used in Linux kernel linker scripts to validate memory layout constraints.

## Changes

- Added `Assert` variant to `Command` and `SectionCommand` enums
- Implemented `parse_assert()` function
- Added handling in `layout_rules.rs` (currently no-op with TODO)
- Added 3 tests covering various ASSERT syntaxes

Evaluation will be implemented in a follow-up PR.

### DIFF
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -338,13 +338,13 @@ fn parse_expression<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> 
     parse_comparison.parse_next(input)
 }
 
-/// Parse comparison operators: <, >, <=, >=, ==, !=
+/// Parse comparison expression: expression < expression, expression == expression, etc.
 fn parse_comparison<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
     let mut left = parse_additive.parse_next(input)?;
 
     multispace0.parse_next(input)?;
 
-    while let Some(op) = opt(alt((
+    if let Some(op) = opt(alt((
         "<=".map(|_| CompOp::LessEqual),
         ">=".map(|_| CompOp::GreaterEqual),
         "==".map(|_| CompOp::Equal),
@@ -421,7 +421,7 @@ fn parse_primary<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
 
     alt((
         // Parentheses - parse expression inside
-        delimited('(', parse_comparison, ')'),
+        delimited('(', parse_expression, ')'),
         // Hex numbers (0x or 0X prefix)
         preceded(alt(("0x", "0X")), hex_uint::<_, u64, _>).map(Expression::Number),
         // Decimal numbers
@@ -465,29 +465,29 @@ fn parse_identifier_or_function<'a>(input: &mut &'a BStr) -> winnow::Result<Expr
                 Ok(Expression::Addr(arg))
             }
             b"ALIGN" => {
-                let arg_expr = parse_comparison.parse_next(input)?;
+                let arg_expr = parse_expression.parse_next(input)?;
                 multispace0.parse_next(input)?;
                 ')'.parse_next(input)?;
                 Ok(Expression::Align(Box::new(arg_expr)))
             }
             b"MIN" => {
                 // MIN takes two expressions separated by comma
-                let first = parse_comparison.parse_next(input)?;
+                let first = parse_expression.parse_next(input)?;
                 multispace0.parse_next(input)?;
                 ','.parse_next(input)?;
                 multispace0.parse_next(input)?;
-                let second = parse_comparison.parse_next(input)?;
+                let second = parse_expression.parse_next(input)?;
                 multispace0.parse_next(input)?;
                 ')'.parse_next(input)?;
                 Ok(Expression::Min(Box::new(first), Box::new(second)))
             }
             b"MAX" => {
                 // MAX takes two expressions separated by comma
-                let first = parse_comparison.parse_next(input)?;
+                let first = parse_expression.parse_next(input)?;
                 multispace0.parse_next(input)?;
                 ','.parse_next(input)?;
                 multispace0.parse_next(input)?;
-                let second = parse_comparison.parse_next(input)?;
+                let second = parse_expression.parse_next(input)?;
                 multispace0.parse_next(input)?;
                 ')'.parse_next(input)?;
                 Ok(Expression::Max(Box::new(first), Box::new(second)))


Blazearth:
The AArch64 test appears to be failing during the wild-action installation step (exit code 2), not during actual test execution. All other checks pass and tests pass locally. Let me know if I should investigate further or if this is a known CI issue.

libwild/src/linker_script.rs
```diff
@@ -69,6 +69,11 @@ pub(crate) enum Command<'a> {
         value: &'a [u8],
     },
     Provide(ProvideSymbolDefinition<'a>),
+    Assert {
```
davidlattimore: I'd probably suggest making the assertion be a struct

libwild/src/linker_script.rs
```diff
@@ -69,6 +69,11 @@ pub(crate) enum Command<'a> {
...
+        /// ASSERT(expression, "message") — parsed but not yet evaluated.
+        expression: &'a [u8],
```
davidlattimore: Could be good to parse the expression, otherwise we'll need to do more parsing work when we go to evaluate the expression.

libwild/src/linker_script.rs
Blazearth: I have added the assertion be struct.

libwild/src/linker_script.rs
Blazearth: Good point, would you prefer adding the expression parsing to this PR only or should i add up in a follow up PR. Either is fine by me.

davidlattimore: I don't have a strong preference, although parsing the expression fits nicely with the title of this PR, so maybe add it to this PR?

Blazearth: Okay

Blazearth:
## Update: Expression Parsing Implemented

I've implemented full expression parsing as requested. The parser now handles:

**Supported features:**
- Arithmetic operators: `+`, `-`, `*`, `/` with correct precedence
- Comparison operators: `<`, `>`, `<=`, `>=`, `==`, `!=`
- Functions: `SIZEOF()`, `ADDR()`, `ALIGN()`
- Numbers (hex/decimal), symbols, location counter (`.`)
- Parentheses for grouping
- Trailing input validation to catch parsing errors

**Tests added:**
- Operator precedence: `1 + 2 * 3 == 7`
- Nested parentheses: `(1 + 2) * (3 + 4) == 21`
- Functions in comparisons: `SIZEOF(.data) + SIZEOF(.bss) < 0x2000`
- Hex number parsing: `0xDEADBEEF`
- ALIGN function with expressions

**Scope documentation:**
Added inline documentation listing what's currently supported and what's not yet implemented (unary operators, bitwise/logical ops, additional functions). This covers the common ASSERT patterns used in Linux kernel linker scripts.

libwild/src/linker_script.rs
```diff
@@ -251,6 +301,338 @@ fn parse_provide<'input>(
     })
 }
 
+fn parse_assert<'input>(input: &mut &'input BStr) -> winnow::Result<AssertCommand<'input>> {
+    '('.parse_next(input)?;
+    skip_comments_and_whitespace(input)?;
+
+    // Parse expression (everything until comma)
+    let expr_bytes = take_while(1.., |b| b != b',').parse_next(input)?;
```
davidlattimore: I think this approach would hit problems if the expression itself is able to contain a ','. e.g. perhaps 'MIN' and 'MAX'. It's probably best if `parse_expression` itself consumes as much as it can and can stop when it reaches something that can't be an extension of the expression.

libwild/src/linker_script.rs
```diff
@@ -251,6 +301,338 @@ fn parse_provide<'input>(
...
+    let expr_bytes = expr_bytes.trim_ascii();
+
+    // Parse the expression
+    let expression = match parse_expression(expr_bytes) {
```
davidlattimore: Winnow parse functions usually return a `winnow::Result` rather than a `Result`. You then call them as `parse_expression.parse_next(input)?`. I think this leads to better error reporting. e.g. if at some point in the expression, something unexpected is encountered, winnow can report where in the overall input the unexpected token was found.

libwild/src/linker_script.rs
```diff
@@ -251,6 +301,338 @@ fn parse_provide<'input>(
...
+            *pos += 1;
+        }
+        let hex_str =
+            std::str::from_utf8(&input[start..*pos]).map_err(|_| error!("Invalid hex number"))?;
+        let num = u64::from_str_radix(hex_str, 16).map_err(|_| error!("Invalid hex number"))?;
+        return Ok(Expression::Number(num));
+    }
+
+    // Handle decimal numbers
+    if input[*pos].is_ascii_digit() {
```
davidlattimore: Winnow provides functions for parsing numbers. It might be worthwhile looking through the winnow tutorial to get a feel for the things it offers - https://docs.rs/winnow/latest/winnow/_tutorial/chapter_0/index.html

libwild/src/linker_script.rs
Blazearth: I’m changing parse_assert so it doesn’t stop at the first comma. Instead of slicing until b',', I let the expression parser consume input using the full grammar and then advance the cursor by what it parsed. After that, I expect the comma at the ASSERT level.
This matches how GNU ld handles it, the expression parser deals with nested parentheses and commas inside function calls, and ASSERT just parses the outer structure. It keeps the responsibilities clean and avoids fragile delimiter matching.

libwild/src/linker_script.rs
Blazearth: Okay will look into it. Thanks.

libwild/src/linker_script.rs
```diff
@@ -251,6 +307,238 @@ fn parse_provide<'input>(
...
+}
+
+/// Parse an expression - entry point for expression parsing
+fn parse_expression<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
+    parse_comparison.parse_next(input)
+}
+
+/// Parse comparison operators: <, >, <=, >=, ==, !=
+fn parse_comparison<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
+    let mut left = parse_additive.parse_next(input)?;
```
davidlattimore, line 343: I'd expect that the left hand side of a comparison would itself be an expression

libwild/src/linker_script.rs
```diff
@@ -251,6 +307,238 @@ fn parse_provide<'input>(
...
+
+    Ok(AssertCommand {
+        expression,
+        message,
+    })
+}
+
+/// Parse an expression - entry point for expression parsing
+fn parse_expression<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
+    parse_comparison.parse_next(input)
```
davidlattimore, line 338: I know for an assertion that an expression must be a comparison, but not all expressions need to be comparisons. I guess there are two options. You could change `parse_assert` to call `parse_comparison`, or you could parse an arbitrary expression, and then at some point we validate that that expression is indeed a comparison.

libwild/src/linker_script.rs
```diff
@@ -251,6 +307,238 @@ fn parse_provide<'input>(
...
+}
+
+/// Parse comparison operators: <, >, <=, >=, ==, !=
+fn parse_comparison<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
+    let mut left = parse_additive.parse_next(input)?;
+
+    multispace0.parse_next(input)?;
+
+    while let Some(op) = opt(alt((
```
davidlattimore: It seems strange to me that this is a loop. Does that mean that `a <= b == c` is valid?

libwild/src/linker_script.rs
```diff
@@ -251,6 +307,238 @@ fn parse_provide<'input>(
...
+    Ok(left)
+}
+
+/// Parse primary expressions: numbers, symbols, functions, parentheses
+fn parse_primary<'a>(input: &mut &'a BStr) -> winnow::Result<Expression<'a>> {
+    multispace0.parse_next(input)?;
+
+    alt((
+        // Parentheses - parse expression inside
+        delimited('(', parse_comparison, ')'),
```
davidlattimore: Why is a comparison what we'd expect inside parenthesis? What about other expressions like `a * (b - c)`? Would a comparison as anything but the top-level be valid?

libwild/src/linker_script.rs
```diff
@@ -251,6 +307,238 @@ fn parse_provide<'input>(
...
+            }
+            b"ALIGN" => {
+                let arg_expr = parse_comparison.parse_next(input)?;
+                multispace0.parse_next(input)?;
+                ')'.parse_next(input)?;
+                Ok(Expression::Align(Box::new(arg_expr)))
+            }
+            b"MIN" => {
+                // MIN takes two expressions separated by comma
+                let first = parse_comparison.parse_next(input)?;
```
davidlattimore: Not sure that a comparison makes sense as an argument to something like `MAX`

davidlattimore (commented):
Sorry about the slow review. Looks like I started reviewing and had some draft comments, but forgot to send them and I only just noticed.

Blazearth:
Thanks for reviewing will take my time to read more about this and resolve the issues.

libwild/src/linker_script.rs
Blazearth: In gnu ld it treats any expression that resolve to non zero value as valid in assertion and if it reolves to zero it stops. So do you recommend i should check for comparison later after parsing the expression or i follow the same behaviour as gnu ld

libwild/src/linker_script.rs
Blazearth: Yes i have changed from while to if

libwild/src/linker_script.rs
Blazearth: You're right, MIN/MAX should accept any expression as arguments, not just comparisons. I'll change them to use parse_expression instead of parse_comparison

libwild/src/linker_script.rs
Blazearth: Will change from parse_comparison to parse_expression

libwild/src/linker_script.rs
davidlattimore: Yes, I'd say that type checking should be a separate pass and that we shouldn't try to bake type checking into parsing.

davidlattimore (approved):
~~Could you test that chaining and mixing of~~ Edit: Never mind. This was a half finished comment that I changed my mind on and forgot to delete.

PR #1609, opened 2026-02-28
chore: cargo clippy issues

marxin:


mati865:
Dunno if I would call that a fix. There are no behaviour differences, so sounds more like a `chore:` to me.

marxin:
Good point, renamed.

mati865 (approved):

PR #1612, opened 2026-03-01
feat: Add relocatable linking (-r) and kernel linker script support

levietduc0712:
This PR adds initial support for relocatable linking (-r) and the linker script features required to build Linux kernel modules.

What This Adds

## 1. Partial-Link (-r) Mode

Implements true relocatable output (`ET_REL`):

- Adds `-r` CLI support
- Produces `ET_REL` instead of `ET_EXEC` / `ET_DYN`
- Disables:
  - `gc_sections`
  - `relro`
  - internal linker-only symbol emission
- Skips program header generation (no PHDRs for `ET_REL`)
- Preserves relocation sections and types
- Adjusts layout/resolution flow for partial linking

## 2. Relocations & Offsets

Ensures correct cross-object behavior in relocatable output:

- Adjust relocation offsets by section base offsets during emission
- Preserve original relocation types
- Maintain correct `sh_link` / `sh_info`
- Handle `__`-prefixed `.rela__*` sections (e.g. `.rela__ex_table`)
- Add robust `SymbolId → output index` mapping for cross-object references

## 3. Linker Script Raw Data (BYTE/SHORT/LONG/QUAD)

Adds support for linker-script data commands used by kernel scripts:

- Parse and store raw section data (`section_datas`)
- Count sizes early during `finalise_sizes`
- Mark linker-script-only sections as `must-keep`
- Ensure alignment consistency between:
  - part-id selection
  - buffer allocation
  - emission
- Emit raw script bytes during ELF write

## 4. /DISCARD/ and Discard Semantics

- Implement proper `/DISCARD/` matcher handling
- Allow discard-only sections in partial-link mode
- Avoid fatal errors when referenced discard sections exist (e.g. `.discard.addressable`)

## 5. Section Size / Keep / Ordering Fixes

Ensure linker-script-only sections:

- Contribute to size computation
- Are preserved in output order
- Fix alignment consistency when allocating buffers

Ensure correct flow through:

- `finalise_sizes`
- `finalise_layout`
- emission

## 6. Symbol Table Handling

- Avoid emitting linker-internal symbols in `-r` mode
- Use global identity maps to resolve symbols across objects
- Fix `.symtab`, `.strtab`, `.shstrtab` handling when linker scripts are present
- Avoid duplicate section table artifacts

## 7. Program Segment Handling

For `ET_REL` output:

- Disable program segment creation entirely
- Ensure sections requiring addressability remain valid
- Keep NULL / FILE_HEADER section at index 0 and correctly zeroed

## 8. Stability & Defensive Fixes

- Prevent unwrap panics when linker-script-only sections have no buffers
- Ensure section index 0 always exists
- Improve internal consistency in layout/emission transitions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1806,18 +1806,18 @@ checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
 
 [[package]]
 name = "strum"
-version = "0.27.2"
+version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "af23d6f6c1a224baef9d3f61e287d2761385a5b88fdab4eb4c6f11aeb54c4bcf"
+checksum = "9628de9b8791db39ceda2b119bbe13134770b56c138ec1d3af810d045c04f9bd"
 dependencies = [
  "strum_macros",
 ]
 
 [[package]]
 name = "strum_macros"
-version = "0.27.2"
+version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7695ce3845ea4b33927c055a39dc438a45b059f7c1b3d91d38d10355fb8cbca7"
+checksum = "ab85eea0270ee17587ed4156089e10b9e6880ee688791d45a905f5b1ca36f664"
 dependencies = [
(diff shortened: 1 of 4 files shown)


davidlattimore:
Would it be possible to split this into multiple PRs, one for each feature?

PR #1614, opened 2026-03-01
feat: Support SORT() and glob wildcard patterns in linker scripts

levietduc0712:
Add SORT(...) wrapper support in linker script section matchers so that patterns like SORT(___ksymtab+*) are expanded and matched as globs.

Also generalise SectionRule to support full glob character-class patterns (e.g. .bss.[0-9a-zA-Z_]*) by compiling them with the glob crate instead of using simple prefix matching.

### DIFF
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -454,9 +454,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
 }
 
 fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
-    let pattern = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
-    skip_comments_and_whitespace(input)?;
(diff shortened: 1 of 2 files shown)


libwild/src/linker_script.rs
```diff
@@ -454,9 +454,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
+            if b == b'(' {
+                depth += 1;
+            } else if b == b')' {
+                depth -= 1;
+            }
+        }
+        if depth != 0 {
+            return Err(ContextError::new());
+        }
+        // Return the entire SORT(...) token including closing ')'
```
davidlattimore, line 473: Returning the whole "SORT(...)" as a byte array seems weird to me. I'd expect that we'd parse to some structure that indicates that we're applying sorting, otherwise something elsewhere will have to reparse the string to see if it starts with "SORT(".

davidlattimore (commented):
Can you add a test that uses SORT in a linker script and verifies that it works as expected?

PR #1615, opened 2026-03-01
feat: Support /DISCARD/ section in linker scripts

levietduc0712:
Add handling for /DISCARD/ sections in linker scripts, which instructs the linker to discard matched input sections.

Also add implicit discard rules for .symtab, .strtab, .shstrtab and .group sections when a linker script is active, since the linker generates its own versions and the input-object copies should be dropped.


davidlattimore (commented):
Looks good to me. Thanks! It'd be good to have a test that passes such a linker script to the linker and then verifies that the sections aren't present in the output.

PR #1616, opened 2026-03-01
feat: Support BYTE/SHORT/LONG/QUAD commands in linker scripts

levietduc0712:
Add parsing and emission of raw inline byte data in linker scripts:

    .plt 0 : { BYTE(0) }         -- 1 byte, decimal value
    .data : { SHORT(0x1234) }     -- 2 bytes, little-endian
    .data : { LONG(0x12345678) }  -- 4 bytes, little-endian
    .data : { QUAD(1) }           -- 8 bytes, little-endian

The bytes flow through: linker_script parser → ProcessedLinkerScript
→ ResolvedLinkerScript → LinkerScriptLayoutState, where:
  - activate() marks affected sections as must-keep
  - finalise_sizes() counts bytes so buffer space is allocated
  - finalise_layout() advances memory offsets
  - write_linker_script_state() copies bytes into the output buffers

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -739,7 +739,7 @@ pub(crate) struct LinkerScriptLayoutState<'data> {
     input: InputRef<'data>,
     symbol_id_range: SymbolIdRange,
     pub(crate) internal_symbols: InternalSymbols<'data>,
-    pub(crate) section_datas: Vec<(crate::output_section_id::OutputSectionId, Vec<u8>)>,
+    pub(crate) section_data: Vec<(crate::output_section_id::OutputSectionId, Vec<u8>)>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


libwild/src/layout.rs
```diff
@@ -5715,7 +5716,16 @@ impl<'data> LinkerScriptLayoutState<'data> {
         resources: &FinaliseLayoutResources<'_, 'data>,
     ) -> Result {
         self.internal_symbols
-            .finalise_layout(memory_offsets, resolutions_out, resources)
+            .finalise_layout(memory_offsets, resolutions_out, resources)?;
+
+        // Allocate space for any raw bytes that the linker script requested via
```
davidlattimore, line 5721: This isn't actually allocating space, that happened earlier. This is computing addresses. Or perhaps more accurately, bumping past such data so that other things can compute addresses.

davidlattimore (commented):
I've only had an initial look so far. Could you add a test that uses these commands in a linker script? Perhaps use a few different BYTE/LONG etc commands and then check at runtime that the section contains the correct bytes in the correct order.

PR #1617, opened 2026-03-01
feat: Add support for partial/relocatable linking (-r flag)

levietduc0712:
Implement partial linking support via the -r / --relocatable flag:

- args.rs: Add -r/--relocatable flag (sets output_relocatable=true,
  gc_sections=false, relro=false)
- output_kind.rs: New OutputKind::Relocatable variant with is_partial_link(),
  needs_dynamic(), base_address() adjustments; is_executable() fixed
- elf_{x86_64,aarch64,loongarch64,riscv64}.rs: arch-specific partial-link stubs
- gc_stats.rs: minor adjustments
- resolution.rs: preserve .rela/.crel sections in relocatable output;
  also propagate section_datas from ParsedLinkerScript (needed by layout.rs)
- layout.rs: skip GOT/PLT allocation; zero out section VMAs; no program
  segments; no linker-internal symbols; preserve undefined symbols;
  SectionSlot::Discard non-fatal; better symtab index tracking per group
- elf_writer.rs: ET_REL file header; build global symbol output map;
  write .rela* output sections with section-offset fixups; skip program
  headers for ET_REL; preserve undefined globals in symbol table
- linker_script.rs / layout_rules.rs: include BYTE/SHORT/LONG/QUAD and
  all linker-script features needed by kernel module link scripts
- wild/tests/integration_tests.rs: add test_kernel_module_relocatable_output
- linker-diff/src/asm_diff.rs: adjustments for relocatable output diffing

### DIFF
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -113,7 +113,7 @@ impl<'data> LayoutRulesBuilder<'data> {
         output_sections: &mut OutputSections<'data>,
     ) -> Result<ProcessedLinkerScript<'data>> {
         let mut symbol_defs = Vec::new();
-        let mut section_datas: Vec<(OutputSectionId, Vec<u8>)> = Vec::new();
+        let mut section_data: Vec<(OutputSectionId, Vec<u8>)> = Vec::new();
 
         for cmd in &input.script.commands {
             if let linker_script::Command::Provide(provide) = cmd {
@@ -234,7 +234,7 @@ impl<'data> LayoutRulesBuilder<'data> {
                                     }
                                     ContentsCommand::Bytes(b) => {
                                         // Append bytes for this output section.
-                                        section_datas.push((primary_section_id, b.clone()));
+                                        section_data.push((primary_section_id, b.clone()));
                                     }
                                 }
                             }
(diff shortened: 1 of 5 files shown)


libwild/src/layout_rules.rs
```diff
@@ -147,10 +152,27 @@ impl<'data> LayoutRulesBuilder<'data> {
                                 .unwrap_or(alignment::MIN)
                                 .max(replace(&mut extra_min_alignment, alignment::MIN));
 
+                            // Support `/DISCARD/ : { ... }` which indicates matched sections should
```
davidlattimore, line 155: It looks like part of one of your other PRs has made it into this one. Could you either remove it from this PR or mark this PR as a draft until the other one is submitted?

wild/tests/integration_tests.rs
```diff
@@ -3821,6 +3821,73 @@ fn read_test_config() -> Result<TestConfig> {
     Ok(config)
 }
 
+// Kernel-module / relocatable output tests
+#[cfg(test)]
+mod kernel_module_tests {
```
davidlattimore, line 3826: This style of test is very different to our existing tests. Do you think it might be possible to meet the needs of testing this feature via our existing testing framework? One possibility that comes to mind is to do something similar to the "//#Archive:" directive, which allows us to create an archive from some object files that are then passed to the linker. "//#Shared:" is similar. Similarly, we could have a "//#Relocatable:" directive that partially links.

davidlattimore (commented):
I haven't looked at this in detail yet.

PR #1618, opened 2026-03-01
docs: Write up LLM use policy

davidlattimore:


davidlattimore:
You're right that we should talk about run-time performance. Probably testing too. I'll do that in a separate PR, since it's not LLM/AI specific.

Issue #1462, opened 2026-01-16
Support for running on non-Unix hosts


bjorn3:
There are a bunch of places where Unix-only apis are used. While Wild is currently an ELF-only linker and thus mostly usable for Unix targets (though there are a bunch of non-Unix OSes that use ELF), it would still be useful to use Wild on non-Unix hosts when cross-compiling from Windows. Or in my case compiling Wild to WASI. Would changes to support this be accepted? When compiling for WASI, I saw the following issues:

* [x] https://github.com/davidlattimore/perfetto-recorder/issues/9 (not relevant for Windows, fix merged)
* [ ] Bunch of `std::os::unix::ffi::OsStrExt` uses.
* [ ] `default_target_arch` doesn't compile on wasm32.
* [ ] Rayon fails to initialize on WASI unless `.use_current_thread()` is used.
* [x] `O_CLOEXEC` (https://github.com/davidlattimore/wild/pull/1463)
* [ ] `make_executable` (can be a no-op outside of Unix hosts)
* [ ] `std::os::unix::fs::symlink`
* [ ] `fork` (The Windows kernel technically supports `fork`, but it is even slower than spawning a new process, which is already slow. Probably no benefit to gain from using the fork trick on Windows)

davidlattimore:
That all sounds good to me. `O_CLOEXEC` is https://github.com/davidlattimore/wild/pull/1463

I'll be interested to see what the implications of `.use_current_thread` are. So long as we can have all CPUs busy with one thread each, I'm happy.

There's https://github.com/davidlattimore/wild/pull/1407 that was aiming to build on windows. It hasn't seen activity for a few weeks though, but if you wanted, I could push changes to their branch and then merge it.

Once it's all done, we should add CI jobs to verify that we don't break it.

bjorn3:
Wasi doesn't have threads at all right now, so `.use_current_thread()` is used to disable usage of threads. Something that works on wasi and I think wouldn't affect performance would be to use that option when the thread pool would have a single thread. The jobserver crate doesn't return any tokens on wasi, so the implicit token results in a single worker thread being requested by default on wasi.

fahdfady:
I'm interested in working on this. not sure that i'll complete the whole checklist but i'll probably be starting with the `std::os::unix` usage within the codebase and see what i could do :)

PR #1629, opened 2026-03-03
feat: Support building on Windows

fahdfady:
should help us finish some of the checklist in #1462

### DIFF
--- a/libwild/src/save_dir.rs
+++ b/libwild/src/save_dir.rs
@@ -352,36 +352,41 @@ impl SaveDirState {
     }
 }
 
-#[cfg(unix)]
 fn create_symlink(target: &Path, dest_path: &Path) -> Result {
-    std::os::unix::fs::symlink(target, dest_path).with_context(|| {
-        format!(
-            "Failed to symlink {} to {}",
-            dest_path.display(),
-            target.display()
-        )
-    })?;
-    Ok(())
-}
-
-#[cfg(windows)]
-fn create_symlink(target: &Path, dest_path: &Path) -> Result {
-    let is_dir = std::fs::metadata(target)
-        .map(|meta| meta.is_dir())
-        .unwrap_or(false);
-    let result = if is_dir {
-        std::os::windows::fs::symlink_dir(target, dest_path)
-    } else {
-        std::os::windows::fs::symlink_file(target, dest_path)
-    };
-    result.with_context(|| {
-        format!(
-            "Failed to symlink {} to {}",
-            dest_path.display(),
-            target.display()
-        )
-    })?;
-    Ok(())
+    #[cfg(unix)]
+    {
+        std::os::unix::fs::symlink(target, dest_path).with_context(|| {
+            format!(
+                "Failed to symlink {} to {}",
+                dest_path.display(),
+                target.display()
+            )
+        })?;
+        Ok(())
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::fs::FileTypeExt as _;
+        let is_dir = std::fs::metadata(target)
+            .map(|meta| meta.is_dir())
+            .unwrap_or(false);
+        let is_symlink_dir = std::fs::symlink_metadata(target)
+            .map(|meta| meta.file_type().is_symlink_dir())
+            .unwrap_or(false);
+        let result = if is_dir || is_symlink_dir {
+            std::os::windows::fs::symlink_dir(target, dest_path)
+        } else {
+            std::os::windows::fs::symlink_file(target, dest_path)
+        };
+        result.with_context(|| {
+            format!(
+                "Failed to symlink {} to {}",
+                dest_path.display(),
+                target.display()
+            )
+        })?;
+        Ok(())
+    }
 }
 
 fn make_linker_script_relative(bytes: &[u8], source_path: &Path) -> Result<Vec<u8>> {


fahdfady:
@davidlattimore is this good? also, should we start having Windows in integration tests and CI?

libwild/src/save_dir.rs
```diff
@@ -358,6 +352,38 @@ impl SaveDirState {
...
+            dest_path.display(),
+            target.display()
+        )
+    })?;
+    Ok(())
+}
+
+#[cfg(windows)]
+fn create_symlink(target: &Path, dest_path: &Path, source_path: &Path) -> Result {
+    let is_dir = std::fs::metadata(source_path)
```
davidlattimore: Why is this checking `source_path` rather than `target`?

davidlattimore:
We use git-clif for generating our release changelogs, so we like our PR titles to reflect what will go into the release notes.

What are your thoughts on scope for this work? Are you looking to get wild building for windows? I think there probably isn't too much that needs changing in order to build for windows. Separate PRs is fine, but we probably want one line in the release notes. e.g. `feat: Support building on Windows`. If we're doing separate PRs, then we can skip putting all but the last one in the release notes. To skip putting something in the release notes, prefix it with `chore:`. Realistically, I'll probably skip putting refactoring PRs in the release notes for our next release, since I've made way too many refactoring PRs for it to be anything but noise.

Regarding checking the build in CI, yes we should check it in CI once it's working.

libwild/src/save_dir.rs
mati865: On Windows there are symlinks (files) and junctions (directories). So, if you want to create unix-like symlink pointing to a directory, you need to create junction.

davidlattimore: That explains why we need to know whether we're working with a directory or not, but I'm not sure it explains why we're checking `source_path` rather than `target`.

fahdfady: >  Why is this checking source_path rather than target?

@davidlattimore i'm always skeptical dealing with relative paths.. `target` is mutated in the code to be the relative path of a `PathBuf` version of `source_path`
I think validating with an absolute path makes more sense. lmk if I'm getting this wrong.

fahdfady:
> Are you looking to get

@davidlattimore I think I can make this PR about Windows. as you said It shouldn't take a long time to finish windows support. I guess I can push the tests of windows too. I'm changing the PR title now. thanks.

libwild/src/save_dir.rs
davidlattimore: We're dealing with relative paths whether we want to or not, since we're linking to a relative path. I guess I'd think about this from the perspective of what a comment describing the behaviour of this function would look like. Right now, it might look something like:

```rust
/// Create a symbolic link `dest_path` that points to `target`. If `target` is a directory then `source_path` must also be a directory. If `target` is a regular file then `source_path` must be a regular file. Otherwise, we'll fail to create the link when running on Windows.
```

That description just sounds weird and fragile to me.

fahdfady: > On Windows there are symlinks (files) and junctions (directories). So, if you want to create unix-like symlink pointing to a directory, you need to create junction.

I just learned that today. Thanks for making me read about it! pushing that update now.

fahdfady: > That description just sounds weird and fragile to me.

That's frankly the exact behavior i intended to create. can you please tell me why you think it's weird and fragile? I wanted to double-check with the absolute path being the source of truth.

davidlattimore: It just seems unnecessary complex. What we care about is whether `target` is a directory or not, so why pass an extra path that has to share characteristics with `target`? It's fragile because if we accidentally pass an unrelated path, we'll attempt to create the wrong kind of link and fail.

fahdfady: makes sense. I'll remove the argument and validate against `target` instead.

lapla-cogito:
yaml syntax error: https://github.com/wild-linker/wild/actions/runs/22697566474

.github/workflows/ci.yml
```diff
@@ -2,16 +2,16 @@ name: CI
 
 on:
   push:
-    branches: [ 'main' ]
+    branches: ["main"]
```
davidlattimore: I'm assuming this is the result of running some automatic formatter on this file. I'm all for automatic formatters, but we should probably do that as a separate PR that says what automatic formatter we use and ideally enforce that formatting somehow.

fahdfady: I was just about to revert that now. This happened because i use `yamlls` on my neovim setup.
Are you open to use something like [yamlfmt](https://github.com/google/yamlfmt) in this repository and maybe have it run in pre-commit? just to standarize fmt for every contributor working on this project. If so i can work on it. for now i'll revert this.

davidlattimore: Using an automatic formatter SGTM. I'm not familiar with any, but from a quick search yamlfmt seems like a reasonable choice. I'm unsure about using commit hooks if that's what you mean by pre-commit. None of our existing formatting checks use them. Usually we just check the formatting from a github action. For clang formatting of C test files, we have a test that checks them. That way so long as people run the tests before they upload their change, they'll see if anything needs reformatting. When the test fails, it's good if it prints the command to run to fix the formatting.

fahdfady:
@mati865 what do you think about this?
https://github.com/wild-linker/wild/actions/runs/22749331708/job/65980378072#step:5:362

davidlattimore:
Sounds like the API to create a junction isn't stable yet. I'd say perhaps just use `symlink_dir` like I think you originally had. It's interesting (and weird) that creating symbolic links on windows is privileged. Makes me wonder if we might eventually want to do something other than symlinking stuff, but I guess this is part of the code that's generally only used by developers anyway.

BTW, what platform are you testing on? If you're developing on Linux and want to test building for windows, it should be possible. I just did `sudo apt install gcc-mingw-w64-x86-64-win32` and was then able to run `cargo check --target x86_64-pc-windows-gnu` and see various windows / non-unix specific compilation errors.

fahdfady:
Nice. didn't know i can do that. I'm on linux dualbooted with Windows. I was going to boot windows and test but that seems so much better.

fahdfady:
should i extract `path_from_bytes` to another place and make modules use it? It does the same functionality twice in the codebase. but i'm also skeptical of moving it to something like `utils.rs` because it's _only_ used two time in the codebase

davidlattimore:
I'd say probably move it to `fs.rs`

.github/workflows/ci.yml
```diff
@@ -85,6 +85,26 @@ jobs:
...
+      - uses: actions/cache@v5
+        with:
+          path: |
+            C:\Users\runneradmin\.cargo\registry\index\
+            C:\Users\runneradmin\.cargo\registry\cache\
+            C:\Users\runneradmin\.cargo\git\db\
+            target\
+          key: ${{ runner.os }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.lock') }}
+      - run: cargo build --profile ci --workspace --no-default-features
+      - run: cargo test --profile ci -p libwild --lib --no-default-features
```
davidlattimore: Probably not much point running the tests on Windows at this stage. Unless we make most of them linux-only, they'll fail.

fahdfady:
for `make_executable` i didn't find what can i do to achieve the same utility function for windows. and working with the LLM, it told me that we don't want to set permissions  windows, and is eventually unnecessary because it's baked into the format itself and only a `.exe` extension would do. so the function should have no use on windows. tell me what do you think because I feel like i don't quite understand.

I'm disabling `fork` for windows because all of the `subprocess.rs` is using UNIX primitives (literally `fork`). We can acheive this in windows but I don't think we have the same  primitives that gives us the safety mentioned in https://github.com/wild-linker/wild/blob/main/libwild/src/subprocess.rs#L20-L22. tell me what you think about that. I think `fork` should be Unix-only.

mati865:
On Windows file is made executable by the extensions, like `.exe`. There is nothing to do here.

`fork` is beneficial on Linux, but Windows alternatives might be neutral or even regress the performance. I'd expect the latter to actually be the case. IMO it should stay disabled unless proven to improve performance.

libwild/src/save_dir.rs
```diff
@@ -358,6 +352,38 @@ impl SaveDirState {
...
+            target.display()
+        )
+    })?;
+    Ok(())
+}
+
+#[cfg(windows)]
+fn create_symlink(target: &Path, dest_path: &Path) -> Result {
+    let is_dir = std::fs::metadata(target)
+        .map(|meta| meta.is_dir())
```
mati865: I'm not sure about this one, but if I remember correctly you cannot create symlink to a junction, in which case this condition should test for both `is_dir()` and `is_symlink_dir()`.

mati865:
Posting this in case not everybody here is aware, but on Windows you cannot create symlinks as a regular user. To do so, you have to enable Developer mode or elevate the process to run as Admin (kind like sudo). So, we might need to also implement a fallback to copy files instead of symlinking.

n1ght-hunter:
> Posting this in case not everybody here is aware, but on Windows you cannot create symlinks as a regular user. To do so, you have to enable Developer mode or elevate the process to run as Admin (kind like sudo). So, we might need to also implement a fallback to copy files instead of symlinking.

It should probably be handled in a similar way to uv which outputs a warning on fallback. 
https://github.com/astral-sh/uv/issues/7509

libwild/src/save_dir.rs
fahdfady: don't think that this is necessary.

> The original path must be a directory or a symlink to a directory, otherwise the symlink will be broken. Use [symlink_file](https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_file.html) for other files.
https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_dir.html

If I understand correctly, this means that it works out-of-the-box to check on if it's a symlink dir. Also, I didn't quite get the junction part? We previously removed it because it's not stable yet.

mati865: Oh, sorry my bad about junctions. This is using a newer API rather than junctions.

Your check `is_dir()` will return false for symlinks to directories and looking at symlink_file description it cleary says you cannot use it on symlinks to directories:
> The original path should not be a directory or a symlink to a directory.

https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_file.html

So, you need to check for path being either dir or symlink to dir.

fahdfady: Makes sense 👍

mati865 (approved):

Issue #1405, opened 2025-12-23
wild --version > /dev/full panics


oech3:
```
$ wild --version > /dev/full

thread 'main' (12789) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

https://github.com/uutils/coreutils/issues/9769

Noxyntious:
it's full

oech3:
```
$ run0 dmsetup create ioerr --table "0 8 error"
$ chmod 777 /dev/mapper/ioerr
$ wild --help > /dev/mapper/ioerr
thread 'main' (2225) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: No space left on device (os error 28)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

davidlattimore:
Looks like the other linkers silently ignore the error that the write to stdout failed, although my guess is that's more to do with people not bothering to check the return value of printf than a deliberate decision. While not a high priority, I figure we might as well fix this. I don't think we need to match the behaviour of the other linkers, where they silently ignore the error, but avoiding the panic would be nice. The error message is already pretty good - `failed printing to stdout: No space left on device (os error 28)`, we just don't want the panic stuff.

It should just be a matter of looking for `println!` calls and changing them to lock stdout then write to that, returning an error if that fails. Not sure if it's worth changing all `println!` usages, some are more for development purposes (e.g. `--time`). But we should at least change `--version` and `--help` to not panic. If the write to stderr fails, then I guess we can either silently ignore it or panic. Nobody will see the panic anyway, since it can't be written to stderr.

PritamP20:
Hi! I’d like to work on this issue.

After going through the codebase, these are the main places where `println!` could panic if stdout fails:

* **--version** (`libwild/src/lib.rs`, around lines 184–196)
* **--help** (`libwild/src/args.rs`, around lines 1715–1728)

My plan is to replace the `println!` calls with a locked stdout and use `writeln!` so that write errors return properly instead of panicking, something like:

```rust
use std::io::Write;
let mut stdout = std::io::stdout().lock();
writeln!(stdout, "{}", linker_identity())?;
```

For `--help`, I might need to adjust the closure’s return type or handle the error inline
I’ll leave the other `println!` calls (like `--time`, `--sym-info`, etc.) as they are for now.

[...]

PR #1631, opened 2026-03-03
fix: prevent panic when writing to stdout fails in --version and --help

PritamP20:
This PR replaces println! calls with explicit writeln! to a locked stdout in the --version and --help command handlers.
This ensures that if writing to stdout fails (e.g., when redirected to /dev/full), the linker returns a proper error instead of panicking.

Changes made:
- Updated libwild/src/lib.rs (in --version mode) to use std::io::stdout().lock() + writeln!.
- Updated libwild/src/args.rs (in --help execution closure) similarly.

fixed: issue #1405 

Preserved existing output formatting and behavior.

Screenshot 2026-03-03 at 10 28 20 PM

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -1717,12 +1717,14 @@ fn setup_argument_parser() -> ArgumentParser {
         .long("help")
         .help("Show this help message")
         .execute(|_args, _modifier_stack| {
+            use std::io::Write as _;
             let parser = setup_argument_parser();
-            println!("{}", parser.generate_help());
+            let mut stdout = std::io::stdout().lock();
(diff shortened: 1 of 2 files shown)


marxin:
> Preserved existing output formatting and behavior.

Thank you for your contribution. When it comes to console output samples, use rather a [code block](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code), rather than a screenshot.

PritamP20:
@marxin sure totally forgot about code blocks, will take care of it from next time!

PR #1313, opened 2025-11-17
chore: port to latest object crate

marxin:
Let's wait for a proper `object` crate release: https://github.com/gimli-rs/object/issues/821.

PR #1632, opened 2026-03-03
feat: implement R_X86_64_CODE_* relocations with relaxations

### LINKED ISSUE #1299: Support R_X86_64_CODE_6_GOTTPOFF relocations
I didn't see this used in the wild yet, but that seems to be a fresh thing: https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg334504.html
For example, GCC configuration scripts check whether the linker supports it (see `configuration` changes in the link above).

vishruth-thimmaiah:
closes #1299

This PR adds support for the `R_X86_64_CODE_*` relocations, with relaxations wherever possible.

### DIFF
--- a/wild/tests/sources/tls-apx-relocs.s
+++ b/wild/tests/sources/tls-apx-relocs.s
@@ -1,7 +1,6 @@
 //#Object:runtime.c
 //#RunEnabled:false
 //#LinkArgs:-z noexecstack
-//#EnableLinker:lld
 //#Arch: x86_64
 
 .section .tbss,"awT",@nobits
@@ -30,9 +29,6 @@ _start:
 	addq tls@GOTTPOFF(%rip), %r28, %r28
 	addq %r16, tls@GOTTPOFF(%rip), %r16
 	addq tls@GOTTPOFF(%rip), %r16, %r16
-	{nf} addq %r8, tls@GOTTPOFF(%rip), %r16
-	{nf} addq tls@GOTTPOFF(%rip), %rax, %r12
-	{nf} addq tls@GOTTPOFF(%rip), %r12
 
 # R_X86_64_CODE_4_GOTPCRELX
 	mov foo@GOTPCREL(%rip), %r16


vishruth-thimmaiah:
I've raised this PR as a draft as the latest stable release of object.rs doesnt include these relocs, which is why Ive used their hex values instead.

libwild/src/elf_x86_64.rs
```diff
@@ -161,8 +161,8 @@ impl<'data> crate::platform::Platform<'data> for ElfX86_64 {
         let offset = offset_in_section as usize;
 
         match relocation_kind {
-            object::elf::R_X86_64_REX_GOTPCRELX => {
-                if offset < 3 {
+            object::elf::R_X86_64_REX_GOTPCRELX | 0x2b => { // R_X86_64_CODE_4_GOTPCRELX
```
davidlattimore: If the object crate doesn't define the constant yet, it'd be good to define our own constant.

davidlattimore (commented):
Thanks for working on this. Can you add a test that uses these relaxations? It might need to be written in asm if you can't get the C compiler to produce them.

libwild/src/elf_x86_64.rs
lapla-cogito: While released object crate doesn't define these constants, upstream does. We can specify the git hash in `Cargo.toml` like [this](https://github.com/wild-linker/wild/pull/1313/changes/4c9695eb5bc91ab6ca4ab5fb3269f4355f62b224). This way, when a new version of the object crate is released, you only need to update your dependencies.

vishruth-thimmaiah: Done!

lapla-cogito:
In Wild, we run mold tests via CI, and tests related to APX have started passing (see "Check tests that should fail still fail" in [this](https://github.com/wild-linker/wild/actions/runs/22670673920/job/65713798435)). Could you remove these tests from `wild/tests/external_tests/mold_skip_tests.toml`? (BTW, in such cases, CI isn't marked as failed because false positives occasionally occur)

linker-diff/src/lib.rs
```diff
@@ -204,8 +204,10 @@ impl Config {
                 // GNU ld sometimes makes this writable sometimes not. Presumably this depends on
                 // whether there are relocations or some flags.
                 "section.eh_frame.flags",
+                "rel.R_X86_64_32.R_X86_64_32",
```
vishruth-thimmaiah: I'm not sure if this is the ideal way to handle this case, I get a `symbol is to far away` error when validating `R_X76_64_CODE_4_GOTPCRELX`, which is relaxed into a `R_X86_64_32.R_X86_64_32`.

davidlattimore: We wouldn't want to add such a broad rule as a default ignore. I think at least part of the problem might be in your test. If I run `readelf -Ws` on the input object file, to check the address of `foo`:

```
readelf -Ws /home/d/work/wild/wild/tests/build/tls-apx-relocs.s/default-host/tls-apx-relocs.s.ld
...
    2: 0000000000000400     0 TLS     LOCAL  DEFAULT    3 foo
```

It shows that indeed the address is a long way from the address that the move instruction used. It also shows that the symbol is a TLS variable. I think perhaps you just put `foo` in the wrong section?

vishruth-thimmaiah: Moving it to `.text` fixed it, thanks!

lapla-cogito (approved):
Thanks!

PR #1636, opened 2026-03-03
feat: Support lui deletion in RISC-V HI20 relaxation when %hi is zero

lapla-cogito:
While a relaxation to convert `lui` to `c.lui` was already implemented, when `%hi(value)` is 0, it's possible to remove `lui` and change the `rs1` value of the following `addi` to `x0`. This would achieve a 4-byte reduction.

part of #874

### DIFF
--- a/wild/tests/external_tests/mold_skip_tests.toml
+++ b/wild/tests/external_tests/mold_skip_tests.toml
@@ -134,10 +134,8 @@ tests = [
 reason = "RISC-V specific tests"
 tests = [
   "arch-riscv64-global-pointer.sh",
-  "arch-riscv64-obj-compatible.sh",
   "arch-riscv64-relax-align.sh",
   "arch-riscv64-relax-got.sh",
-  "arch-riscv64-relax-hi20.sh",
   "arch-riscv64-reloc-overflow.sh",
 ]
 
@@ -157,6 +155,7 @@ tests  = ["tls-common.sh", "tls-le-error.sh"]
 [skipped_groups.ignore]
 reason = "We ignore these tests for some reasons"
 tests = [
+  "arch-riscv64-obj-compatible.sh", # Different message formats
   "arch-riscv64-relax-j.sh",        # Passes when `--no-gc-sections` is passed.
   "arch-x86_64-empty-arg.sh",       # Different message formats
   "arch-x86_64-note.sh",            # The order of sections is different.


wild/tests/external_tests/mold_skip_tests.toml
```diff
@@ -157,6 +155,7 @@ tests  = ["tls-common.sh", "tls-le-error.sh"]
 [skipped_groups.ignore]
 reason = "We ignore these tests for some reasons"
 tests = [
+  "arch-riscv64-obj-compatible.sh", # Different message formats
```
lapla-cogito, line 158: While this test itself had the corresponding functionality implemented in #1526 , I didn't realize at the time that the test was virtually passing, so I moved it this time.

libwild/src/layout.rs
```diff
@@ -4687,6 +4859,10 @@ impl Resolution {
 /// happens in 2–3 passes.
 const MAX_RELAXATION_ITERATIONS: usize = 5;
 
+/// Sentinel value stored in `SymbolOutputInfos::addresses` for symbols whose output address is
+/// unknown.
+const SYMBOL_ADDRESS_UNRESOLVED: u64 = u64::MAX;
```
lapla-cogito, line 4864: Previously, symbols with value=0 were treated as unresolved, causing relaxation to be skipped.

libwild/src/elf_riscv64.rs
```diff
@@ -387,38 +417,43 @@ fn collect_relaxation_deltas<R: Relocation>(
...
-                                // 0x1000 change in value.
-                                let hi20 = value.wrapping_add(0x800) >> 12;
-                                let hi20_signed = hi20 as i64;
-                                let margin = if hi20_signed > 31 {
-                                    ((hi20_signed - 31) as u64) * 0x1000
-                                } else if hi20_signed < -32 {
-                                    // Address decreases from relaxation push hi20 further negative.
-                                    // This can never be fixed.
-                                    u64::MAX
+                    {
```
davidlattimore: Why is this scope here? It seems unnecessary

davidlattimore (approved):
I haven't really thought through the logic behind the bulk of the riscv-specific parts of this change, but given that there's a new integration test that checks it, I'm happy with it if you are.

libwild/src/elf_riscv64.rs
lapla-cogito: These are changes made during implementation that were later unnecessary. Thanks for pointing this out.

PR #1649, opened 2026-03-06
fix: Don't use nop for calls to undefined functions

### LINKED ISSUE #1102: Make linker-diff more tolerant of differences when a call is made to an undefined function
This is motivated by [this comment](https://github.com/davidlattimore/wild/discussions/832#discussioncomment-14482542). It seems that on aarch64, GNU ld sometimes replaces a call to an undefined function with a `nop` instruction and sometimes doesn't.

Wild currently attempts to match the behaviour that we've observed, which is to replace with a `nop`. We should perhaps remove this behaviour and leave the `bl 0` instruction, which will segfault if executed, but that IMO is probably better than executing the `nop` and continuing. The code in Wild that does this is [here](https://github.com/davidlattimore/wild/blob/main/libwild/src/aarch64.rs#L182). Once we remove it however, we'll need to make our tests pass despite having a difference of behaviour from GNU ld (at least on some platforms).

Steps:

1) Remove the behaviour from wild
2) Run the tests and confirm that the unresolved-symbols test fails.
3) Copy and paste the linker-diff command line that will have been printed when the test fails. Running this directly means you can see tracing output if you add `tracing::trace!` lines to linker-diff.

The code that does this diffing is in `asm_diff.rs`.

PritamP20:
**Fixes #1102**

GNU ld replaces `bl foo` with `nop` when `foo` is an undefined symbol (i.e. resolves to address `0`). Wild was copying this behavior. This change removes it.

Emitting `bl 0x0` (which segfaults if executed) is preferable to silently replacing the instruction with `nop`. A crash at the call site makes the issue obvious, while a `nop` allows the program to continue executing in an undefined state.

Wild already has the infrastructure to handle the edge case where the 26-bit branch offset overflows, so the `nop` replacement is unnecessary.

Wild will still emit a warning when a call to an undefined symbol is encountered:

```
WARNING: wild: Undefined symbol foo, referenced by
    wild/tests/build/unresolved-symbols-object.c/warn-unresolved-symbols-host/unresolved-symbols-object.9b626dcec291312b.o
```

### Changes

* **libwild/src/elf_aarch64.rs**
  Removed the `non_zero_address` branch for `R_AARCH64_CALL26` / `R_AARCH64_JUMP26`.
  Wild now always emits a direct call via `RelaxationKind::NoOp`.

* **libwild/src/elf_loongarch64.rs**
  Applied the same change for `R_LARCH_B26`.
  Also removed the now-unused `rel_info_from_type!` macro and its import.

* **wild/tests/sources/unresolved-symbols-object.c**
  Updated `DiffIgnore` from `extra-opt` to `missing-opt.*`, since GNU ld still emits `nop` but Wild no longer does.

* **wild/tests/sources/shared.c**
  Applied the same `DiffIgnore` change.

### DIFF
--- a/libwild/src/elf_loongarch64.rs
+++ b/libwild/src/elf_loongarch64.rs
@@ -11,7 +11,6 @@ use linker_utils::elf::SIZE_2KB;
 use linker_utils::elf::loongarch64_rel_type_to_string;
 use linker_utils::elf::shf;
 use linker_utils::loongarch64::RelaxationKind;
-
 use linker_utils::relaxation::RelocationModifier;
 use linker_utils::utils::or_from_slice;
 
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -290,6 +290,7 @@ impl Config {
                     // bl 0x0 so that if the call site is reached, it will crash rather than
                     // silently continuing execution.
                     "rel.missing-opt.R_AARCH64_CALL26.ReplaceWithNop.*",
+                    "rel.missing-opt.R_AARCH64_JUMP26.ReplaceWithNop.*",
                 ]
                 .into_iter()
                 .map(ToOwned::to_owned),


PritamP20:
I’ll move the DiffIgnore rules into apply_wild_defaults in linker-diff (separately for AArch64 and LoongArch64) and remove the per-test DiffIgnore entries from the test files.

I’ll also update the PR title to something like "Fix: Don't use nop for calls to undefined functions" so the behavior change is clearer.

Makes sense to handle it once centrally instead of patching it in individual tests. Thanks for pointing it out.

davidlattimore (approved):

Issue #588, opened 2025-03-22
Implement proper ordering of .init_array, .ctors etc


davidlattimore:
Currently, we just fill `.init_array` in whatever order the input sections appear in. Same for `.fini_array`. `.ctors` is also added to `.init_array` as though it were the same. However, `.ctors` is supposed to have reversed ordering. Also, both of these sections can have an optional priority. e.g. `.init_array.00005`. Much the same things apply to `.fini_array` and `.dtors`. See [Maskray's blog on this stuff](https://maskray.me/blog/2021-11-07-init-ctors-init-array) 

Actually sorting these sections will require some refactoring, possibly substantial. Most likely, it'd require adding the ability to transfer responsibility for a section and its relocations to the epilogue (or the prelude). All input sections destined for `.init_array` etc would be thus transferred and assigned a priority. The epilogue would then handle processing of relocations, sorting by priority and copying the sections to the output.

I've written some code to make linker-diff check the ordering of these sections, which I'll upload shortly.

marxin:
Can you please share a sample output from the linker-diff if there's a mismatch in a `.init_array` or `.fini_array`?

davidlattimore:
Sure. This is how libc-integration test fails if I comment out the diff-ignore lines:

```
wild: /home/david/work/wild/wild/tests/build/libc-integration.c-gcc-dynamic-no-pie-host.wild
ld: /home/david/work/wild/wild/tests/build/libc-integration.c-gcc-dynamic-no-pie-host.ld
init_array
  ┌─────────────┬─────────────┐
  │ wild        │ ld          │
  ├─────────────┼─────────────┤
  │ frame_dummy │ frame_dummy │
  │ init1       │ init3       │
  │ init2       │ init2       │
  │ init3       │ init1       │
  └─────────────┴─────────────┘
```

And this is rust-integration:

```
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
wild: /home/david/work/wild/wild/tests/build/rust-integration.rs-llvm-dynamic-host.wild
ld: /home/david/work/wild/wild/tests/build/rust-integration.rs-llvm-dynamic-host.ld
init_array
  ┌─────────────────────────────────────────────────┬────────────────────────────────────────────────┐
  │ wild                                            │ ld                                             │
  ├─────────────────────────────────────────────────┼────────────────────────────────────────────────┤
[...]

PR #1654, opened 2026-03-06
fix: make three more x86_64 relaxations reqired for static binaries

mati865:
This change fixes linking of C++ hello world with `-static -Wl,--no-relax` and addresses one of linker-diff's errors about missing relaxation.

Wild still applies too many relaxations as shown in linker-diff output, but disabling some of them results in broken binaries, so I gave up on touching that. Also, in the output there are not yet implemented `endbr64` relaxations.

### DIFF
--- a/wild/tests/sources/cpp-integration.cc
+++ b/wild/tests/sources/cpp-integration.cc
@@ -23,21 +23,24 @@
 //#LinkArgs:-no-pie -Wl,-z,now
 //#EnableLinker:lld
 
-//#Config:static:default
+//#Config:static-no-relax:default
 //#CompArgs:-fmerge-constants
 //#LinkerDriver:g++
 //#LinkArgs:-static -Wl,-z,now,-no-relax
 //#DiffIgnore:rel.extra-got-plt-got
+//#DiffIgnore:section.rela.plt.link
+// TODO: This is likely an important issue.
 //#DiffIgnore:init_array
+// TODO: Missing `endbr64` relaxations.
 //#DiffIgnore:rel.match_failed.R_X86_64_GOTPCRELX
 //#DiffIgnore:rel.match_failed.R_X86_64_REX_GOTPCRELX
 //#DiffIgnore:rel.match_failed.R_X86_64_PLT32
 //#DiffIgnore:literal-byte-mismatch
+// TODO: Some conditions for required relaxations are wrong.
 //#DiffIgnore:rel.extra-opt.R_X86_64_GOTPCRELX.JmpIndirectToRelative*
 //#DiffIgnore:rel.extra-opt.R_X86_64_REX_GOTPCRELX.RexCmpIndirectToAbsolute*
 //#DiffIgnore:rel.extra-opt.R_X86_64_REX_GOTPCRELX.RexMovIndirectToAbsolute*
 //#DiffIgnore:rel.missing-opt.R_X86_64_GOTTPOFF.RexMovIndirectToAbsolute*
-//#DiffIgnore:section.rela.plt.link
 //#Arch: x86_64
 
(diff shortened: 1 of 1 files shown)


marxin (approved):

mati865:
My changes fixed the issue on Arch Linux, but all other test runners still hit it:
```
wild: error: Undefined symbol __tls_get_addr, referenced by 
    /usr/lib/gcc/x86_64-linux-gnu/15/libstdc++.a @ eh_globals.o
```
Welp, at least I didn't regress anything, although failing to link is a bar set pretty low.

mati865:
Quite a bumpy ride but added test that now passes the CI.

wild/tests/sources/cpp-integration.cc
```diff
@@ -23,6 +23,23 @@
 //#LinkArgs:-no-pie -Wl,-z,now
 //#EnableLinker:lld
 
+//#Config:static:default
```
davidlattimore: Perhaps call the config "static-no-relax" to better capture what we're testing?

wild/tests/sources/cpp-integration.cc
```diff
@@ -23,6 +23,23 @@
...
+//#CompArgs:-fmerge-constants
+//#LinkerDriver:g++
+//#LinkArgs:-static -Wl,-z,now,-no-relax
+//#DiffIgnore:rel.extra-got-plt-got
+//#DiffIgnore:init_array
```
davidlattimore, line 33: Do you think any of these ignores are things that we should look at in future? If so, perhaps add a TODO. The init_array diff especially stands out to me

mati865: I thought `init_array` diff is a known defect, I get it even when linking dynamic executables with linker-diff enabled. For example, when linking Release profile (without debug info) Clang.
FWIW, this is the diff from this test:
```
    wild: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc/static-no-relax-host/cpp-integration.cc.wild
    ld: /home/mateusz/Projects/wild/wild/tests/build/cpp-integration.cc/static-no-relax-host/cpp-integration.cc.ld
    init_array
      ┌──────────────────────────────────────┬──────────────────────────────────────┐
      │ wild                                 │ ld                                   │
      ├──────────────────────────────────────┼──────────────────────────────────────┤
      │ _GLOBAL__sub_I.00090_globals_io.cc   │ _GLOBAL__sub_I.00090_globals_io.cc   │
      │ frame_dummy                          │ frame_dummy                          │
      │ _GLOBAL__sub_I_eh_alloc.cc           │ _GLOBAL__sub_I_cxx11_wlocale_inst.cc │
      │ _GLOBAL__sub_I_cxx11_locale_inst.cc  │ _GLOBAL__sub_I_locale_inst.cc        │
      │ _GLOBAL__sub_I_cxx11_wlocale_inst.cc │ _GLOBAL__sub_I_wlocale_inst.cc       │
      │ _GLOBAL__sub_I_locale_inst.cc        │ _GLOBAL__sub_I_eh_alloc.cc           │
      │ _GLOBAL__sub_I_wlocale_inst.cc       │ _GLOBAL__sub_I_cxx11_locale_inst.cc  │
      │ _IO_stdfiles_init                    │ _IO_stdfiles_init                    │
      └──────────────────────────────────────┴──────────────────────────────────────┘
```

davidlattimore: It was a known issue, but I thought it was fixed in #588, but it sounds like there might be more to do.

mati865: For some reason I thought it wasn't supposed to be a full fix.

PR #1656, opened 2026-03-06
doc: Add governance documentation

davidlattimore:


GOVERNANCE.md
```diff
@@ -0,0 +1,100 @@
...
+reach consensus on issues, however the actual decision / voting still happens via the relevant PRs.
+
+### Zulip
+
+The Wild project has a [Zulip instance](http://wild.zulipchat.com/) where discussions can also take
+place.
+
+## Releasing
+
+Releases should be done every 1 to 2 months following the procedure in RELEASING.md.
```
lapla-cogito:
```suggestion
Releases should be done every 1 to 2 months following the procedure in [RELEASING.md](https://github.com/wild-linker/wild/blob/main/RELEASING.md).
```

davidlattimore: Thanks. Done, although I made it a relative link, which I think should work too.

GOVERNANCE.md
```diff
@@ -0,0 +1,100 @@
...
+* [davidlattimore](https://github.com/davidlattimore) (lead maintainer)  
+* [lapla-cogito](https://github.com/lapla-cogito)
+* [marxin](https://github.com/marxin)
+* [mati865](https://github.com/mati865)
+
+### Contributors
+
+People who contribute to Wild can be added to the GitHub org. This can be done when an individual
+requests to be added. They should have already made a few contributions.
+
```
mati865, line 19: Maybe it should require some activity to keep that status? Rust's compiler team does that, but I don't think there are strict rules.

GOVERNANCE.md
```diff
@@ -0,0 +1,100 @@
...
+reach consensus on issues, however the actual decision / voting still happens via the relevant PRs.
+
+### Zulip
+
+The Wild project has a [Zulip instance](http://wild.zulipchat.com/) where discussions can also take
+place.
+
+## Releasing
+
+Releases should be done every 1 to 2 months following the procedure in [RELEASING.md](RELEASING.md).
```
mati865: BTW, we nine days away from two months mark since the last release.
Do we want to stick to that time frame even if there are no significant changes?

GOVERNANCE.md
```diff
@@ -0,0 +1,100 @@
+# Project Governance
+
+This document describes how the Wild project is governed.
+
+## Roles
+
+### Maintainers
```
lapla-cogito, line 7: While this is outside the scope of this PR, I believe we should also update [this part](https://github.com/wild-linker/wild/blob/main/CONTRIBUTING.md#options-for-communicating) (as the explicitly listed maintainer should be the target of communication, and inconsistencies don't seem desirable I think).

GOVERNANCE.md
davidlattimore: Sure. I've added a sentence saying that we might remove people after they've been inactive for more than a year.

GOVERNANCE.md
davidlattimore: Hmm, good point. I've added the word "ideally" and a sentence saying that it's flexible based on circumstances. I agree that we probably don't need to do a release right now.

GOVERNANCE.md
davidlattimore: Could you expand on what you mean? I'm not sure I understand

lapla-cogito: Shouldn't the people listed in the "Maintainers" section match those under "Options for communicating"? You mentioned on Zulip that if someone is "you're happy for people to contact you to talk about linkers / wild," they can be added to this communicating list. Since they're already maintainers, wouldn't that condition naturally be satisfied? In other words, any inconsistency between these two could potentially cause confusion.

davidlattimore: Ah, OK, so you're suggesting changing "Options for communicating" to invite people to directly reach out to any of the maintainers if they'd like to chat about linkers?

lapla-cogito: Yes, pointing to the governance-side description would likely eliminate future sync discrepancies.

PR #1664, opened 2026-03-08
chore: Make CI fail when a new external test passes

lapla-cogito:
Previously, we didn't configure CI to fail when new external tests passed, but since I think few false positives recently, we can now make CI fail when new tests pass. This helps keep the skip list up-to-date.


davidlattimore (approved):
Sounds good to me :)

PR #1666, opened 2026-03-08
fix: Emit error when shared objects required symbols are not provided

mati865:
Noticed this regression (it works fine with Wild 0.8) when building libwebp. Wild doesn't error when undefined symbols in SO are provided by other SO that we don't link. Because of this false-negative, libwebp's build system "thinks" `-lglut` is sufficient instead of `-lglut -lGL`.

Co-authored-by: David Lattimore <dvdlttmr@gmail.com>

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -2791,10 +2791,17 @@ pub(crate) fn process_relocation<'data, 'scope, A: Arch, R: Relocation>(
             }
 
             queue.send_symbol_request::<A>(symbol_id, resources, scope);
+            // If the canonical symbol belongs to a shared object, use the file ID of symbols that
+            // made the reference.
+            let undefined_check_file_id = symbol_db.file_id_for_symbol(if flags.is_dynamic() {
+                local_symbol_id
+            } else {
+                symbol_id
+            });
             if should_emit_undefined_error::<A::Platform>(
                 object.object.symbol(local_sym_index)?,
                 object.file_id,
-                symbol_db.file_id_for_symbol(symbol_id),
+                undefined_check_file_id,
                 flags,
                 args,
                 symbol_db.output_kind,


mati865:
Bisected with jj:
> The first bad revision is: vxoxllzt bdedeb2c refactor: Several small refactorings to help support linker plugins (#1531)

I think I have inspected all the changes in that PR and they are fine (don't change the behaviour), apart from that inverse sorting change.
@davidlattimore I'll need your help here. So far we have either broken regular case or broken LTO case.

Those are `--sym-info` outputs with the revert:
```
Global name `def3` refers to: None
Definitions / references with name `def3`:
  29: Global Undefined: ABSOLUTE | NON_INTERPOSABLE
    def3
    #5 in File #256 (1/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined.702970f99e44e608.o (LOADED)
  37 -> 29: Global Undefined: ABSOLUTE | DYNAMIC | FUNCTION
    def3
    #1 in File #768 (3/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined-4.cdb5bb1d19a8b2c4.wild.so (LOADED)

wild: error: Undefined symbol def3, referenced by
    /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined.702970f99e44e608.o
```
And before it:
```
Global name `def3` refers to: None
Definitions / references with name `def3`:
  29 -> 37: Global Undefined: ABSOLUTE
    def3
    #5 in File #256 (1/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined.702970f99e44e608.o (LOADED)
  37: Global Undefined: ABSOLUTE | DYNAMIC | NON_INTERPOSABLE | FUNCTION | DIRECT
    def3
    #1 in File #768 (3/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined-4.cdb5bb1d19a8b2c4.wild.so (LOADED)
```
If I go back to the revision before #1531, it's slightly different but that might be irrelevant:
```
Definitions / references with name `def3`:
  19: Global Undefined: ABSOLUTE | NON_INTERPOSABLE | DIRECT
    def3
    #5 in File #256 (1/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined.702970f99e44e608.o (LOADED)
  27 -> 19: Global Undefined: ABSOLUTE | DYNAMIC | FUNCTION
    def3
    #1 in File #768 (3/0) /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined-4.cdb5bb1d19a8b2c4.wild.so (LOADED)
wild: error: Undefined symbol def3, referenced by
    /tmp/bisect/home/mateusz/Projects/wild/wild/tests/build/shlib-undefined.c/transitive-host/shlib-undefined.702970f99e44e608.o
```

davidlattimore:
Thanks for tracking that down. I found the test a bit hard to read with the ifdefs, so made myself a stand-alone test. I also wanted to test the effect of different orderings between the shared library and the object that references the symbol, so I moved the reference out to a separate object. Here's a diff of the test I ended up with:

```patch
diff --git a/wild/tests/integration_tests.rs b/wild/tests/integration_tests.rs
index 97f732b7bf..a9481e6f68 100644
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3539,6 +3539,7 @@
         "trivial_asm.s",
         "non-alloc.s",
         "gnu-unique.c",
+        "undef-transitive.c",
         "symbol-versions.c",
         "simple-version-script.c",
         "mixed-verdef-verneed.c",
diff --git a/wild/tests/sources/undef-transitive-1.c b/wild/tests/sources/undef-transitive-1.c
new file mode 100644
index 0000000000..0938236b74
--- /dev/null
+++ b/wild/tests/sources/undef-transitive-1.c
@@ -0,0 +1,5 @@
+int foo(void);
+
+int bar(void) {
+    return foo();
+}
diff --git a/wild/tests/sources/undef-transitive-2.c b/wild/tests/sources/undef-transitive-2.c
new file mode 100644
index 0000000000..f9472b204b
--- /dev/null
+++ b/wild/tests/sources/undef-transitive-2.c
@@ -0,0 +1,7 @@
+//#Shared:undef-transitive-3.c
+
+int foo(void);
+
+int call_foo(void) {
+    return foo();
+}
diff --git a/wild/tests/sources/undef-transitive-3.c b/wild/tests/sources/undef-transitive-3.c
new file mode 100644
index 0000000000..c1be8d0531
--- /dev/null
+++ b/wild/tests/sources/undef-transitive-3.c
@@ -0,0 +1,3 @@
+int foo(void) {
+    return 42;
+}
diff --git a/wild/tests/sources/undef-transitive.c b/wild/tests/sources/undef-transitive.c
new file mode 100644
index 0000000000..576bf18dc8
--- /dev/null
+++ b/wild/tests/sources/undef-transitive.c
@@ -0,0 +1,25 @@
+// Disallow linking against shared object that doesn't provide our required
+// symbol by itself, but depends on another shared object providing that symbol.
+
+//#AbstractConfig:default
+//#Object:runtime.c
+//#Mode:dynamic
+
+//#Config:object-first:default
+//#Object:undef-transitive-1.c
+//#Shared:undef-transitive-2.c
+//#ExpectError:foo
+
+//#Config:shlib-first:default
+//#Shared:undef-transitive-2.c
+//#Object:undef-transitive-1.c
+//#ExpectError:foo
+
+#include "runtime.h"
+
+int bar(void);
+
+void _start(void) {
+  runtime_init();
+  exit_syscall(bar());
+}
```

This fails with config `object-first`, however if I revert the ordering change you identified, then it fails with config `shlib-first`. So the bug was already there, the commit just inverted which case fails.

I think the issue is that if we select a shared object as the canonical owner of an undefined symbol, then we don't emit errors for it, unless the conditions are met for emitting undefined symbol errors for that shared object. One possible fix might be, when we select the canonical source for an undefined symbol to ignore shared objects, and only use a shared object if no regular objects list the symbol.

An alternative, probably better fix might be to look at the code that decides whether to emit an undefined symbol.  In `process_relocation`, we have the following two variables:

```rust
        let local_symbol_id = object.symbol_id_range.input_to_id(local_sym_index);
        let symbol_id = symbol_db.definition(local_symbol_id);
```

So `local_symbol_id` is the symbol ID in the object that made the reference, while `symbol_id` is whatever it resolved to. In the case of an undefined symbol, that would be the selected canonical undefined symbol, which might be from a shared object. When we decide whether to emit the error, it looks like we currently use `symbol_id`. Perhaps we should be using `local_symbol_id` instead.

mati865:
Thanks!

mati865:
Reworking that is definitely beyond me. I barely figured out the current logic. Could you create an issue to track that?

PR #1669, opened 2026-03-09
refactor: Rename Platform to Arch

davidlattimore:
Issue #1538


davidlattimore:
The name "Platform" just didn't seem right.

PR #1670, opened 2026-03-09
Add support for windows linker args

n1ght-hunter:
Requires #1629  to be merged before this can.

This pr adds windows style flags to the args parsing. 
It also moves linux flags to linux modules.
This introduces target-lexicon as a new dep. for parsing -target which follows the llvm target triple spec.

Questions to be answered.
- do we want the new sub modules. args/linux and args/windows. i like having the modules.
- should they be named after the os or the output type. i.e. switch linux to elf and windows to coff or pe
- the args/consts needs more thought. these mainly apply to elf only so perhaps should move removed and consts moved into either args.rs or linux.rs
- do we want to use target-lexicon types more heavily. currently they are mapped to local types i.e. target_lexicon::Architecture -> lib_wild::Architecture. these could be removed from lib_wild and just use target-lexicon but that would mean matching on Architecture will have to match for Architecture's not supported by wild

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -32,6 +32,7 @@ use std::path::PathBuf;
 
 pub mod elf;
 pub mod macho;
+pub mod pe;
 
 use crate::error::Warning;
 use crate::platform;
@@ -117,6 +118,7 @@ impl Args {
         match PlatformKind::host() {
             PlatformKind::Elf => Ok(Args::Elf(elf::ElfArgs::new()?)),
             PlatformKind::MachO => Ok(Args::MachO(macho::MachOArgs::new()?)),
+            PlatformKind::Pe => Ok(Args::Pe(pe::PeArgs::new()?)),
         }
     }
 
@@ -137,6 +139,7 @@ impl Args {
         match self {
             Args::Elf(args) => args.parse(input),
             Args::MachO(args) => args.parse(input),
+            Args::Pe(args) => args.parse(input),
         }
     }
 
@@ -156,15 +159,17 @@ impl Args {
 
     pub(crate) fn common(&self) -> &CommonArgs {
         match self {
-            Args::Elf(elf_args) => &elf_args.common,
-            Args::MachO(macho_args) => &macho_args.common,
+            Args::Elf(args) => &args.common,
+            Args::MachO(args) => &args.common,
+            Args::Pe(args) => &args.common,
         }
     }
 
     pub(crate) fn common_mut(&mut self) -> &mut CommonArgs {
         match self {
-            Args::Elf(elf_args) => &mut elf_args.common,
-            Args::MachO(macho_args) => &mut macho_args.common,
+            Args::Elf(args) => &mut args.common,
+            Args::MachO(args) => &mut args.common,
+            Args::Pe(args) => &mut args.common,
         }
     }
 }
@@ -172,12 +177,15 @@ impl Args {
 enum PlatformKind {
     Elf,
     MachO,
+    Pe,
 }
 
 impl PlatformKind {
     fn host() -> Self {
         if cfg!(target_os = "macos") {
             PlatformKind::MachO
+        } else if cfg!(target_os = "windows") {
+            PlatformKind::Pe
         } else {
             PlatformKind::Elf
         }
@@ -426,6 +434,7 @@ pub struct ThreadPool {
 pub enum Args {
     Elf(elf::ElfArgs),
     MachO(macho::MachOArgs),
+    Pe(pe::PeArgs),
 }
 
 impl std::fmt::Debug for Args {
@@ -433,6 +442,7 @@ impl std::fmt::Debug for Args {
         match self {
             Args::Elf(args) => args.fmt(f),
             Args::MachO(args) => args.fmt(f),
+            Args::Pe(args) => args.fmt(f),
         }
     }
 }
@@ -550,10 +560,46 @@ pub(crate) enum UnresolvedSymbols {
     IgnoreAll,
 }
 
+#[derive(Clone, Copy)]
+enum OptionStyle {
+    /// `--flag=value`, `-flag`, case-sensitive
+    Unix,
+    /// `/FLAG:value`, case-insensitive
+    Windows,
+}
+
+impl OptionStyle {
+    fn strip_option(self, arg: &str) -> Option<&str> {
+        match self {
+            Self::Unix => arg.strip_prefix("--").or(arg.strip_prefix('-')),
+            Self::Windows => arg.strip_prefix('/').or(arg.strip_prefix('-')),
+        }
+    }
+
+    fn has_option_prefix(self, arg: &str) -> bool {
+        match self {
+            Self::Unix => arg.starts_with('-'),
+            Self::Windows => arg.starts_with('/') || arg.starts_with('-'),
+        }
+    }
+
+    fn find_separator(self, stripped: &str) -> Option<usize> {
+        match self {
+            Self::Unix => stripped.find('='),
+            Self::Windows => stripped.find(':'),
+        }
+    }
+
+    fn is_case_insensitive(self) -> bool {
+        matches!(self, Self::Windows)
+    }
+}
+
 struct ArgumentParser<T> {
     options: HashMap<&'static str, OptionHandler<T>>, // Long option lookup
     short_options: HashMap<&'static str, OptionHandler<T>>, // Short option lookup
     prefix_options: HashMap<&'static str, PrefixOptionHandler<T>>, // For options like -L, -l, etc.
+    style: OptionStyle,
 }
 
 impl<T: platform::Args> Default for ArgumentParser<T> {
@@ -569,6 +615,17 @@ impl<T: platform::Args> ArgumentParser<T> {
             options: HashMap::new(),
             short_options: HashMap::new(),
             prefix_options: HashMap::new(),
+            style: OptionStyle::Unix,
+        }
+    }
+
+    #[must_use]
+    fn new_windows() -> Self {
+        Self {
+            options: HashMap::new(),
+            short_options: HashMap::new(),
+            prefix_options: HashMap::new(),
+            style: OptionStyle::Windows,
         }
     }
 
@@ -608,6 +665,22 @@ impl<T: platform::Args> ArgumentParser<T> {
         }
     }
 
+    fn get_option_handler(&self, option_name: &str) -> Option<&OptionHandler<T>> {
+        if self.style.is_case_insensitive() {
+            if let Some(handler) = self.options.get(option_name) {
+                return Some(handler);
+            }
+            for (key, handler) in &self.options {
+                if key.eq_ignore_ascii_case(option_name) {
+                    return Some(handler);
+                }
+            }
+            None
+        } else {
+            self.options.get(option_name)
+        }
+    }
+
     fn handle_argument<S: AsRef<str>, I: Iterator<Item = S>>(
         &self,
         args: &mut T,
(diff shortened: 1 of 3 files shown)


wild/src/main.rs
```diff
@@ -1,30 +1,45 @@
...
-
-    if args.should_fork() {
-        // Safety: We haven't spawned any threads yet.
-        unsafe { libwild::run_in_subprocess(args) };
-    } else {
-        // Run the linker in this process without forking.
-        libwild::run(args)
-    }
-}
+#[cfg(feature = "mimalloc")]
```
davidlattimore: GitHub is showing a diff for lines that look like they haven't changed. This makes me wonder if the file might have been changed to have Windows line endings or something. Could you check?

n1ght-hunter: yup i noticed that as well. I think my jj setup is doing something weird will fix it.

libwild/src/args.rs
```diff
@@ -345,574 +289,267 @@ pub(crate) enum UnresolvedSymbols {
...
-];
-
-impl Default for Args {
+impl<T: Default> Default for Args<T> {
     fn default() -> Self {
         Args {
-            arch: default_target_arch(),
+            // Infrastructure
+            should_fork: true,
+            arch: const { Architecture::from_target_lexicon(target_lexicon::HOST.architecture) },
```
n1ght-hunter: the default should probably be the build target architecture not the build host

libwild/src/args/consts.rs
davidlattimore, line 1: I'd prefer not to have a separate file for consts. We don't do that kind of thing elsewhere. It's a bit like having a module called "traits". It looks like these constants are a mix of ELF-specific constants and constants that should just be in the top-level args module.

libwild/src/args/elf.rs
davidlattimore, line 1: Github is showing the entire file as new, which means we can't see what has changed. I'd say either leave all the linux argument parsing in args.rs, or we could do a PR before this one where we just copy the file and don't edit it.

libwild/src/args.rs
```diff
@@ -1325,1315 +981,509 @@ impl<'a> OptionDeclaration<'a, WithOptionalParam> {
...
-                "none" => {}
-                other => warn_unsupported(&format!("--icf={other}"))?,
+/// Pre-scan CLI arguments to determine the output format and architecture.
+///
+/// Recognizes:
+/// - `--target=<triple>` / `-target=<triple>` / `/TARGET:<triple>` — primary (parsed by target-lexicon)
+/// - `-m <emulation>` — overrides format to ELF when present
+///
+/// Priority: `-m` overrides format from `--target`. Architecture comes from `--target` only.
+pub(crate) fn detect_target(args: &[String]) -> Result<DetectedTarget> {
```
davidlattimore: I think the --target flag might be too much for this PR, especially since it's non-standard. I worry that the discussion might get bogged down. What about splitting that to a later PR and for this one just accepting `--flavor gnu` or `--flavor link` as the first argument like LLD does?

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+        .long("ALLOWISOLATION")
+        .help("/ALLOWISOLATION - Specifies behavior for manifest lookup.")
+        .execute(|_, _| unimplemented_option("/ALLOWISOLATION"));
+    // /APPCONTAINER - Specifies whether the app must run within an appcontainer process environment.
+    parser
+        .declare()
+        .long("APPCONTAINER")
+        .help("/APPCONTAINER - Specifies whether the app must run within an appcontainer process environment.")
+        .execute(|_, _| unimplemented_option("/APPCONTAINER"));
+    // /ARM64XFUNCTIONPADMINX64 - Specifies the minimum number of bytes of padding between x64 functions in ARM64X images. 17.8
```
davidlattimore: The comments seem to duplicate the help text. Also, they seem to be copied from Microsoft. I suspect for copyright reasons, we need to come up with our own descriptions of each flag. For unimplemented flags, it'd be fine to have the help text just say "unimplemented" and we can fill in our own descriptions as we implement them.

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+    use crate::args::InputSpec;
+    use std::path::Path;
+
+    // Example Windows linker flags from Rust compilation
+    const WINDOWS_LINKER_ARGS: [&str; 28] = [
+        "--target=x86_64-pc-windows-msvc",
+        r#"C:\Users\Samuel\AppData\Local\Temp\rustc7RL5Io\symbols.o"#,
+        "dummy.dummy.6cfbe55db138f4b-cgu.0.rcgu.o",
+        "dummy.3wxfnlvokcqcl6j45c8xeicgz.rcgu.o",
+        r#"C:\Users\Samuel\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\x86_64-pc-windows-msvc\lib\libstd-efa6c7783284bd31.rlib"#,
```
davidlattimore: Perhaps edit these lines to make them a bit shorter.

wild/src/main.rs
mati865: @n1ght-hunter is your git configured to `core.autocrlf true`? I think the intentions for this default were good, but in reality this setting just causing a lot of troubles.

n1ght-hunter: i use jj not git. i changed after this to use lf on windows

mati865: JJ still calls git binary underneath for network operations like cloning, fetching and pushing. So, if your git is configured to convert line endings, you will get CRLF in repositories you clone with jj.

n1ght-hunter: its a fully separate setting with jj. anyways i have it fixed now

n1ght-hunter:
im pretty busy with work stuff but will get to this over the weekend

libwild/src/args/pe.rs
```diff
@@ -0,0 +1,1502 @@
...
+    pub(crate) manifest_file: Option<PathBuf>,
+    pub(crate) map_file: Option<PathBuf>,
+    pub(crate) pdb_file: Option<PathBuf>,
+    pub(crate) version: Option<String>,
+    pub(crate) large_address_aware: bool,
+    pub(crate) dynamic_base: bool,
+    pub(crate) nx_compat: bool,
+    pub(crate) terminal_server_aware: bool,
+    pub(crate) high_entropy_va: bool,
+    pub(crate) no_default_libs: Vec<String>,
```
mati865, line 55: This looks like GNU's `-nodefaultlibs` but works very differently. I'd suggest renaming it to `disabled_default_libs` or something like that.

libwild/src/args/pe.rs
```diff
@@ -0,0 +1,1502 @@
...
+    EfiRom,
+    EfiRuntimeDriver,
+}
+
+/// PE/COFF-specific linker arguments. Common fields (output, arch, inputs, etc.)
+/// live on `Args<PeArgs>`. Access them via direct field access on `Args`,
+/// and PE-specific fields are accessible via `Deref`/`DerefMut`.
+#[derive(Debug)]
+pub struct PeArgs {
+    // Windows-specific fields
```
mati865, line 37: This comment doesn't seem useful - looks like it applies to all struct fields.

libwild/src/args/linux.rs
```diff
@@ -0,0 +1,1852 @@
...
+    args.save_dir = SaveDir::new(&input)?;
+
+    let mut input = input();
+
+    let mut modifier_stack = vec![Modifiers::default()];
+
+    if std::env::var(REFERENCE_LINKER_ENV).is_ok() {
+        args.write_layout = true;
+        args.write_trace = true;
+    }
```
mati865: Leftover?

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+        .map(|s| s.parse())
+        .transpose()?;
+
+    if let Some(x) = files_per_group {
+        ensure!(
+            x <= MAX_FILES_PER_GROUP,
+            "{FILES_PER_GROUP_ENV}={x} but maximum is {MAX_FILES_PER_GROUP}"
+        );
+    }
+
```
mati865: Leftover?

libwild/src/args/pe.rs
```diff
@@ -0,0 +1,1502 @@
...
+    if !args.unrecognized_options.is_empty() {
+        let options_list = args.unrecognized_options.join(", ");
+        bail!("unrecognized option(s): {}", options_list);
+    }
+
+    Ok(args)
+}
+
+pub(crate) fn setup_windows_argument_parser() -> ArgumentParser<PeArgs> {
+    // Helper function for unimplemented options
```
mati865: LLMs create unnecessary verbose comments for obvious things like this one. They only needlessly make files longer.

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+        "kernel32.lib",
+        "ntdll.lib",
+        "userenv.lib",
+        "ws2_32.lib",
+        "dbghelp.lib",
+        "/defaultlib:msvcrt",
+        "/NXCOMPAT",
+        "/OUT:dummy.exe",
+        "/DEBUG",
+    ];
```
mati865: Make at least one arg use the alternative prefix `-` and one path using forward slashes.

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+    parser
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::args::InputSpec;
+    use std::path::Path;
+
+    // Example Windows linker flags from Rust compilation
```
mati865: This is misleading, Rust doesn't pass `--target` to the linkers. Using synthetic command line arguments rather than real-world will be easier to deal with.

libwild/src/args/windows.rs
```diff
@@ -0,0 +1,1502 @@
...
+    }
+
+    #[test]
+    fn test_parse_windows_linker_args() {
+        let args =
+            unwrap_pe(crate::args::Args::parse(|| WINDOWS_LINKER_ARGS.into_iter()).unwrap());
+
+        // Test that key flags were parsed correctly
+        assert!(args.debug_info); // /DEBUG flag
+        assert!(args.nx_compat); // /NXCOMPAT flag
```
mati865: Comments like that are not useful.

libwild/Cargo.toml
```diff
@@ -35,6 +35,7 @@ libloading = { workspace = true, optional = true }
 linker-layout = { path = "../linker-layout", version = "0.8.0" }
 linker-trace = { path = "../linker-trace", version = "0.8.0" }
 linker-utils = { path = "../linker-utils", version = "0.8.0" }
+target-lexicon = { workspace = true }
```
mati865: Move it near to other `t..` deps.

Cargo.toml
```diff
@@ -58,6 +58,7 @@ jobserver = "0.1.30"
 leb128 = "0.2.5"
 libc = "0.2.171"
 libloading = "0.9.0"
+target-lexicon = "0.13.5"
```
mati865: Ditto

libwild/src/args.rs
```diff
@@ -345,574 +289,267 @@ pub(crate) enum UnresolvedSymbols {
...
+            case_insensitive: false,
+            has_option_prefix: |arg| arg.starts_with('-'),
+            strip_option: |arg| arg.strip_prefix("--").or(arg.strip_prefix('-')),
+            find_separator: |stripped| stripped.find('='),
         }
     }
 
-    fn declare(&mut self) -> OptionDeclaration<'_, NoParam> {
+    #[must_use]
+    pub fn new_case_insensitive() -> Self {
```
mati865: This is `link.exe` variant, not just case-insensitive.

mati865 (commented):
Few things I noticed during a quick skim through this PR.

n1ght-hunter:
> Few things I noticed during a quick skim through this PR.

thanks so much. will should be able to clean it all up tm. i was hoping to have more time but work had some major issues that took up my time

n1ght-hunter:
@davidlattimore @mati865 i have created a new pr just moving elf args and making args more generic. I have tried to keep the diff pretty minimal. 
https://github.com/wild-linker/wild/pull/1692
My idea is get that in where is easier to review and add nothing windows specific. then will redo this pr after

davidlattimore:
Any plan to return to this, or should we close it?

n1ght-hunter:
> Any plan to return to this, or should we close it?

Yup sorry unexpectedly job change came up and moving to other side of the world. This is still a work in progress for me. If you would like feel free to close and i will open a new one when its better. this should be pretty close to ready if i remember correct i just needed to review it myself. 

This is something i really want to support so i can use wild as a linker for my own language.

marxin:
Given the number of changes recently landed in the argument parsing logic (WASM, macOS ports), and the fact the PR is pretty stall, I am going to close it for now. Feel free to open a new one once you have more spare cycles.

n1ght-hunter:
sorry was going to close it soon anyways. have had some unexpected work changes that im still working out. sorry for the mess around i will make a new pr soonish

PR #1671, opened 2026-03-09
test: Verify that all source files use unix line endings

davidlattimore:
Also check for a newline at the end of the file.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3887,4 +3887,79 @@ mod tidy {
             )
         }
     }
+
+    #[test]
+    fn check_text_files() -> Result {
+        const EXCLUDE_DIR: &[&str] = &[
+            "target",
+            "build",
+            "external_test_suites",
+            "fakes-debug",
+            "fakes",
+        ];
+
+        fn verify_path(path: &Path, problems: &mut Vec<String>) -> Result {
+            if EXCLUDE_DIR.iter().any(|e| path.ends_with(e)) {
+                return Ok(());
+            }
+
+            if path.is_dir() {
+                for entry in read_dir(path)
+                    .with_context(|| format!("Failed to read directory {}", path.display()))?
+                {
+                    let entry = entry?;
+                    let file_name = entry.file_name();
+                    let Some(file_name) = file_name.to_str() else {
+                        continue;
+                    };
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -3887,4 +3887,66 @@ mod tidy {
             )
         }
     }
+
+    #[test]
+    fn check_text_files() -> Result {
+        const EXCLUDE_DIR: &[&str] = &["target", "build", "external_test_suites"];
+        const EXTENSIONS: &[&str] = &[
```
mati865: Do you know if we can somehow tell if file is binary based on magic bytes range or something similar? I'm afraid the whitelist would go out of date without noticing it.

davidlattimore: Sure. I extended the test to check any files that have extensions on the list and also any files that are valid UTF-8, but don't have null bytes. I excluded a couple of json files since they get generated by certain tools, so we don't control them (they're unix format, but don't have trailing newlines). We also now validate that all files with extension on the list are valid UTF-8.

mati865: My intention was not having to deal with keeping extensions list up-to-date by using some heuristic.

wild/tests/integration_tests.rs
```diff
@@ -3887,4 +3887,112 @@ mod tidy {
...
+            "md",
+            "nix",
+            "rs",
+            "s",
+            "toml",
+            "txt",
+            "yml",
+        ];
+
+        const IGNORE_FILES: &[&str] = &["dhat-heap.json", "profile.json"];
```
mati865: They wouldn't have been matched anyway because the list doesn't contain json, but it won't hurt ;)

davidlattimore: They match due to the heuristic

wild/tests/integration_tests.rs
davidlattimore: Sure. I've gotten rid of the list and now just rely on the heuristic.

wild/tests/integration_tests.rs
mati865: Ah, the second comparison with `is_known_text_format` had logic OR instead of AND. You got me there.

PR #1673, opened 2026-03-09
fix: Print the correct message in --sym-info

mati865:
It broke in #1621, only printing the `update` was never called.


mati865:
No worries.

> New software has bugs, right? It's the state of nature of software. Software is absolutely full of bugs all the time.

Richard Jones from Red Hat in https://www.youtube.com/watch?v=aoag03mSuXQ

PR #1674, opened 2026-03-09
linker_script: emit "unclosed comment" error for unterminated block comments

plasmaDestroyer:
When a linker script contains an unclosed `/*` comment, the parser previously returned a generic winnow parse error with no meaningful message.

This adds an `UnclosedComment` variant to `LinkerScriptError` and returns it when `take_until("*/")` fails due to EOF, producing an error message containing "unclosed comment".

Fixes the mold test `linker-script-error.sh`.

### DIFF
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -12,10 +12,13 @@ use winnow::BStr;
 use winnow::Parser as _;
 use winnow::ascii::dec_uint;
 use winnow::ascii::hex_digit1;
+use winnow::ascii::hex_uint;
 use winnow::ascii::multispace0;
 use winnow::combinator::alt;
+use winnow::combinator::delimited;
 use winnow::combinator::eof;
 use winnow::combinator::opt;
+use winnow::combinator::preceded;
 use winnow::combinator::repeat_till;
 use winnow::error::ContextError;
 use winnow::error::FromExternalError;
@@ -69,6 +72,7 @@ pub(crate) enum Command<'a> {
         value: &'a [u8],
     },
     Provide(ProvideSymbolDefinition<'a>),
+    Assert(AssertCommand<'a>),
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -81,6 +85,7 @@ pub(crate) enum SectionCommand<'a> {
     Section(Section<'a>),
     SetLocation(Location),
     Align(Alignment),
+    Assert(AssertCommand<'a>),
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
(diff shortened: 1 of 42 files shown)


libwild/src/linker_script.rs
```diff
@@ -510,6 +515,7 @@ impl std::fmt::Display for LinkerScriptError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             LinkerScriptError::InvalidAlignment => write!(f, "Invalid alignment"),
+            LinkerScriptError::UnclosedComment => write!(f, "unclosed comment"),
```
mati865: Let's make the case consistent:
```suggestion
            LinkerScriptError::UnclosedComment => write!(f, "Unclosed comment"),
```

mati865 (commented):
Please create a test (in addition to mold one) for it. You can let us know if you need help with that.

mati865:
> Fixes the mold test `linker-script-error.sh`.

Looks like this test was already passing before your changes.

plasmaDestroyer:
> > Fixes the mold test `linker-script-error.sh`.
> 
> Looks like this test was already passing before your changes.

It seems the test is listed in `mold_skip_tests.toml` under `skipped_groups.ignore` with reason `"We ignore these tests for some reasons"`, so it may have been getting skipped rather than passing. Running with WILD_IGNORE_SKIP=mold confirmed it failed before this fix.

Regarding the capitalization, the mold test greps for the exact string `'unclosed comment'` in lowercase, so capitalizing it would cause the test to fail, that is why i kept it that way.

i will create and add the test you asked for shortly.

plasmaDestroyer:
> You can let us know if you need help with that.

hey, I've written a test based on the pattern of the tests i saw, but it doesn't seem to be getting discovered by the test runner. Could you help me with this?

davidlattimore:
> > You can let us know if you need help with that.
> 
> hey, I've written a test based on the pattern of the tests i saw, but it doesn't seem to be getting discovered by the test runner. Could you help me with this?

It doesn't look like you've pushed the change that includes the test, so it's hard to say, but at a guess, I'd say you might have not known to add the test to the list of tests in `integration_tests.rs`.

plasmaDestroyer:
> 
> It doesn't look like you've pushed the change that includes the test, so it's hard to say, but at a guess, I'd say you might have not known to add the test to the list of tests in `integration_tests.rs`.

yeah i forgot to push that but got it now, thanks for the feedback.

libwild/src/linker_script.rs
davidlattimore: I agree. The mold test can have a comment added next to it - "Different message formats". There's plenty of other tests that we do that for.

plasmaDestroyer: capitalized the error message, updated the integration test to match, and resolved the merge conflict.

Issue #1679, opened 2026-03-10
COMMON symbols left undefined in shared libraries instead of being allocated into BSS


alexandergunnarson:
## Summary

When linking a shared library that contains object files with COMMON symbols (e.g. `.gomp_critical_user_.var` emitted by Clang for `#pragma omp critical`), wild leaves them as undefined (`U`) in the output instead of allocating them into BSS (`B`). Both `lld` and GNU `ld` correctly handle this case.

## Version

wild 0.8.0 (`d789d5d6716de3f665ff3da0c53d02dc65e49e3b`)

Also verified the bug is present on current `main` (`62a18be`).

## Reproduction

```c
// test.c
#include <omp.h>
#include <stdio.h>

void do_work(int* data, int n) {
    int error_count = 0;
    #pragma omp parallel for reduction(+:error_count)
    for (int i = 0; i < n; i++) {
        if (data[i] < 0) {
            #pragma omp critical
            {
                fprintf(stderr, "Error at %d\n", i);
                error_count++;
            }
        }
    }
}
```

```sh
# Compile the object file
clang -c -fopenmp=libomp test.c -o test.o

# Confirm the COMMON symbols exist in the object file
$ nm test.o | grep gomp
0000000000000020 C .gomp_critical_user_.reduction.var
0000000000000020 C .gomp_critical_user_.var

# Link with lld — COMMON symbols correctly become BSS
$ clang -shared -fopenmp=libomp test.o -o test_lld.so -fuse-ld=lld
$ nm -D test_lld.so | grep gomp
0000000000003fb8 B .gomp_critical_user_.reduction.var
0000000000003f98 B .gomp_critical_user_.var

# Link with wild — COMMON symbols left undefined
$ clang -shared -fopenmp=libomp test.o -o test_wild.so -fuse-ld=wild
$ nm -D test_wild.so | grep gomp
                 U .gomp_critical_user_.reduction.var
                 U .gomp_critical_user_.var
```

## Expected behavior

COMMON symbols (`C`) in input object files should be allocated into BSS and appear as defined (`B`) in the output shared library, matching `lld` and GNU `ld` behavior.

## Actual behavior

COMMON symbols remain undefined (`U`) in the output shared library, causing `dlopen` to fail at runtime:

```
undefined symbol: .gomp_critical_user_.var
```

## Impact

[...]

PR #1685, opened 2026-03-12
fix: COMMON/ABS symbols are now defined in .dynsym

ostylk:
Fixes https://github.com/wild-linker/wild/issues/1679 .

In https://github.com/ostylk/wild/blob/e9f0b316e7ac2157a53bc66e307b6c77e19ce691/libwild/src/elf_writer.rs#L3886 every dynamic symbol got set to zero if it was not associated to a "concrete" section (meaning shndx < SHN_LORESERVE).
However, every COMMON symbol is in a virtual section SHN_COMMON with an index higher than SHN_LORESERVE.
So all COMMON symbols got written as undefined into the .dynsym section.

The fix is to introduce a small special case like in https://github.com/ostylk/wild/blob/e9f0b316e7ac2157a53bc66e307b6c77e19ce691/libwild/src/elf_writer.rs#L1597 .
Now COMMON symbols are correctly written into the .dynsym section with the BSS/TBSS section.

This PR is still a draft because I have two open questions and I am unsure about them:

1. The test provided in https://github.com/wild-linker/wild/issues/1679 by @davidlattimore reproduces the issues (as in: the shared library has undefined dynamic symbol) but it doesn't fail when I do ``cargo test``. Is my dev environment weird or does the test allow undefined symbols?
2. https://github.com/ostylk/wild/blob/e9f0b316e7ac2157a53bc66e307b6c77e19ce691/libwild/src/elf_writer.rs#L1597 also handles ``SHN_ABS``. Do we also need absolute symbols in .dynsym? If yes, that would also need to be added here. (But I'm still researching that question, just wanted to get the PR out).

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -3884,6 +3884,15 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
             .with_context(|| {
                 format!("Failed to copy dynamic {}", layout.symbol_debug(symbol_id))
             })?;
+    } else if sym.is_absolute(LittleEndian) {
+        dynamic_symbol_writer
+            .copy_absolute_symbol(sym, name, ValueFlags::empty())
+            .with_context(|| {
+                format!(
+                    "Failed to absolute {}",
+                    layout.symbol_debug(sym_def.symbol_id)
+                )
+            })?;
     } else {
         dynamic_symbol_writer
             .copy_symbol_shndx(sym, name, 0, 0, ValueFlags::empty())
--- a/wild/tests/sources/absolute-symbol.s
+++ b/wild/tests/sources/absolute-symbol.s
@@ -1,0 +1,17 @@
+/* For some reason, GAS on riscv64 does not support '//' comments.
+//#LinkArgs:-shared -z now
+//#RunEnabled:false
+//#DiffIgnore:section.got
+//#ExpectSym:abs_sym address=0xCAFECAFE
+//#ExpectDynSym:abs_sym address=0xCAFECAFE
+
+// TODO: checkout those differences later
+//#DiffIgnore:segment.RISCV_ATTRIBUTES.alignment
+//#DiffIgnore:segment.RISCV_ATTRIBUTES.flags
+//#DiffIgnore:riscv_attributes..riscv.attributes
+//#DiffIgnore:riscv_attributes.arch
+//#DiffIgnore:riscv_attributes.stack_align
+*/
+
+.global abs_sym
+.set abs_sym, 0xCAFECAFE
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3548,6 +3548,7 @@ fn integration_test(
         "gnu-unique.c",
         "symbol-versions.c",
         "common-shared.c",
+        "absolute-symbol.s",
         "simple-version-script.c",
         "mixed-verdef-verneed.c",
         "copy-relocations.c",


libwild/src/elf_writer.rs
```diff
@@ -3857,6 +3857,36 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
...
+        let symbol_id = sym_def.symbol_id;
+        let resolution = layout.local_symbol_resolution(symbol_id).with_context(|| {
+            format!(
+                "Tried to write dynamic symbol definition without a resolution: {}",
+                layout.symbol_debug(symbol_id)
+            )
+        })?;
+
+        // As common symbols are denoted by setting shndx=0xfff2 which is a special section,
+        // we need to put them manually into BSS/TBBS sections depending on whether they are thread
```
lapla-cogito: typo?
```suggestion
        // we need to put them manually into BSS/TBSS sections depending on whether they are thread
```

libwild/src/elf_writer.rs
```diff
@@ -3857,6 +3857,36 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
...
+            output_section_id::BSS
+        };
+        let section_id = layout.output_sections.primary_output_section(section_id);
+
+        dynamic_symbol_writer
+            .copy_symbol(
+                sym,
+                name,
+                section_id,
+                resolution.value(),
```
lapla-cogito: For TLS COMMON symbols, I think we need to subtract `tls_start_address()` from `resolution.value()` (like [this](https://github.com/wild-linker/wild/blob/9027437a8776afea1a1f0840fcda3fd6895f6f55/libwild/src/elf_writer.rs#L1634)).

In fact, if you run code like the following:
```c
//#Object:tls-common-shared-1.c
//#LinkArgs:-shared -z now
//#RunEnabled:false

extern __thread int tvar[];
__thread int tvar[10] __attribute__((common));
```

```c
extern __thread int tvar[];
__thread int tvar[100] __attribute__((common));
```

I get the following output (note that the value of tvar should be 0):

```
（▰╹◡╹）❯  readelf --dyn-syms wild/tests/build/tls-common-shared.c/default-host/tls-common-shared.c.*ld

File: wild/tests/build/tls-common-shared.c/default-host/tls-common-shared.c.ld

Symbol table '.dynsym' contains 2 entries:
     1: 0000000000000000   400 TLS     GLOBAL DEFAULT    6 tvar

File: wild/tests/build/tls-common-shared.c/default-host/tls-common-shared.c.wild

Symbol table '.dynsym' contains 2 entries:
     1: 0000000000002610   400 TLS     GLOBAL DEFAULT    6 tvar
```

ostylk: Done, I added a check in the test to make sure the tls var is at address 0. I hope that is sufficient and no linker ever tries to put something before that symbol if only one tls var is used.

ostylk:
> 1. Probably `run_all_diffs` is not configured in the configuration file described at https://github.com/wild-linker/wild/blob/main/CONTRIBUTING.md#configuration-file-for-tests. I recommend running `cp test-config-ci.toml test-config.toml`.

Yeah, that solved it. Thanks.

> 2. I think we also should implement that.

I added this plus a test case. Is there an easier way to emit an ABS symbol? I used inline assembly.

> Additionally, CI appears to be showing differences for `section.got`, which should be ignored. You can resolve this by adding directives like `//#DiffIgnore:section.got`.

Done.

Thank you for your comments. I marked this PR as ready.

libwild/src/elf_writer.rs
```diff
@@ -3857,6 +3857,42 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
...
+        let resolution = layout.local_symbol_resolution(symbol_id).with_context(|| {
+            format!(
+                "Tried to write dynamic symbol definition without a resolution: {}",
+                layout.symbol_debug(symbol_id)
+            )
+        })?;
+
+        let mut sym_value = resolution.value();
+
+        // As common symbols are denoted by setting shndx=0xfff2 which is a special section,
```
mati865: `shndx=0xfff2` is just SHN_COMMON which is far easier to understand.

libwild/src/elf_writer.rs
```diff
@@ -3857,6 +3857,42 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
...
+        } else {
+            output_section_id::BSS
+        };
+        let section_id = layout.output_sections.primary_output_section(section_id);
+
+        dynamic_symbol_writer
+            .copy_symbol(sym, name, section_id, sym_value, ValueFlags::empty())
+            .with_context(|| {
+                format!("Failed to copy dynamic {}", layout.symbol_debug(symbol_id))
+            })?;
```
mati865, line 3886: This branch seems to duplicate a lot of code from `write_symbols`. Would it be possible to extract common parts?

ostylk: Hmm, this refactoring looked a bit tricky for me. I attempted it here: https://github.com/ostylk/wild/commit/00abc82a33ec8c1e87d03f64a2bd2a03b1d08da0 . I tried extracting the section_id/symbol_value resolution logic.

Do you think this approach looks good? I am currently a bit unsure (as some special case logic is now also applied for COMMON symbols but that may be desirable?) so I didn't include that commit in this PR.
At least the unit tests pass.

davidlattimore: It seems to me that there's probably already a bit of duplication between writing of dynamic symbols and writing of debug symbols even prior to this PR. Given that, if we're going to attempt to deduplicate it, it'd probably make sense to do that as a separate PR. So it's fine with me to leave this bit as-is for now.

mati865: Thanks for the attempt. Since this proves hard we can do it separately.

davidlattimore:
Those riscv diffs are fine to `DiffIgnore`. They probably require some investigation, but aren't really related to this PR. If you want to try running the riscv tests locally, there's instructions in `CONTRIBUTING.md`.

ostylk:
Then we'll postpone the deduplication of symbol table logic. I ignored the riscv diffs, but added a TODO.

CI should hopefully pass now.

wild/tests/sources/absolute-symbol.s
lapla-cogito, line 1: You can simply write this test as an assembly file (there are other test files with `*.s`).

ostylk: done

lapla-cogito: Assembly tests that also run on RISC-V must enclose comments in `/* */` (see https://github.com/wild-linker/wild/blob/main/wild/tests/sources/exclude-section.s).

ostylk: Well that's a surprising gimmick :)

Issue #1508, opened 2026-02-04
Babl fails to link with Wild due to global symbols


mati865:
Babl's build system (tested version 0.1.118) fails with the error:

<details>

```
[165/277] /usr/bin/python3 /var/tmp/portage/media-libs/babl-0.1.118/work/babl-0.1.118/tools/defcheck.py babl/babl.def
FAILED: [code=255] check-def-files
/usr/bin/python3 /var/tmp/portage/media-libs/babl-0.1.118/work/babl-0.1.118/tools/defcheck.py babl/babl.def

Problem found in babl/babl.def
  the following symbols are in the library,
  but are not listed in the .def-file:
     + add_check_overflow
     + babl_base_destroy_generic
     + babl_base_destroy_x86_64_v2
     + babl_base_init
     + babl_base_init_generic
     + babl_base_init_x86_64_v2
     + babl_base_model_cmyk_generic
     + babl_base_model_cmyk_x86_64_v2
     + babl_base_model_gray_generic
     + babl_base_model_gray_x86_64_v2
     + babl_base_model_rgb_generic
     + babl_base_model_rgb_x86_64_v2
     + babl_base_model_ycbcr_generic
     + babl_base_model_ycbcr_x86_64_v2
     + babl_base_type_float_generic
     + babl_base_type_float_x86_64_v2
     + babl_base_type_half_generic
     + babl_base_type_half_x86_64_v2
     + babl_base_type_u15_generic
     + babl_base_type_u15_x86_64_v2
     + babl_base_type_u16_generic
     + babl_base_type_u16_x86_64_v2
     + babl_base_type_u32_generic
     + babl_base_type_u32_x86_64_v2
     + babl_base_type_u8_generic
     + babl_base_type_u8_x86_64_v2
     + babl_calloc
     + babl_component_class_for_each
     + babl_component_db
     + babl_component_from_id
     + babl_conversion
     + babl_conversion_cost
     + babl_conversion_create_name
     + babl_conversion_db
     + babl_conversion_error
     + babl_conversion_find
     + babl_conversion_from_id
     + babl_core_init
     + babl_cpu_accel_set_use
     + babl_db_count
     + babl_db_exist
     + babl_db_insert
     + babl_die
     + babl_doc
     + babl_dup
     + babl_extension
     + babl_extension_base
     + babl_extension_class_for_each
     + babl_extension_db
     + babl_extension_deinit
     + babl_extension_from_id
     + babl_extension_load_dir_list
[...]

PR #1686, opened 2026-03-12
fix: Hide linker-defined symbols in shared libraries

Blazearth:
Addresses https://github.com/wild-linker/wild/issues/1508#issuecomment-4016404588

GNU ld suppresses linker-defined layout symbols (etext, edata, end, etc.) when producing shared objects. Wild previously emitted them as GLOBAL DEFAULT, causing them to appear as exported ABI symbols in .dynsym. This change hides them when linking shared libraries so they remain local in .symtab but are not exported.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -629,27 +629,23 @@ impl platform::Platform for Elf {
             .hide();
 
         // GNU ld doesn't emit these symbols in shared libraries, so we hide them
-        if output_kind.is_shared_object() {
-            symbols.section_end(output_section_id::TEXT, "etext").hide();
-            symbols
-                .section_end(output_section_id::TEXT, "_etext")
-                .hide();
-            symbols
-                .section_end(output_section_id::TEXT, "__etext")
-                .hide();
-        } else {
-            symbols.section_end(output_section_id::TEXT, "etext");
-            symbols.section_end(output_section_id::TEXT, "_etext");
-            symbols.section_end(output_section_id::TEXT, "__etext");
-        }
+        let hidden = output_kind.is_shared_object();
+        symbols
+            .section_end(output_section_id::TEXT, "etext")
+            .set_hidden(hidden);
+        symbols
+            .section_end(output_section_id::TEXT, "_etext")
+            .set_hidden(hidden);
+        symbols
+            .section_end(output_section_id::TEXT, "__etext")
+            .set_hidden(hidden);
 
-        if output_kind.is_shared_object() {
-            symbols.section_end(output_section_id::BSS, "end").hide();
-            symbols.section_end(output_section_id::BSS, "_end").hide();
-        } else {
-            symbols.section_end(output_section_id::BSS, "end");
-            symbols.section_end(output_section_id::BSS, "_end");
-        }
+        symbols
+            .section_end(output_section_id::BSS, "end")
+            .set_hidden(hidden);
+        symbols
+            .section_end(output_section_id::BSS, "_end")
+            .set_hidden(hidden);
         symbols.section_end(output_section_id::BSS, "__end").hide();
 
         // TODO: define the symbol only on RISC-V target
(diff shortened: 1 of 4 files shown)


libwild/src/elf.rs
```diff
@@ -628,12 +628,28 @@ impl platform::Platform for Elf {
             .section_group_end(output_section_id::FINI_ARRAY, "__fini_array_end")
             .hide();
 
-        symbols.section_end(output_section_id::TEXT, "etext");
-        symbols.section_end(output_section_id::TEXT, "_etext");
-        symbols.section_end(output_section_id::TEXT, "__etext");
+        // GNU ld doesn't emit these symbols in shared libraries, so we hide them
+        if output_kind.is_shared_object() {
+            symbols.section_end(output_section_id::TEXT, "etext").hide();
```
davidlattimore: I think it might read slightly nicer if we had a `set_hidden` function that took a `bool`, then we could do something like the following and avoid repeating all the symbols:

```rust
symbols.section_end(output_section_id::TEXT, "etext").set_hidden(hidden);`
```

mati865:
Those were symtab entries:
```
4d89610e445f ~/wild-proxy # readelf -Ws /tmp/x86-64-v3-gggl-lies.so.wild | grep -E 'edata|end|etext|:$'
Symbol table '.dynsym' contains 12 entries:
Symbol table '.symtab' contains 65 entries:
     3: 00000000000080f8     0 NOTYPE  LOCAL  DEFAULT   16 __init_array_end
     5: 0000000000008100     0 NOTYPE  LOCAL  DEFAULT   17 __fini_array_end
     6: 00000000000093c9     0 NOTYPE  LOCAL  DEFAULT   23 __end
    51: 00000000000070c0     0 NOTYPE  GLOBAL DEFAULT   13 etext
    52: 00000000000070c0     0 NOTYPE  GLOBAL DEFAULT   13 _etext
    53: 00000000000070c0     0 NOTYPE  GLOBAL DEFAULT   13 __etext
    54: 00000000000093c9     0 NOTYPE  GLOBAL DEFAULT   23 end
    55: 00000000000093c9     0 NOTYPE  GLOBAL DEFAULT   23 _end
    57: 00000000000093c8     0 NOTYPE  GLOBAL DEFAULT   21 edata
    58: 00000000000093c8     0 NOTYPE  GLOBAL DEFAULT   21 _edata
```

After some testing I see that GNU ld always outputs `_edata` and `_end` into symtab for executables but GCs them for shared objects. LLD always GCs them.
GC in this case means removal of the symbols which are not referenced by any exported symbol.

Secondly, Wild never puts them into `.dynsym` and with this change they will become local even when referenced.
Consider this example:
```
❯ bat -p a.c
extern int _edata;

int foo() {
  return _edata;
}

❯ gcc a.c -fpic -shared -o a.ld

❯ gcc a.c -fpic -shared -o a.lld -fuse-ld=lld

❯ gcc a.c -fpic -shared -o a.wild -B ~/Projects/wild/fakes-debug/

❯ readelf -Ws ./a.ld | rg 'edata$|end$|etext$|:$'
Symbol table '.dynsym' contains 7 entries:
     5: 0000000000004008     0 NOTYPE  GLOBAL DEFAULT   19 _edata
Symbol table '.symtab' contains 26 entries:
    21: 0000000000004008     0 NOTYPE  GLOBAL DEFAULT   19 _edata

❯ readelf -Ws ./a.lld | rg 'edata$|end$|etext$|:$'
Symbol table '.dynsym' contains 7 entries:
     6: 00000000000037f0     0 NOTYPE  GLOBAL DEFAULT   20 _edata
Symbol table '.symtab' contains 23 entries:
    22: 00000000000037f0     0 NOTYPE  GLOBAL DEFAULT   20 _edata

❯ readelf -Ws ./a.wild | rg 'edata$|end$|etext$|:$'
Symbol table '.dynsym' contains 6 entries:
Symbol table '.symtab' contains 33 entries:
     3: 0000000000002d18     0 NOTYPE  LOCAL  DEFAULT   14 __init_array_end
     5: 0000000000002d20     0 NOTYPE  LOCAL  DEFAULT   15 __fini_array_end
     6: 0000000000001cdf     0 NOTYPE  LOCAL  DEFAULT   11 etext
     7: 0000000000001cdf     0 NOTYPE  LOCAL  DEFAULT   11 _etext
     8: 0000000000001cdf     0 NOTYPE  LOCAL  DEFAULT   11 __etext
     9: 0000000000003f99     0 NOTYPE  LOCAL  DEFAULT   21 end
    10: 0000000000003f99     0 NOTYPE  LOCAL  DEFAULT   21 _end
    11: 0000000000003f99     0 NOTYPE  LOCAL  DEFAULT   21 __end
    12: 0000000000003f98     0 NOTYPE  LOCAL  DEFAULT   19 edata
    13: 0000000000003f98     0 NOTYPE  LOCAL  DEFAULT   19 _edata
```
So, we might want to settle on least bad workaround for now, and rework the concept later.

Blazearth:
Fair enough i will add an intergration test that links a shared object and see linker defined symbols dont appear in .dynsym unless they are referenced and  i will also look into extending the symbol assertions to support checking visibility if needed.
@mati865 
Thanks for the detailed explanation and example.
You're right — with the current change these symbols become LOCAL in shared objects even if referenced, which differs from GNU ld / LLD. For now this mainly avoids leaking unused layout symbols into the exported ABI (which caused the babl failure). A more complete fix would likely involve proper GC of these linker-defined symbols.

Issue #1687, opened 2026-03-13
Add support for loading and running shared objects to test framework


davidlattimore:
Most of our tests currently produce executables, sometimes dynamically linked against shared objects. We have a few tests where we just produce a shared object and currently we always disable execution with `//#RunEnabled:false`. It'd be good to be able to write tests that produce a shared object and then have the test framework dlopen that shared object, find a particular function and call it.

dlopen can be done with the `libloading` crate, which we already have a dependency on.

Two possible design options. We could change the `RunEnabled` command, which currently accepts `true` or `false` to accept a third option `open`. If that's set, then when we run it, we do so via dlopen and dlsym. The symbol to load can just be hard-coded. e.g. `entry` or we could use `_start` the same as executables use.

Alternatively, we could leave `RunEnabled` alone and have a command that sets the name of the symbol to load. If that's set, then when we we run, we'd do so via dlopen and dlsym.

I don't have a preference for which we do - but probably don't do both.

If you'd like to work on this, please let us know so that we can assign it to you.

ostylk:
I'd like to work on it.

Should we expect the function that is called to return an integer to indicate a status code?

PR #1690, opened 2026-03-13
test: integration tests can now call dynamic libraries

ostylk:
Implements https://github.com/wild-linker/wild/issues/1687 .

Using `//#RunDynSym:foo` the output binary is treated as a shared library and loaded into the current test process.
The function `foo` needs to return an integer and it succeeds if it returns `42`.

Cross compiled libraries cannot be tested that way.

Is there one test in particular where you want to use this feature? I enabled it for `shared.c` and called `foo()`.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -94,6 +94,9 @@
 //! Arch:{arch1}[,{arch2}...] Specifies which architectures this test should be run with. Defaults
 //! to all supported architectures.
 //!
+//! SkipArch:{arch1}[,{arch2}...] The opposite of Arch. Tests for all architectures that aren't in
+//! the supplied list.
+//!
 //! RequiresGlibc:{bool} Defaults to false. Set to true to disable this test if we're running on a
 //! system without glibc.
 //!
@@ -409,6 +412,10 @@ impl Architecture {
 
     fn get_cross_sysroot_path(&self) -> String {
         format!("/usr/{self}-linux-gnu")
+    }
+
+    fn parse(name: &str) -> Result<Architecture> {
+        Self::from_str(name).map_err(|_| error!("Unknown arch '{name}'"))
     }
 }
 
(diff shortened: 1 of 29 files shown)


wild/tests/sources/shared.c
lapla-cogito: You can remove these lines.

wild/tests/integration_tests.rs
lapla-cogito: Currently, specifying `//#Mode:dynamic` implicitly sets `//#RunEnabled:false` (introduced in #1033), so I think this needs to be corrected.

ostylk: I think that has been reverted in https://github.com/wild-linker/wild/commit/b88576e77112b5a4ac2f56ca715de7dd8159b1b6 ?

davidlattimore:
Thanks for doing this!

From a quick look at the test failures, there seem to be two causes. Tests fail on Alpine. Alpine uses musl libc. Although musl libc supports dynamic linking, it's possible that this isn't happening. dlopen doesn't work with statically linked binaries. Probably the easiest option would be to skip the test if `RunDynSym` is set and `is_musl_used` returns true.

The other failures look like they're from trying to use qemu to run the binary. It'd be possible to support this, but we'd need to have the test framework compile an executable for the target arch where that executable loaded the shared object, ran it, then exited with the appropriate code. This could perhaps be done as a follow-up. For now, I'd probably suggest just skipping cross-compilation tests when `RunDynSym` is set.

ostylk:
> From a quick look at the test failures, there seem to be two causes. Tests fail on Alpine. Alpine uses musl libc. Although musl libc supports dynamic linking, it's possible that this isn't happening. dlopen doesn't work with statically linked binaries. Probably the easiest option would be to skip the test if `RunDynSym` is set and `is_musl_used` returns true.
> 
> The other failures look like they're from trying to use qemu to run the binary. It'd be possible to support this, but we'd need to have the test framework compile an executable for the target arch where that executable loaded the shared object, ran it, then exited with the appropriate code. This could perhaps be done as a follow-up. For now, I'd probably suggest just skipping cross-compilation tests when `RunDynSym` is set.

Thank you for looking into my CI failures. I skipped the test runs for now when cross compiling or ``is_musl_used`` is true.

Seems like providing such a small "call wrapper binary" would fix both of those issues as the binary could be linked dynamically even with musl. If the binary expects two arguments (shared library and function name to call) it would only need to be compiled once for each arch and could be reused for each test. But I'll postpone it for now.

davidlattimore:
> Is there one test in particular where you want to use this feature? I enabled it for `shared.c` and called `foo()`.

That's good for now. We can enable it in other tests and in particular in new tests later.

My intended to merge your change before #1743, but forgot, so this PR got a merge conflict. Since I felt bad about unnecessarily creating a merge conflict, I resolved it through the github UI.

PR #1692, opened 2026-03-15
refactor: move and splitup args.rs into args.rs and args/elf.rs in prep to suppport more targets

n1ght-hunter:
move elf specific argument code into args/elf.rs.
make argument parser and dependent code generic over T so it can be used with ElfArgs but also other args later. 
create shared Args to be used for common args between targets.
create Target args enum to hold any target args and branch at runtime

### DIFF
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -347,8 +347,8 @@ impl LoadedPlugin {
 
         let output_kind = if args.should_output_executable {
             match args.relocation_model {
-                crate::args::RelocationModel::NonRelocatable => OutputFileType::Exec,
-                crate::args::RelocationModel::Relocatable => OutputFileType::Pie,
+                crate::args::elf::RelocationModel::NonRelocatable => OutputFileType::Exec,
+                crate::args::elf::RelocationModel::Relocatable => OutputFileType::Pie,
             }
         } else {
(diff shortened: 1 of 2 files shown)


n1ght-hunter:
> Thanks for reducing the scope. I think we should be able to merge this once the CI failures are fixed. Sorry for the delay running CI. I was away on the weekend. Once you've had your first PR merged, subsequent PRs shouldn't need approval before CI runs.

No worries. Hope you had a good weekend away :). 
ci should be fixed now

davidlattimore:
Have you managed to reproduce the failures locally? Doing so is likely to result in faster iteration. If you haven't got a `test-config.toml`, then copying `test-config-ci.toml` is a good place to start. There's also further details in `CONTRIBUTING.md` about setting up testing for aarch64, riscv64 etc in case you're hitting issues with those.

n1ght-hunter:
> Have you managed to reproduce the failures locally? Doing so is likely to result in faster iteration. If you haven't got a `test-config.toml`, then copying `test-config-ci.toml` is a good place to start. There's also further details in `CONTRIBUTING.md` about setting up testing for aarch64, riscv64 etc in case you're hitting issues with those.

ya some of them i have got locally. tests don't work for me as I'm on windows at the moment. i should probably setup wsl for building the tests in

PR #1694, opened 2026-03-15
chore: sort dependencies lists

mati865:


libwild/Cargo.toml
```diff
@@ -8,17 +8,20 @@ rust-version.workspace = true
 edition.workspace = true
 
 [dependencies]
+linker-layout = { path = "../linker-layout", version = "0.8.0" }
+linker-trace = { path = "../linker-trace", version = "0.8.0" }
+linker-utils = { path = "../linker-utils", version = "0.8.0" }
```
mati865, line 13: `linker-diff/Cargo.toml` puts them first, so made this file follow the same pattern.

davidlattimore (approved):

PR #1695, opened 2026-03-15
feat: evaluate linker script ASSERT commands

Blazearth:
Implements #1607 
Added the implementation for ASSERT cmd along with the integration tests.
ALIGN(n) aligns 0 since the location counter is not threaded through evaluation.
If an ASSERT expression uses a symbol, Wild currently doesn’t know how to evaluate it yet, instead of failing the build it prints a warning and ignores that ASSERT.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -14,6 +14,13 @@ use crate::output_section_id::SectionName;
 use crate::output_section_map::OutputSectionMap;
 use crate::platform::Platform;
 
+/// Compute 1-based line number by counting newlines before `remainder` in `file_bytes`.
+fn line_number(file_bytes: &[u8], remainder: &[u8]) -> u32 {
+    let parsed_len = file_bytes.len().saturating_sub(remainder.len());
+    let consumed = &file_bytes[..parsed_len];
+    consumed.iter().filter(|&&b| b == b'\n').count() as u32 + 1
+}
+
 /// Evaluate all ASSERT commands from all processed linker scripts.
 /// Must be called after layout is complete so section sizes/addresses are known.
 pub(crate) fn evaluate_assertions<'data, P: Platform>(
@@ -28,15 +35,16 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
         for script in scripts {
             let parsed = &script.parsed;
             for assertion in &parsed.assertions {
+                let line = line_number(parsed.file_bytes, assertion.remainder);
                 let result =
                     evaluate_expression(&assertion.expression, section_layouts, output_sections)
                         .with_context(|| {
-                            format!("Failed to evaluate assertion from {}", parsed.input)
+                            format!("{}:{}: Failed to evaluate ASSERT", parsed.input, line)
                         })?;
 
                 if result == 0 {
                     let msg = String::from_utf8_lossy(assertion.message);
-                    bail!("{msg}\n  in assertion from {}", parsed.input);
+                    bail!("{}:{}: {msg}", parsed.input, line);
                 }
             }
         }
@@ -474,6 +482,7 @@ mod tests {
                 input: crate::input_data::InputRef { file, entry: None },
                 symbol_defs: Vec::new(),
                 assertions,
+                file_bytes: b"",
             },
             symbol_id_range: SymbolIdRange::empty(),
             file_id: FileId::new(0, 0),
@@ -490,6 +499,7 @@ mod tests {
                 Box::new(Expression::Number(1)),
             ),
             message: b"should pass",
+            remainder: b"",
         }]);
         assert!(evaluate_assertions::<Elf>(&[group], &layouts, &sections).is_ok());
     }
@@ -500,6 +510,7 @@ mod tests {
         let group = make_group(vec![AssertCommand {
             expression: Expression::Number(0),
             message: b"intentional failure",
+            remainder: b"",
         }]);
         let err = evaluate_assertions::<Elf>(&[group], &layouts, &sections).unwrap_err();
         assert!(err.to_string().contains("intentional failure"));
--- a/libwild/src/linker_script.rs
+++ b/libwild/src/linker_script.rs
@@ -120,11 +120,22 @@ pub(crate) struct ProvideSymbolDefinition<'a> {
     pub(crate) hidden: bool,
 }
 
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, Clone)]
 pub(crate) struct AssertCommand<'a> {
     pub(crate) expression: Expression<'a>,
     pub(crate) message: &'a [u8],
+    /// Remaining input at the point this ASSERT was parsed. Used to lazily compute
+    /// the line number only when an error occurs.
+    pub(crate) remainder: &'a [u8],
 }
+
+impl<'a> PartialEq for AssertCommand<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        self.expression == other.expression && self.message == other.message
+    }
+}
+
+impl<'a> Eq for AssertCommand<'a> {}
 
 /// Represents a parsed expression in linker scripts (e.g., in ASSERT commands).
 ///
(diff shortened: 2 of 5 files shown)


libwild/src/assert_eval.rs
```diff
@@ -0,0 +1,488 @@
...
+            Ok(1)
+        }
+
+        Expression::Add(l, r) => Ok(eval!(l)?.wrapping_add(eval!(r)?)),
+        Expression::Subtract(l, r) => Ok(eval!(l)?.wrapping_sub(eval!(r)?)),
+        Expression::Multiply(l, r) => Ok(eval!(l)?.wrapping_mul(eval!(r)?)),
+        Expression::Divide(l, r) => {
+            let divisor = eval!(r)?;
+            if divisor == 0 {
+                bail!("Division by zero in linker script ASSERT expression");
```
davidlattimore: I'd suggest removing the "in..." part of the message. It's likely that expression evaluation will eventually be used for other purposes, e.g. symbols can be assigned expressions. If we want to provide context, that's better done at a higher level. e.g. when we parse the assertion, we could store the line number and file it came from, then `evaluate_assertion` above could add that context. e.g.

```rust
evaluate_expression(&assertion.expression, section_layouts, output_sections)
   .with_context(|| format!("Failed to evaluate assertion from {filename}:{line_number}))?
```

or something like that.

libwild/src/expression_eval.rs
davidlattimore, line 1: Consider naming this module to not be assertion-specific, possibly something like "expression_eval"

libwild/src/assert_eval.rs
```diff
@@ -0,0 +1,488 @@
...
+        Expression::Equal(l, r) => Ok(u64::from(eval!(l)? == eval!(r)?)),
+        Expression::NotEqual(l, r) => Ok(u64::from(eval!(l)? != eval!(r)?)),
+
+        Expression::Sizeof(name) => Ok(section_size(name, section_layouts, output_sections)),
+        Expression::Addr(name) => section_address(name, section_layouts, output_sections),
+
+        Expression::Align(expr) => {
+            let align = eval!(expr)?;
+            if align == 0 {
+                bail!("ALIGN(0) is invalid in linker script expression");
```
davidlattimore: Perhaps just "ALIGN(0) is invalid"?

libwild/src/assert_eval.rs
```diff
@@ -0,0 +1,488 @@
...
+}
+
+fn section_size<'data, P: Platform>(
+    name: &[u8],
+    section_layouts: &OutputSectionMap<OutputRecordLayout>,
+    output_sections: &OutputSections<'data, P>,
+) -> u64 {
+    // GNU ld returns 0 for SIZEOF of a section that doesn't exist in the output.
+    // We match that behavior to avoid breaking scripts that guard with SIZEOF.
+    let Some(id) = output_sections.custom_name_to_id(SectionName(name)) else {
```
davidlattimore: I would assume that we need to be able to handle sections other than custom sections here.

libwild/src/assert_eval.rs
```diff
@@ -0,0 +1,488 @@
...
+    section_layouts.get(id).mem_size
+}
+
+fn section_address<'data, P: Platform>(
+    name: &[u8],
+    section_layouts: &OutputSectionMap<OutputRecordLayout>,
+    output_sections: &OutputSections<'data, P>,
+) -> Result<u64> {
+    let id = output_sections
+        .custom_name_to_id(SectionName(name))
```
davidlattimore: Likely need to handle non-custom sections here too

libwild/src/layout.rs
```diff
@@ -309,6 +309,26 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     update_defsym_symbol_resolutions(&symbol_db, &mut symbol_resolutions.resolutions)?;
     crate::gc_stats::maybe_write_gc_stats(&group_layouts, symbol_db.args)?;
 
+    // Evaluate ASSERT commands from all linker scripts now that layout is complete.
+    let all_assertions: Vec<crate::parsing::ProcessedLinkerScript> = symbol_db
+        .groups
+        .iter()
+        .flat_map(|group| {
+            if let crate::grouping::Group::LinkerScripts(scripts) = group {
+                itertools::Either::Left(scripts.iter().map(|s| {
```
davidlattimore: I think this could probably be simplified. The `clone()` call should be able to go. Rather than constructing a new `ProcessedLinkerScript`, just return a `&[ProcessedLinkerScript]`. The else branch can then return `&[]` which has the same type, so `itertools::Either` wouldn't be needed.

The function that this code is being added to is already really large, so it'd be good to try to avoid adding more than we need to it. One possibility is to just pass `symbol_db.groups` and locate the linker scripts in the callee. Alternatively, perhaps extract the code that gets all the expressions out into a separate function.

wild/tests/sources/linker-script-assert-fail.c
```diff
@@ -0,0 +1,6 @@
+//#LinkArgs:-T ./linker-script-assert-fail.ld
+//#RunEnabled:false
```
davidlattimore: `ExpectError` sets both `RunEnabled` and `DiffEnabled` to false, so these two lines aren't needed

wild/tests/sources/linker-script-assert-fail.c
```diff
@@ -0,0 +1,6 @@
...
+//#DiffEnabled:false
+//#SkipLinker:ld
```
davidlattimore: Does GNU ld not also emit an assertion failure? If we can leave GNU ld enabled, then it provides nice validation that our test is correct. If we can't for some reason, then commenting as to why would be good.

wild/tests/sources/linker-script-assert-pass.ld
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/* A trivially true assertion - should always pass */
+ASSERT(1 == 1, "1 == 1 must always be true");
+
+/* Arithmetic assertion */
+ASSERT(1 + 1 == 2, "arithmetic must work");
+
+/* SIZEOF of a known section - just check it is >= 0 (always true for u64) */
+ASSERT(SIZEOF(.text) >= 0, "text section size must be non-negative");
```
davidlattimore: It'd be good to check that it's > 0, which it should be, since the `_start` function will be in it. Related to comment above about looking up non-custom section names.

wild/tests/sources/linker-script-assert-fail.c
Blazearth: Yes, GNU ld also emit an assertion failure, i have removed it and validated the test still passes.

libwild/src/assert_eval.rs
Blazearth: Added line field to AssertCommand and refactored small changes in `linker_script.rs` to propogate the line, `parse_command` and `parse_section_command` compute the line number before calling `parse_assert`. since linker scripts are generally small so i dont think it would cost more to calculate the line for each expression.

libwild/src/assert_eval.rs
Blazearth: Yes, i have added a fn to get the full set of sections, we iterate over `section_infos` and compare names using `output_sections.name(id)`. This ensures we consider all sections rather than just the custom ones.

libwild/src/linker_script.rs
```diff
@@ -241,31 +244,41 @@ pub(crate) fn skip_comments_and_whitespace(input: &mut &BStr) -> winnow::Result<
     }
 }
 
-fn parse_paren_group<'input>(input: &mut &'input BStr) -> winnow::Result<Vec<Command<'input>>> {
+fn parse_paren_group<'input>(
+    input: &mut &'input BStr,
+    original: &'input [u8],
```
davidlattimore: Rather than passing the original file contents through all the parsing and computing the line number of every assert, what about storing the remainder of the input into the `AssertCommand`, then computing the line number only if we end up needing it? i.e. only compute the line number if we fail to evaluate the expression or if the assertion fails. e.g. in `AssertCommand`, instead of `line`, store:

```rust
pub(crate) remainder: &'data [u8],
```

Alternatively, you could store the length the remaining input, which effectively gives you the offset in bytes from the end of the linker script.

Blazearth:
That makes sense ,I initially computed the line number since linker scripts are typically small, but I agree that storing the remainder and computing it lazily is a cleaner approach.
I have updated the implementation to store the remaining input slice in `AssertCommand` and derive the line number only when reporting errors.

PR #1697, opened 2026-03-16
fix: Register linker plugin message callback first

mati865:
Otherwise, linker plugin has no way to indicate invalid arguments.

### DIFF
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -355,6 +355,9 @@ impl LoadedPlugin {
 
         let mut transfer_vector = Vec::new();
 
+        // Linker plugins handle entries of this vector serially, which means the message callback
+        // should be registered first. Otherwise, they won't be able to indicate the problem with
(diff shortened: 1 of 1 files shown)


libwild/src/linker_plugins.rs
```diff
@@ -355,6 +355,8 @@ impl LoadedPlugin {
 
         let mut transfer_vector = Vec::new();
 
+        transfer_vector.push(LdPluginTv::fn_ptr2(Tag::Message, message));
```
davidlattimore, line 361: Consider adding a comment - something along the lines of "Register message callback first so that the plugin can report problems with subsequent registrations".

davidlattimore (approved):

libwild/src/linker_plugins.rs
mati865: Added a bit more verbose one. Feel free to tune it as you wish (even with the review suggestion).

Issue #1229, opened 2025-10-18
Add support for -r/--relocatable (generating static relocatable binaries)


p4vook:

davidlattimore:
I think this would be good to have. I haven't heard of it being used much, but I can imagine uses for it, so it'd be good to support it. It'd be nice to hear of actual uses of this flag.

It's a moderately involved feature to implement. When active, we need to skip various parts of linking and also do some additional things that we don't do during normal linking. Some other things need to be done in a similar, but not identical way.

A few thoughts on implementation:

* When `-r` or equivalent is specified, we'd need default to `--no-gc-sections`. Whether we'd support `--gc-sections` at all in conjunction with `-r` would remain to be seen.
* As far as I can tell, most of the regular section merging rules should be skipped. e.g. where normally we'd merge `.text.foo` and `.text.bar` into `.text`, with this mode, we'd emit them as two separate sections. In a large codebase compiled with `-ffunction-sections` or equivalent, that could mean that we have a heap of output sections, which is not something we're used to dealing with. So this could be problematic from a performance perspective.
* The output symbol tables needs to be a merge of all the input symbol tables.
  * For globals, all symbols with the same name would need to be unified, probably via the normal symbol resolution phase.
  * All local symbols would, I think, need to pass through 1:1. We'd need to update the section and offset-in-section of each symbol.
* Skip the normal processing of relocations, both during the layout phase and during the writing phase.
* Instead, merge the relocations from all the input sections, grouped by their output section.
* Relocations would need updating when they're copied over.
  * The offset at which the relocation is applied would be different if other input sections were merged into the same output section.
  * The symbol and addend would also be different and would need updating.
* String merging would still need to be done as usual. Where a relocation referenced a string-merge section, we'd need to use the new offset of the merged string in the relocation.

vishruth-thimmaiah:
Can I work on this?

davidlattimore:
For testing, have a look at the "Archive" and "Shared" commands in our integration tests. I think we might want to do something similar here. e.g. add a "Relocatable" command that first uses the linker to link to a relocatable object, then feeds that back into the linker to produce the actual binary.

p4vook:
> I think this would be good to have. I haven't heard of it being used much, but I can imagine uses for it, so it'd be good to support it. It'd be nice to hear of actual uses of this flag.

I'm ashamed to understand very little about the linkers in general, representing the worst of open-source here: opening the issue without a real understanding of the problem.

[...]

PR #1698, opened 2026-03-16
feat: add support for generating static relocatables

vishruth-thimmaiah:
Closes: #1229

This PR implements support for generating static relocatble binaries, enabled with the -r/--relocatable flag.

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -363,7 +363,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(
     }
 
     // Copy relocations are only permitted when building executables.
-    if !args.should_output_executable && !args.should_output_partial_object {
+    if !args.should_output_executable {
         args.copy_relocations =
             CopyRelocations::Disallowed(CopyRelocationsDisabledReason::SharedObject);
     }
@@ -1740,12 +1740,14 @@ impl platform::Args for ElfArgs {
         &self.output
     }
 
+    // TODO: Some linkers like ld and mold cleanup debug symbols when linking with -r. For now, we
+    // ignore --strip-all and --strip-debug in partial link mode.
     fn should_strip_debug(&self) -> bool {
-        matches!(self.strip, Strip::All | Strip::Debug)
+        !self.should_output_partial_object() && matches!(self.strip, Strip::All | Strip::Debug)
     }
 
     fn should_strip_all(&self) -> bool {
-        matches!(self.strip, Strip::All)
+        !self.should_output_partial_object() && matches!(self.strip, Strip::All)
     }
 
     fn should_strip_symbol_named(&self, name: &[u8]) -> bool {
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1836,7 +1836,7 @@ impl platform::Platform for Elf {
         builder.build()
     }
 
-    fn will_emit_section_symbol(
+    fn will_emit_section_symbol_for_partial_objects(
         output_sections: &OutputSections<Self>,
         section_id: OutputSectionId,
     ) -> bool {
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -230,8 +230,8 @@ fn write_file_contents<'data, A: Arch<Platform = Elf>>(
     timing_phase!("Write data to file");
     let mut section_buffers = split_output_into_sections(layout, &mut sized_output.out);
 
-    let global_sym_index_map = if layout.args().should_output_partial_object() {
-        build_global_sym_index_map(layout)
+    let sym_index_map = if layout.args().should_output_partial_object() {
+        build_sym_index_map(layout)
     } else {
         Vec::new()
     };
@@ -243,15 +243,10 @@ fn write_file_contents<'data, A: Arch<Platform = Elf>>(
         .try_for_each(|(group, mut buffers)| -> Result {
             verbose_timing_phase!("Write group");
 
-            let group_total_locals =
-                (*group.mem_sizes.get(part_id::SYMTAB_LOCAL) / elf::SYMTAB_ENTRY_SIZE) as u32;
-
             let mut table_writer = TableWriter::from_layout(
                 layout,
                 group.dynstr_start_offset,
                 group.strtab_start_offset,
-                group.symtab_local_start_index,
-                group_total_locals,
                 &mut buffers,
                 group.format_specific.eh_frame_start_address,
             );
@@ -263,7 +258,7 @@ fn write_file_contents<'data, A: Arch<Platform = Elf>>(
                     &mut table_writer,
                     layout,
                     &sized_output.trace,
-                    &global_sym_index_map,
+                    &sym_index_map,
                 )
                 .with_context(|| format!("Failed copying from {file} to output file"))?;
             }
@@ -454,17 +449,12 @@ fn write_file<'data, A: Arch<Platform = Elf>>(
     table_writer: &mut TableWriter,
     layout: &ElfLayout<'data>,
     trace: &TraceOutput,
-    global_sym_index_map: &[Option<u32>],
+    sym_index_map: &[Option<u32>],
 ) -> Result {
     match file {
-        FileLayout::Object(s) => write_object::<A>(
-            s,
-            buffers,
-            table_writer,
-            layout,
-            trace,
-            global_sym_index_map,
-        )?,
+        FileLayout::Object(s) => {
+            write_object::<A>(s, buffers, table_writer, layout, trace, sym_index_map)?;
+        }
         FileLayout::Prelude(s) => write_prelude::<A>(s, buffers, table_writer, layout)?,
         FileLayout::Epilogue(s) => write_epilogue::<A>(s, buffers, table_writer, layout)?,
         FileLayout::SyntheticSymbols(s) => write_synthetic_symbols::<A>(s, table_writer, layout)?,
@@ -597,9 +587,6 @@ struct TableWriter<'layout, 'out> {
 
     dynamic: DynamicEntriesWriter<'out>,
     version_writer: VersionWriter<'out>,
-
-    symtab_local_start_index: u32,
-    group_total_locals: u32,
 }
 
 impl<'layout, 'out> TableWriter<'layout, 'out> {
@@ -607,8 +594,6 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
         layout: &'layout ElfLayout,
         dynstr_start_offset: u32,
         strtab_start_offset: u32,
-        symtab_local_start_index: u32,
-        group_total_locals: u32,
         buffers: &mut OutputSectionPartMap<&'out mut [u8]>,
         eh_frame_start_address: u64,
     ) -> TableWriter<'layout, 'out> {
@@ -624,8 +609,6 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
             dynsym_writer,
             debug_symbol_writer,
             eh_frame_start_address,
-            symtab_local_start_index,
-            group_total_locals,
         )
     }
 
@@ -636,8 +619,6 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
         dynsym_writer: SymbolTableWriter<'layout, 'out>,
         debug_symbol_writer: SymbolTableWriter<'layout, 'out>,
         eh_frame_start_address: u64,
-        symtab_local_start_index: u32,
-        group_total_locals: u32,
     ) -> TableWriter<'layout, 'out> {
         let eh_frame = buffers.take(part_id::EH_FRAME);
         let eh_frame_hdr = buffers.take(part_id::EH_FRAME_HDR);
@@ -664,8 +645,6 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
             eh_frame_hdr,
             dynamic,
             version_writer,
-            symtab_local_start_index,
-            group_total_locals,
         }
     }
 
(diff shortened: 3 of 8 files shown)


vishruth-thimmaiah:
I've opened this PR as a draft as there are are a few tests from mold that still fail.

davidlattimore:
Thanks for working on this. It'd be good to resolve the merge conflicts before we start reviewing too. Most of what was in `args.rs` was moved to `args/elf.rs`.

libwild/src/output_section_id.rs
```diff
@@ -819,6 +849,32 @@ impl<'data, P: Platform> OutputSections<'data, P> {
...
+            return false;
+        }
+
+        if matches!(section_id, FILE_HEADER | PROGRAM_HEADERS | SECTION_HEADERS) {
+            return false;
+        }
+
+        let section_attr = self.output_info(section_id).section_attributes;
+        let segment_type = section_id
+            .opt_built_in_details::<Elf>()
```
davidlattimore: Can you avoid ELF-specific code here? It's probably necessary to make `will_emit_section_symbol` generic over `Platform`

libwild/src/platform.rs
```diff
@@ -196,7 +196,7 @@ pub(crate) trait Platform: Copy + Send + Sync + Sized + std::fmt::Debug + 'stati
 
     type SectionIterator<'data>: Iterator<Item = &'data Self::SectionHeader>;
     type DynamicTagValues<'data>: DynamicTagValues<'data>;
-    type RelocationList<'data>: Send + Sync + 'data;
+    type RelocationList<'data>: RelocationList<'data> + Send + Sync + 'data;
```
davidlattimore: Since we've got a trait here now, could you move the bounds (`Send` etc) to the trait itself? See other traits in this file for examples of how this looks.

wild/tests/sources/elf/relocatables/relocatable-2.c
```diff
@@ -0,0 +1 @@
+int multiply(int a, int b) { return a * b; }
```
davidlattimore, line 3: I think it'd be good to have at least one reference from one of the files to the other. At the moment, these two files define symbols, but likely don't contain any relocations themselves.

libwild/src/layout_rules.rs
```diff
@@ -473,6 +474,44 @@ impl<'data> SectionRules<'data> {
 
         SectionRuleOutcome::Custom
     }
+
+    #[inline(always)]
+    pub(crate) fn lookup_for_partial_link(
```
davidlattimore: The implementation of this function is ELF-specific. Could you make this a function on `Platform` and move the implementation to `libwild/src/elf.rs`?

libwild/src/platform.rs
```diff
@@ -880,8 +888,12 @@ pub(crate) trait SectionAttributes:
 
     fn flags(&self) -> <Self::Platform as Platform>::SectionFlags;
 
+    fn ty(&self) -> <Self::Platform as Platform>::SectionType;
+
     /// Called for custom sections that return true to `is_null`.
     fn set_to_default_type(&mut self);
+
+    fn new_relocation_type() -> Self;
```
davidlattimore: The name makes it sound to me like this is creating a "type". How about "new_for_relocations" or something like that?

libwild/src/resolution.rs
```diff
@@ -697,9 +704,34 @@ fn assign_section_ids<'data, P: Platform>(
...
+                            continue;
+                        }
+
+                        let section_name = s
+                            .common
+                            .object
+                            .section_name(s.common.object.section(section_index)?)
+                            .unwrap_or_default();
+
+                        let mut rela_name = Vec::with_capacity(section_name.len() + 5);
```
davidlattimore: The input file should already have a section with the required name. e.g. if our section is ".text", and that section has relocations, they'll be in ".rela.text". So we shouldn't need to regenerate the section name.

Do you think it would work to change the rule for relocation sections to not discard them? i.e. rather than working off sections that have relocations, allow the relocation sections themselves to pass through as custom sections, with appropriate processing of those relocations occurring.

davidlattimore:
#1754 moved all test sources into separate subdirectories, so the test files added by this PR will need moving too, otherwise the test won't get picked up. Note that it's no longer necessary to list the test in `integration_tests.rs`.

libwild/src/platform.rs
vishruth-thimmaiah: I've removed this function entirely, we no longer need it if we are retaining relocation sections.

vishruth-thimmaiah:
@davidlattimore The mold test `many-input-sections2.sh` is failing because wild doesn’t yet generate a `.symtab_shndx` section when `shnum` > 65,536. Since this isn’t strictly related to relocatable output support, should it be addressed in a separate PR?

davidlattimore:
Yeah, we definitely wouldn't want to solve that in the same PR. I'd suggest just listing the test in `mold_skip_tests.toml`, ideally with a short comment next to it saying why it fails.

libwild/src/args/elf.rs
```diff
@@ -361,7 +363,7 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(
     }
 
     // Copy relocations are only permitted when building executables.
-    if !args.should_output_executable {
+    if !args.should_output_executable && !args.should_output_partial_object {
```
davidlattimore: Is this change needed? It looks like you set `args.should_output_executable = false` when `-r` is passed.

libwild/src/platform.rs
```diff
@@ -1183,4 +1209,6 @@ pub(crate) trait Args: std::fmt::Debug + Send + Sync + 'static {
         }
         Ok(())
     }
+
+    fn should_output_partial_object(&self) -> bool;
```
davidlattimore: I think it'd be reasonable to have a default implementation here that returns false. You could then remove the implementation for MachO. I think MachO does have an equivalent concept, but until we add support for it, returning false is reasonable. On Windows, I don't think there's an equivalent at all.

wild/tests/integration_tests.rs
```diff
@@ -416,6 +419,44 @@ impl Linker {
...
+            let mut config = config.clone();
+            config.expect_stderr = Vec::new();
+            config.expect_stdout = Vec::new();
+            config.should_error = false;
+
+            command.run(&config)?;
+            command.write_input_hashes()?;
+        }
+
+        Ok(LinkerInput::with_command(so_path.to_owned(), command))
```
davidlattimore: This bit of code is the same as for `link_shared`. Could be good to deduplicate it. Perhaps a function called `link_intermediate` or something like that.

libwild/src/elf_writer.rs
```diff
@@ -1351,7 +1405,248 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
     }
 
     if !layout.args().should_strip_all() {
-        write_symbols(object, &mut table_writer.debug_symbol_writer, layout)?;
+        if layout.args().should_output_partial_object() {
```
davidlattimore: I assume `--strip-all` and `-r` are incompatible, but it seems odd to me having this if-statement inside the if-statement for strip-all. How about something like the following?

```rust
if layout.args().should_output_partial_object() {
   ...
} else if !layout.args().should_strip_all() {
   write_symbols(object, &mut table_writer.debug_symbol_writer, layout)?;
}
```

I just tried running both GNU ld and lld with both -r and --strip-all and neither emitting an error, which is interesting. From what I can see, it looks like they ignore `--strip-all`, even if it's specified after `-r`. I just tried with wild and we currently error, but the error isn't one that would be of much help to the user:

```
wild: error: validate_empty failed for <prelude>
  Caused by:
    Didn't use up all allocated symtab/strtab space. local=7 global=0 strings=67
```

libwild/src/elf.rs
```diff
@@ -1808,6 +1835,72 @@ impl platform::Platform for Elf {
...
+            secnames::NOTE_GNU_PROPERTY_SECTION_NAME => return SectionRuleOutcome::NoteGnuProperty,
+            secnames::NOTE_ABI_TAG_SECTION_NAME => {
+                return SectionRuleOutcome::Section(crate::layout_rules::SectionOutputInfo::keep(
+                    output_section_id::NOTE_ABI_TAG,
+                ));
+            }
+            _ => {}
+        }
+
+        SectionRuleOutcome::Custom
```
davidlattimore, line 1902: Wild's performance relies on not having a very large number of output sections. I just tried linking wild itself with `-r` and it failed:

```
thread 'main' (632584) panicked at libwild/src/layout.rs:3012:18:
output section count must fit in a u16: TryFromIntError(())
```

But even if we support more than 65k output sections, the performance would be terrible, since we track space allocation for each output section for each input object group. So there would per-group bookkeeping for each output section even though likely only one group contributes to the output section.

I think we'll need some way for objects to define one-off output sections. i.e. output sections that are effectively private to that input object. But given that this PR is already pretty large, I think we should do that separately.

libwild/src/platform.rs
```diff
@@ -825,6 +841,10 @@ pub(crate) trait SectionHeader: std::fmt::Debug + Send + Sync + 'static {
 pub(crate) trait SectionType:
     Default + Copy + Send + Sync + std::fmt::Debug + 'static
 {
+    fn is_rela(&self) -> bool;
```
davidlattimore, line 844: Rela and rel are two different ELF-specific representations for relocations. How about having a single `is_relocations` function that returns whether the section contains relocations of any kind?

Alternatively, given that it looks like this and the other functions here are only called from ELF-specific code, you could delete these and just use the elf types directly from the elf code.

libwild/src/layout.rs
```diff
@@ -4503,24 +4559,48 @@ fn layout_section_parts<P: Platform>(
...
-                            mem_offset = alignment.align_up(mem_offset);
-
-                            let file_size = if output_sections.has_data_in_file(merge_target) {
-                                mem_size as usize
+                            if args.should_output_partial_object() {
+                                let file_size = if output_sections.has_data_in_file(merge_target) {
+                                    mem_size as usize
+                                } else {
+                                    0
+                                };
```
davidlattimore, line 4567: These lines appear in both branches. Could we move them before the if-statement?

libwild/src/elf_writer.rs
```diff
@@ -1351,7 +1405,248 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
+    let num_all_locals = (layout
+        .section_part_layouts
+        .get(part_id::SYMTAB_LOCAL)
+        .file_size
+        / size_of::<elf::SymtabEntry>()) as u32;
+
+    let total_syms = layout.symbol_db.num_symbols();
+    let mut map: Vec<Option<u32>> = vec![None; total_syms];
+
+    for group in &layout.group_layouts {
```
davidlattimore, line 1409: I'm pretty sure this loop could be done in parallel (with the use of `ShardedVecWriter`), but it's lower priority than the performance issues related to output sections, so I'd say a TODO should be fine for now.

libwild/src/elf_writer.rs
```diff
@@ -1351,7 +1405,248 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
+        let group_global_base = num_all_locals + group.symtab_global_start_index;
+        let mut group_global_offset: u32 = 0;
```
davidlattimore: Do we need both variables? It looks like perhaps we could just have one variable - perhaps called something like `output_sym_index`.

libwild/src/output_section_id.rs
```diff
@@ -755,6 +768,10 @@ impl<'data, P: Platform> OutputSections<'data, P> {
         found
     }
 
+    pub(crate) fn will_emit_section_symbol(&self, section_id: OutputSectionId) -> bool {
```
davidlattimore: This method and the method of the same name on `Platform` seem to be specific to partial linking, but it's not currently clear that that's the case from the name. Perhaps rename to make it clearer and maybe add a comment saying what the method is for.

libwild/src/elf_writer.rs
```diff
@@ -1351,7 +1405,248 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
+                    group_global_offset += 1;
+                }
+            }
+        }
+    }
+
+    map
+}
+
+fn build_section_sym_indices(layout: &ElfLayout<'_>) -> HashMap<OutputSectionId, u32> {
```
davidlattimore: Any reason not to use an `OutputSectionMap` rather than a `HashMap`?

libwild/src/elf_writer.rs
```diff
@@ -1351,7 +1405,248 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
+    buffers: &mut OutputSectionPartMap<&mut [u8]>,
+    layout: &ElfLayout<'data>,
+    global_sym_index_map: &[Option<u32>],
+    section_sym_indices: &HashMap<OutputSectionId, u32>,
+    obj_local_start: u32,
+) -> Result {
+    let e = LittleEndian;
+
+    let num_syms = object.object.symbols.len();
+    let mut local_sym_map: Vec<Option<u32>> = vec![None; num_syms];
```
davidlattimore: Could we store the local symbol index mapping in the same large Vec as we store the globals in? i.e. compute them in `build_global_sym_index_map` (with possible rename of that function).

libwild/src/platform.rs
mati865: Indeed there is no such thing on Windows. Wasm equivalent seems similar to ELF but much simpler (one could say saner).

davidlattimore:
If any of the changes I suggested are going to take a while and you'd like to leave them for a later PR, just say so (and add a TODO). It'd be good to get this merged before more merge conflicts come up.

libwild/src/elf_writer.rs
vishruth-thimmaiah: lld ignores `--strip-all`, but ld appears to remove debug sections along with their relocs/symbols.

Perhaps for now we can do what lld does and keep all sections/symbols?

vishruth-thimmaiah:
With the exception of the performance optimisations and handling of `--strip-all/debug`, which we ignore for now, I've updated the PR with the rest of the required changes.

PR #1700, opened 2026-03-16
fix: preserve hidden visibility when merging symbols and localize dynsym exports

### LINKED ISSUE #1693: Wild exports symbols to dynsym that should remain local after visibility merging
Wild exports symbols to `dynsym` that should remain local after visibility merging.

When a symbol has mixed visibility definitions, the final merged visibility should be enforced before export filtering. Right now Wild doesn't seem to apply the merged visibility correctly, so some symbols still end up exported.

**Expected behavior**
After merging, the most restrictive visibility should win:
```
hidden > protected > default
```
`dynsym` export should follow this final visibility, so symbols that resolve to `hidden` or `protected` should not be exported.

but Wild still exports some symbols that should be local.

From the test results:
* **Wild dynsym:** `data1`, `data4`, `data5` appear as exports
* **BFD dynsym:** `data1`, `data4`, `data5` remain local

I’d like to work on this if it’s okay

PritamP20:
When a symbol has mixed visibility definitions across object files, the most restrictive visibility should win (`hidden > protected > default`). Wild was computing the merged visibility correctly but silently discarding it, so symbols that should be local were still leaking into `dynsym`.

Fixes #1693

### Root cause

Two issues in `handle_non_default_visibility`:

1. The merged visibility was never passed in, so hidden and protected were treated identically
2. Only `NON_INTERPOSABLE` was set, never `DOWNGRADE_TO_LOCAL` — the flag that actually prevents export into dynsym

A third issue: `data5` is declared hidden in one object but defined (with default visibility) in another. Since one is undefined and the other defined, they never appear in the same alternative-definition group, so the merged visibility path never ran.
This is fixed by tracking hidden undefined references during symbol loading and downgrading their definitions after resolution.

```
Before (Wild dynsym):           After (Wild dynsym):
  data1  ← wrong, should local    data2
  data2                           data3
  data3                           get_data1
  data4  ← wrong, should local    get_data5
  data5  ← wrong, should local
  get_data1
  get_data5

BFD dynsym (correct):
  data2, data3, get_data1, get_data5
```

### DIFF
--- a/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility.c
+++ b/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility.c
@@ -1,0 +1,25 @@
+//#AbstractConfig:default
+//#Mode:dynamic
+//#RunEnabled:false
+//#DiffEnabled:false
+//#CompArgs:-fPIC
+//#LinkArgs:-shared -z now
+//#Archive:hidden-weak-archive-visibility-1.c
+//#Object:hidden-weak-archive-visibility-2.c
+//#NoDynSym:get_value
+//#ExpectDynSym:public_func
+
+//#Config:vacant:default
+
+//#Config:occupied:default
+//#Object:hidden-weak-archive-visibility-3.c
+//#ExpectDynSym:other_func
+
+__attribute__((weak, visibility("hidden"))) extern int get_value(void);
+
+int public_func(void) {
+    if (get_value) {
+        return get_value();
+    }
+    return 0;
+}
--- a/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-1.c
+++ b/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-1.c
@@ -1,0 +1,1 @@
+int get_value(void) { return 1; }
--- a/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-2.c
+++ b/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-2.c
@@ -1,0 +1,1 @@
+int get_value(void) { return 42; }
--- a/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-3.c
+++ b/wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility-3.c
@@ -1,0 +1,8 @@
+__attribute__((weak, visibility("protected"))) extern int get_value(void);
+
+int other_func(void) {
+    if (get_value) {
+        return get_value();
+    }
+    return -1;
+}


libwild/src/symbol_db.rs
```diff
@@ -1164,6 +1175,16 @@ pub(crate) fn resolve_alternative_symbol_definitions<'data, P: Platform>(
 
     symbol_db.buckets = buckets;
 
+    for name in &symbol_db.hidden_undefined_names {
```
davidlattimore: Why does this need to be done from the main thread? Perhaps it could be done in parallel from each object or from each hash bucket? Better yet, could we avoid storing a collection of hidden undefined names at all and update the flags during the resolution phase? - i.e. when we resolve undefined symbols

libwild/src/symbol_db.rs
```diff
@@ -1607,16 +1639,24 @@ trait SymbolLoader<'data, P: Platform> {
...
 
             if symbol.is_undefined() || self.should_ignore_symbol(symbol) {
+                if symbol.is_undefined()
+                    && !self.object().is_dynamic()
+                    && !symbol.is_interposable()
+                    && !symbol.is_local()
+                {
+                    let info = self.get_symbol_name_and_version(symbol, local_index)?;
+                    outputs
+                        .add_hidden_undefined_name(UnversionedSymbolName::prehashed(info.name()));
```
davidlattimore: Why are we storing symbol names? Would SymbolIds work?

davidlattimore (commented):
Thanks for working on this. I've benchmarked the change and unfortunately it incurs an 8% slowdown on linking librustc-driver and a 4% slowdown linking the Zed editor.

davidlattimore:
BTW, to what extent did you use LLMs/AI when writing the PR description? I don't know your usual writing style, but it has an unusual feel to it that reminds me of how LLMs write

PritamP20:
@davidlattimore So, I initially I write the PR description myself explaining the issue and the fix, and then use an llm to help polish the wording and formatting

davidlattimore:
OK, thanks. Going forward, would you be able to just post what you write? Don't worry if it's not polished - if an LLM can understand what you wrote, then I'm sure we can too

PritamP20:
@davidlattimore Sure, noted from next time I will take care of it, 

I just had a doubt while working on this the hidden visibility, handling was being skipped in the test files, so it was clear why wild's behaviour was wrong for the hidden merge. But the protected visibility seems to be skipped both in the tests and in `resolve_alternative_symbol_definitions`,  Should that also be handled in this PR?

Screenshot 2026-03-17 at 8 02 32 AM

davidlattimore:
It's good to keep PRs small and focused on one issue where possible. So I'd say fixing hidden symbols in this PR and protected in a separate PR would be best. But certainly good to think about how protected visibility might be handled and take that into consideration when deciding how to solve hidden. The exception, where I'd say it might be reasonable to fix both in the same PR would be if that actually made the PR simpler, but otherwise, probably keep them separate.

libwild/src/symbol_db.rs
PritamP20: yeah, Didn't think about it, so the hidden visibility merge cab be taken care during alternative resolution phase itself, reducing the main thread iteration

libwild/src/symbol_db.rs
PritamP20: we can't store it in symbolid because if the symbol is hidden and undefined then there won't be any entry in symbol_definitions vec and can't find the definition ID so the same behaviour might repeat

libwild/src/symbol_db.rs
davidlattimore: I think it could probably be done during resolution itself, which is after alternative symbol selection. Resolution is where undefined symbols are looked up. For performance reasons, we want to do just a single hash lookup of each symbol. For defined symbols this is done when building the symbol DB. For undefined symbols, it's during resolution. I suspect you can handle both definitions and undefined symbols in resolution though, since for defined symbols, we know what symbol ID they resolve to without having to do hashmap lookup.

PritamP20:
@davidlattimore any review on this?

libwild/src/resolution.rs
```diff
@@ -1383,6 +1422,17 @@ fn resolve_symbol<'data, 'scope, P: Platform>(
     match resources.symbol_db.get(&prehashed_name, allow_dynamic) {
         Some(symbol_id) => {
             *definition_out = symbol_id;
+            // If the undefined reference has non-default visibility, the definition must be
+            // downgraded so it cannot leak into dynsym
+            if !local_symbol_attributes.default_visibility {
+                let visibility = resources.symbol_db.input_symbol_visibility(local_symbol_id);
+                if visibility != Visibility::Default {
+                    resources
```
davidlattimore: I think in order to get acceptable performance, we need to not store into any SegQueues or Vecs. Would it work to get access to the AtomicPerSymbolFlags here and directly update it?

PritamP20: Yeah, we can do it I thought doing it sequentially would affect the efficiency but doing it in parallel can save a lot of time. Thanks for telling me about the atomic symbols, they do exactly what we're trying to do

davidlattimore:
#1754 moved all test sources into separate subdirectories. The test files were mostly unchanged, so hopefully rebasing should only require moving your changes to the file in the new location.

PritamP20:
@davidlattimore I rebased the branch, git handled the rebase itself, so everything is in the right place

libwild/src/resolution.rs
```diff
@@ -903,6 +911,16 @@ fn canonicalise_undefined_symbols<'data, P: Platform>(
                                 per_symbol_flags
                                     .set_flag(undefined.symbol_id, ValueFlags::NON_INTERPOSABLE);
                             }
+
+                            if visibility != Visibility::Default
+                                && let Some(def_id) = symbol_db.get_unversioned(&pre_hashed)
+                            {
+                                symbol_db::apply_visibility_to_definition(
```
davidlattimore, line 918: If I comment out both calls to `apply_visibility_to_definition`, all the tests still pass. Do you think you can come up with an integration test that is fixed by this part of the change?

PritamP20: understood, so you want me to create a integration test, which tests part of the code is working or not, I will do that

davidlattimore: Thanks. It looks like the new test successfully fails if this call is commented out, but commenting out the other call below still results in the tests passing. Do you think the test can be extended to cover that as well?

BTW, the failing test:

```
/usr/bin/ld.bfd: /__w/wild/wild/wild/tests/build/elf/x86_64/hidden-weak-archive-visibility/default/hidden-weak-archive-visibility.c.o: relocation R_X86_64_32 against hidden symbol `get_value' can not be used when making a shared object
/usr/bin/ld.bfd: failed to set dynamic section sizes: bad value
```

I think might be fixed by adding `CompArgs:-fPIC`, although I haven't tried.

PritamP20: Thank you @davidlattimore , i did try extending the test for the second call but couldn’t make it fail when the second call is commented out i added that call just to be safe thinking the same symbol might show up twice in undefined_symbols but the first call already covers it, i will remove the 2nd call

davidlattimore: What about having two objects with references to an undefined symbol, both weak references but only one hidden, then a configuration where the object containing the hidden symbol is first and a configuration where it's second?

PritamP20: damn, I didn't think in that way, Now i have created one more ref of get_value which is weak and protected which goes through the 2nd function call

wild/tests/sources/elf/hidden-weak-archive-visibility/hidden-weak-archive-visibility.c
```diff
@@ -0,0 +1,23 @@
...
+//#Mode:dynamic
+//#RunEnabled:false
+//#DiffEnabled:false
+//#CompArgs:-fPIC
+//#LinkArgs:-shared -z now
+//#Archive:hidden-weak-archive-visibility-1.c
+//#Object:hidden-weak-archive-visibility-2.c
+//#NoDynSym:get_value
+//#ExpectDynSym:public_func
+//#Config:vacant:default
```
davidlattimore, line 12: Could you put a blank line before each of the configs? That's how all our other tests are formatted since it makes it easier to see where they start

Issue #1702, opened 2026-03-17
Improve eh_frame validation logic in linker-diff


davidlattimore:
A few of our tests are disabled on some architectures due to our eh_frame validation logic not being quite right.

In particular, in `libc-integration.c`:

```c
// Both ld and lld complain about a duplicate eh_frame info (loongarch64).
//#SkipArch: loongarch64,riscv64
```

See the function `verify_frames` in `eh_frame_diff.rs`.

Given that both GNU ld and LLD are triggering this error from linker-diff, it's almost certain that the logic in linker diff isn't right.

If anyone would like to have a go at fixing this, let us know and we'll assign it.

Probably start by reading up on .eh_frame and .eh_frame_hdr. `objdump --dwarf=frames` may also help to show what's in these sections.

PritamP20:
Hi @davidlattimore, I would like to take this up I will look into the .eh_frame validation logic and try to figure out what is going wrong and fix it

PR #1703, opened 2026-03-17
test: Add support for cross-testing with clang

davidlattimore:


davidlattimore:
Note that I haven't actually enabled use of this for any tests due to the tests where it would be useful being affected by #1702

PR #1707, opened 2026-03-18
refactor: Make Args an associated type on Platform

davidlattimore:
Issue #1538


davidlattimore:
Having `Args` being a type that's generic over the kind of args is holds didn't work out so had to go. We now have an `Args` trait and code generally takes `&P::Args`, `&impl platform::Args` or, if it's ELF-specific code, `&ElfArgs`.

PR #1708, opened 2026-03-18
perf: Reuse initial address computation for first relaxation scan pass

lapla-cogito:
Before this change, `perform_iterative_relaxation` would redundantly recompute all section and symbol output addresses on its first iteration, even though `layout_section_parts` had just produced the layout from which those addresses are derived.

Compute the addresses once immediately after `layout_section_parts` and pass them into the first relaxation scan pass, avoiding a full traversal of every section and symbol. Subsequent iterations still recompute addresses internally, since the layout changes after each round of relaxations.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -222,6 +222,12 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     );
 
     if symbol_db.args.relax && A::supports_size_reduction_relaxations() {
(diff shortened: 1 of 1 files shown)


lapla-cogito:
Sorry, let me consider this for a while (skipping address calculation only in the first iteration was a strategy I'd been considering for a while, but since that was before our platform abstraction refactoring, the current situation might be different and it may not work as easily).

PR #1717, opened 2026-03-18
Support glob character classes in section name patterns

### LINKED ISSUE #1691: Character class globs not supported in section name patterns
Wild does not support character class globs like `[0-9A-Za-z]` in linker script section name patterns. Patterns containing character classes are treated as literal strings, so sections that should match are silently missed.

for example, in patterns like `*(.text.[0-9a-zA-Z_]*)`, section name patterns only look for an exact match or trailing `*` prefix match.

The `glob` crate is already used elsewhere in Wild for pattern matching and could potentially be applied here too.

I'd like to work on this.

plasmaDestroyer:
Section name patterns containing character classes like `[0-9A-Za-z]` were previously treated as literal strings, causing sections to be silently missed. This PR adds glob pattern support for section names using the existing `glob` crate, similar to how `input_file_pattern` 
is handled.

Fixes #1691

@marxin you mentioned subtle deficiencies in the glob crate, would appreciate any pointers on those, and whether this implementation handles them correctly. Also happy to add more tests for `?` and `*` combinations if that would be useful.

### DIFF
--- a/wild/tests/sources/linker-script-glob.c
+++ b/wild/tests/sources/linker-script-glob.c
@@ -9,9 +9,9 @@
 //#ExpectSym:val_A section="other"
 //#ExpectSym:val_a section="other"
 //#ExpectSym:val_underscore section="other"
-//#ExpectSym:val_foo_bar section="foobar"
-//#ExpectSym:val_foo_dash_bar section="foobar"
-//#ExpectSym:val_foo_baz section="other"
+//#ExpectSym:val_hello_world section="helloworld"
+//#ExpectSym:val_hello_dash_world section="helloworld"
+//#ExpectSym:val_hello_baz section="other"
 //#DiffIgnore:section.riscv.attributes
 //#DiffIgnore:segment.RISCV_ATTRIBUTES.*
 // GNU ld emits `.riscv.attributes`, but Wild does not
@@ -26,7 +26,8 @@ static int val_A __attribute__((used, section(".mydata.A"))) = 10;
 static int val_a __attribute__((used, section(".mydata.a"))) = 11;
 static int val_underscore __attribute__((used, section(".mydata._"))) = 12;
 
-static int val_foo_bar __attribute__((used, section("foo_bar"))) = 100;
-static int val_foo_dash_bar __attribute__((used, section("foo-bar"))) = 101;
+static int val_hello_world __attribute__((used, section("hello_world"))) = 100;
+static int val_hello_dash_world __attribute__((used, section("hello-world"))) =
+    101;
 
-static int val_foo_baz __attribute__((used, section("foo_baz"))) = 102;
+static int val_hello_baz __attribute__((used, section("hello_baz"))) = 102;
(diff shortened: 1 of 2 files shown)


libwild/src/layout_rules.rs
```diff
@@ -274,10 +278,33 @@ impl<'data> SectionRule<'data> {
...
+            && (idx < pattern.len() - 1 || pattern[idx] != b'*')
+        {
+            let s = std::str::from_utf8(pattern)
+                .map_err(|_| crate::error!("Invalid UTF-8 in section pattern"))?;
+
+            let compiled_pattern =
+                Pattern::new(s).map_err(|_| crate::error!("Invalid glob pattern '{}'", s))?;
+
+            return Ok(Self {
+                name: &pattern[..idx],
```
davidlattimore: It seems weird to me that we're putting the prefix of the pattern in here. It looks to me like that means a pattern of "foo[_-]bar" would then match all sections that start with "foo", which doesn't seem right. I'd suggest replacing `name`, `is_prefix` and `section_name_pattern` with a single field that holds an enum with three variants `Exact`, `Prefix` and `Glob`.

plasmaDestroyer: yeah, I went with the simpler approach initially but the enum makes much more sense, working on it now.

plasmaDestroyer:
hey @davidlattimore, I implemented the `SectionNameMatcher` enum as you suggested, but when I went to write the integration test for your `foo[_-]bar` example, it immediately hit this existing panic inside `SectionRules::from_rules`:

```rust
let hash = section_name_prefix_hash(rule.name_matcher.prefix_bytes())
    .expect("Prefixes of length less than 4 not yet supported");
```

since the literal prefix `foo` before the wildcard is only 3 bytes long.

How should I handle this? Should I implement some kind of fallback logic as part of this PR, or would you prefer I work around the panic for now (e.g., using something like `hello[_-]world` in the test instead) and we tackle the hash table issue in a dedicated follow-up issue?

davidlattimore:
Yeah, using a longer prefix like "hello" seems like the right thing for now

libwild/src/layout_rules.rs
```diff
@@ -274,10 +293,29 @@ impl<'data> SectionRule<'data> {
             })
             .transpose()?;
 
+        let wildcard_idx = pattern
```
marxin, line 296: I might be a bit late to the party, but I think the code should be re-using the existing glob pattern analysis and parsing logic in: https://github.com/wild-linker/wild/blob/main/libwild/src/version_script.rs#L709-L724.
The place also contains a logic for replacement of the expected negation `^` (while the `glob` crate expects `!`).

Can you prepare a follow-up PR?

plasmaDestroyer: > Can you prepare a follow-up PR?

Sure, working on it.

One more thing i can do is make a `glob_match.rs` module with `compile_glob_pattern` that implements the  `[^` -> `[!` replacement. Can we use that in both `layout_rules.rs` and `version_script.rs` instead of copying the logic?

marxin: > Can we use that in both layout_rules.rs and version_script.rs instead of copying the logic?

Yes, we should reuse as much as possible in between these 2 places.

PR #1723, opened 2026-03-19
Support running on wasi

bjorn3:
Fixes https://github.com/wild-linker/wild/issues/1462

I have tested this with:

```shell
cat > wasm-wild.sh <<EOF
#!/bin/sh
wasmtime run --dir . --dir / $(pwd)/target/wasm32-wasip1/debug/wild.wasm "\$@"
EOF
chmod u+x wasm-wild.sh

cargo build --target wasm32-wasip1
wasmtime run target/wasm32-wasip1/debug/wild.wasm --help # precompile the wasm module

# Run the following two commands in a loop until everything builds. Wasi doesn't allow setting the executable bit.
RUSTFLAGS="-Clinker=clang -Clink-arg=-fuse-ld=$(pwd)/wasm-wild.sh" cargo build
chmod u+x target/debug/build/*/build-script-build
```

### DIFF
--- a/libwild/src/save_dir.rs
+++ b/libwild/src/save_dir.rs
@@ -277,7 +277,13 @@ impl SaveDirState {
                 self.copy_file(&absolute_target, parsed_args)?;
             }
 
-            create_symlink(&target, &dest_path)?;
+            if let Err(error) = create_symlink(&target, &dest_path) {
+                // If we can't create a symlink, then fall back to copying. If that fails, then
+                // return the error from when we tried to create the symlink.
+                if std::fs::copy(&target, &dest_path).is_err() {
+                    return Err(error);
+                }
+            }
         } else {
             if let Ok(data) = FileData::new(source_path, false) {
                 match FileKind::identify_bytes(&data) {
@@ -394,7 +400,7 @@ fn create_symlink(target: &Path, dest_path: &Path) -> Result {
     #[cfg(target_os = "wasi")]
     {
         let _ = (target, dest_path);
-        panic!("creating symlinks on wasi not supported on stable rust");
+        bail!("creating symlinks on wasi not supported on stable rust");
     }
 }
 


Cargo.toml
```diff
@@ -72,7 +72,7 @@ os_info = "3.0.0"
 paste = "1.0.15"
 # 0.4.9 contains breaking changes.
 perf-event = "=0.4.8"
-perfetto-recorder = "0.3.0"
+perfetto-recorder = "0.3.2"
```
bjorn3: `cargo update -Zdirect-minimal-versions` fails with 0.3.1 too. 0.3.0 does not compile on wasi.

bjorn3: Updating the `libc` dependency does work. There are other crates that don't work at their minimal version on wasm, so I've reverted all Cargo.toml changes. It works with the current versions in the lockfile.

libwild/src/input_data.rs
```diff
@@ -787,46 +792,69 @@ impl FileData {
     }
 
     fn open(path: &Path, prepopulate_maps: bool) -> Result<(Self, std::fs::File)> {
```
mati865, line 813: Not a blocker, but maybe it would be cleaner if we split `FIleData` into distinct mmap and no-mmap variants, each its with own `open` variant?

libwild/src/save_dir.rs
```diff
@@ -391,6 +391,11 @@ fn create_symlink(target: &Path, dest_path: &Path) -> Result {
         })?;
         Ok(())
     }
+    #[cfg(target_os = "wasi")]
+    {
+        let _ = (target, dest_path);
+        panic!("creating symlinks on wasi not supported on stable rust");
```
mati865: Is it likely to be stabilised? Given that the symlinks are a problem for Windows as well, we might create "copy rather than symlink" version.

mati865:
Love to see that!

Regarding the arch, I wonder how bad it would be to add variant `Unsupported` and just bail out if it doesn't get overriden by `-m`. GCC and Clang will always call the linker with explicit architecture.

libwild/src/save_dir.rs
bjorn3: No clue

libwild/src/input_data.rs
bjorn3: You mean using an enum?

mati865: I shouldn't have used "variant" in that context.

I was thinking about something like:
```rust
struct FileData {
    bytes: FileBytes
    modification_time: std::time::SystemTime,
}

impl FileData {
    fn open(path: Path, prepopulate_maps: bool) { 
        let modification_time = ...;
        let bytes = FileBytes::read(..);
        ...
    }
}

#[cfg(target_family = "wasm")]
struct FileBytes(Vec<u8>);

impl FileBytes {
    fn read(path: Path, _prepopulate_maps: bool) { .. }
}

#[cfg(not(target_family = "wasm"))]
struct FileBytes(Vec<u8>);

impl FileBytes {
    fn read(path: Path, prepopulate_maps: bool) { .. }
}
```
But an enum would also work.

davidlattimore: An enum would allow switching at runtime, which we don't really need

libwild/src/save_dir.rs
davidlattimore: I'd be inclined to bail here rather than panic. We can then change the caller of `create_symlink` to fall back to copying if it gets an error.

davidlattimore:
We should also add something to CI to verify that we can build for wasi, otherwise we'll almost certainly break it at some point. I'm happy to add that afterwards though if you don't want to.

libwild/src/save_dir.rs
bjorn3: I only half follow the `copy_file` logic. Which steps are necessary for the fallback?

libwild/src/input_data.rs
bjorn3: Something like https://github.com/wild-linker/wild/pull/1723/commits/5fe8d84b35e568692428e48da617b81a839f2eaa?

bjorn3:
I've added a build config to CI. For testing it is fairly easy to setup by using `wasmtime --dir .` as runner. integrations_tests.rs doesn't compile due to depending on wait-timeout though, 4 tests fail due to needing to create a temporary file and `test_no_disallowed_overlaps` fails as it doesn't override `Architecture::Unsupported`.

.github/workflows/ci.yml
```diff
@@ -206,6 +206,24 @@ jobs:
...
+    name: wasip1 build
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    container:
+      image: ubuntu:25.10
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - run: apt-get install --update -y build-essential gcc clang curl
```
davidlattimore: I think the CI failure might be fixed by adding `gcc-multilib`

libwild/src/save_dir.rs
davidlattimore: I was thinking something like this:

```patch
diff --git a/libwild/src/save_dir.rs b/libwild/src/save_dir.rs
index 42511e5ec2..fd353b6e8b 100644
--- a/libwild/src/save_dir.rs
+++ b/libwild/src/save_dir.rs
@@ -277,7 +277,13 @@
                 self.copy_file(&absolute_target, parsed_args)?;
             }

-            create_symlink(&target, &dest_path)?;
+            if let Err(error) = create_symlink(&target, &dest_path) {
+                // If we can't create a symlink, then fall back to copying. If that fails, then
+                // return the error from when we tried to create the symlink.
+                if std::fs::copy(&target, &dest_path).is_err() {
+                    return Err(error);
+                }
+            }
         } else {
             if let Ok(data) = FileData::new(source_path, false) {
                 match FileKind::identify_bytes(&data) {
@@ -394,7 +400,7 @@
     #[cfg(target_os = "wasi")]
     {
         let _ = (target, dest_path);
-        panic!("creating symlinks on wasi not supported on stable rust");
+        bail!("creating symlinks on wasi not supported on stable rust");
     }
 }
```

.github/workflows/ci.yml
bjorn3: Indeed. Thanks!

libwild/src/input_data.rs
mati865: Yeah, thanks.

PR #1735, opened 2026-03-20
refactor: Extract shared glob pattern compilation into glob_match module

plasmaDestroyer:
Follow-up to #1717, on advice of @marxin

Extracts glob pattern compilation logic from `version_script.rs` into `glob_match.rs` module, which is then reused in both `version_script.rs` and `layout_rules.rs`.

Additional improvements i was able to get in section name pattern matching:
- `[^` to `[!` negation fix now applies to section name patterns too
- Escape sequences handled correctly
- `*` can now appear anywhere in a pattern, not just at the end (which removed the need for `ensure` in `layout_rules.rs`)

PS: The `foo[_-]bar` example from earlier discussion now works correctly :)

### DIFF
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -291,15 +291,10 @@ impl<'data> SectionRule<'data> {
                 SectionNameMatcher::Exact(Cow::Owned(unescape_pattern(pattern)))
             }
             GlobPatternType::Star | GlobPatternType::NonStar => {
-                let wildcard_idx = pattern
-                    .iter()
-                    .position(|&b| b == b'*' || b == b'?' || b == b'\\' || b == b'[' || b == b']')
-                    .unwrap();
-
                 let compiled_pattern =
                     compile_glob_pattern(pattern).map_err(|e| crate::error!("{}", e))?;
 
-                SectionNameMatcher::Glob(&pattern[..wildcard_idx], compiled_pattern)
+                SectionNameMatcher::Glob(pattern, compiled_pattern)
             }
         };
 
@@ -546,4 +541,18 @@ fn test_glob_section_character_class() {
     assert!(!rule.matches(b"foobar", None));
     assert!(!rule.matches(b"foo_barbaz", None));
     assert!(!rule.matches(b"fooxbar", None));
+
+    // [a-z] alphabet range match
+    let range_rule = SectionRule::new(b"foo[a-z]bar", None, SectionRuleOutcome::Discard).unwrap();
+    assert!(range_rule.matches(b"fooabar", None));
+    assert!(range_rule.matches(b"foozbar", None));
+    assert!(range_rule.matches(b"foombar", None));
+    assert!(!range_rule.matches(b"fooAbar", None));
+    assert!(!range_rule.matches(b"foo1bar", None));
+
+    // escaped character match
+    let escape_rule = SectionRule::new(b"foo\\*bar", None, SectionRuleOutcome::Discard).unwrap();
+    assert!(escape_rule.matches(b"foo*bar", None));
+    assert!(!escape_rule.matches(b"fooxbar", None));
+    assert!(!escape_rule.matches(b"foobar", None));
 }


libwild/src/layout_rules.rs
```diff
@@ -277,52 +279,32 @@ impl<'data> SectionRule<'data> {
...
-            })
+            .map(|pattern| compile_glob_pattern(pattern).map_err(|e| crate::error!("{e}")))
             .transpose()?;
 
-        let wildcard_idx = pattern
-            .iter()
-            .position(|&b| b == b'*' || b == b'[' || b == b'?');
+        let name_matcher = match analyze_glob_pattern(pattern) {
+            GlobPatternType::Exact => SectionNameMatcher::Exact(pattern),
+            GlobPatternType::Star | GlobPatternType::EscapedExact | GlobPatternType::NonStar => {
```
marxin: Can't you just make dispatch based like this:
```rust
match analyze_glob_pattern(pattern) {
  GlobPatternType::Exact => SectionNameMatcher::Exact(pattern),
  GlobPatternType::EscapedExact => SectionNameMatcher::Exact(unescape_pattern(pattern)),
  GlobPatternType::Star | GlobPatternType::NonStar  => SectionNameMatcher::Glob(compile_glob_pattern(pattern)...)
}
```

and get rid of a special-case `SectionNameMatcher::Prefix`?

plasmaDestroyer: yeah, i actually tried that but `unescape_pattern` returns an allocated `Vec<u8>`, but `SectionNameMatcher::Exact` requires a `&'data [u8]` reference. I couldn't figure out how to deal with that change so i implemented it this way. Can you suggest how i should deal with that?

marxin: What about replacing `Exact(&'data [u8]),` with `Exact(Cow<'data, [u8]>),`:
https://doc.rust-lang.org/std/borrow/enum.Cow.html
?

plasmaDestroyer: > What about replacing `Exact(&'data [u8]),` with `Exact(Cow<'data, [u8]>),`: https://doc.rust-lang.org/std/borrow/enum.Cow.html ?

yeah, i guess that could help and make the dispatch the way you mentioned, working on it.

One thing, though, getting rid of `Prefix` doesn't seem viable because `DEFAULT_SECTION_RULES` in `elf.rs` is a const static array, it must be constructed at compile time using const fn and the `prefix` and `prefix_section` helpers that build these rules are const fn that construct `SectionNameMatcher::Prefix(name)` directly.

is there any way we could use `compile_glob_pattern` with it since that allocates a heap `Pattern`?

plasmaDestroyer: Implemented `Exact(Cow<'data, [u8]>)` as suggested. `Prefix` is still in the enum because of `DEFAULT_SECTION_RULES` in `elf.rs`.

marxin: Good job! Have one more nit, but I like it now.

libwild/src/layout_rules.rs
```diff
@@ -277,58 +282,38 @@ impl<'data> SectionRule<'data> {
...
+                let compiled_pattern =
+                    compile_glob_pattern(pattern).map_err(|e| crate::error!("{}", e))?;
 
-            return Ok(Self {
-                name_matcher: SectionNameMatcher::Glob(&pattern[..idx], compiled_pattern),
-                input_file_pattern: compiled_file_pattern,
-                outcome,
-            });
-        }
+                SectionNameMatcher::Glob(&pattern[..wildcard_idx], compiled_pattern)
```
marxin: Note the first element of `SectionNameMatcher::Glob` is used only for hashing purposes (where first 4 characters are hashed). That said, you don't have to split it at `wildcard_idx` and just use the entire `pattern`.

libwild/src/layout_rules.rs
```diff
@@ -544,3 +529,21 @@ fn test_glob_section_matching() {
...
+}
+
+#[test]
+fn test_glob_section_character_class() {
+    let rule = SectionRule::new(b"foo[_-]bar", None, SectionRuleOutcome::Discard).unwrap();
+    assert!(rule.matches(b"foo_bar", None));
+    assert!(rule.matches(b"foo-bar", None));
+    assert!(!rule.matches(b"foobar", None));
+    assert!(!rule.matches(b"foo_barbaz", None));
+    assert!(!rule.matches(b"fooxbar", None));
```
marxin, line 543: Please also include some `[a-z]` type, and maybe escape character (`\`)?

PR #1737, opened 2026-03-21
fix: Emit undefined symbol error even if primary reference is GCed

davidlattimore:
Issue #1739


davidlattimore:
Previously, we only emitted undefined symbol errors when the reference came from the file that "owned" that symbol. This kind of worked because if multiple files all referenced the undefined symbol, we'd emit an undefined symbol error for one of them. If the section containing the reference got GCed though, then we'd not emit the error.

In order to fix this, we now don't check if the reference came from the canonical file for that symbol. Removing that check required that we fix how we identify if a symbol is undefined. We were relying on `ValueFlags::ABSOLUTE` being set, however that flag is also set for absolute symbols, so wasn't sufficient. We now read the symbol from the object file that owns it so we can check if it's actually undefined.

PR #1741, opened 2026-03-21
fix: Don't report undefined symbols when writing executables with -z undefs

davidlattimore:


wild/tests/sources/undefined_symbols.c
```diff
@@ -1,18 +1,29 @@
 //#AbstractConfig:default
+//#EnableLinker:lld
```
mati865, line 2: What was the reason for LLD here?

davidlattimore: Mostly to verify that LLD has the same behaviour as GNU ld. i.e that it errors and doesn't error in the same circumstances. I wonder if we should change linker-diff's default behaviour to match-all rather than match-any, then we could enable match-any just for tests that need it. It doesn't affect this test much, since diffing isn't so important for this test - all that really matters is whether we're failing or succeeding the link.

mati865: Ah, right. Expect error makes it so all the linkers must fail similar way.
I don't have strong opinion on whether running all tests with additional linker is worthwhile. It might prove time-consuming, but on high core count machines it likely won't be very noticeable since c++/libc/rust integration tests dwarf all the other tests.

PR #1743, opened 2026-03-21
test: Report file and line number when rejecting test directives

davidlattimore:


marxin:
Really like the idea!

PR #1745, opened 2026-03-22
test: Remove command hash from object filenames

davidlattimore:


davidlattimore:
I've wanted to do this for a while in order to make the object files easier to locate, when checking what's in them. Without the hash, there was one test that would have written the same object file twice, once from a .c file and once from a .s file. I added validation so that we get a nice error message telling us what went wrong - that we tried to write the same file twice two different command-lines. That problem was then fixed by appending the suffix instead of replacing the existing extension. i.e. our object file is now called `preinit-array.c.o` and `preinit-array.s.o` rather than `preinit-array.o`.

Issue #757, opened 2025-05-13
MachO support


djc:
Would be awesome if this project at some point grew support for Darwin-based systems like macOS.

lu-zero:
Would be nice to provide a checklist on what is needed and see if some quick preparation work could be contributed.

davidlattimore:
On Linux, our trivial binaries are statically linked executables, but given that Mac (and Windows) don't as far as I know have a stable interface to the kernel, it might be better for the trivial binaries to be dylibs. That way the dylib can just define a function that returns a number. Adding support for this kind of test (a shared object that gets opened and executed) is something that could be done before everything else - whoever is doing it would need Linux though.

After that, the next next step would be to move as much of the elf-specific code as possible behind a trait. Working out what is and isn't elf-specific is likely to be a bit of guesswork, so the trait would certainly need adjusting as it starts to actually get a second implementation.

Then work on producing a trivial dylib test that passes, probably in conjunction with adding basic macho support to linker-diff.

marxin:
I've **started working** on the MachO port, where I collected the early observations and wrote them down in a document:
https://github.com/marxin/wild/blob/macos/macOS.md

marxin:
@davidlattimore I think I'll need a bit of `Platform` refactoring related to the `build_output_order_and_program_segments`  (and `DEFAULT_SECTION_RULES`) functionality where we define the single part sections (and their corresponding rules). I would like to use the very same concept, but with a different section names and segment names.  Can you work on that?

davidlattimore:
Happy to help. Could you elaborate a bit?

davidlattimore:
I moved `build_output_order_and_program_segments` - #1750.

`DEFAULT_SECTION_RULES` was already previously moved. You can implement `default_layout_rules` on `Platform`.

RossSmyth:
What is the current support for running Wild on Darwin? Trying to figure out what to mark Wild as supporting for hosts when packaging 0.9.0 in Nixpkgs, and while it is mentioned in the release notes as being tested in CI, there is not a mention of the level of confidence.

davidlattimore:
The release notes say "We've also started work on porting to Mach-O and WebAssembly. These aren't yet ready for use, but if you'd like to help out with porting, get in touch."

djc:
> The release notes say "We've also started work on porting to Mach-O and WebAssembly. These aren't yet ready for use, but if you'd like to help out with porting, get in touch."

[...]

PR #1748, opened 2026-03-22
port(MachO): initial scaffolding and input file parsing

marxin:
It's a first step for the #757, where I implemented the `Platform` trait and provided a basic input parsing functionality.
Apart from a small change to the trait, the `Platform` trait guides the implementation pretty smoothly. The top-level dispatch is currently driven by `target_os`.

### DIFF
--- a/libwild/src/macho.rs
+++ b/libwild/src/macho.rs
@@ -10,6 +10,7 @@ use object::Endianness;
 use object::macho;
 use object::macho::Section64;
 use object::read::macho::MachHeader;
+use object::read::macho::Nlist;
 use object::read::macho::Section;
 use object::read::macho::Segment;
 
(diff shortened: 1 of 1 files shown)


davidlattimore:
Actually, one thought. Do you think we should use "port" rather than "feat"?

marxin:
> Actually, one thought. Do you think we should use "port" rather than "feat"?

Works for me.

libwild/src/args.rs
```diff
@@ -94,19 +95,29 @@ impl Args {
...
+        #[cfg(target_os = "macos")]
+        {
+            let macho_args = macho::parse(|| all_args.iter())?;
+            Ok(Args::MachO(macho_args))
+        }
+        #[cfg(not(target_os = "macos"))]
+        {
+            let elf_args = elf::parse(|| all_args.iter())?;
+            Ok(Args::Elf(elf_args))
+        }
```
bjorn3, line 107: You may want to support overriding this option at runtime just like lld. (lld supports `-flavor` and detecting based on arg0) As written it is impossible to cross-link from macOS to ELF based platforms and vice versa.

mati865: Indeed, we plan to support LLDs `-flavor` at some point (discussed in Windows support PRs). All the porting is still an early stage.

PR #1754, opened 2026-03-23
test: Move integration tests to one dir per test

davidlattimore:


davidlattimore:
Also renamed tests with `_` in their name to have `-` instead. Only did this for the top-level test file, not the supporting files, since updating the supporting files would have required edits to the files and I didn't want to do that at this stage.

We now discover tests without needing to list them in `integration_tests.rs`.

In a later PR, I intend to switch to libtest-mimic rather than rstest.

davidlattimore:
Oh, and since I was moving all the files anyway, I put them all into a `elf` subdirectory, since it's likely we'll want to write separate tests for non-ELF platforms.

PR #1755, opened 2026-03-24
refactor: Split argument parsing from Args creation

davidlattimore:


davidlattimore:
My initial motivation for doing this was that I want to be able to set a field on `Args` before the arguments get parsed. Specifically, I want to be able to set a callback that gets run when there are warnings. I think I got some nice deduplication in the process of doing this though.

PR #1758, opened 2026-03-24
test: Speed up running single arch with nextest

davidlattimore:


davidlattimore:
This PR detects nextest and avoids registering non-host arch tests if the arch isn't enabled. This is a common mode of running tests and currently saves about 600 process spawns, reducing `cargo nextest run` from around 1.1 seconds to 0.9 seconds.

PR #1759, opened 2026-03-24
test: Support running unit tests on wasi

bjorn3:
I was able to get integration_tests.rs to compile with a bunch of patches, but it doesn't work as it needs to spawn a bunch of external processes. It is probably a better idea to compile integration_tests.rs for native and then instead have integration_tests.rs run the wild processes it starts in Wasmtime or another wasm engine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,6 +229,9 @@ jobs:
           targets: wasm32-wasip1
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --target wasm32-wasip1
+      # -p libwild prevents integration tests from running. they depend on
+      # spawning processes, which isn't supported on wasi.
+      - run: cargo test --profile ci -p libwild
 
   spelling:
     name: Spell Check with Typos


mati865:
Oh, my bad I didn't check that previously. We are generating changelogs from PR titles using these filters: https://github.com/wild-linker/wild/blob/f0a57890c196027339d2c32c13c7d38a3469f427/cliff.toml#L61-L80 also mentioned here https://github.com/wild-linker/wild/blob/f0a57890c196027339d2c32c13c7d38a3469f427/CONTRIBUTING.md#commit-messages

Mind updating the title?
Also, we squash PRs (mentioned in other CONTRIBUTING.md section). So, you might want to split the commits into separate PRs if you want them to end up as distinct ones.

davidlattimore:
Changes look good to me. I'm happy to squash them as-is if that's what you want. Or if you'd prefer the commit "Simplify integration tests Program::run" to be separate, it could be a separate PR. Our reason for squashing BTW, is that we like to have PR numbers in all commit messages, since it makes the release notes nicer.

Do you think we should update ci.yml so that the tests get run in CI?

PR #1772, opened 2026-03-28
feat: Support `R_RISCV_ALIGN` relaxation

lapla-cogito:
The assembler emits `R_RISCV_ALIGN` relocations with a NOP sled whose byte length is stored in the addend. The linker must delete excess NOPs so that the instruction after the sled lands on the required alignment boundary.

part of #874

### DIFF
--- a/libwild/src/elf_riscv64.rs
+++ b/libwild/src/elf_riscv64.rs
@@ -155,6 +155,21 @@ impl crate::platform::Arch for ElfRiscV64 {
         ]
     }
 
+    fn fill_nop_padding(buf: &mut [u8], offset: usize, len: usize) {
+        // Fill with 4-byte NOP.
+        let mut i = 0;
+        while i + 4 <= len {
+            buf[offset + i..offset + i + 4].copy_from_slice(&0x0000_0013u32.to_le_bytes());
+            i += 4;
+        }
+        // Fill a remaining 2-byte slot with c.nop. This is safe because a 2-byte
+        // remainder only arises when the assembler used c.nop in the original sled,
+        // which implies the C extension is enabled.
+        if i + 2 <= len {
+            buf[offset + i..offset + i + 2].copy_from_slice(&0x0001u16.to_le_bytes());
+        }
+    }
+
     fn is_symbol_variant_pcs(
         object: &<Self::Platform as Platform>::File<'_>,
         symbol_index: object::SymbolIndex,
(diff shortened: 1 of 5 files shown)


libwild/src/elf_writer.rs
```diff
@@ -2540,20 +2540,11 @@ fn apply_relocation<
         RelocationKind::None => return Ok(RelocationModifier::Normal),
         RelocationKind::Alignment => {
             let addend = addend as u64;
-            let address = section_address + rel.offset();
-            ensure!(
-                addend.is_power_of_two(),
-                "A power of 2 expected for Alignment relocation: {}",
-                addend
-            );
```
lapla-cogito, line 2548: This assertion was incorrect. The addend here represents the number of NOP bytes to be inserted, not an alignment value, so there's no guarantee it's a power of two.

libwild/src/platform.rs
```diff
@@ -142,6 +142,9 @@ pub(crate) trait Arch: Send + Sync + 'static {
         relax_deltas: Option<&SectionRelaxDeltas>,
     ) -> Option<Self::Relaxation>;
 
+    /// Fill `len` bytes of NOP padding at `offset` in `buf`.
+    fn fill_nop_padding(_buf: &mut [u8], _offset: usize, _len: usize) {}
```
lapla-cogito, line 146: While RISC-V is currently the only architecture using `RelocationKind::Alignment`, it's defined generically to allow for future use by other architectures.

libwild/src/elf_riscv64.rs
```diff
@@ -155,6 +155,17 @@ impl crate::platform::Arch for ElfRiscV64 {
...
+    fn fill_nop_padding(buf: &mut [u8], offset: usize, len: usize) {
+        let mut i = 0;
+        while i + 4 <= len {
+            buf[offset + i..offset + i + 4].copy_from_slice(&0x0000_0013u32.to_le_bytes());
+            i += 4;
+        }
+        if i + 2 <= len {
+            buf[offset + i..offset + i + 2].copy_from_slice(&0x0001u16.to_le_bytes());
+        }
+    }
```
lapla-cogito, line 171: This fills the padding area with as many 4-byte NOP instructions as possible, then fills the remaining bytes with c.nop. Since only scenarios where two bytes remain would occur when C extensions are enabled, this method should remain valid even when supporting architectures other than rv64gc in the future.

davidlattimore: That seems like a useful comment. Do you think it'd be worthwhile adding something along those lines as a comment in the code?

libwild/src/elf_riscv64.rs
```diff
@@ -384,9 +395,47 @@ fn collect_relaxation_deltas<R: Relocation>(
...
+                    } else {
+                        d.deltas()[idx - 1].cumulative_deleted
+                    }
+                });
+                let output_offset = input_offset - existing_cum - new_cumulative_delta;
+                let p = section_output_address + output_offset;
+
+                // The alignment the assembler requested.
+                let alignment = addend.next_power_of_two();
+                let desired = (p + alignment - 1) & !(alignment - 1); // align_up
```
davidlattimore: Might be more readable to use `next_multiple_of`

davidlattimore (approved):

PR #1773, opened 2026-03-28
chore: Bump MSRV to 1.94

lapla-cogito:
~~Plus, starting from this version, Cargo now supports TOML 1.1, enabling line breaks within inline tables.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,11 +46,11 @@ git-version = "0.3.9"
 glob = "0.3.2"
 hashbrown = "0.16.0"
 hex = "0.4.0"
-iced-x86 = {
-    version = "1.17.0",
-    default-features = false,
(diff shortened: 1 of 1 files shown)


lapla-cogito:
Until `builtins.fromTOML` in Nix is updated, this line cannot parse TOML 1.1 files properly.

https://github.com/wild-linker/wild/blob/23e45ecd23c25a40a390d3c675c20e77ac77d967/nix/default.nix#L19

What it does is simply extract Wild's version, so you could resolve this issue by either hardcoding the version or parsing files that ignore everything after `[workspace.dependencies]`. However, both solutions are hacky. For this reason, I've reverted the change to `Cargo.toml`.

davidlattimore (approved):

Issue #1662, opened 2026-03-07
Allow linker plugins to print full log messages


davidlattimore:
Split off from https://github.com/wild-linker/wild/issues/1572. Currently when the linker plugin logs a message, we only print the format string. It'd be good if we could print the whole message. First step is to investigate options. The two options mentioned in the other issue are to compile a bit of C code or inline assembly. But there might be other options, so a bit of investigation would be worthwhile.

mati865:
Hmm, I found https://crates.io/crates/va_list and it looks promising. I might give it a try somewhat soonish on x86_64 and perhaps other architectures via QEMU.

mati865:
Also worth mentioning that `c_variadic` feature may be near stabilisation, so we'd only need a stopgap solution.

davidlattimore:
Interesting. I guess what we ideally need is a way to forward the valist to `snprintf`, although in a pinch I guess we could implement a partial printf implementation that only processed a few basics lik %s and %d.

mati865:
I've looked into it a bit and looks like C-shim is unavoidable for now. There is no way (even with nightly) to create function that accepts varags (like `args: ...`) and turns that into `VaList`. Which rules out `va_list` entirely.

Then I tried nightly's `c_varadic` to accept varags and just pass them down to `libc::sprintf`, but that seemed to replace format specifiers with Data Link Exchange character. So, `LLVM gold plugin has failed to create LTO module: %s` would turn into `Linker plugin fatal error: LLVM gold plugin has failed to create LTO module:` with the final bytes being (decimal values): `58, 32, 16, 0`. So, unless I messed up something, varargs might have some compatibility issues.

PritamP20:
Hi @davidlattimore @mati865 , I was looking through this issue and it seems interesting. I’d like to investigate the possible approaches for forwarding the variadic arguments (maybe via a small C shim with `vsnprintf`) and try prototyping something

PritamP20:
Hi @davidlattimore @mati865, I looked into this a bit.

Based on what @mati865 found the va_list crate can't capture varargs, and c_variadic on nightly produces corrupted output when forwarding to vsnprintf I tried the C shim approach locally. The idea is to capture the va_list in C, format the message with vsnprintf, and then pass the formatted string back to Rust.

I implemented it and tested it, and it works plugin messages now show the actual values instead of the raw %s placeholders

mati865:
[...]

PR #1774, opened 2026-03-28
fix: format linker plugin messages via C shim and prints full log messages.

Blazearth:
#1662 fixes linker plugin message handling by correctly formatting printf-style variadic arguments.
Introduced a small C shim that accepts the variadic arguments and formats the message via vsnprintf and sends the formatted string back to rust.
Works on stable rust and avoids c_variadics until its stable.
Added a test to verify fomatting correctness.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,6 +863,7 @@ dependencies = [
  "blake3",
  "bumpalo-herd",
  "bytesize",
+ "cc",
  "colored",
  "colosseum",
  "crossbeam-queue",
--- a/libwild/Cargo.toml
+++ b/libwild/Cargo.toml
@@ -56,6 +56,9 @@ zstd = { workspace = true }
 [target.'cfg(all(target_os = "linux", any(target_arch = "x86_64", target_arch = "aarch64")))'.dependencies]
 perf-event = { workspace = true }
 
+[build-dependencies]
+cc = "1.1.12"
+
 [dev-dependencies]
 ar = { workspace = true }
 tempfile = { workspace = true }
--- a/libwild/build.rs
+++ b/libwild/build.rs
@@ -1,0 +1,7 @@
+fn main() {
+    if std::env::var("CARGO_FEATURE_PLUGINS").is_ok() {
+        cc::Build::new()
+            .file("src/plugin_message_shim.c")
+            .compile("plugin_message_shim");
+    }
+}
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -361,7 +361,12 @@ impl LoadedPlugin {
         // Linker plugins handle entries of this vector serially, which means the message callback
         // should be registered first. Otherwise, they won't be able to indicate the problem with
         // entries preceding the callback and, for example, silently skip invalid arguments.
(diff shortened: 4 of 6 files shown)


Blazearth:
Should I remove  `test_plugin_message_shim.c` since the existing integration test covers the fix? Or is there value in keeping it as a standalone C-level unit test?"

davidlattimore:
Thanks for implementing this. The PR looks good to me. Regarding the C test, I'd say given that it's not being executed when we run `cargo test`, that we should perhaps delete it. You could however copy its contents into a comment on this PR for easy access if we want it in future.

Blazearth:
```c
/*
 * Standalone test for wild_plugin_message_callback.
 * Simulates what a linker plugin does: calls the callback with printf-style args.
 * Verifies that wild_handle_plugin_message receives the fully-formatted string.
 *
 * Compile and run (from workspace root):
 *   gcc -o /tmp/test_shim libwild/tests/test_plugin_message_shim.c libwild/src/plugin_message_shim.c && /tmp/test_shim
 */

#include <stdio.h>
#include <stdlib.h>
#include <string.h>

/* Track what Rust would receive */
static int received_level = -1;
static char received_message[256];

/* Stub for the Rust function */
void wild_handle_plugin_message(int level, const char *message) {
    received_level = level;
    strncpy(received_message, message, sizeof(received_message) - 1);
    received_message[sizeof(received_message) - 1] = '\0';
}

/* The function under test */
void wild_plugin_message_callback(int level, const char *fmt, ...);

static int passed = 0;
static int failed = 0;

static void check(const char *test, int level, const char *expected) {
    if (received_level == level && strcmp(received_message, expected) == 0) {
        printf("PASS: %s\n", test);
        passed++;
    } else {
        printf("FAIL: %s\n  expected level=%d msg=\"%s\"\n  got     level=%d msg=\"%s\"\n",
               test, level, expected, received_level, received_message);
        failed++;
    }
    /* reset */
    received_level = -1;
    received_message[0] = '\0';
}

int main(void) {
    /* Test 1: plain string, no format args */
    wild_plugin_message_callback(0, "hello world");
    check("plain string", 0, "hello world");

    /* Test 2: %s substitution */
    wild_plugin_message_callback(1, "failed at %s", "main.c");
    check("string substitution", 1, "failed at main.c");

    /* Test 3: %s and %d substitution */
    wild_plugin_message_callback(2, "error at %s:%d", "foo.c", 42);
    check("string+int substitution", 2, "error at foo.c:42");

    /* Test 4: error level (2) */
    wild_plugin_message_callback(2, "undefined symbol: %s", "my_func");
    check("error level", 2, "undefined symbol: my_func");

    /* Test 5: fatal level (3) */
    wild_plugin_message_callback(3, "fatal: %s", "out of memory");
    check("fatal level", 3, "fatal: out of memory");

    /* Test 6: format string with no specifiers (safe, no args) */
    wild_plugin_message_callback(0, "no specifiers here");
    check("no specifiers", 0, "no specifiers here");

    printf("\n%d passed, %d failed\n", passed, failed);
    return failed > 0 ? 1 : 0;
}
```
Test that can be used to verify the variadic handling

PR #1776, opened 2026-03-29
fix: null-terminate linker identity string in .comment section

### LINKED ISSUE #1775: Linker identity is written without null-termination
I just noticed this while reading the code... we write the linker name and version into the .comment sections. The comment in `indentity.rs` says that it returns a "null-terminated string", however it doesn't and nothing else is adding that null byte. It's made less obvious because the comment we add tends to end up as the last string in the `.comment` section, but we should properly null-terminate the string.

The fix should be trivial, but we should also extend our testing to pick this up. One place this check could be added is in `Assertions::check_path`  in `integration_tests.rs`.

Let us know if you'd like to work on this so that we can assign it to you.

F0RREALTHO:
Fixes #1775

The `linker_identity()` function in `identity.rs` was documented as 
returning a null-terminated string, but was not actually adding the 
null byte.

Changes:
- Added `\0` at the end of the format string in `identity.rs`
- Added a test in `verify_comment_section` in `integration_tests.rs` 
  that verifies the `.comment` section ends with a null byte when 
  linked with Wild

Tested locally with `cargo test comment` — all tests pass.

### DIFF
--- a/libwild/src/identity.rs
+++ b/libwild/src/identity.rs
@@ -1,6 +1,6 @@
 use git_version::git_version;
 
-/// Returns a string that identifies this linker.
+/// Returns a string that identifies this linker. This is written into the .comment
 /// section which usually also contains the versions of compilers that were used.
 pub(crate) fn linker_identity() -> String {
     let mut git_hash = git_version!(


libwild/src/identity.rs
```diff
@@ -12,7 +12,7 @@ pub(crate) fn linker_identity() -> String {
         git_hash = format!("({git_hash}) ");
     }
     format!(
-        "Wild version {} {}(compatible with GNU linkers)",
+        "Wild version {} {}(compatible with GNU linkers)\0",
```
lapla-cogito: This means returning a `String` containing a null char. Since the caller already adds a "Linker:" prefix, I think `linker_identity()` should just return the identity str without null-termination. How about shifting the null-termination responsibility to [the caller in `layout.rs`](https://github.com/wild-linker/wild/blob/f7441c81e0471bf4a45ba12204edb5189c49c303/libwild/src/layout.rs#L2607)?  (Alternatively, returning a `CString` could also work.) If the former is adopted, the doc comment should also be updated.

wild/tests/integration_tests.rs
```diff
@@ -3054,6 +3054,16 @@ impl Assertions {
         if obj.kind() == ObjectKind::Relocatable {
             return Ok(());
         }
+        // Verify that the linker identity string is null-terminated.
+        if linker_used.is_wild() {
+            let comment_section = obj.section_by_name(".comment");
+            if let Some(section) = comment_section {
+                let data = section.data()?;
+                if !data.is_empty() && *data.last().unwrap() != 0 {
```
lapla-cogito: This might be more idiomatic.
```suggestion
                if !data.is_empty() && data.last() != Some(&0) {
```

libwild/src/identity.rs
F0RREALTHO: Good point! Moved the null-termination to `layout.rs` where the identity 
string is constructed, and updated the doc comment in `identity.rs`. I also 
traced the write path through `elf_writer.rs` and confirmed that `prelude.identity` 
is written directly to the binary via `copy_from_slice`, so this is the 
correct place to ensure null-termination.

wild/tests/integration_tests.rs
F0RREALTHO: Updated, thanks!

libwild/src/identity.rs
```diff
@@ -1,6 +1,6 @@
 use git_version::git_version;
 
-/// Returns a null-terminated string that identifies this linker. This is written into the .comment
+/// Returns a string that identifies this linker.
 /// section which usually also contains the versions of compilers that were used.
```
lapla-cogito, line 4: The comment is broken. "This is written ~~" should be restored.

lapla-cogito (commented):
Only one nit. Otherwise LGTM.

libwild/src/identity.rs
F0RREALTHO: Fixed!

PR #1778, opened 2026-03-30
build: Change version format

davidlattimore:
* Release builds now don't include git hash
* We now indicate if git info was not available
* Git hash is now always an origin/main ancestor
* The version is now supplied via the API rather than built-in to libwild


davidlattimore:
Example version strings:

```
Wild 0.8.1 (compatible with GNU linkers)

Wild 0.8.0 c32cbfc220b28ac7831c8515565260c01511ebb5-modified (compatible with GNU linkers)

Wild 0.8.0 non-git-build (compatible with GNU linkers)

Wild integration-test (compatible with GNU linkers)
```

My initial motivation for this change was that I noticed that when editing test files, cargo test was still rebuilding libwild, wild and the integration test. This turned out to be because the git hash had changed because I had edited a file. The git hash was likely changing because my editor has a JJ extension active, which was querying JJ, causing it to snapshot the changes, thus changing the git hash.

With this change, we no longer rerun in the above case. If we do rerun, it's now much faster since we at most rebuild the main binary, which is a lot smaller and thus a lot faster to build than libwild.

I think also, that showing a commit hash that's an ancestor of origin/main and thus part of our main repo is likely to be more useful than a commit hash local to the user's clone that we have no idea where it sits in the wild commit history.

Issue #1769, opened 2026-03-26
Support missing built-in functions (`ALIGNOF`, `LOADADDR`)


plasmaDestroyer:
Wild currently handles only `SIZEOF` and `ADDR` but is missing `ALIGNOF`, `LOADADDR`, `LENGTH`, and `ORIGIN`. I'd like to add support for these.

This issue tracks:
- [x] `ALIGNOF` - #1770 
- [x] `LOADADDR` -  #1780 - currently implemented as an alias for `ADDR` (returns VMA). Full implementation requires `AT(expr)` support, which is not yet implemented.

davidlattimore:
Given that these might have a bit of logic beyond parsing, I'd suggest considering doing one per PR

plasmaDestroyer:
> Given that these might have a bit of logic beyond parsing, I'd suggest considering doing one per PR

Another thing I realised while looking into these is that we require parsing `MEMORY` blocks first, which doesn't exist yet. Should I open a separate issue for that, or just include it in the same PR?

davidlattimore:
> Should I open a separate issue for that, or just include it in the same PR?

Definitely shouldn't be in the same PR, but I suspect you meant "issue" not PR?

I don't have strong opinions about granularity of issues. But are you sure those things can't be implemented without `MEMORY`? It seems to me like they probably could. Why don't you try writing a linker script that uses them without a `MEMORY` block and see if GNU ld accepts it?

plasmaDestroyer:
I tested it, GNU ld errors with "memory region not declared" when `LENGTH` is used without a `MEMORY` block, so it does seem to me like `MEMORY` parsing is a prerequisite.

davidlattimore:
You're right that LENGTH and ORIGIN require memory regions to be defined, but it looks like ALIGNOF and LOADADDR both take sections. Maybe it would make sense to reduce the scope of this issue to just implementing the section-based functions and leave the memory region functions for a separate issue.

plasmaDestroyer:
hey @davidlattimore, one issue currently is that `LOADADDR` returns the LMA of a section, which differs from `ADDR` only when `AT(expr)` is used and since `AT(expr)` isn't implemented yet, would it be acceptable to implement `LOADADDR` as an alias for `ADDR` for now, emitting a warning if the result may be inaccurate, to keep scope limited for now? It could be updated to return the actual LMA once `AT(expr)` is supported.

[...]

PR #1780, opened 2026-03-31
feat: support LOADADDR in linker scripts

plasmaDestroyer:
Implements `LOADADDR` as an alias for `ADDR` for now. Since `AT(expr)` is not yet implemented, `LOADADDR` and `ADDR` return the same value (the VMA of the section), I have left a comment in the code that this needs to be updated when `AT(expr)` is supported. Part of #1769.


davidlattimore:
Thanks for doing this. Where possible, can you prefix your PR titles based on which section of the release notes it should be in? See `CONTRIBUTING.md` for details. I try to update the title when I merge, but I sometimes forget.

PR #1783, opened 2026-03-31
feat: generate a .symtab_shndx section when shnum > 65k

vishruth-thimmaiah:
This PR adds support for generating a `.symtab_shndx` when the number of sections crosses SHN_LORESERVE. Additionaly we also set the header's `shnum` to 0 and `shstrndx` to SHN_XINDEX, using the first section's `sh_size` and `sh_link` fields respectively instead.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1896,6 +1896,33 @@ impl platform::Platform for Elf {
             crate::elf::NON_PIE_START_MEM_ADDRESS
         }
     }
+
+    fn requires_symtab_shndx(num_sections: usize) -> bool {
+        num_sections >= object::elf::SHN_LORESERVE as usize
+    }
+
+    fn compute_symtab_shndx_section_size(
+        group_sizes: &mut OutputSectionPartMap<u64>,
+        total_sizes: &mut OutputSectionPartMap<u64>,
+    ) {
+        let symtab_entry_size = size_of::<Self::SymtabEntry>() as u64;
+        let symtab_shndx_entry_size = size_of::<Self::SymtabShndxEntry>() as u64;
+        let locals = group_sizes.get(part_id::SYMTAB_LOCAL) / symtab_entry_size;
+        let globals = group_sizes.get(part_id::SYMTAB_GLOBAL) / symtab_entry_size;
+
+        let mut extra_sizes = OutputSectionPartMap::with_size(group_sizes.num_parts());
+        extra_sizes.increment(
+            part_id::SYMTAB_SHNDX_LOCAL,
+            locals * symtab_shndx_entry_size,
+        );
+        extra_sizes.increment(
+            part_id::SYMTAB_SHNDX_GLOBAL,
+            globals * symtab_shndx_entry_size,
+        );
+
+        group_sizes.merge(&extra_sizes);
+        total_sizes.merge(&extra_sizes);
+    }
 }
 
 impl<'data> platform::ObjectFile<'data> for File<'data> {
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -1732,7 +1732,11 @@ fn compute_total_section_part_sizes<'data, P: Platform>(
         .expect("we should have computed header info by now")
         .num_output_sections_with_content;
 
-    compute_shndx_section(group_states, &mut total_sizes, num_sections);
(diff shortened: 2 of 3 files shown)


libwild/src/elf.rs
```diff
@@ -1294,9 +1295,20 @@ impl platform::Platform for Elf {
...
     fn finalise_sizes_all<'data>(
         mem_sizes: &mut OutputSectionPartMap<u64>,
+        output_sections: &OutputSections<Self>,
         symbol_db: &SymbolDb<'data, Self>,
     ) {
         finalise_gnu_version_size(mem_sizes, symbol_db);
+        if output_sections.has_data_in_file(SYMTAB_SHNDX_LOCAL) {
+            mem_sizes.increment(
+                part_id::SYMTAB_SHNDX_LOCAL,
+                *mem_sizes.get(part_id::SYMTAB_LOCAL) / 24 * 4,
```
davidlattimore: Nicer if we can use existing constants here. e.g. `SYMTAB_ENTRY_SIZE`. Then I guess the 4 possibly needs a new constant.

libwild/src/layout.rs
```diff
@@ -5125,3 +5148,32 @@ impl OutputRecordLayout {
 impl<'data, P: Platform> Drop for Layout<'data, P> {
     fn drop(&mut self) {}
 }
+
+fn compute_shndx_section<'data, P: Platform>(
+    group_states: &mut [GroupState<'data, P>],
+    total_sizes: &mut OutputSectionPartMap<u64>,
+    num_sections: u32,
+) {
+    if num_sections < u32::from(object::elf::SHN_LORESERVE) {
```
davidlattimore: We want to avoid references to `object::elf` from platform-generic code. Probably the whole shndx mechanism is elf-specific, so maybe this function should be moved to `elf.rs`, most likely by adding a new method to the `Platform` trait.

libwild/src/layout.rs
```diff
@@ -2991,7 +3010,11 @@ impl<'data, P: Platform> PreludeLayoutState<'data, P> {
             }
         }
 
-        let num_sections = keep_sections.values_iter().filter(|p| **p).count();
+        let mut num_sections = keep_sections.values_iter().filter(|p| **p).count();
+        if num_sections >= object::elf::SHN_LORESERVE as usize {
```
davidlattimore: Can this bit of elf-specific code be moved to `elf.rs` by adding a new method to the `Platform` trait?

libwild/src/elf_writer.rs
```diff
@@ -1283,39 +1300,64 @@ impl<'layout, 'out> SymbolTableWriter<'layout, 'out> {
...
+            if let Some(s) = symtab_shndx_entries {
+                *s = 0;
+            }
+            shndx as u16
+        } else {
+            symtab_shndx_entries.map(|s| *s = shndx).with_context(|| {
+                format!("Expected .symtab_shndx section when writing symbol {} with shndx set to SHN_XINDEX.", String::from_utf8_lossy(name))
+            })?;
+            object::elf::SHN_XINDEX
+        };
```
philipc, line 1360: I'm not sure that this logic is correct (and I'm not sure how to go about testing the u16 section overflow to verify). The problem is that we're using a u32 to encode two different kinds of values: a section index (which can logically be any value in the u32 range), and reserved index values such as `SHN_ABS` which have values in the 0xfff0-0xffff range, and these value ranges overlap. So if we happened to have over 0xfff0 sections, we would wrongly encode one of them as `SHN_ABS` instead of `SHN_XINDEX`.

davidlattimore: Thanks for pointing that out. You're right, both that this isn't right as it currently stands and that testing it is likely to be hard. From a testing perspective, we probably need some way to define tests that compile generated code, which we don't currently support. That way, we can generate an input file with a sufficiently large number of sections. I guess the other problem is that the sections aren't especially used at runtime, so there's not much we can do at runtime to check that they're right. Probably we need to rely on test assertions to verify that it's correct.

vishruth-thimmaiah: We can probably use macros in c to generate >65k sections. To verify the symbols' section indexes, we can probably use the `ExpectSym`? I can take a look at this and submit a PR.

davidlattimore: That sounds like it would work. Thanks!

PR #1788, opened 2026-04-01
fix: do not allocate space for uninitialized tls variables

ostylk:
I noticed the tls-common test failed on my machine depending on which environment and versions of compilers I used. It seems like the ``TLS`` program header is generated incorrectly for zero initialized tls variables.

E.g. in the tls-common we have a variable `tvar` that is in `.tbss`. However, the TLS program header is `filesz=4,memsz=4` which doesn't zero-initialize it but loads bytes from the elf binary. It just so happens that all CIs and your local setup contained zeros in the binary at the exact place for this to nevertheless work. The correct TLS program header would be `filesz=0,memsz=4`.

So this pull request does/will do two things:

- [x] Implement program header assertions in testing so this bug actually errors out deterministically. Now a test can contain `//#ExpectProgramHeader:PT_TLS filesz=0,memsz=4` and it will only accept if a program header with at least these properties exists. Implemented by https://github.com/wild-linker/wild/pull/2190 .
- [x] Fix the actual bug.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2367,8 +2367,6 @@ impl<'data> platform::ObjectFile<'data> for File<'data> {
 
         if let Some((compression, _, _)) = section.compression(LittleEndian, self.data)? {
             decompress_into(compression, &data[COMPRESSION_HEADER_SIZE..], out)?;
-        } else if section.sh_type(LittleEndian) == object::elf::SHT_NOBITS {
-            out.fill(0);
         } else {
             copy_section_data(data, out);
         }
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -1947,22 +1947,9 @@ fn compute_segment_layout<'data, P: Platform>(
                         rec.mem_start = rec.mem_start.min(section_layout.mem_offset);
                         rec.lma_start = rec.lma_start.min(section_layout.lma_offset);
 
-                        // `has_data_in_file` reports that .tbss has data in file, since this makes
-                        // address allocation easier - see comment in `has_data_in_file`. However,
-                        // we don't want .tbss to increase the file-size of the program segment even
-                        // though we do actually have zeros in the file. If we ever change the
-                        // behaviour of `has_data_in_file`, then this can likely be cleaned up.
-                        let section_file_size = if program_segments.is_tls_segment(rec.segment_id)
-                            && section_info.section_attributes.is_no_bits()
-                        {
-                            0
-                        } else {
-                            section_layout.file_size
-                        };
-
                         rec.file_end = rec
                             .file_end
-                            .max(section_layout.file_offset + section_file_size);
+                            .max(section_layout.file_offset + section_layout.file_size);
                         rec.mem_end = rec
                             .mem_end
                             .max(section_layout.mem_offset + section_layout.mem_size);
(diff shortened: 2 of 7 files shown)


davidlattimore:
Hi @ostylk. Do you intend to return to this PR?

ostylk:
Hi, thank you for the reminder, this PR slipped my mind.
I do intend on returning to it, preferably in this month. Am I impeding anything by working on this slowly?

davidlattimore:
All good, I was just checking in on PRs that hadn't been touched in a while. I don't think there's any particular hurry on this though.

ostylk:
Ready for review but I have some comments/questions:

1. The CI failure is weird, it only happens on opensuse (I could reproduce it locally with a container). Any pointers on how to approach that?
2. There was a comment in output_section_id.rs about how uninitialized TLS data is padded with zeroes in the output file. However, I couldn't find where that happens in the code? Maybe it was forgotten and that caused this bug in the first place?
3. Funnily enough GNU ld seems to have a bug (to see for yourself: delete the SkipLinker instruction in tls-custom.c test) where uninitialized TLS sections overlap initialized TLS sections. LLD seems to handle the case correctly though.

davidlattimore:
Welcome back!

Regarding the CI failure - RexMovIndirectToAbsolute vs MovIndirectToLea, that sounds like something that could be caused by different defaults as to whether the binary is position independent or not. So possibly the difference might go away if you override those defaults - e.g. by passing a flag to make both linkers produce a position-independent binary.

The PR in which we started putting zero bytes in the file for .tbss was https://github.com/wild-linker/wild/pull/489. It unfortunately looks like I didn't add a test for it, so it's not impossible that something around there could have regressed.

ostylk:
Okay, so the linker failure seems to be due to lld behaving differently to GNU ld on openSUSE. While I am unsure what the *exact* reason is, the startup files (`crtbegin.o` `crtend.o`) on openSUSE cannot be linked with ``--pie``. This leads to GNU ld and lld applying different relaxations. GNU ld and also wild choose to relax like this:
```
0000000000401020 <_start>:
  <snip>
  401038:       48 c7 c7 16 11 40 00    mov    $0x401116,%rdi
  40103f:       ff 15 93 2f 00 00       call   *0x2f93(%rip)        # 403fd8 <__libc_start_main@GLIBC_2.34>
```
while lld relaxes to this (effectively ignoring pie/no-pie stuff)
```
0000000000201570 <_start>:
  <snip>
  201588:       48 8d 3d d7 00 00 00    lea    0xd7(%rip),%rdi        # 201666 <main>
  20158f:       ff 15 03 1b 00 00       call   *0x1b03(%rip)        # 203098 <__libc_start_main@GLIBC_2.34>
```
So ultimately I think it is best to ignore that difference. GNU ld and wild switch to absolute addresses with no-pie files while lld ignores that fact completely.

davidlattimore (commented):
Thanks for working on this. The change looks good, so I'll merge it. Before I do though, how do you want this to appear in the release notes? Can you update the PR title to read as a line from the release notes, then prefix it with "fix:" or "feat:" depending on whether you want it in the bug fixes section or the features section.

ostylk:
Done and thank you for your patience with this PR.

davidlattimore (approved):

Issue #1781, opened 2026-03-31
Support `MEMORY` block parsing and `LENGTH`/`ORIGIN` linker script functions


plasmaDestroyer:
Wild is missing `LENGTH(region)` and `ORIGIN(region)`, which return the length and origin address of a named `MEMORY` region respectively. These also required `MEMORY` block parsing, which is also not yet implemented.

Tracking:
- [x] `MEMORY` block parsing
- [x] `LENGTH` and `ORIGIN` functions

plasmaDestroyer:
Right now, `evaluate_expression` takes `section_layouts` and `output_sections` as parameters. To evaluate `LENGTH` and `ORIGIN`, it needs to know the values populated by the `MEMORY` block commands. After parsing `MEMORY` blocks into `MemoryRegion` structs, I can think of 2 ways we could go around this - either we add a `&[MemoryRegion]` that is passed into `evaluate_expression`, or we can expand the `OutputSections` struct to track available memory regions in some way. The first option seems more reasonable to me but i want to check if there's a better implementation for this that I might be missing.

davidlattimore:
I'd say don't worry too much about `evaluate_expression` until after `MEMORY` has been implemented - these should be separate PRs.

PR #1792, opened 2026-04-02
feat: add MEMORY block parsing and LENGTH/ORIGIN expression variants

plasmaDestroyer:
Adds parsing support for `MEMORY` blocks in linker scripts, along with `Expression::Length` and `Expression::Origin` variants. The parser handles the full `MEMORY { ... }` syntax including all keyword aliases (`org`/`o`, `len`/`l`). This is the first step toward fully implementing these functions (part of #1781).

Evaluation is currently `todo!()` in `evaluate_expression`, it will need access to the parsed memory regions to implement it. As mentioned earlier, I can think of 2 ways we could go around this - either we add a &[MemoryRegion] that is passed into evaluate_expression, or we can expand the OutputSections struct to track available memory regions in some way. Happy to proceed with whichever seems better in this case.


marxin:
@plasmaDestroyer can we drop part of the following comment?
https://github.com/wild-linker/wild/blob/e6e8f983227732f478780743384962cd68aee10d/libwild/src/linker_script.rs#L161-L163

PR #1795, opened 2026-04-03
port(MachO): section/segment mapping and first final output

marxin:
The second Mach-O port PR implements the concept of sections, segments and I'm able to reach the point a final output file is produced. So far, the designed `Platform` trait has guided me pretty well.

First, let's start a bit with the Mach-O layout explanation and how I map it to the Wild's essential concepts (`PartId` that is 1:1 mapping to `OutputSectionId`; and the `ProgramSegmentDef`):
mach64_structure

The mapping is following:
- `Mach Header` is `part_id::FILE_HEADER`
- `__TEXT_SEGMENT` segment_command_64 is mapped to `part_id::TEXT_SEGMENT`
- `__DATA_SEGMENT` segment_command_64 is mapped to `part_id::DATA_SEGMENT`
- `__text` section data is mapped to `part_id::TEXT`
- `__data` section data is mapped to `part_id::DATA`

and the following program segments are created (with the clarification how are they used):
- `SegmentType::Header` - contains only the `part_id::FILE_HEADER`
- `SegmentType::LoadCommand` - contains all load commands (`__TEXT_SEGMENT`, `__DATA_SEGMENT`, `__PAGEZERO_SEGMENT`) - we need the information about the number of load commands (and their total size)
- `SegmentType::Text` - contains `__text` section content (and will include all other sections defined in __TEXT segment in input files) - based on that we can fill-up the `__TEXT_SEGMENT` segment_command_64 (number of sections, names, their offset in memory and file)

The suggested approach works fine, except I feel the section iteration of a segment is a bit cumbersome and it heavily depends on the `ProgramSegmentDef`. I'm going to mark methods in the code that might be improved.

After a small tweak to the ELF part, I was able to introduce an extra Mach-O specific `part_id`s and so we can basically live with a sparse output section maps. Plus, a lifetime of a method was adjusted based on Mach-O needs.

Right now, I am pretty strict when it comes to the supported segments and sections, but I believe we can get pretty far as the number of used sections is limited. At least to a point, where a simple hello world program will run. Once having that, we can fill the missing gaps.

There are still many cut corners, but it's pretty neat the PR reaches a point where a final output file scaffolding is laid out:
```
❯ otool -l a.out
a.out:
Load command 0
      cmd LC_SEGMENT_64
  cmdsize 72
  segname __PAGEZERO
   vmaddr 0x0000000000000000
   vmsize 0x0000000100000000
  fileoff 0
 filesize 0
  maxprot 0x00000000
 initprot 0x00000000
   nsects 0
    flags 0x0
Load command 1
      cmd LC_SEGMENT_64
  cmdsize 232
  segname __TEXT
   vmaddr 0x00000000004001e8
   vmsize 0x0000000000000049
  fileoff 488
 filesize 73
  maxprot 0x00000000
 initprot 0x00000000
   nsects 2
    flags 0x0
Section
  sectname __text
   segname __TEXT
      addr 0x00000000004001e8
      size 0x0000000000000038
    offset 488
     align 2^0 (1)
    reloff 0
    nreloc 0
     flags 0x00000000
 reserved1 0
 reserved2 0
Section
  sectname __cstring
   segname __TEXT
      addr 0x0000000000400220
      size 0x0000000000000011
    offset 544
     align 2^0 (1)
    reloff 0
    nreloc 0
     flags 0x00000000
 reserved1 0
 reserved2 0
Load command 2
      cmd LC_SEGMENT_64
  cmdsize 152
  segname __DATA
   vmaddr 0x0000000000400280
   vmsize 0x0000000000000100
  fileoff 640
 filesize 256
  maxprot 0x00000000
 initprot 0x00000000
   nsects 1
    flags 0x0
Section
  sectname __data
   segname __DATA
      addr 0x0000000000400280
      size 0x0000000000000100
    offset 640
     align 2^0 (1)
    reloff 0
    nreloc 0
     flags 0x00000000
 reserved1 0
 reserved2 0
```

Issue #757

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -242,7 +242,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Spell Check Repo
-        uses: crate-ci/typos@v1.44.0
+        uses: crate-ci/typos@v1.45.0
 
   calc-matrix:
     name: Find Nix checks
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,9 +196,9 @@ checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
 
 [[package]]
 name = "blake3"
-version = "1.8.3"
+version = "1.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2468ef7d57b3fb7e16b576e8377cdbde2320c60e1491e961d11da40fc4f02a2d"
+checksum = "4d2d5991425dfd0785aed03aedcf0b321d61975c9b5b3689c774a2610ae0b51e"
 dependencies = [
  "arrayref",
  "arrayvec",
@@ -259,9 +259,9 @@ checksum = "6bd91ee7b2422bcb158d90ef4d14f75ef67f340943fc4149891dcce8f8b972a3"
 
 [[package]]
 name = "cc"
-version = "1.2.56"
+version = "1.2.59"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aebf35691d1bfb0ac386a69bac2fde4dd276fb618cf8bf4f5318fe285e821bb2"
+checksum = "b7a4d3ec6524d28a329fc53654bbadc9bdd7b0431f5d65f1a56ffb28a1ee5283"
 dependencies = [
  "find-msvc-tools",
  "jobserver",
@@ -390,9 +390,9 @@ dependencies = [
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.17"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
 dependencies = [
  "libc",
 ]
@@ -624,9 +624,9 @@ checksum = "e629b9b98ef3dd8afe6ca2bd0f89306cec16d43d907889945bc5d6687f2f13c7"
 
 [[package]]
 name = "gimli"
-version = "0.33.0"
+version = "0.33.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0bf7f043f89559805f8c7cacc432749b2fa0d0a0a9ee46ce47164ed5ba7f126c"
+checksum = "19e16c5073773ccf057c282be832a59ee53ef5ff98db3aeff7f8314f52ffc196"
 dependencies = [
  "fnv",
  "hashbrown 0.16.1",
@@ -712,9 +712,9 @@ checksum = "3d3067d79b975e8844ca9eb072e16b31c3c1c36928edf9c6789548c524d0d954"
 
 [[package]]
 name = "indexmap"
-version = "2.13.0"
+version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7714e70437a7dc3ac8eb7e6f8df75fd8eb422675fc7678aff7364301092b1017"
+checksum = "45a8a2b9cb3e0b0c1803dbb0758ffac5de2f425b23c28f518faabd9d805342ff"
 dependencies = [
  "equivalent",
  "hashbrown 0.16.1",
@@ -796,9 +796,9 @@ checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
-version = "0.2.183"
+version = "0.2.184"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
+checksum = "48f5d2a454e16a5ea0f4ced81bd44e4cfc7bd3a507b61887c99fd3538b28e4af"
 
 [[package]]
 name = "libloading"
@@ -851,7 +851,7 @@ dependencies = [
  "derive_more",
  "flate2",
  "foldhash 0.2.0",
- "gimli 0.33.0",
+ "gimli 0.33.1",
  "glob",
  "hashbrown 0.16.1",
  "hex",
@@ -892,7 +892,7 @@ dependencies = [
  "anyhow",
  "clap",
  "colored",
- "gimli 0.33.0",
+ "gimli 0.33.1",
  "hashbrown 0.16.1",
  "iced-x86",
  "itertools",
@@ -1651,9 +1651,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "1.1.0"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "876ac351060d4f882bb1032b6369eb0aef79ad9df1ea8bc404874d8cc3d0cd98"
+checksum = "6662b5879511e06e8999a8a235d848113e942c9124f211511b16466ee2995f26"
 dependencies = [
  "serde_core",
 ]
@@ -1885,9 +1885,9 @@ dependencies = [
 
 [[package]]
 name = "toml"
-version = "1.1.0+spec-1.1.0"
+version = "1.1.2+spec-1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8195ca05e4eb728f4ba94f3e3291661320af739c4e43779cbdfae82ab239fcc"
+checksum = "81f3d15e84cbcd896376e6730314d59fb5a87f31e4b038454184435cd57defee"
 dependencies = [
  "indexmap",
  "serde_core",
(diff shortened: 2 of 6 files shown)


libwild/src/elf_writer.rs
```diff
@@ -1485,7 +1485,7 @@ fn build_sym_index_map(layout: &ElfLayout<'_>) -> Vec<Option<u32>> {
 }
 
 fn build_section_sym_indices(layout: &ElfLayout<'_>) -> OutputSectionMap<u32> {
-    let mut map = OutputSectionMap::with_size(layout.output_order.len());
+    let mut map = OutputSectionMap::with_size(layout.output_sections.num_sections());
```
marxin, line 1488: Now the sections are sparse for ELF, that's why not all `OutputSectionId`s are present in the `output_order`.

libwild/src/macho.rs
```diff
@@ -476,125 +529,212 @@ impl platform::SectionAttributes for SectionAttributes {
...
-        todo!()
+        false
     }
 
     fn order_key(self) -> usize {
-        todo!()
+        self.segment_type as usize
     }
 
     fn should_include_section(
```
marxin, line 643: This is the main section to segment assignment place. Dunno if we can do any better?

libwild/src/macho.rs
```diff
@@ -476,125 +529,212 @@ impl platform::SectionAttributes for SectionAttributes {
...
+    defs[output_section_id::DATA.as_usize()] = BuiltInSectionDetails {
+        kind: SectionKind::Primary(SectionName(b"__data")),
+        ..DEFAULT_DEFS
+    };
+
+    defs
+};
+
+// TODO: a better approach?
+fn count_sections_for_segment_type(
```
marxin: Needed for allocation of the `SegmentCommand` - can we do any better?

libwild/src/macho.rs
```diff
@@ -1045,6 +1215,48 @@ impl platform::Platform for MachO {
...
+
+// TODO: sort properly
+const DEFAULT_SECTION_RULES: &[SectionRule<'static>] = &[
+    SectionRule::exact_section_keep(b"__text", crate::output_section_id::TEXT),
+    SectionRule::exact_section_keep(b"__cstring", crate::output_section_id::CSTRING),
+    SectionRule::exact_section_keep(b"__data", crate::output_section_id::DATA),
+    // SectionRule::exact_section_keep(b"__compact_unwind", crate::output_section_id::EH_FRAME),
+];
+
+const PROGRAM_SEGMENT_DEFS: &[ProgramSegmentDef] = &[
```
marxin, line 1242: Wanted to point out the section -> segment mapping works as needed:
```
2026-04-03T12:47:43.399540Z TRACE Link:Layout: libwild::layout: Output order:
START(Header)
  `FILE_HEADER`
END(Header)
START(LoadCommand)
  `__PAGEZERO`
  `__TEXT`
  `__DATA`
END(LoadCommand)
START(Text)
  `__text`
  `__cstring`
END(Text)
START(Data)
  `__data`
END(Data)
```

Am I using the right approach for section to segment assignment?

libwild/src/macho_writer.rs
```diff
@@ -0,0 +1,299 @@
...
+            .map_err(|_| error!("Invalid PAGEZERO segment allocation"))?
+            .0;
+    write_pagezero_command::<A>(pagezero_command);
+
+    write_segment_commands::<A>(layout, buffers)?;
+
+    Ok(())
+}
+
+fn get_segment_sections<'data>(
```
marxin: Here similarly I need an ability to iterate sections that live in a specific segment (based on `SegmentType`).
Do we have any existing helper method I can rely on?

libwild/src/macho_writer.rs
```diff
@@ -0,0 +1,299 @@
...
+    let load_commands_size = layout
+        .segment_layouts
+        .segments
+        .iter()
+        .find(|seg| {
+            layout.program_segments.segment_def(seg.id).segment_type
+                == crate::macho::SegmentType::LoadCommand
+        })
+        .map_or(0, |seg| seg.sizes.file_size);
+    let commands = layout
```
marxin: Similar issue here.

libwild/src/output_section_id.rs
```diff
@@ -88,16 +88,19 @@ pub(crate) const NOTE_GNU_PROPERTY: OutputSectionId =
...
 pub(crate) const SYMTAB_LOCAL: OutputSectionId = part_id::SYMTAB_LOCAL.output_section_id();
-#[allow(dead_code)]
 pub(crate) const SYMTAB_GLOBAL: OutputSectionId = part_id::SYMTAB_GLOBAL.output_section_id();
 pub(crate) const RELA_DYN_RELATIVE: OutputSectionId =
     part_id::RELA_DYN_RELATIVE.output_section_id();
 pub(crate) const RELA_DYN_GENERAL: OutputSectionId = part_id::RELA_DYN_GENERAL.output_section_id();
 pub(crate) const RISCV_ATTRIBUTES: OutputSectionId = part_id::RISCV_ATTRIBUTES.output_section_id();
 pub(crate) const RELRO_PADDING: OutputSectionId = part_id::RELRO_PADDING.output_section_id();
+// Mach-O specific sections
+pub(crate) const PAGEZERO_SEGMENT: OutputSectionId = part_id::PAGEZERO_SEGMENT.output_section_id();
```
marxin, line 99: For `PAGEZERO_SEGMENT`, we should probably create an artificial `part_id::PAGEZERO` section that will be `1 << 32` of size, while not keeping any file size. The section will be then responsible for shifting the VM addresses. So something similar to BSS. How can I create such fixed empty section?

libwild/src/macho_writer.rs
davidlattimore: Nothing currently, since we don't ever need to do that with ELF. It's fine to add a helper to do that

libwild/src/output_section_id.rs
davidlattimore: Do we need a section in order to shift the start address. This sounds a bit like our `NON_PIE_START_MEM_ADDRESS` (in `elf.rs`). Or perhaps MachO doesn't have a header field for load address and just inserts a section of the appropriate size in order to skip to the desired address?

libwild/src/elf_writer.rs
davidlattimore: I guess that's one way to make sure ELF doesn't pay any cost for the Mac port. However, mac would still be paying the cost of the ELF-specific IDs. I might investigate alternatives at some point, but this seems reasonable for now.

libwild/src/macho.rs
davidlattimore: Wild used to have code that said explicitly built up the list of `OrderEvent`s from code. This was changed to the current scheme in order to accomodate linker scripts, since we needed to be able to accept sections in any order and put them into segments appropriately. Given that AFAIK, linker scripts aren't something that exists on other platforms, we could if you wanted bypass the code that calls `should_include_section` and just allow the platform to build up the list of `OrderEvent`s via whatever means is appropriate for that platform.

I don't think building the list of order events is in any way performance sensitive, so what you've got here is probably fine from that perspective.

libwild/src/macho.rs
davidlattimore: Discussed on the comment on `should_include_section`

libwild/src/macho_writer.rs
```diff
@@ -0,0 +1,289 @@
...
+    header: &mut FileHeader,
+) {
+    // TODO: can we do better?
+    let load_commands_size = layout
+        .segment_layouts
+        .segments
+        .iter()
+        .find(|seg| {
+            layout.program_segments.segment_def(seg.id).segment_type
+                == crate::macho::SegmentType::LoadCommand
```
davidlattimore: Is the intention here to just find the size of the first load command? From the variable name ("commands" rather than "command"), I'd have expected this to be filtering, then perhaps summing the sizes.

libwild/src/macho_writer.rs
davidlattimore: From the code, it looks like it's finding the first segment that isn't a LoadCommands segment, then counting the number of events until a SegmentEnd is reached. But it's not 100% obvious to me why that is what's wanted. Perhaps add a comment?

libwild/src/elf_writer.rs
marxin: Agree, right approach for now.

libwild/src/macho.rs
marxin: Good, good, I added a note to my TODO list and we can continue with what we've got right now. I also don't expect that being an expensive operation.

libwild/src/macho.rs
marxin: Keeping for now.

libwild/src/macho_writer.rs
marxin: Fine, so far I'm using 2 helper functions defined in `macho.rs`: `count_sections_for_segment_type` and `get_segment_sections`. Could be easily revisited (or introduced to the `Platform` trait) in the future.

libwild/src/macho_writer.rs
marxin: No no, it's correct. It's suppose to find all the load commands (each being a section in the Wild's nomenclature). The lookup is implemented by putting all of the commands into the `SegmentType::LoadCommand`. I'm gonna rename it to `LoadCommands` and it would better reflect there are multiple commands in the segment. Good point.

libwild/src/macho_writer.rs
marxin: Simplified by using the new helper method `get_segment_sections`.

libwild/src/output_section_id.rs
marxin: Good observation, we don't need it. That said, I introduce a new platform entry point `P::start_memory_address`.

Issue #1561, opened 2026-02-18
wild: error: unrecognized option(s): --use-android-relr-tags, --pack-dyn-relocs=relr


agrieve:
Thanks for fixing my previous Chrome bug! I've made it easier to try out non-lld linkers when building chrome with target_os="android"

With a [chromium checkout](https://chromium.googlesource.com/chromium/src/+/main/docs/android_build_instructions.md)

```
gn gen out/Debug --args 'target_os="android" is_component_build=false experimental_linker_path="/code/wild/ld"'
autoninja -C out/Debug libchrome.so
```

This fails with:

```
[...]

PR #1805, opened 2026-04-04
feat: Support `--use-android-relr-tags`

mati865:
cc https://github.com/wild-linker/wild/issues/1561


mati865:
Removed GitHub link to avoid it closing the issue.

davidlattimore (approved):

agrieve:
Thanks for your work on this! 

Just noticed that there's still a flag warning in chrome:

```
wild: warning: --pack-dyn-relocs=android+relr is not yet supported
```

We don't pass this flag, so it's clang that's adding it.

mati865:
Yeah, Clang does that when target is set to Android.

PR #1826, opened 2026-04-08
fix: import just `GLIBC_ABI_DT_RELR` version

mati865:
A self-contained mess to import just the version and not the whole symbol for #1817. If we want to go this route I'll try to clean this up, but there is not much that can be done.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -918,15 +918,6 @@ impl platform::Platform for Elf {
             name: b"_TLS_MODULE_BASE_",
             symbol: elf_symbol,
         });
-
-        // When `-z pack-relative-relocs` is used, Glibc requires this special version to be
-        // defined.
-        if args.z_pack_relative_relocs {
-            symbols.add_symbol(InternalSymDefInfo::new(
-                SymbolPlacement::ImportDynamicSymbol,
-                b"GLIBC_ABI_DT_RELR",
-            ));
-        }
     }
 
     fn built_in_section_infos<'data>()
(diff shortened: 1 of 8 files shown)


davidlattimore:
What about looking up the symbol, then loading the version for that symbol, but not loading the symbol itself? Something like the following:

```rust
        if symbol_db.args.is_relr_enabled()
            && let Some(symbol_id) =
                symbol_db.get_unversioned(&UnversionedSymbolName::prehashed(b"GLIBC_ABI_DT_RELR"))
        {
            let file_id = symbol_db.file_id_for_symbol(symbol_id);
            if let layout::FileLayoutState::Dynamic(state) =
                &mut groups[file_id.group()].files[file_id.file()]
            {
                let symbol_index = state.symbol_id_range.id_to_offset(symbol_id);
                let version_index = state.object.versym[symbol_index].0.get(LittleEndian);
                state.format_specific_state.symbol_versions_needed[version_index as usize - 1] = true;
            }
        }
```

This could be added somewhere like `finalise_find_required_sections`. It'd need to have the `SymbolDb` added as an argument and would also need the groups as a mutable reference.

mati865:
I'm going to sleep right now. You can post a new PR or push it here, I don't know how my availability later today (April 9 UTC time) will look like.

davidlattimore:
I pushed changes along the lines of what I suggested above to this PR

davidlattimore:
Given that you might have limited availability, I decided to just merge it, since I was happy with the state of it. Hope that's OK. Can always follow-up in a separate PR.

mati865:
Thanks for taking care of it and for the good solution.

PR #1837, opened 2026-04-13
fix: Don't bypass dynamic symbol lookup due to unused hidden in archive

davidlattimore:


mati865:
Whether purposely or accidentally, this fixes x86_64 error in https://github.com/wild-linker/wild/issues/1561.

davidlattimore:
In retrospect, I should have realised, but didn't. I was working on range extension thunks and thought I'd test on chrome, hit an issue that at first I thought was related to the thunk work, but then realised it wasn't, so separated it out. Sorry if I ended up taking the problem you were working on!

mati865:
No problem, it's good to have to it working.
I was a bit confounded when the issue stopped reproducing, but that's all.

PR #1838, opened 2026-04-13
fix: proper handling of .group sections in partial link mode

vishruth-thimmaiah:
Retains .group sections when linking partially. We also have to retain multiple sections with the same name, if they are in different groups.


marxin:
Given that the PR is quite stale, could we close it, please?

PR #1841, opened 2026-04-15
refactor: Introduce a global InputSectionId

davidlattimore:


davidlattimore:
I've been wanting to have an `InputSectionId` for at least a year or two. The final motivation for it here is that for range-extension thunks, I need to know, during the GC phase what PartId a particular symbol is defined in. PartIds are already all decided by the time we reach the GC phase, but I previously couldn't access this, since they were stored in `SectionSlot`, which each object had exclusive access to.

PR #1842, opened 2026-04-15
refactor: Remove section index from layout::Section

davidlattimore:


davidlattimore:
We always know or can compute the SectionIndex when we access a `Section`, so storing it is wasteful.

PR #1844, opened 2026-04-15
tools: Add some colour to --sym-info output

davidlattimore:


davidlattimore:
Sample output.

image

Mostly I wanted to be able to visually separate different parts and to be able to more quickly see the canonical symbol (ID is green). It also distinguishes loaded vs non-loaded files with different colours. Probably plenty of room for improvement.

PR #1847, opened 2026-04-17
feat: Implement range-extension thunks for aarch64

### LINKED ISSUE #1763: Add support for range-extension thunks
Needed to fix #1753 (may not be sufficient if there are other unrelated issues).

Range-extension thunks are mostly needed when producing large binaries for RISC architectures e.g. aarch64.

Where a branch instruction has limited range, e.g. +/- 128MiB, but the target of the branch is beyond that, we need to insert a thunk somewhere within the range that branches to the actual location.

The tricky part is that in order to know whether a thunk is needed, we need to know addresses. In order to figure out addresses, we need to know sizes, but the sizes of the thunks depends on how many we need. So there's basically a cycle in the state dependencies.

One option (call it A) would be to have an iterative algorithm e.g.:

* Determine sizes
* Determine addresses
* Scan relocations, looking of any that are out of range. Allocate thunks as needed.
* Adjust sizes based on allocated thunks
* If any adjustments were made, then go back to determining addresses

Given that determining addresses currently takes about 5% of link time, this could get expensive.

An alternative approach (B) would be to pessimistically allocate a thunk for every branch that might possibly require one. For an initial implementation, that might be sufficient. We could later add a pass where we eliminate unneeded thunks. An elimination approach has the advantage that we can probably stop after one iteration. While eliminating some thunks might reduce sizes enough that other thunks can then be eliminated, I suspect this would be of limited benefit, so probably isn't worth it. Most of the the gains would come from the first pass.

It might be tempting to create a thunk per input symbol ID rather than per relocation. This might work, but there are definitely circumstances in which it wouldn't work. e.g. if a single input symbol is referenced from two different functions that are placed in very different parts of the file - e.g. one function is marked as "cold" and placed with other cold functions.

It would also be tempting to create one thunk per symbol per input section, however the bookkeeping for doing that is likely to be expensive.

Even creating a thunk per eligible relocation requires a bit of bookkeeping. We'd need a count of thunks per input section. Storing any data per input section tends to slow us down. Fortunately there are currently a few free bytes that we can use without increasing the size of `SectionSlot`.

davidlattimore:
Fixes #1763

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -62,6 +62,7 @@ use crate::platform::SectionFlags as _;
 use crate::platform::SectionHeader as _;
 use crate::platform::SectionType as _;
 use crate::platform::Symbol as _;
+use crate::platform::ThunkConfig;
 use crate::platform::VerneedTable as _;
 use crate::program_segments::ProgramSegments;
 use crate::resolution::LoadedMetrics;
@@ -144,6 +145,8 @@ use zerocopy::KnownLayout;
 pub const NON_PIE_START_MEM_ADDRESS: u64 = 0x400_000;
 
 pub(crate) const GLOBAL_POINTER_SYMBOL_NAME: &str = "__global_pointer$";
+
+pub(crate) const THUNK_SYMBOL_PREFIX: &str = "__thunk_";
 
 pub(crate) type FileHeader = object::elf::FileHeader64<LittleEndian>;
 pub(crate) type ProgramHeader = object::elf::ProgramHeader64<LittleEndian>;
@@ -557,6 +560,10 @@ impl platform::Platform for Elf {
         memory_offsets.increment(part_id::EH_FRAME, object.format_specific.eh_frame_size);
     }
 
+    fn file_thunk_config<'data>(file: &File<'data>) -> Option<ThunkConfig> {
+        thunk_config_for_object(file)
+    }
+
     fn finalise_layout_dynamic<'data>(
         state: &mut layout::DynamicLayoutState<'data, Self>,
         memory_offsets: &mut OutputSectionPartMap<u64>,
@@ -669,7 +676,7 @@ impl platform::Platform for Elf {
     }
 
     fn load_object_section_relocations<'data, 'scope, A: Arch<Platform = Self>>(
-        state: &layout::ObjectLayoutState<'data, Self>,
+        state: &mut layout::ObjectLayoutState<'data, Self>,
         common: &mut layout::CommonGroupState<'data, Self>,
         queue: &mut layout::LocalWorkQueue,
         resources: &'scope layout::GraphResources<'data, '_, Self>,
@@ -1524,6 +1531,25 @@ impl platform::Platform for Elf {
         Ok(())
     }
 
+    fn allocate_thunk_symbol_sizes(
+        sizes: &mut OutputSectionPartMap<u64>,
+        symbols: &[SymbolId],
+        symbol_db: &SymbolDb<Self>,
+    ) {
+        let total_name_bytes: usize = symbols
+            .iter()
+            .map(|&sym_id| {
+                let name_len = symbol_db.symbol_name(sym_id).map_or(0, |n| n.bytes().len());
+                elf::THUNK_SYMBOL_PREFIX.len() + name_len + 1
+            })
+            .sum();
+        sizes.increment(
+            part_id::SYMTAB_LOCAL,
+            symbols.len() as u64 * size_of::<elf::SymtabEntry>() as u64,
+        );
+        sizes.increment(part_id::STRTAB, total_name_bytes as u64);
+    }
+
     fn allocate_prelude(common: &mut CommonGroupState<Self>, symbol_db: &SymbolDb<Self>) {
         // The first entry in the symbol table must be null. Similarly, the first string in the
         // strings table must be empty.
@@ -1802,10 +1828,12 @@ impl platform::Platform for Elf {
         builder.add_sections(&custom.ro);
 
         builder.add_section(output_section_id::PLT_GOT);
-        builder.add_section(output_section_id::TEXT);
         builder.add_section(output_section_id::INIT);
         builder.add_section(output_section_id::FINI);
         builder.add_sections(&custom.exec);
+        // We want ThunkConfig::primary_function_part_id to be more or less last, since we only
+        // support generating thunks before the primary_function_part, not after.
+        builder.add_section(output_section_id::TEXT);
 
         builder.add_section(output_section_id::TDATA);
         builder.add_sections(&custom.tdata);
@@ -2571,6 +2599,7 @@ fn process_eh_frame_relocations<'data, 'scope, A: Arch<Platform = Elf>, R: Reloc
                     common,
                     rel,
                     eh_frame_section,
+                    output_section_id::EH_FRAME.base_part_id(),
                     resources,
                     queue,
                     false,
@@ -2677,9 +2706,10 @@ fn process_section_exception_frames<'data, 'scope, A: Arch<Platform = Elf>, R: R
                 common,
                 &rel,
                 eh_frame_section,
+                output_section_id::EH_FRAME.base_part_id(),
                 resources,
                 queue,
-                false,
+                true,
                 scope,
             )?;
         }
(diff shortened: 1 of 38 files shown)


libwild/src/elf.rs
```diff
@@ -1796,10 +1822,12 @@ impl platform::Platform for Elf {
         builder.add_sections(&custom.ro);
 
         builder.add_section(output_section_id::PLT_GOT);
-        builder.add_section(output_section_id::TEXT);
         builder.add_section(output_section_id::INIT);
         builder.add_section(output_section_id::FINI);
         builder.add_sections(&custom.exec);
+        // We want ThunkConfig::primary_function_part_id to be more or less last, since we only
+        // support generating thunks before the primary_function_part, not after.
+        builder.add_section(output_section_id::TEXT);
```
marxin, line 1836: Don't know which fraction of linker scripts we support, but can't one reorder the section in a linker script? How will it influence then the range-extension thunks?

libwild/src/elf_aarch64.rs
```diff
@@ -296,6 +298,54 @@ impl crate::platform::Arch for ElfAArch64 {
...
+    fn generate_thunk(thunk_address: u64, target_address: u64, buf: &mut [u8]) {
+        debug_assert_eq!(buf.len(), 16);
+
+        // Use ADRP+ADD+BR - a PC-relative thunk that works in PIE binaries.
+        //
+        // ADRP computes the page-aligned base address of the target, then ADD adds the
+        // page offset. Together they can reach any address within +/-4GB of the thunk.
+
+        // Page-aligned addresses (each page is 4096 bytes)
+        let thunk_page = thunk_address & !0xFFF;
```
marxin: What about using `elf::PAGE_MASK_4KB` existing constant?

libwild/src/elf_aarch64.rs
```diff
@@ -296,6 +298,54 @@ impl crate::platform::Arch for ElfAArch64 {
...
+        let page_diff_bytes = (target_page as i64).wrapping_sub(thunk_page as i64);
+        // ADRP encodes (page_diff_bytes / 4096) as a 21-bit signed integer
+        let page_count = page_diff_bytes >> 12;
+        let imm21 = page_count as i32;
+
+        // ADRP x16: encoding = 1|immhi(19bits)|10000|immlo(2bits)|Rd(5bits)
+        // where Rd = x16 = 16, immlo = imm21[1:0], immhi = imm21[20:2]
+        let immlo = (imm21 & 0x3) as u32;
+        let immhi = ((imm21 >> 2) & 0x7_FFFF) as u32;
+        let adrp: u32 = 0x9000_0010 | (immlo << 29) | (immhi << 5);
```
marxin: What about using the existing encoding functionality in `write_to_value` for `AArch64Instruction::Adr` and other instructions in the function. And filling it up similarly to how we do it for PLT entries, see `PLT_ENTRY_TEMPLATE`. See also `write_plt_entry`.

libwild/src/elf_aarch64.rs
```diff
@@ -296,6 +298,54 @@ impl crate::platform::Arch for ElfAArch64 {
             offset_in_section,
         )
     }
+
+    fn thunk_config() -> Option<crate::platform::ThunkConfig> {
+        Some(crate::platform::ThunkConfig {
+            primary_function_part_id: const {
+                output_section_id::TEXT.part_id_with_alignment(Alignment { exponent: 2 })
+            },
+            min_branch_range: 128 * 1024 * 1024,
```
marxin: Maybe a const in elf.rs?

libwild/src/elf_aarch64.rs
```diff
@@ -296,6 +298,54 @@ impl crate::platform::Arch for ElfAArch64 {
...
+            // ADRP x16 + ADD x16, x16 + BR x16 = 3 * 4 bytes, padded to 16 for alignment
+            thunk_size: 16,
```
marxin: Could be `size_of` of the `THUNK_ENTRY_TEMPLATE` (an equivalent to `PLT_ENTRY_TEMPLATE`)?

libwild/src/layout.rs
```diff
@@ -268,6 +294,11 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
         .map(|group| res_writer.take_shard(group.num_symbols))
         .collect_vec();
 
+    let thunk_block_addresses_out: Vec<Mutex<BTreeMap<SymbolId, u64>>> =
```
marxin: Is the type annotation needed here?

libwild/src/layout.rs
```diff
@@ -311,6 +344,11 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
         &resources.symbol_db.args.common().warning_callback,
     )?;
 
+    let thunk_block_addresses: Vec<BTreeMap<SymbolId, u64>> = thunk_block_addresses_out
```
marxin: Same here.

libwild/src/thunks.rs
```diff
@@ -0,0 +1,546 @@
...
+        max_branch_range,
+        |fid, bid, is_owner| {
+            if let FileLayoutState::Object(obj) = &mut group_states[fid.group()].files[fid.file()] {
+                obj.thunk_block_id = bid;
+                obj.owns_thunk_block = is_owner;
+            }
+        },
+    );
+
+    let mut block_builders: Vec<ThunkBlockBuilder<'data, 'state, P>> = (0..num_blocks.max(1))
```
marxin: Is type annotation needed? Consider using `collect_vec`.

linker-diff/src/aarch64.rs
```diff
@@ -192,6 +192,84 @@ impl Arch for AArch64 {
...
+            // LDR x16, <pc_rel_literal> / BR x16.
+            // LDR Xt (literal): bits[31:30]=01(64-bit), bits[29:27]=011, bit[26]=V=0,
+            //   bits[25:5]=imm19 (PC-relative word offset), bits[4:0]=Rt=x16(=0b10000)
+            // The target absolute address is stored as a 64-bit literal at PC + imm19*4.
+            const LDR_X16_LITERAL_MASK: u32 = 0xFF00001F;
+            const LDR_X16_LITERAL: u32 = 0x58000010;
+            if bytes.len() >= 16 {
+                let insn1 = u32::from_le_bytes(bytes[4..8].try_into().ok()?);
+                if insn0 & LDR_X16_LITERAL_MASK == LDR_X16_LITERAL && insn1 == 0xD61F0200 {
+                    let imm19 = ((insn0 >> 5) & 0x7FFFF) as i32;
```
marxin: `AArch64Instruction::read_value` might help you decoding the immediates?

linker-diff/src/aarch64.rs
```diff
@@ -192,6 +192,84 @@ impl Arch for AArch64 {
...
+                // Check ADD x16, x16, #imm (no shift): bits [31:29]=100, [28:24]=10001,
+                //   [23:22]=00, [9:5]=10000(Rn=x16), [4:0]=10000(Rd=x16)
+                const ADD_MASK: u32 = 0xFFC003FF;
+                const ADD_X16_X16: u32 = 0x91000210;
+
+                if insn0 & ADRP_MASK == ADRP_X16
+                    && insn1 & ADD_MASK == ADD_X16_X16
+                    && insn2 == BR_X16
+                {
+                    // Decode ADRP immediate: immlo=[30:29], immhi=[23:5], imm = immhi:immlo
```
marxin: Similarly here, the same logic already exists for the purpose of linker-diff.

marxin (commented):
Just made a light review of the PR - I haven't managed to get into nitty gritty. I am really happy we'll receive a feature parity on AArch64 with other linkers.

libwild/src/elf.rs
davidlattimore: At the moment, I'm pretty sure it would be possible to break the assumptions made by this PR using a linker script. My guess is that linker scripts are most often used when building for embedded targets and kernel work, both of which would, I assume generally be well under 128 MiB of .text. So I'm somewhat inclined that this is just something we don't support at the moment. If someone does do it, they should get an error that a relocation is out-of-range but we didn't generate a thunk.

libwild/src/elf_aarch64.rs
davidlattimore: It's aarch64-specific, so I added the constant to this file.

davidlattimore (commented):
Thanks for the review. I'm pretty sure I addressed all comments or at least replied to them.

libwild/src/elf.rs
marxin: Works for me - thanks for the clarification.

marxin (approved):
LGTM

PR #1851, opened 2026-04-18
feat: support SEGMENT_START function in Linker Script

### LINKED ISSUE #1098: linker-scripts: Support SEGMENT_START
https://sourceware.org/binutils/docs/ld/Builtin-Functions.html#index-SEGMENT_005fSTART_0028segment_002c-default_0029

We currently support assigning symbols in linker scripts, but at the moment, those symbols are always assigned the value `.`, which is the current location. This causes them to be bound to either the start or the end of a section. Look for references to `ContentsCommand::SymbolAssignment`. There are lots of other things that can be put on the RHS of a symbol definition in a linker script, including absolute values. `SEGMENT_START` is one such thing. It's probably a good idea to read through the other builtin functions listed on the page linked above, since that might influence the chosen design.

It may be necessary to add new variants to `SymbolPlacement`.

I'll tag this as a good-first-issue, but it's a more ambitious one that will likely require a bit of iteration. Please don't hesitate to ask questions.

Blazearth:
Implements SEGMENT_START("name", default) in linker script parsing and resolution. Fixes linking of binaries that use __executable_start = SEGMENT_START("text", 0) in their linker scripts.

Closes  #1098

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -1920,6 +1920,15 @@ impl platform::Args for ElfArgs {
         }
     }
 
+    fn segment_start_override(&self, name: crate::parsing::SegmentName) -> Option<u64> {
+        match name {
+            crate::parsing::SegmentName::Text => self.ttext,
+            crate::parsing::SegmentName::Data => self.tdata,
+            crate::parsing::SegmentName::Bss => self.tbss,
+            crate::parsing::SegmentName::Rodata => None,
+        }
+    }
+
     fn version_script_path(&self) -> Option<&Path> {
         self.version_script_path.as_deref()
     }
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -1925,7 +1925,7 @@ impl platform::Args for ElfArgs {
             crate::parsing::SegmentName::Text => self.ttext,
             crate::parsing::SegmentName::Data => self.tdata,
             crate::parsing::SegmentName::Bss => self.tbss,
-            crate::parsing::SegmentName::Rodata => None,
+            crate::parsing::SegmentName::Rodata | crate::parsing::SegmentName::Other => None,
         }
     }
 
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3304,17 +3304,20 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
-fn segment_name_matches(name: &str, def: impl crate::platform::ProgramSegmentDef) -> bool {
+fn segment_name_matches(name: &[u8], def: impl crate::platform::ProgramSegmentDef) -> bool {
     match name {
-        "text" => def.is_executable() && !def.is_writable(),
-        "data" | "bss" => def.is_writable() && !def.is_executable(),
-        "rodata" => !def.is_writable() && !def.is_executable(),
+        b"text" => def.is_executable() && !def.is_writable(),
+        b"data" | b"bss" => def.is_writable() && !def.is_executable(),
+        // rodata lives in a dedicated RO segment when one exists, but in a typical Linux
+        // executable it shares the RX segment with .text. Match any non-writable loadable
+        // segment so we find whichever one actually contains read-only data.
+        b"rodata" => !def.is_writable(),
         _ => false,
     }
 }
 
-fn is_known_segment_name(name: &str) -> bool {
-    matches!(name, "text" | "data" | "bss" | "rodata")
+fn is_known_segment_name(name: &[u8]) -> bool {
+    matches!(name, b"text" | b"data" | b"bss" | b"rodata")
 }
 
 fn create_start_end_symbol_resolution<'data, P: Platform>(
@@ -3385,7 +3388,8 @@ fn create_start_end_symbol_resolution<'data, P: Platform>(
         SymbolPlacement::SegmentStart(name, default) => {
             if !is_known_segment_name(name) {
                 resources.symbol_db.args.warning(format!(
-                    "SEGMENT_START: unknown segment name '{name}', using default 0x{default:x}"
+                    "SEGMENT_START: unknown segment name '{}', using default 0x{default:x}",
+                    String::from_utf8_lossy(name)
                 ));
             }
             resources
@@ -3394,7 +3398,7 @@ fn create_start_end_symbol_resolution<'data, P: Platform>(
                 .iter()
                 .find(|seg| {
                     let def = resources.program_segments.segment_def(seg.id);
-                    def.is_loadable() && segment_name_matches(name, *def)
+                    def.is_loadable() && seg.sizes.mem_size > 0 && segment_name_matches(name, *def)
                 })
                 .map_or(default, |seg| seg.sizes.mem_offset)
         }
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3304,22 +3304,6 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
-fn segment_name_matches(name: &[u8], def: impl crate::platform::ProgramSegmentDef) -> bool {
-    match name {
-        b"text" => def.is_executable() && !def.is_writable(),
-        b"data" | b"bss" => def.is_writable() && !def.is_executable(),
-        // rodata lives in a dedicated RO segment when one exists, but in a typical Linux
-        // executable it shares the RX segment with .text. Match any non-writable loadable
-        // segment so we find whichever one actually contains read-only data.
-        b"rodata" => !def.is_writable(),
-        _ => false,
-    }
-}
-
-fn is_known_segment_name(name: &[u8]) -> bool {
-    matches!(name, b"text" | b"data" | b"bss" | b"rodata")
-}
-
 fn create_start_end_symbol_resolution<'data, P: Platform>(
     memory_offsets: &mut OutputSectionPartMap<u64>,
     resources: &FinaliseLayoutResources<'_, 'data, P>,
@@ -3385,23 +3369,11 @@ fn create_start_end_symbol_resolution<'data, P: Platform>(
             .find(|seg| resources.program_segments.segment_def(seg.id).is_loadable())
             .map(|seg| seg.sizes.mem_offset)?,
 
-        SymbolPlacement::SegmentStart(name, default) => {
-            if !is_known_segment_name(name) {
-                resources.symbol_db.args.warning(format!(
-                    "SEGMENT_START: unknown segment name '{}', using default 0x{default:x}",
-                    String::from_utf8_lossy(name)
-                ));
-            }
-            resources
-                .segment_layouts
-                .segments
-                .iter()
-                .find(|seg| {
-                    let def = resources.program_segments.segment_def(seg.id);
-                    def.is_loadable() && seg.sizes.mem_size > 0 && segment_name_matches(name, *def)
-                })
-                .map_or(default, |seg| seg.sizes.mem_offset)
-        }
+        SymbolPlacement::SegmentStart(name, default) => resources
+            .symbol_db
+            .args
+            .segment_start_override(name)
+            .unwrap_or(default),
     };
 
     Some(P::create_resolution(
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3313,6 +3313,10 @@ fn segment_name_matches(name: &str, def: impl crate::platform::ProgramSegmentDef
     }
 }
 
+fn is_known_segment_name(name: &str) -> bool {
+    matches!(name, "text" | "data" | "bss" | "rodata")
+}
+
 fn create_start_end_symbol_resolution<'data, P: Platform>(
     memory_offsets: &mut OutputSectionPartMap<u64>,
     resources: &FinaliseLayoutResources<'_, 'data, P>,
@@ -3378,15 +3382,22 @@ fn create_start_end_symbol_resolution<'data, P: Platform>(
             .find(|seg| resources.program_segments.segment_def(seg.id).is_loadable())
             .map(|seg| seg.sizes.mem_offset)?,
 
-        SymbolPlacement::SegmentStart(name, default) => resources
-            .segment_layouts
-            .segments
-            .iter()
-            .find(|seg| {
-                let def = resources.program_segments.segment_def(seg.id);
-                def.is_loadable() && segment_name_matches(name, *def)
-            })
-            .map_or(default, |seg| seg.sizes.mem_offset),
+        SymbolPlacement::SegmentStart(name, default) => {
+            if !is_known_segment_name(name) {
+                resources.symbol_db.args.warning(format!(
+                    "SEGMENT_START: unknown segment name '{name}', using default 0x{default:x}"
+                ));
+            }
+            resources
+                .segment_layouts
+                .segments
+                .iter()
+                .find(|seg| {
+                    let def = resources.program_segments.segment_def(seg.id);
+                    def.is_loadable() && segment_name_matches(name, *def)
+                })
+                .map_or(default, |seg| seg.sizes.mem_offset)
+        }
     };
 
     Some(P::create_resolution(
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3304,6 +3304,15 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
+fn segment_name_matches(name: &str, def: impl crate::platform::ProgramSegmentDef) -> bool {
+    match name {
+        "text" => def.is_executable() && !def.is_writable(),
+        "data" | "bss" => def.is_writable() && !def.is_executable(),
+        "rodata" => !def.is_writable() && !def.is_executable(),
+        _ => false,
+    }
+}
+
 fn create_start_end_symbol_resolution<'data, P: Platform>(
     memory_offsets: &mut OutputSectionPartMap<u64>,
     resources: &FinaliseLayoutResources<'_, 'data, P>,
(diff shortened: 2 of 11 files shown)


libwild/src/layout_rules.rs
```diff
@@ -26,6 +26,77 @@ use hashbrown::HashTable;
...
+fn placement_from_symbol_definition_value<'data>(
+    name: &[u8],
+    value: &'data [u8],
+) -> crate::error::Result<Option<SymbolPlacement<'data>>> {
+    use crate::linker_script::Expression;
+    use winnow::BStr;
+
+    // Try parsing as a full expression first.
+    let mut input = BStr::new(value);
+    if let Ok(expr) = crate::linker_script::parse_expression_pub(&mut input) {
```
lapla-cogito: After `parse_expression_pub()` successfully parses a prefix of the value (e.g. `SEGMENT_START`), any trailing tokens in the same expression are silently ignored because the caller does not verify that the input was fully consumed.

libwild/src/layout_rules.rs
```diff
@@ -26,6 +26,77 @@ use hashbrown::HashTable;
...
+        let placement = match expr {
+            Expression::SegmentStart(seg_name, default_expr) => {
+                let default = eval_constant_expr(&default_expr).unwrap_or(0);
+                let name_str = std::str::from_utf8(seg_name)
+                    .map_err(|_| crate::error!("SEGMENT_START: segment name is not valid UTF-8"))?;
+                Some(SymbolPlacement::SegmentStart(name_str, default))
+            }
+            Expression::Number(n) => Some(SymbolPlacement::DefsymAbsolute(n)),
+            // For other expressions fall through to the legacy path below.
+            _ => None,
```
lapla-cogito: Using legacy paths results in double parsing by both `parse_expression_pub()` and `parse_symbol_expression()`. This is inherently inefficient, and if the two parsing results conflict, it creates a breeding ground for difficult-to-debug bugs. Ideally, we should consolidate this into a single parsing path.

libwild/src/layout.rs
```diff
@@ -3174,6 +3174,15 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
+fn segment_name_matches(name: &str, def: impl crate::platform::ProgramSegmentDef) -> bool {
+    match name {
+        "text" => def.is_executable() && !def.is_writable(),
+        "data" | "bss" => def.is_writable() && !def.is_executable(),
+        "rodata" => !def.is_writable() && !def.is_executable(),
+        _ => false,
```
lapla-cogito, line 3185: It may be better to emit a warning and then ignore the unknown segment name rather than silently ignoring it.

lapla-cogito (commented):
Also, please update the status of `SEGMENT_START` in [`LINKER_SCRIPT_SUPPORT.md`](https://github.com/wild-linker/wild/blob/main/LINKER_SCRIPT_SUPPORT.md).

libwild/src/layout_rules.rs
Blazearth: Thanks for pointing this out , you're right that the fallback introduced double parsing and could lead to inconsistent interpretations.
I have removed the legacy `parse_symbol_expression` fallback entirely. Now uses only single parser.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
```diff
@@ -0,0 +1,13 @@
+//#LinkerScript:linker-script-segment-start.ld
+//#RunEnabled:false
+//#DiffEnabled:false
```
lapla-cogito: Why should `DiffEnabled:false` be necessary? The diff detection by linker-diff should be verified unless there's a valid reason.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
```diff
@@ -0,0 +1,13 @@
...
+//#Mode:dynamic
+//#LinkArgs:-shared -z now
```
lapla-cogito: Is there a reason you're linking this as a shared library? IIRC, `SEGMENT_START` is typically used for regular executables more often. If you link this test as a regular executable, you won't need `RunEnabled:false` either.

lapla-cogito (dismissed):
I have some questions about the test, but otherwise it seems good.

libwild/src/layout_rules.rs
```diff
@@ -209,17 +308,35 @@ impl<'data> LayoutRulesBuilder<'data> {
...
+                                                    SymbolPlacement::SectionStart(
+                                                        primary_section_id,
+                                                    )
+                                                }
+                                            }
+                                            Expression::SegmentStart(name, default_expr) => {
+                                                let default =
+                                                    eval_constant_expr(default_expr).unwrap_or(0);
+                                                let name_str = std::str::from_utf8(name).map_err(|_| {
+                                                    crate::error!("SEGMENT_START: segment name is not valid UTF-8")
```
davidlattimore: I'd probably be inclined to just keep the name as `&[u8]`. That way we don't need to worry about UTF-8 validation. When we're comparing it with things, we can just compare it with b"text" etc. That's how names are handled throughout most of the rest of the linker.

libwild/src/layout_rules.rs
```diff
@@ -26,6 +26,108 @@ use hashbrown::HashTable;
 use std::borrow::Cow;
 use std::mem::replace;
 
+/// Evaluate a linker script expression that must be a compile-time constant.
+/// Returns `Err` for any expression that requires runtime context (symbols, location counter,
+/// etc.).
+fn eval_constant_expr(expr: &crate::linker_script::Expression<'_>) -> crate::error::Result<u64> {
```
davidlattimore: I feel like this probably belongs in `expression_eval.rs`

libwild/src/layout_rules.rs
```diff
@@ -147,10 +249,7 @@ impl<'data> LayoutRulesBuilder<'data> {
                         .with_hidden(provide.hidden),
                 );
             } else if let linker_script::Command::SymbolDefinition { name, value } = cmd {
```
davidlattimore, line 206: If `value` is a string that needs parsing, then we should probably parse it when we parse the rest of the linker script. That way if there's an error while parsing, we can report the line number in the linker script where the error occurred.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.ld
```diff
@@ -0,0 +1,8 @@
+__executable_start = SEGMENT_START("text", 0);
```
davidlattimore: It could be good to define a symbol for each of the supported segment types, then have some code (in `_start`) that checks that those values appear correct relative to variables in the respective segments.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
Blazearth: DiffEnabled:false was originally added because the test was linked as a shared object, which introduced differences (e.g relocations/GOT) that made the diff noisy.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
Blazearth: You are right, linking as a shared object was also not ideal here, I took the easy path. I’ll rework this test to use a regular executable.

Blazearth:
```
//#DiffIgnore:segment.LOAD.RWX.alignment
//#DiffIgnore:segment.LOAD.RX.alignment
```
Added these in the test since ld produces a single RWE LOAD segment when no PHDRS directive is present in the linker script. Wild always uses separate RO/RX/RW segments. Took me while to resolve the test.
I have addressed all the issues.

libwild/src/layout.rs
```diff
@@ -3174,6 +3174,22 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
+fn segment_name_matches(name: &[u8], def: impl crate::platform::ProgramSegmentDef) -> bool {
+    match name {
+        b"text" => def.is_executable() && !def.is_writable(),
+        b"data" | b"bss" => def.is_writable() && !def.is_executable(),
+        // rodata lives in a dedicated RO segment when one exists, but in a typical Linux
+        // executable it shares the RX segment with .text. Match any non-writable loadable
+        // segment so we find whichever one actually contains read-only data.
```
davidlattimore, line 3183: This doesn't seem right to me. Maybe executables on Linux used to put RO data into the RX segment, but modern executables don't do this since it would make the binary slightly less secure. But in any case, what's important is what Wild does, which is emit a separate read-only non-executable segment.

libwild/src/layout.rs
```diff
@@ -3174,6 +3174,22 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
     }
 }
 
+fn segment_name_matches(name: &[u8], def: impl crate::platform::ProgramSegmentDef) -> bool {
+    match name {
+        b"text" => def.is_executable() && !def.is_writable(),
+        b"data" | b"bss" => def.is_writable() && !def.is_executable(),
```
davidlattimore, line 3180: What does GNU ld return when you request the start of BSS? Does it return the start of the combined data/bss segment or does it return where bss starts within that segment?

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
```diff
@@ -0,0 +1,53 @@
...
+/* Variables in each segment to verify the segment start is <= them */
+static const int rodata_var = 42;
+static int data_var = 1;
+static int bss_var;
+
+void _start(void) {
+  runtime_init();
+
+  /* text_start must be <= _start (both in the text segment) */
+  if (ptr_to_int(&text_start) > ptr_to_int(&_start)) {
```
davidlattimore, line 33: Given that the linker script explicitly says to put .text at 0x600000, do you think we should confirm that it's at that address?

libwild/src/layout.rs
```diff
@@ -3174,6 +3174,22 @@ impl<'data, P: Platform> InternalSymbols<'data, P> {
...
+        // rodata lives in a dedicated RO segment when one exists, but in a typical Linux
+        // executable it shares the RX segment with .text. Match any non-writable loadable
+        // segment so we find whichever one actually contains read-only data.
+        b"rodata" => !def.is_writable(),
+        _ => false,
+    }
+}
+
+fn is_known_segment_name(name: &[u8]) -> bool {
+    matches!(name, b"text" | b"data" | b"bss" | b"rodata")
```
davidlattimore, line 3190: It seems a shame to duplicate the list here and in the function above. What about adding an enum SegmentMatcher, then parsing that from the string. Bonus if we can parse when we parse the linker script so that that an unknown value gets reported with the correct line number.

I guess that might result in an unknown segment name being an error rather than a warning. That's possibly not a bad thing though.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.ld
```diff
@@ -0,0 +1,16 @@
+ENTRY(_start)
+
+text_start = SEGMENT_START("text", 0x600000);
```
davidlattimore, line 3: The default value here is identical to the `. = 0x600000` used for the .text section below. That makes it hard to determine which was actually used. I tried replacing the default values for all four symbols and it appears that GNU ld ended up using those default values for all of them - e.g. it didn't actually use the start of the text segment at all. It could be good to do some experimentation to determine when the default value gets used and when it doesn't then try to match that.

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.c
```diff
@@ -0,0 +1,53 @@
...
+// ld merges all sections into a single RWE segment when no PHDRS are specified,
+// while wild uses separate RO/RX/RW segments. Ignore the resulting layout
+// diffs.
+//#DiffIgnore:segment.LOAD.RWX.alignment
+//#DiffIgnore:segment.LOAD.RX.alignment
+
+#include <stddef.h>
+
+#include "../common/runtime.h"
+#include "ptr_black_box.h"
```
davidlattimore, line 16: Could you change this to "../common/ptr_black_box.h"? I'll be doing a PR soon to update the includes in the other files and also to change the test runner to not have "../common" in the include path.

libwild/src/layout.rs
Blazearth: Sure i will revert it back.

libwild/src/layout.rs
Blazearth: Testing with GNU ld, `SEGMENT_START("bss", ...)` returns the start of the loadable writable segment, not the start of the .bss section within that segment, since `.data` and `.bss` are typically placed in the same LOAD segment.

Blazearth:
>The default value here is identical to the . = 0x600000 used for the .text section below. That makes it hard to determine which was actually used. I tried replacing the default values for all four symbols and it appears that GNU ld ended up using those default values for all of them - e.g. it didn't actually use the start of the text segment at all. It could be good to do some experimentation to determine when the default value gets used and when it doesn't then try to match that.

Hey @davidlattimore , I ran a series of experiments to understand the real behavior before deciding how Wild should match it.

What the docs say

The official GNU ld docs state:

SEGMENT_START(segment, default) — If an explicit value has already been given for this segment with a command-line -T option then that value will be returned, otherwise the value will be default.

Case 1 — no -T, default differs from actual layout:

```
text_start = SEGMENT_START("text", 0x1);
. = 0x600000;  /* .text actually lands here */
Result: text_start = 0x1 (the default, not the actual segment address)
```

Case 2 — -Ttext=0x700000 override:

```
text_start = SEGMENT_START("text", 0x1);
. = 0x600000;
ld -Ttext=0x700000 ...
Result: text_start = 0x700000 (the CLI override wins, layout moves too)
```

Case 3 — no -T, default happens to match layout (our current test):

```
text_start = SEGMENT_START("text", 0x600000);
. = 0x600000;
Result: text_start = 0x600000 — ambiguous, can't tell if it used the default or the segment
```

Case 4 — no -T, default is different from layout:

```
text_start = SEGMENT_START("text", 0x500000);
. = 0x600000;
Result: text_start = 0x500000 — default again, not the actual segment base
``` 

In these cases, GNU ld appears to return the -T* override if provided, otherwise the default value — even when the actual segment lands elsewhere in the layout. This suggests SEGMENT_START does not derive its value from scanning PT_LOAD segments, at least in these scenarios.

What Wild currently does

Wild scans PT_LOAD segments by flag combinations (RX for text, RO for rodata, RW for data/bss) and returns the actual segment base address. This is more "layout-aware" but fundamentally different from GNU ld's behavior.

Now ,there are two options:

Option A — Match GNU ld strictly: SEGMENT_START only returns a value when the corresponding -T flag was passed on the command line (e.g. -Ttext, -Tdata, -Tbss). Otherwise return the default. Wild would need to store -T overrides and look them up here.

Option B — Keep Wild's current behavior: scan actual PT_LOAD segments by flags and return their base address. More useful in practice, but diverges from GNU ld when no -T is passed.

davidlattimore:
I think for things like this, we want to match GNU ld or lld's behaviour, especially if both GNU ld and lld have the same behaviour.

Blazearth:
I have changed it to match the behaviour of GNU ld where it overrides the address with `-T` for `bss`, `data` and `text`. As for `rodata` it always returns the default. 
The test now has two configs to verify both paths: one without any -T flags (all four symbols return their linker script defaults), and one with `-Ttext/-Tdata/-Tbss` overrides (verifying the override values are returned, while rodata still returns its default).

libwild/src/args/elf.rs
```diff
@@ -1264,6 +1273,36 @@ fn setup_argument_parser() -> ArgumentParser<ElfArgs> {
         .prefix("T")
         .help("Use linker script")
         .execute(|args, _modifier_stack, value| {
+            // -Ttext=ADDR, -Tdata=ADDR, -Tbss=ADDR are segment start overrides,
+            // not linker script paths. Handle them here since they share the -T prefix.
+            // The prefix handler gives us the part after "-T", which may be:
```
davidlattimore, line 1294: Thanks for figuring this out. I hadn't realised this particular oddity in GNU linker parsing semantics.

davidlattimore (commented):
I'm a bit worried that it appears this implements parsing of `-Ttext=` etc and hooks it up to `SEGMENT_START`, but doesn't actually affect the actual start of the segments. That would mean that someone linking with `-Ttext=` would now not get an error, but would get a binary where the text segment didn't start where they said it needed to.

Ideally, we'd fully support `-Ttext=` etc before we'd support `SEGMENT_START`. In that regard, the best thing to do would be to separate `-Ttext=` support out into a separate PR, hook it up to layout so that it affects the address of the text segment and make sure that part of it is tested, submit that PR, then proceed with the rest of this PR.

But I also don't want to make you drag this PR out longer than you want to, so I'm open to other options. e.g. we could split the other way. i.e. move the parsing of `-Ttext=` etc to a later PR, including the part of the test you've got that ensures that works with `SEGMENT_START`. That would mean that this PR would support `SEGMENT_START`, but only where it returns the default value. The later PR could add support for `-Ttext=` etc, making it both affect the actual segment starts and the return value from `SEGMENT_START`.

Blazearth:
> I'm a bit worried that it appears this implements parsing of `-Ttext=` etc and hooks it up to `SEGMENT_START`, but doesn't actually affect the actual start of the segments. That would mean that someone linking with `-Ttext=` would now not get an error, but would get a binary where the text segment didn't start where they said it needed to.

First of all, sorry for overlooking that — you're right that -Ttext= currently only affects the SEGMENT_START return value and doesn't actually move the segment in the output binary.

>Ideally, we'd fully support -Ttext= etc before we'd support SEGMENT_START. In that regard, the best thing to do would be to separate -Ttext= support out into a separate PR, hook it up to layout so that it affects the address of the text segment and make sure that part of it is tested, submit that PR, then proceed with the rest of this PR.

I'd like to try and go with this option first — implement proper -Ttext/-Tdata/-Tbss support that actually affects the segment layout, then this PR's with-T-overrides test will be fully correct. I'll open a separate PR for that.



If I find it's too complex to get right, I'll fall back to Option B — strip the -Ttext parts from this PR so it only supports SEGMENT_START returning the default value, and leave the override support for a later PR.

Blazearth:
Now `SEGMENT_START` move the section at the correct place in output binary with `-Ttext` override from #1877

wild/tests/sources/elf/linker-script-segment-start/linker-script-segment-start.ld
davidlattimore: Possibly you forgot this comment? It looks like it's still the case that the linker script starts .text at the same address as the default value. I'd suggest Changing the default values for all four SEGMENT_START calls to distinct values - e.g. 0x10, 0x11, 0x12, 0x13.

Blazearth: Sorry I forgot about that. I hope its all fine now.

davidlattimore (approved):

PR #1852, opened 2026-04-19
fix: always handle elf::SHN_XINDEX when reading

philipc:
Delete Symbol::section_index and use ObjectFile::symbol_section instead.


lapla-cogito (approved):
Looks good, thanks!

PR #1854, opened 2026-04-19
fix: use correct addend for relocs referencing STT_SECTION

vishruth-thimmaiah:
When linking with the -r flag, we now correctly increment the addend of relocations referencing `STT_SECTION`, when multiple input sections are merged.


lapla-cogito (approved):
Thanks!

PR #1857, opened 2026-04-22
feat: Implement SORT() directive for linker scripts

### LINKED ISSUE #1661: Implement linker-script SORT(...) for input sections
I’d like to work on adding support for the SORT(...) linker‑script keyworrd for input sections.

In GNU‑style scripts, SORT(pattern) matches the same input sections as pattern, but then sorts those matched sections by name before placing them in the output section.

My parser already represents SORT(...) as a wrapper around an inner pattern. The behaviour I want to implement is: use the inner pattern for matching during section assignment, and then, at layout time, sort the matched input sectionss by name before writing them (reusing the existing SortedSection machinery).

Does this approach fit how you’d like SORT to work in Wild? I would appreciate guidance

Abh10705:
- Introduces Harvester engine in layout.rs to collect and sort sections alphabetically.

- Implements physical section reordering in elf_writer.rs with O(1) skip logic for optimized performance.

- Adds comprehensive integration test and 50k-section stress test verification.

- Verified physical byte layout is contiguous and alphabetical via objdump.

Benchmark :

- GNU ld: 0.221s

- wild: 0.093s

Fixes #1661

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -50,6 +50,7 @@ use crate::resolution::NotLoaded;
 use crate::resolution::ResolvedGroup;
 use crate::resolution::SectionSlot;
 use crate::resolution::UnloadedSection;
+use crate::resolution::ScriptSortedSectionDetail;
 use crate::sharding::ShardKey;
 use crate::string_merging::MergedStringStartAddresses;
 use crate::string_merging::MergedStringsSection;
@@ -148,12 +149,14 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
 
     let mut dynamic_symbol_definitions =
         merge_dynamic_symbol_definitions(&group_states, &symbol_db)?;
+        let script_sorted_sections = harvest_and_sort_script_sections(&mut group_states, &output_sections, &symbol_db.section_part_ids);
 
     group_states.push(GroupState {
         files: vec![FileLayoutState::Epilogue(EpilogueLayoutState::new(
             symbol_db.args,
             symbol_db.output_kind,
             &mut dynamic_symbol_definitions,
+            script_sorted_sections.clone(),
         ))],
         queue: LocalWorkQueue::new(epilogue_file_id.group()),
         common: CommonGroupState::new(&output_sections),
@@ -249,14 +252,34 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     )?;
 
     let mem_offsets: OutputSectionPartMap<u64> = starting_memory_offsets(&section_part_layouts);
-    let starting_mem_offsets_by_group = compute_start_offsets_by_group(&group_states, mem_offsets);
+    let starting_mem_offsets_by_group = compute_start_offsets_by_group(&group_states, mem_offsets.clone());
 
     let merged_string_start_addresses = MergedStringStartAddresses::compute(
         &output_sections,
         &starting_mem_offsets_by_group,
         &merged_strings,
     );
+    
+    // --- SECTION SORTING & ADDRESS ASSIGNMENT ---
 
+    // At this stage, sections marked for sorting have been harvested but lack concrete memory addresses.
+    // We perform a two-step finalization:
+    // Sort the harvested sections according to the requested criteria.
+    // Linearize them in memory, starting from the base offset of their respective output section part, and advancing by the size of each section.
+    let mut script_sorted_sections = script_sorted_sections;
+    script_sorted_sections.sort_by(|a, b| a.name.cmp(&b.name));
+
+    let mut harvested_sections_registry = std::collections::HashMap::new();
+    if let Some(first) = script_sorted_sections.first() {
+        let mut current_offset = *mem_offsets.get(first.part_id);
+        for sec in &mut script_sorted_sections {
+            sec.mem_offset = current_offset;     // Assign the real hex address!
+            current_offset += sec.size;          // Advance the address by the function's size
+            // Adding to our instant-lookup HashMap
+            harvested_sections_registry.insert((sec.file_id, sec.section_index.0), sec.clone());
+        }
+    }
+   
     let mut symbol_resolutions = SymbolResolutions {
         resolutions: Vec::with_capacity(symbol_db.num_symbols()),
     };
@@ -271,6 +294,7 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     let resources = FinaliseLayoutResources {
         symbol_db: &symbol_db,
         output_sections: &output_sections,
+        harvested_sections_registry: &harvested_sections_registry,
         output_order: &output_order,
         section_layouts: &section_layouts,
         merged_string_start_addresses: &merged_string_start_addresses,
@@ -660,6 +684,7 @@ pub(crate) struct SyntheticSymbolsLayoutState<'data, P: Platform> {
 
 pub(crate) struct EpilogueLayoutState<P: Platform> {
     format_specific: P::EpilogueLayoutExt,
+    pub(crate) script_sorted_sections: Vec<HarvestedSortedSection>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


libwild/src/linker_script.rs
```diff
@@ -1019,10 +1023,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
 
-fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
-    let pattern = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
-    skip_comments_and_whitespace(input)?;
-    Ok(pattern)
+fn parse_pattern<'input>(
+    input: &mut &'input BStr,
+) -> winnow::Result<SectionPattern<'input>> {
+    // Handling SORT(...) wrapper: SORT is an alias for SORT_BY_NAME in GNU ld.
+    if input.starts_with(b"SORT") {
```
davidlattimore, line 1030: I'm pretty sure there's a function somewhere in this file that can be used to combine this check with the consumption of the relevant prefix.

libwild/src/linker_script.rs
```diff
@@ -1019,10 +1023,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
+        // Consume "SORT"
+        "SORT".parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+        // Expect '('
```
davidlattimore, line 1034: This comment doesn't seem to tell us much we couldn't tell by looking at the next line. Best to remove such comments

libwild/src/linker_script.rs
```diff
@@ -1019,10 +1023,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
+        // Inside SORT(...), accept a single wildcard pattern up to ')'
+        let name = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+        // Closing ')'
+        ')'.parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+
+        Ok(SectionPattern { name, sorted: true })
+    } else {
+        // Original behavior: bare pattern
```
davidlattimore, line 1047: Comments in code should be written for the reader of the code post-merging, not for the reader of the PR. To someone coming back to this code after the PR is merged, talking about "Original behaviour" won't make much sense. The exception is a temporary comment to the reviewer that you intend to remove before merging, but those should generally be clear that that's what they are.

wild/tests/sources/elf/script-sort/script-sort.c
```diff
@@ -0,0 +1,14 @@
+//#LinkArgs:-T ./script-sort.ld
+
+__attribute__((used, section(".text.func_c"))) int func_c() { return 3; }
+
+__attribute__((used, section(".text.func_a"))) int func_a() { return 1; }
+
+__attribute__((used, section(".text.func_b"))) int func_b() { return 2; }
+
+__attribute__((naked)) void _start() {
+  asm volatile(
```
davidlattimore, line 10: Can you look at a few other tests to see how we normally structure them? This test is quite different and for example won't work on anything other than x86_64.

libwild/src/layout.rs
```diff
@@ -5158,3 +5208,52 @@ impl OutputRecordLayout {
...
+    output_sections: &OutputSections<P>,
+    section_part_ids: &[PartId],
+) -> Vec<HarvestedSortedSection> {
+    timing_phase!("Harvest and sort script sections"); 
+    let mut temp = Vec::new();
+    for group in group_states.iter_mut() {
+        for file in &mut group.files {
+            if let FileLayoutState::Object(obj) = file {
+                // Steal the sticky notes
+                let notes = &obj.script_sorted_sections;
```
davidlattimore, line 5234: In the linker, "notes" generally refers to GNU notes sections. Is that what this is working with? If not, then it might be better to use a different name

libwild/src/layout.rs
```diff
@@ -5158,3 +5208,52 @@ impl OutputRecordLayout {
...
+                        // Steal the space from the original object!
+                        // group.common.mem_sizes.decrement(sec.part_id, capacity);
+                        temp.push((note.name, HarvestedSortedSection {
+                            file_id: obj.file_id,
+                            section_index: note.index,
+                            part_id,
+                            size: capacity, 
+                            alignment: part_id.alignment(output_sections),
+                            mem_offset: 0,
+                            name: note.name.to_vec(),
```
davidlattimore, line 5250: For performance reasons, we try to avoid copying things to the heap unless absolutely necessary. I don't think this copy should be necessary

libwild/src/layout.rs
```diff
@@ -3766,20 +3809,27 @@ impl<'data, P: Platform> ObjectLayoutState<'data, P> {
...
                         sec.capacity(part_id, resources.output_sections);
+                        
                     // Collect SFrame section ranges while we're already iterating
                     if part_id.output_section_id() == output_section_id::SFRAME {
                         let offset = (address - sframe_start_address) as usize;
                         let len = sec.size as usize;
                         sframe_ranges.push(offset..offset + len);
                     }
+
+                    if let Some(harvested) = resources.harvested_sections_registry.get(&(self.file_id, sec_idx)) {
```
davidlattimore, line 3829: The vast majority of linking won't be using the SORT feature. In those cases, we want the performance to be basically unchanged from before the sort feature was implemented. Doing a hashmap lookup here will very likely have an effect on performance, so we should look for alternatives.

libwild/src/layout.rs
```diff
@@ -5158,3 +5208,52 @@ impl OutputRecordLayout {
...
+    pub(crate) file_id: FileId,
+    pub(crate) section_index: object::SectionIndex,
+    pub(crate) part_id: PartId,
+    pub(crate) size: u64,
+    pub(crate) alignment: Alignment,
+    pub(crate) mem_offset: u64,
+    pub(crate) name: Vec<u8>,
+}
+
+fn harvest_and_sort_script_sections<'data, P: Platform>(
```
davidlattimore, line 5223: This function looks pretty expensive and runs on a single thread. It looks like it currently always runs, even if sorted sections are not in use. It's OK if linking is slowed down when section sorting is active, but we definitely shouldn't slow down when it's inactive.

libwild/src/layout.rs
```diff
@@ -249,14 +252,34 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
-
+    
+    // --- SECTION SORTING & ADDRESS ASSIGNMENT ---
+
+    // At this stage, sections marked for sorting have been harvested but lack concrete memory addresses.
+    // We perform a two-step finalization:
+    // Sort the harvested sections according to the requested criteria.
+    // Linearize them in memory, starting from the base offset of their respective output section part, and advancing by the size of each section.
+    let mut script_sorted_sections = script_sorted_sections;
+    script_sorted_sections.sort_by(|a, b| a.name.cmp(&b.name));
```
davidlattimore, line 270: I think `sort_by_key` can probably be used here and would be simpler

libwild/src/layout.rs
```diff
@@ -249,14 +252,34 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+
+    let mut harvested_sections_registry = std::collections::HashMap::new();
+    if let Some(first) = script_sorted_sections.first() {
+        let mut current_offset = *mem_offsets.get(first.part_id);
+        for sec in &mut script_sorted_sections {
+            sec.mem_offset = current_offset;     // Assign the real hex address!
+            current_offset += sec.size;          // Advance the address by the function's size
+            // Adding to our instant-lookup HashMap
```
davidlattimore, line 278: Hashmap lookups definitely aren't instant. In the context of the linker, they're actually pretty expensive and something that we only tend to use when we have exhausted less expensive options. When we do use hashmaps, we generally select one of the faster hash algorithms. I haven't yet put too much thought into whether this hashmap is actually needed. Have you considered what alternatives might exist?

davidlattimore:
Thanks for your work on this! Just wanted to check if you intend to continue with it or not? Completely understand if not, in which case we'll close the PR and unassign the issue.

Issue #1829, opened 2026-04-11
Add linker-script support for /DISCARD/ pseudo-section


PritamP20:
Wild doesn’t currently understand the `/DISCARD/` pseudo-section used in linker scripts to explicitly drop input sections. In GNU ld, this is commonly used to strip out debug info, metadata or other unnecessary sections especially in kernel and embedded builds. Right now, if a script includes `/DISCARD/`, wild either fails to parse it or has no way to represent the discard rule which means the script can’t behave the same way it would under GNU ld

The goal of this issue is to add proper support for `/DISCARD/` by extending the parser to recognize it, wiring it into the layout phase so matching input sections are actually removed, and adding tests to ensure both parsing and final output match GNU ld behaviour

i would like to work on this

davidlattimore:
No response heard on open PR, so I've unassigned this issue

PR #1866, opened 2026-04-27
feat: add GNU ld-compatible /DISCARD/ support

PritamP20:
added support for the /DISCARD/ pseudo-section in Wild linker scripts, similar to how GNU ld handles it , we now recognize /DISCARD/ blocks, turn their patterns into discard rules, and make sure any matching sections are intentionally removed instead of being treated as errors in the layout

this helps bring Wild in line with real-world linker scripts where /DISCARD/ is commonly used to strip out debug info or other unnecessary sections.

I also updated the symbol handling so that anything coming from discarded sections isn’t emitted or exported. This avoids some failures while keeping the behavior unchanged for normal sections.

fixiing: #1829

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -3620,10 +3620,7 @@ impl<'data, P: Platform> ObjectLayoutState<'data, P> {
                 self.load_debug_section::<A>(common, *part_id, section_index, resources)?;
             }
             SectionSlot::Discard => {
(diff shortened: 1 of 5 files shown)


davidlattimore:
Thanks for your work on this! Just wanted to check if you intend to continue with it or not? Completely understand if not, in which case we'll close the PR and unassign the issue.

PR #1869, opened 2026-04-28
port(MachO): support relocation handling

marxin:
With the change, I can link a multi-input file test-case like this:
```foo.c
int baz();

int bar() {
  return baz();
}

int foo() {
  return bar();
}
...
int value = 42;

int foo();
long baz() {
  return value;
}

int main() {
  return foo();
}
```

As of now, the change only handles symbol relocations, but I am planning to add also sections as a target of relocations. There's one specialty where Mach-O scales relocation values for the low 12-bits of `MachOLow12` based on the size of the load or a store instruction (needs to be decoded from the instruction stream).

Issue: #757

### DIFF
--- a/linker-utils/src/elf.rs
+++ b/linker-utils/src/elf.rs
@@ -1268,6 +1268,8 @@ pub enum AArch64Instruction {
     TstBr,
     Bcond,
     JumpCall,
+    // Mach-O specific
+    MachOLow12,
 }
 
 #[derive(Clone, Debug, Copy, PartialEq, Eq)]
--- a/linker-utils/src/elf.rs
+++ b/linker-utils/src/elf.rs
@@ -1,3 +1,4 @@
+use crate::bit_misc::BitExtraction;
 use crate::bit_misc::BitRange;
 use anyhow::Result;
 use object::LittleEndian;
@@ -5,6 +6,7 @@ use object::read::elf::ProgramHeader as _;
 use object::read::elf::SectionHeader;
 use std::borrow::Cow;
 use std::fmt;
+use std::io::Cursor;
 
 macro_rules! const_name_by_value {
     ($needle: expr, $( $const:ident ),* $(,)?) => {
@@ -1514,7 +1516,7 @@ pub struct RelocationKindInfo {
 
 impl RelocationKindInfo {
     #[inline(always)]
-    pub fn verify(&self, value: i64) -> Result<()> {
+    fn verify(&self, value: i64) -> Result<()> {
         anyhow::ensure!(
             (value as usize).is_multiple_of(self.alignment),
             "Relocation {value} not aligned to {} bytes",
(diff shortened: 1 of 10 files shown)


linker-utils/src/elf.rs
```diff
@@ -1526,6 +1533,44 @@ impl RelocationKindInfo {
...
+
+        if matches!(self.kind, RelocationKind::PairSubtractionULEB128(..)) {
+            // u64 always fits in 10 bytes in the ULEB format: 64 / 7 = 9.14
+            let mut writer = Cursor::new(vec![0u8; 10]);
+            let n = leb128::write::unsigned(&mut writer, value).expect("Must fit into the buffer");
+            anyhow::ensure!(
+                output.len() >= n,
+                "cannot write encoded ULEB128 value of {n} bytes"
+            );
+            output[..n].copy_from_slice(&writer.into_inner()[..n]);
```
davidlattimore, line 1563: Not really related to this PR, since this code was just moved, but I wonder if this might leave uninitialised bytes if we're updating the file in-place.

davidlattimore (approved):
Sorry, I didn't notice that had a conflict with my thunks change, which I merged today, otherwise I'd have merged your PR and resolved the conflict myself. Looks like the conflict is that we both modified adjacent lines, but didn't modify the same lines, so that's good.

The thunk code also looks like it needs a couple of small changes in order to compile with your changes. Here's a patch if you want it:

```patch
diff --git a/libwild/src/macho_aarch64.rs b/libwild/src/macho_aarch64.rs
index 77faadfa0d..00ceb9d82f 100644
--- a/libwild/src/macho_aarch64.rs
+++ b/libwild/src/macho_aarch64.rs
@@ -114,6 +114,7 @@
             mask,
             range,
             size,
+            thunkable: false,
         })
     }

diff --git a/libwild/src/thunks.rs b/libwild/src/thunks.rs
index b92e9a1757..b2c3af8ce2 100644
--- a/libwild/src/thunks.rs
+++ b/libwild/src/thunks.rs
@@ -322,11 +322,11 @@
     resources: &layout::GraphResources<'_, '_, A::Platform>,
     local_symbol_id: SymbolId,
     symbol_id: SymbolId,
-    r_type: u32,
+    rel: <A::Platform as Platform>::RelocationInfo,
 ) {
     if resources.thunk_layout_builder.is_some()
         && let Some(config) = A::thunk_config()
-        && let Some(rel_info) = A::relocation_from_raw(r_type).ok()
+        && let Some(rel_info) = A::relocation_from_raw(rel).ok()
         && rel_info.thunkable
     {
         if section_part_id == config.primary_function_part_id {
```

linker-utils/src/elf.rs
marxin: Good point! I believe the relocation places are generally places already copied from the relocable files, right? Or do I miss something?

davidlattimore: Good point, it shouldn't be a problem

PR #1871, opened 2026-04-28
fix: disable SFrame support by default

mati865:
SFrame support is so hopelessly broken everywhere right now that making it experimental is the best course of action.
With this PR `.sframe` section is emitted if present in the inputs regardless of sections GC, although it will be broken unless `--wild-experimental-sframe` is passed and system's `as` outputs SFrame version 2.
I made an attempt to never emit it unless explicitly enabled [fix: Emit SFrame section only when enabled](https://github.com/wild-linker/wild/pull/1871/commits/8a1e349915547937bb57f6adbf8a952b34ed302c) but it doesn't seem worth the hassle.

CC https://github.com/wild-linker/wild/issues/1576

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -307,7 +307,7 @@ fn fill_padding(mut section_buffers: OutputSectionMap<&mut [u8]>) {
 }
 
 fn write_sframe_section(sframe_buffer: &mut [u8], layout: &ElfLayout) -> Result {
-    if sframe_buffer.is_empty() {
+    if !layout.args().experimental_sframe || sframe_buffer.is_empty() {
         return Ok(());
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -871,7 +871,7 @@ dependencies = [
  "linker-utils",
  "memchr",
  "memmap2",
- "object 0.39.0",
+ "object 0.39.1",
  "perf-event",
  "perfetto-recorder",
  "rayon",
@@ -906,7 +906,7 @@ dependencies = [
  "linker-utils",
  "memchr",
  "memmap2",
- "object 0.39.0",
+ "object 0.39.1",
  "symbolic-demangle",
  "tabled",
  "tempfile",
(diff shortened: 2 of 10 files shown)


mati865:
Great, there is only CI platform able to test current implementation, and it becomes EoL in 2 months.

libwild/src/elf_writer.rs
```diff
@@ -308,7 +307,7 @@ fn fill_padding(mut section_buffers: OutputSectionMap<&mut [u8]>) {
 }
 
 fn write_sframe_section(sframe_buffer: &mut [u8], layout: &ElfLayout) -> Result {
```
marxin, line 309: Not sure how difficult it would be, but could we avoid loading those sections altogether? Similar to how we handle the debug info section with `--strip-debug`?

marxin (approved):
Thanks for working on it, it's really annoying the feature in such an inconsistent shape.

davidlattimore (approved):

libwild/src/elf_writer.rs
mati865: I attempted that in [`8a1e349` (this PR)](https://github.com/wild-linker/wild/pull/1871/commits/8a1e349915547937bb57f6adbf8a952b34ed302c).

There is no consistency at all in all of this.

Older ld.bfd versions were handling (merging and sorting)  SFrame sections just fine but removed them if `--gc-sections` was enabled. Recent versions preserve SFrame sections with sections GC enabled.

LLD doesn't merge them, and retains one of SFrame sections unless sections GC is enabled (standard ELF behaviour for unknown sections).

Before and with this PR Wild retains one of SFrame sections regardless of sections GC. If I make it retain SFrame sections only when the experimental flag is passed (basically revert last two reverts) we get CI failures like here https://github.com/wild-linker/wild/actions/runs/25107607314/job/73572763695?pr=1871. Although we could ignore SFrame entirely in linker-diff, if you prefer that way.

marxin: Well, given the time and experiments you made, let's go with the currently suggested approach in this PR.
The change is pretty minimal.

PR #1873, opened 2026-04-29
feat: Select platform based on -flavor, arg0 then host

davidlattimore:


libwild/src/args.rs
```diff
@@ -182,6 +207,28 @@ impl PlatformKind {
             PlatformKind::Elf
         }
     }
+
+    fn from_flavor(flavor: &str) -> Result<Self> {
+        match flavor {
+            "gnu" => Ok(PlatformKind::Elf),
```
mati865: I didn't know/remember you can use `gnu`. I always used `-flavor ld` which is missing in this PR.

libwild/src/args.rs
```diff
@@ -182,6 +207,28 @@ impl PlatformKind {
...
+            "link" => bail!("Windows (link flavor) is not yet supported"),
+            _ => bail!(
+                "Unknown flavor '{}'. Valid flavors: gnu, darwin, link",
+                flavor
+            ),
+        }
+    }
+
+    fn from_executable_name(name: &str) -> Option<Self> {
+        let base_name = Path::new(name).file_name().and_then(|n| n.to_str())?;
```
mati865: Perhaps you want to use `file_stem()` here to futureproof.

mati865 (approved):

libwild/src/args.rs
marxin: Btw. is it somewhere documented (the `-flavor` option) on the `lld` side?

libwild/src/args.rs
davidlattimore: I assume you mean so we can call the symlink "ld.anything" rather than just "ld.wild". II can probably see cases where that might be useful, so I've done it.

libwild/src/args.rs
davidlattimore: This is what I found: https://releases.llvm.org/12.0.0/tools/lld/docs/Driver.html

marxin: Pretty old version, but matches the source code:
https://github.com/llvm/llvm-project/blob/6ea383f58ee7ec8e64247ea5106c117ecf5c8527/lld/Common/DriverDispatcher.cpp#L31-L38

libwild/src/args.rs
mati865: Oh sorry, I thought of Windows `.exe` extension for binaries. Now that you mention the symlinks, I'm less certain about the `file_stem`.

davidlattimore: It does mean that we can be invoked as "ld.lld" and work, which I did just yesterday since at the time I didn't know a better way to override the linker used by the chromium build process (I do now). The `.exe` extension is perhaps a good reason to accept it too

mati865: Sure, glad that I could accidentally help.

PR #1875, opened 2026-04-29
chore: Double maximum thunk bytes per block

davidlattimore:


davidlattimore:
This is just a temporary fix until we can come up with a better approach for taking into account the size required for thunks when computing whether thunks are needed. See discussion - https://github.com/wild-linker/wild/issues/1763#issuecomment-4340760880

PR #1876, opened 2026-04-30
test: Add basic support for running MachO tests

davidlattimore:


davidlattimore:
I've only tested this on Linux, so haven't tried running the executable produced by the test.

PR #1877, opened 2026-04-30
feat: add -Ttext/-Tdata/-Tbss segment layout for SEGMENT_START support

Blazearth:
This PR adds support for `-Ttext`, `-Tdata`, and `-Tbss`, wiring them into the layout phase so they affect the actual placement of `.text`, `.data`, and `.bss`. The section addresses are updated according to the provided `-T*` values
This builds on PR #1851, which introduced `SEGMENT_START` support using default values only.
This need to be merged first before PR #1851

### DIFF
--- a/libwild/src/output_section_id.rs
+++ b/libwild/src/output_section_id.rs
@@ -581,6 +581,22 @@ impl<'data, P: Platform> OutputSections<'data, P> {
         }
     }
 
+    /// Applies `--section-start` / `-Ttext` / `-Tdata` / `-Tbss` overrides to the built-in
+    /// sections `.text`, `.data`, and `.bss`. Must be called after `with_base_address` and before
+    /// the layout phase reads `section_info.location`.
+    pub(crate) fn apply_section_start_overrides(&mut self, args: &P::Args) {
+        for (section_id, name) in [
+            (TEXT, SectionName(b".text")),
+            (DATA, SectionName(b".data")),
+            (BSS, SectionName(b".bss")),
+        ] {
+            if let Some(address) = args.start_address_for_section(name) {
+                self.section_infos.get_mut(section_id).location =
+                    Some(linker_script::Location { address });
+            }
+        }
+    }
+
     pub(crate) fn add_named_section(
         &mut self,
         name: SectionName<'data>,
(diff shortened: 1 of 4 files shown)


libwild/src/output_section_id.rs
```diff
@@ -581,6 +581,22 @@ impl<'data, P: Platform> OutputSections<'data, P> {
         }
     }
 
+    /// Applies `--section-start` / `-Ttext` / `-Tdata` / `-Tbss` overrides to the built-in
+    /// sections `.text`, `.data`, and `.bss`. Must be called after `with_base_address` and before
+    /// the layout phase reads `section_info.location`.
+    pub(crate) fn apply_section_start_overrides(&mut self, args: &P::Args) {
+        for (section_id, name) in [
+            (TEXT, SectionName(b".text")),
```
davidlattimore, line 589: Should -Ttext effect the address of the .text section or of the start of the executable segment? For your test, these are the same things, but if we have other executable sections such as .init, .plt etc, these will be before .text.

An alternative way of changing locations would be to alter `build_output_order_and_program_segments`. You'd likely need to pass Args to the function, but then you could add the `OrderEvent::SetLocation` directly before the first executable section. That would make it work in a way that would be more similar to how linker scripts work.

Blazearth: I tested the GNU ld's actual behaviour to confirm and interestingly ! it puts the section specifically at the given addr rather than the whole segment.
I tested With a binary that has .init and -Ttext=0x700000:
```
LOAD  0x401000  R E   ← .init lives here, separate segment
LOAD  0x700000  R E   ← .text starts here
```
_start is at 0x700000, _init is at 0x401000 — they end up in separate LOAD segments. GNU ld's -Ttext sets the address of the .text section specifically, not the start of the RX segment. 
So injecting SetLocation before the first executable section (before .init/PLT_GOT) would actually implement -Ttext-segment which is a different flag.
In GNU ld to set the base address of the whole executable segment, it uses a completely different flag: `-Ttext-segment=addr`.
The current approach of setting section_info.location on .text matches GNU ld/lld behavior exactly.

In GNU ld ,the gap between `.init at 0x401000` and `.text at 0x700000` is handled naturally — they go into separate LOAD segments, so there's no wasted file space, only a gap in virtual address space. Wild does the same.

davidlattimore: Interesting. Thanks for checking this!

davidlattimore (approved):

PR #1882, opened 2026-05-03
refactor: Track primary-part thunk candidates in a queue

lapla-cogito:
We are currently using every bit of the u16 `ValueFlags`. However, `HAS_RANGE_LIMITED_REL` does not really need to live in `ValueFlags` at all: it was only ever used as a transient marker between relocation processing and thunk layout, and the non-primary path already tracked the equivalent information in `non_primary_referenced_symbols` without consuming a flag bit. Move the primary path to use the same pattern via a new `primary_thunkable_locals` queue on `ThunkLayoutBuilder`.

### DIFF
--- a/libwild/src/thunks.rs
+++ b/libwild/src/thunks.rs
@@ -19,8 +19,8 @@
 //! * Non-primary part references non-primary part: Assumed to be in range.
 //! * Non-primary part references primary part: Stored in
 //!   ThunkLayoutBuilder::non_primary_referenced_symbols.
-//! * Prmary part references anything: ValueFlags::HAS_RANGE_LIMITED_REL set for local symbol in the
-//!   object that made the reference.
+//! * Prmary part references anything: the (file_id, local_symbol_id) pair is pushed into
+//!   ThunkLayoutBuilder::primary_thunkable_locals.
 
 use crate::input_data::FileId;
 use crate::layout;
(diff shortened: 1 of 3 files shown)


libwild/src/thunks.rs
```diff
@@ -87,14 +87,20 @@ pub(crate) struct ThunkLayoutBuilder {
     /// Symbols that are defined in primary parts and referenced by range-limited relocations from
     /// non-primary parts.
     non_primary_referenced_symbols: SegQueue<SymbolId>,
+
+    /// `(file_id, local_symbol_id)` pairs recorded each time a thunkable, range-limited
+    /// relocation is encountered in a primary-part section. Filled in concurrently during
+    /// relocation processing and consumed once when building the thunk layout. May contain
+    /// duplicates.
+    primary_thunkable_locals: SegQueue<(FileId, SymbolId)>,
```
davidlattimore, line 95: I did something like this in an earlier iteration of the range-extension thunks implementation. Unfortunately, it's not great for performance..

I just ran a benchmark with this change and linking a debug build of chromium for aarch64 more than doubled - from 3.04 seconds to 7.16 seconds.

If you're wanting to make a change that needs to add another ValueFlag, the easiest option for now might be to increase it to 32 bits. I just ran a benchmark of doing that there was more or less no measurable performance change.

lapla-cogito: > I just ran a benchmark of doing that there was more or less no measurable performance change.

Ah, I'd been under the mistaken impression that switching to 32-bit would degrade performance. Thanks.

PR #1884, opened 2026-05-03
feat: support `--nmagic`

TechnoPorg:
`--nmagic` disables page alignment for loadable sections and prevents linking against shared libraries. Support for `--omagic` will follow once this is cleaned up and merged.

ld will merge the R and RX loadable segments with `--nmagic`, whereas lld will not. I have chosen to match the LLD behaviour.

~~I'm aware of two shortcomings so far in this PR:~~

- ~~Cross testing is disabled because QEMU apparently requires loadable segments to be page-aligned. This is rather unfortunate, as the tests in nmagic.c are passing on my x64 laptop, but failing in aarch64 CI, and I don't have an aarch64 box to test on.~~
- ~~ld and lld both don't write the PHDR program header entry with --nmagic, whereas wild with this PR does. Advice on how to match that behaviour would be appreciated.~~

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -126,6 +126,8 @@ pub struct ElfArgs {
     pub(crate) should_output_executable: bool,
     pub(crate) should_output_partial_object: bool,
 
+    pub(crate) nmagic: bool,
+
     rpath_set: IndexSet<String>,
 
     pub experimental_sframe: bool,
@@ -313,6 +315,8 @@ impl Default for ElfArgs {
             pack_dyn_relocs: PackDynRelocs::None,
             use_android_relr_tags: false,
 
+            nmagic: false,
+
             unresolved_symbols: UnresolvedSymbols::ReportAll,
             error_unresolved_symbols: true,
             allow_multiple_definitions: false,
@@ -412,6 +416,18 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(
         || args.pack_dyn_relocs == PackDynRelocs::AndroidRelr
     {
         args.warn_unsupported("--pack-dyn-relocs=android")?;
+    }
+
+    if args.nmagic {
+        // GNU_RELRO requires segments to start on page boundaries and cover an entire page
+        args.relro = false;
+        if args.max_page_size.is_some() {
+            args.warning("-z max-page-size is incompatible with --nmagic");
+        }
+        args.common_mut()
+            .inputs
+            .iter_mut()
+            .for_each(|input| input.modifiers.allow_shared = false);
     }
 
     Ok(())
@@ -1795,6 +1811,25 @@ fn setup_argument_parser() -> ArgumentParser<ElfArgs> {
             Ok(())
         });
 
+    parser
+        .declare()
+        .short("n")
+        .long("nmagic")
+        .help("Disable page alignment of sections and disable linking against shared libraries")
+        .execute(|args, _modifier_stack| {
+            args.nmagic = true;
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("no-nmagic")
+        .help("Page align sections (default)")
+        .execute(|args, _modifier_stack| {
+            args.nmagic = false;
+            Ok(())
+        });
+
     add_silently_ignored_flags(&mut parser);
     add_default_flags(&mut parser);
 
@@ -1998,6 +2033,10 @@ impl platform::Args for ElfArgs {
     }
 
     fn loadable_segment_alignment(&self) -> Alignment {
+        if self.nmagic {
+            return Alignment { exponent: 0 };
+        }
+
         if let Some(max_page_size) = self.max_page_size {
             return max_page_size;
         }
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -815,6 +815,13 @@ impl platform::Platform for Elf {
                 }
             }
         }
+
+        // The PHDR program header should only be present if --nmagic is not set
+        for (segment_def, keep) in program_segments.into_iter().zip(keep_segments.iter_mut()) {
+            if segment_def.segment_type == pt::PHDR {
+                *keep = !args.nmagic;
+            }
+        }
     }
 
     fn program_segment_defs() -> &'static [ProgramSegmentDef] {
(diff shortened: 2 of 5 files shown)


libwild/src/elf.rs
```diff
@@ -411,7 +411,6 @@ impl platform::Platform for Elf {
             output_section_id::PROGRAM_HEADERS,
             output_section_id::SECTION_HEADERS,
             output_section_id::SHSTRTAB,
-            output_section_id::RELRO_PADDING,
```
davidlattimore: Why is RELRO_PADDING removed?

libwild/src/input_data.rs
```diff
@@ -718,9 +718,16 @@ impl<'data, P: Platform> TemporaryState<'data, P> {
             modifiers: input_ref.file.modifiers,
         };
 
-        let object = ParsedInputObject::new(&input_bytes, self.args);
+        let object = InputRecord::Object(ParsedInputObject::new(&input_bytes, self.args));
 
-        Ok(InputRecord::Object(object))
+        if object.is_dynamic_object() && !input_ref.file.modifiers.allow_shared {
+            bail!(
+                "attempted static link of dynamic object {}",
```
davidlattimore: This looks like it's adding some error detection that we were perhaps missing before. It could be good to do this as a separate PR, ideally with a test to detect this error scenario.

libwild/src/args/elf.rs
```diff
@@ -418,6 +422,18 @@ pub(crate) fn parse<S: AsRef<str>, I: Iterator<Item = S>>(
         args.warn_unsupported("--pack-dyn-relocs=android")?;
     }
 
+    if args.nmagic {
+        // PT_GNU_RELRO requires segments to start on page boundaries and cover an entire page
+        args.relro = false;
+        if args.max_page_size.is_some() {
```
davidlattimore, line 424: Checking this here means that we'll only error if max-page-size is set before nmagic. Probably best to check for invalid flag combinations after parsing is complete.

wild/tests/sources/elf/nmagic/nmagic.c
```diff
@@ -0,0 +1,26 @@
+//#Config:default
+//#LinkArgs:--nmagic -z relro
+//#Object:runtime.c
+//#DoesNotContain:relro
+//#ExpectLoadAlignment:0x8 0x1
+// GNU ld merges segment.LOAD.R and segment.LOAD.RX, which wild and lld do not
+//#DiffIgnore: segment.LOAD.RX.alignment
+//#DiffIgnore:segment.LOAD.RWX.alignment
+// QEMU requires segments to be page-aligned
```
davidlattimore: Given that the binaries are segfaulting when run in CI, it seems likely that the cause is the same reason QEMU was rejecting them - i.e. lack of page alignment. Perhaps see if you can fix the page alignment, then maybe both qemu and CI will be happy.

davidlattimore (commented):
Did you come across some software that uses this flag? It's the sort of flag that if someone is using it in software that's actively maintained, they should probably be working to migrate off of it. In that regard, it wouldn't say this is an essential flag for us to implement. That said, I don't object to implementing it, so long as it doesn't add substantial complexity.

TechnoPorg:
> Did you come across some software that uses this flag? It's the sort of flag that if someone is using it in software that's actively maintained, they should probably be working to migrate off of it. In that regard, it wouldn't say this is an essential flag for us to implement. That said, I don't object to implementing it, so long as it doesn't add substantial complexity.

It's part of a lot of embedded projects for boards with unaligned flash memory ([source](https://github.com/rust-embedded/cortex-m-quickstart/pull/95)). While this is far from the largest blocker to using wild in embedded projects, it's one I felt was realistically within my capabilities to contribute.

libwild/src/elf.rs
TechnoPorg: The short answer is that the x64 binaries segfault if it's not removed. The long answer is that a few lines down, the segment gets conditionally kept if args.relro is set, but since it's force kept above, the relro check doesn't have any effect. Is this something that should be split into another PR?

wild/tests/sources/elf/nmagic/nmagic.c
TechnoPorg: Hmm. Given that --nmagic is there to disable page alignment, I'm not sure how best to test this...

libwild/src/elf.rs
davidlattimore: Ah, good catch. Yes, I think a separate PR for that would be good, that way we can make sure it's tested and list the change in the release notes.

wild/tests/sources/elf/nmagic/nmagic.c
davidlattimore: If unaligned segments are a characteristic of the flag, but the hardware (and emulators) that we run on don't support running with unaligned segments, then perhaps we need "RunEnabled:false", then add whatever assertions we can about the output file to convince ourselves that it's correct.

libwild/src/args/elf.rs
TechnoPorg: I thought that this did run after parsing is complete? `--nmagic -z max-page-size=4096` and `-z max-page-size=4096 --nmagic` both give me the warning.

mati865: This is the right place.
The parsing is done before line 398.

wild/tests/sources/elf/nmagic/nmagic.c
TechnoPorg: I've set up an aarch64 VM and on there, the wild binary segfaults while the ld/lld binaries don't, so there's seemingly something wrong here besides page alignment. I'll look into it and get back to you.

TechnoPorg:
I've disabled running the binaries in `nmagic.c`, since unaligned loadable segments either segfault when running natively or are refused by QEMU. This differs somewhat from ld and LLD, which as far as I can tell, behave like this with --nmagic:

- Sections are never page aligned
- On x86_64, riscv64, and loongarch64, p_vaddr ≡ p_offset (mod 4KiB)
- On AArch64, p_vaddr and p_offset are not congruent except by chance

The last point feels the most correct to me for this flag, especially assuming it's for embedded development. Let me know if you think otherwise.

davidlattimore (approved):

PR #1886, opened 2026-05-04
fix: check for static link of dynamic object

TechnoPorg:
The allow_shared modifier currently only applies if given a library name as an input, but is ignored if given a filename. This PR adds a check to see whether a shared object given as a filename is allowed to be linked.

Split off from #1884.

### DIFF
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -718,9 +718,16 @@ impl<'data, P: Platform> TemporaryState<'data, P> {
             modifiers: input_ref.file.modifiers,
         };
 
-        let object = ParsedInputObject::new(&input_bytes, self.args);
+        let object = InputRecord::Object(ParsedInputObject::new(&input_bytes, self.args));
 
-        Ok(InputRecord::Object(object))
+        if object.is_dynamic_object() && !input_ref.file.modifiers.allow_shared {
+            bail!(
+                "Attempted static link of dynamic object {}",
+                input_ref.file.filename.display()
+            );
+        }
+
+        Ok(object)
     }
 }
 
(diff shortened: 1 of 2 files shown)


wild/tests/sources/elf/no-dynamic-in-static/no-dynamic-in-static.c
```diff
@@ -0,0 +1,5 @@
+//#Mode:static
+//#Shared:force-dynamic-linking.c
+//#ExpectError:attempted static link of dynamic object .*
```
davidlattimore: I don't think the '.*' at the end will do anything. At least from what I recall, the check is that the regex matches anywhere in the line.

libwild/src/input_data.rs
```diff
@@ -718,9 +718,16 @@ impl<'data, P: Platform> TemporaryState<'data, P> {
             modifiers: input_ref.file.modifiers,
         };
 
-        let object = ParsedInputObject::new(&input_bytes, self.args);
+        let object = InputRecord::Object(ParsedInputObject::new(&input_bytes, self.args));
 
-        Ok(InputRecord::Object(object))
+        if object.is_dynamic_object() && !input_ref.file.modifiers.allow_shared {
+            bail!(
+                "attempted static link of dynamic object {}",
```
davidlattimore: For consistency with our error message elsewhere, could you uppercase the "A" in "Attempted"?

davidlattimore (commented):
Thanks for this. Always good to see improvements to our error reporting.

libwild/src/input_data.rs
TechnoPorg: Done. Since GNU ld uses a lowercase, I've made the regex match either.

davidlattimore (approved):

PR #1890, opened 2026-05-05
chore: Add job to update homebrew tap on release

davidlattimore:


.github/workflows/release.yml
```diff
@@ -101,3 +102,34 @@ jobs:
...
           files: |
             wild-linker-*.tar.gz
+
+  update-tap:
+    needs: release
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Tap
+        uses: actions/checkout@v4
```
lapla-cogito:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/release.yml
```diff
@@ -101,3 +102,34 @@ jobs:
           draft: true
           files: |
             wild-linker-*.tar.gz
```
lapla-cogito, line 103: Here, since `*.tar.gz.sha256` isn't included in the release, wouldn't `gh release download $VERSION -R wild-linker/wild -p "*.sha256"` result in no downloads being performed? I think it would be better to directly download and `shasum` the tarball on the `update-tap` side.

.github/workflows/release.yml
```diff
@@ -101,3 +102,34 @@ jobs:
           draft: true
```
lapla-cogito, line 101: `update-tap` runs immediately after this step, but the release is still in draft status because of this, so I assume the asset URLs shouldn't be accessible yet.

lapla-cogito (commented):
Haven't tested it, but these seem like potential concerns.

.github/workflows/release.yml
davidlattimore: Good point! Done

.github/workflows/release.yml
davidlattimore: Another great point! I've moved it to a separate workflow that only triggers when the release gets published.

lapla-cogito (approved):
Thanks! Looks good.

.github/workflows/promote.yml
```diff
@@ -0,0 +1,40 @@
...
+          ARM_SHA=$(shasum -a 256 "wild-linker-$VERSION-aarch64-unknown-linux-gnu.tar.gz" | awk '{print $1}')
+
+          echo "Updating Homebrew formula for $VERSION"
+          echo "x86_64 hash: $X86_SHA"
+          echo "aarch64 hash: $ARM_SHA"
+
+          VERSION=$VERSION X86_SHA=$X86_SHA ARM_SHA=$ARM_SHA ./update.sh
+
+          git config user.name "Wild Release Bot"
+          git config user.email "actions@github.com"
```
lapla-cogito:
```suggestion
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

lapla-cogito (commented):
Ah, there're some nits.

.github/workflows/promote.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/checkout@v6
+        with:
+          repository: wild-linker/homebrew-wild
+          token: ${{ secrets.HOMEBREW_TAP_TOKEN }}
+
+      - name: Update Formula
+        run: |
+          VERSION=${{ github.event.release.tag_name }}
+
+          gh release download $VERSION -R wild-linker/wild \
```
lapla-cogito: considering potential future transfers or something like that
```suggestion
          gh release download $VERSION -R ${{ github.repository }} \
```

lapla-cogito (approved):

PR #1892, opened 2026-05-05
fix: only keep RELRO_PADDING section when relro is enabled

TechnoPorg:
As mentioned in https://github.com/wild-linker/wild/pull/1884#discussion_r3179461807, the check in `apply_force_keep_sections` to only keep the RELRO_PADDING section is overridden by the fact that the section is unconditionally kept a few lines above.

New ExpectProgramHeader/NoProgramHeader test directives have been added here to test this behaviour.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -57,6 +57,12 @@
 //! ExpectLoadAlignment:{alignment} Checks that the first PT_LOAD segment in the output binary has
 //! the specified alignment.
 //!
+//! ExpectProgramHeader:{type} Checks that the output binary contains a program header of the
+//! specified type.
+//!
+//! NoProgramHeader:{type} Checks that the output binary contains no program headers of the
+//! specified type.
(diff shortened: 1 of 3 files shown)


wild/tests/integration_tests.rs
```diff
@@ -3447,6 +3485,12 @@ impl Assertions {
                 if !self.absent_dynamic_entries.is_empty() {
                     bail!("NoDynamic is not supported for MachO",);
                 }
+                if !self.expected_program_headers.is_empty() {
+                    bail!("ExpectedProgramHeader is not supported for MachoO",);
```
lapla-cogito: This seems slightly different from the actual directive name 🤔 (`ExpectedProgramHeader` vs. `ExpectProgramHeader`)

lapla-cogito: Also, s/MachoO/MachO

lapla-cogito (approved):
Thanks!

PR #1898, opened 2026-05-07
SFrame V3 MVP

mati865:
Commits starting with `update` and `account` are what https://sourceware.org/binutils/docs-2.46/sframe-spec.html#Changes-from-Version-2-to-Version-3 is about.

Just like ld.bfd did, this change drops support for SFrame V2 due to how annoying it would be to support both formats at the same time, and the fact that the decisions by the distros to enable SFrame V2 by default were short-lived.

At the time of writing this description, glibc still can only read SFrame V2 (the changes are lingering on the mailing list), so I'm posting this as a draft. Hopefully it doesn't take too long from now before we can test it on the CI.
Can be tested with https://gist.github.com/mati865/5d5fead3a7cbefcbe9a609274a0d27eb if your system has Binutils 2.46.

This PR only ports existing code from SFrame V2 to V3. It doesn't add new `SFRAME_FDE_TYPE_FLEX` or `SFRAME_FDE_TYPE_DEFAULT`.

Fixes https://github.com/wild-linker/wild/issues/1576

PR #1899, opened 2026-05-07
Remove nonexistent SFrame flag

mati865:
`SFRAME_AARCH64_PAUTH_KEY` is a bit in `sfda_func_info` rather than `sfp_flags`. Since we copy `sfda_func_info` as-is, we don't need to care about it.

It has not changed in SFrame V3, it was already wrong in the previous implementation.

Tried to stack it on top of #1898, but you cannot do it from a fork. Only the last commit matters here.

### DIFF
--- a/libwild/src/sframe.rs
+++ b/libwild/src/sframe.rs
@@ -123,12 +123,12 @@ fn read_u32(data: &[u8], offset: usize) -> u32 {
     u32::from_le_bytes(data[offset..offset + 4].try_into().unwrap())
 }
 
-fn read_i32(data: &[u8], offset: usize) -> i32 {
-    i32::from_le_bytes(data[offset..offset + 4].try_into().unwrap())
+fn read_i64(data: &[u8], offset: usize) -> i64 {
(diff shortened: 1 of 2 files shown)


mati865:
Apparently you still cannot, but now `gh` will copy the branches and mess everything up..
```
❯ gh stack submit
Checking stack state...
? Multiple remotes found. Which remote should be used? upstream
? Save "upstream" as the default remote for all gh stack operations? Yes
✓ Saved "upstream" as the default remote for gh stack
To change later, run: git config gh-stack.remote <other-remote>
To clear, run:        git config --unset gh-stack.remote
Pushing to upstream...
PR #1898 (https://github.com/wild-linker/wild/pull/1898) for push-psuukqllkpkt is up to date
✓ Updated base branch for PR #1899 (https://github.com/wild-linker/wild/pull/1899) to push-psuukqllkpkt
⚠ Could not create stack: Pull requests from forks cannot be added to stacks.
✓ Pushed and synced 2 branches
```

PR #1902, opened 2026-05-08
fix: Implement plugin callsbacks required for thin LTO

### LINKED ISSUE #1571: lto: wild's LTO plugin doesn't work with LLVM's thin LTO (`wild: error: Linker plugin error: Failed to release file information`)
Wild like mold (https://github.com/rui314/mold/issues/1358 See-also: https://bugs.gentoo.org/947697#c7 ) doesn't natively understand LLVM's IR e.g if you use LTO without passing it in `$LDFLAGS` you get:
```sh
$ CC=clang RUSTFLAGS="-C linker=clang -C linker-plugin-lto -C link-arg=-fuse-ld=wild" cargo build --verbose --release
error: linking with `clang` failed: exit status: 1
[...]
  = note: wild: error: Input file /tmp/portage/www-apps/zola-0.22.1-r1/work/zola-0.22.1/target/release/build/quote-434f99b422b32673/build_script_build-434f99b422b32673.build_script_build.3766f06489545ff0-cgu.0.rcgu.o contains LLVM-IR, but linker plugin was not supplied
```

So during linking stage you also have to pass LTO (e.g via `$LDFLAGS` with `-flto` or in rust with `RUSTFLAGS="-C link-arg=-flto"`). But unlike mold wild isn't accepting the -flto=thin option? When passing that I get:
```sh
CC=clang RUSTFLAGS="-C linker=clang -C linker-plugin-lto -C link-arg=-flto=thin -C link-arg=-fuse-ld=wild" cargo build --verbose --release
[...]

= note:  "clang" "-m64" "/tmp/rustcYjC6Y9/symbols.o" "<2 object files omitted>" "-Wl,--as-needed" "-Wl,-Bstatic" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib/{libstd-*,libpanic_unwind-*,libobject-*,libmemchr-*,libaddr2line-*,libgimli-*,libcfg_if-*,librustc_demangle-*,libstd_detect-*,libhashbrown-*,librustc_std_workspace_alloc-*,libminiz_oxide-*,libadler2-*,libunwind-*,liblibc-*,librustc_std_workspace_core-*,liballoc-*,libcore-*,libcompiler_builtins-*}.rlib" "-Wl,-Bdynamic" "-lgcc_s" "-lc" "-L" "/tmp/rustcYjC6Y9/raw-dylibs" "-Wl,--eh-frame-hdr" "-Wl,-z,noexecstack" "-Wl,-plugin-opt=O0,-plugin-opt=mcpu=x86-64" "-L" "<sysroot>/lib/rustlib/x86_64-unknown-linux-musl/lib" "-o" "/tmp/portage/www-apps/zola-0.22.1-r1/work/zola-0.22.1/target/release/build/quote-12df9f7f6ab9afbc/build_script_build-12df9f7f6ab9afbc" "-Wl,--gc-sections" "-pie" "-Wl,-z,relro,-z,now" "-Wl,--strip-debug" "-nodefaultlibs" "-flto=thin" "-fuse-ld=wild"
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: wild: error: Linker plugin error: Failed to release file information
          Linker plugin fatal error: Failed to get file information
          Linker plugin fatal error: Failed to get file information
          Linker plugin fatal error: Failed to release file information
          Linker plugin fatal error: Failed to release file information
```

Wild works fine if I only use `-flto` (not `-flto=thin`). Also `-C link-plugin-lto=/path/to/LLVMgold.so` mentioned in https://bugs.gentoo.org/947697#c10 also works.

davidlattimore:
Fixes #1571

### DIFF
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -124,6 +124,10 @@ struct FileHandle<'data> {
 
     /// This isn't known initially because we allocate file IDs later.
     file_id: AtomicCell<Option<FileId>>,
+
+    fd: RawFd,
+    offset: u64,
+    name: &'data CStr,
 }
 
 #[derive(Default)]
@@ -249,14 +253,22 @@ impl<'data> LinkerPlugin<'data> {
                 wrap_symbols: self.wrap_symbols,
             };
 
+            let name = CString::new(input_ref.file.filename.as_os_str().as_encoded_bytes())?;
+            let name = CStr::from_bytes_with_nul(
+                self.herd.get().alloc_slice_copy(name.as_bytes_with_nul()),
+            )
+            .unwrap();
+
             let handle = FileHandle {
                 data,
+                name,
+                fd,
+                offset,
                 file_id: AtomicCell::new(None),
             };
 
             let handle = self.herd.get().alloc(handle);
 
-            let name = CString::new(input_ref.file.filename.as_os_str().as_encoded_bytes())?;
             let file = LdPluginInputFile {
                 name: name.as_ptr(),
                 fd,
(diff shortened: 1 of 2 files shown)


libwild/src/linker_plugins.rs
```diff
@@ -249,14 +253,22 @@ impl<'data> LinkerPlugin<'data> {
...
+                fd,
+                offset,
                 file_id: AtomicCell::new(None),
             };
 
             let handle = self.herd.get().alloc(handle);
 
-            let name = CString::new(input_ref.file.filename.as_os_str().as_encoded_bytes())?;
             let file = LdPluginInputFile {
                 name: name.as_ptr(),
```
mati865, line 273: Wasn't that use-after-free previously?
CString created allocation and unless I'm missing something due to viewing this on my phone, we only used pointer to it. So the name would be dropped when going out of scope.

With the new code name is stored within the struct, so the allocation is kept alive.

Borrow checker doesn't catch such issues because it's legal and safe to create dangling pointers. The part requiring unsafe code is reading from the pointers.

mati865 (approved):
Such a lovely interface.

libwild/src/linker_plugins.rs
davidlattimore: The pointer to name would previously have been valid for the duration of the call into the plugin's claim-file hook. But if the claim-file hook stored the pointer and used it later, then yes, it would have been a use-after-free. AFAIK though it doesn't, but the linker plugin API docs are pretty absent of any mentions of required lifetimes, so who knows.

mati865: Ah, this function returns `LtoInputInfo`, not `LdPluginInputFile`. Line wrapping in the mobile view tricked me.

PR #1910, opened 2026-05-08
port(MachO): fat binary support

theoparis:
This PR adds FAT binary identification to libwild. Currently it is not fully handled. In order to use wild with rustc and clang, `-arch arm64` and other macos-specific flags need to be implemented.

### DIFF
--- a/libwild/src/file_kind.rs
+++ b/libwild/src/file_kind.rs
@@ -18,6 +18,7 @@ pub(crate) enum FileKind {
     ElfObject,
     ElfDynamic,
     MachOObject,
+    FatMachOObject,
     WasmObject,
     Archive,
     ThinArchive,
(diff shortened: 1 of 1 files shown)


libwild/src/file_kind.rs
```diff
@@ -18,6 +18,7 @@ pub(crate) enum FileKind {
     ElfObject,
     ElfDynamic,
     MachOObject,
+    FatMachO,
```
marxin: I would name it rather `FatMachOObject`

marxin (changes requested):

marxin:
Can you please rebase and address the comment?

marxin:
Please address the CI failure.

marxin (approved):

Issue #1431, opened 2026-01-09
Feature request: wasm support


Ddystopia:
Hello, it would be great to eventually get incremental linking on wasm too! Do you plan this too? I didn't see it mentioned in README so I opened this issue.

lapla-cogito:
I've begun conducting preliminary research for implementing WASM porting. However, before proceeding with full implementation, it would be beneficial to abstract ELF-specific handling as much as possible through traits (and since part of this work overlaps with the Mach-O port, some coordination will be necessary). Considering current priorities, completing the WASM port may take some time but remains within our scope.

marxin:
@lapla-cogito given the recent active development and you being able to link a numerous non-trivial projects, isn't it a right time to enable it by default and somewhat promote it as functional (and really fast 🚀)? What do you think about it?

lapla-cogito:
I believe we can make the wasm feature default once Wild supports TLS in a multi-threaded environment and debug info.

PR #1912, opened 2026-05-09
port(wasm): Initial scaffolding for WebAssembly support

lapla-cogito:
Part of #1431

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ edition = "2024"
 anyhow = "1.0.97"
 ar = "0.9.0"
 atomic-take = "1.0.0"
-bitflags = "2.4.0"
+bitflags = "2.4.1"
 blake3 = { version = "1.0.0", features = ["rayon"] }
 bumpalo-herd = "0.1.2"
 bytesize = "2.0.0"
@@ -50,7 +50,7 @@ iced-x86 = { version = "1.17.0", default-features = false, features = [
     "decoder",
     "gas",
 ] }
-indexmap = "2.11.4"
+indexmap = "2.14.0"
 indicatif = "0.18.3"
 itertools = "0.14.0"
 jobserver = "0.1.30"
@@ -101,6 +101,7 @@ tracing-subscriber = { version = "0.3.16", default-features = false, features = 
 uuid = { version = "1.0.0", features = ["v4"] }
 wait-timeout = "0.2.0"
 wait4 = "0.1.3"
+wasmparser = "0.248.0"
 which = "8.0.0"
 winnow = { version = "1.0.0", features = ["simd"] }
 zerocopy = { version = "0.8.27", features = ["derive"] }
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -32,7 +32,6 @@ use std::path::PathBuf;
 
 pub mod elf;
 pub mod macho;
-#[cfg(feature = "wasm")]
 pub mod wasm;
 
 use crate::error::Warning;
@@ -134,7 +133,6 @@ impl Args {
         let mut args = match platform {
             PlatformKind::Elf => Args::Elf(elf::ElfArgs::new()?),
             PlatformKind::MachO => Args::MachO(macho::MachOArgs::new()?),
-            #[cfg(feature = "wasm")]
             PlatformKind::Wasm => Args::Wasm(wasm::WasmArgs::new()?),
         };
 
@@ -166,7 +164,6 @@ impl Args {
         match self {
             Args::Elf(args) => args.parse(input),
             Args::MachO(args) => args.parse(input),
-            #[cfg(feature = "wasm")]
             Args::Wasm(args) => args.parse(input),
         }
     }
@@ -189,7 +186,6 @@ impl Args {
         match self {
             Args::Elf(elf_args) => &elf_args.common,
             Args::MachO(macho_args) => &macho_args.common,
-            #[cfg(feature = "wasm")]
             Args::Wasm(wasm_args) => &wasm_args.common,
         }
     }
@@ -198,7 +194,6 @@ impl Args {
         match self {
             Args::Elf(elf_args) => &mut elf_args.common,
             Args::MachO(macho_args) => &mut macho_args.common,
-            #[cfg(feature = "wasm")]
             Args::Wasm(wasm_args) => &mut wasm_args.common,
         }
     }
@@ -207,7 +202,6 @@ impl Args {
 enum PlatformKind {
     Elf,
     MachO,
-    #[cfg(feature = "wasm")]
     Wasm,
 }
 
@@ -225,12 +219,7 @@ impl PlatformKind {
             "gnu" | "ld" => Ok(PlatformKind::Elf),
             "darwin" | "ld64" => Ok(PlatformKind::MachO),
             "link" => bail!("Windows (link flavor) is not yet supported"),
-            #[cfg(feature = "wasm")]
             "wasm" | "ld-wasm" => Ok(PlatformKind::Wasm),
-            #[cfg(not(feature = "wasm"))]
-            "wasm" | "ld-wasm" => {
-                bail!("Wasm (link flavor) requires the `wasm` feature to be enabled")
-            }
             _ => bail!(
                 "Unknown flavor '{}'. Valid flavors: gnu, darwin, link",
                 flavor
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -28,7 +28,7 @@ edition = "2024"
 anyhow = "1.0.97"
 ar = "0.9.0"
 atomic-take = "1.0.0"
-bitflags = "2.4.0"
+bitflags = "2.4.1"
```
lapla-cogito, line 31: This is required for the `minimal-versions` check (the same applies to `indexmap`).

libwild/src/args/wasm.rs
```diff
@@ -0,0 +1,164 @@
...
+}
+
+impl Default for WasmArgs {
+    fn default() -> Self {
+        Self {
+            common: CommonArgs::default(),
+
+            // TODO: move to CommonArgs
+            relocation_model: RelocationModel::NonRelocatable,
+            output: Arc::from(Path::new("a.out")),
```
lapla-cogito, line 51: wasm-ld does not specify a default output filename, making it mandatory to specify one using `-o`. To comply with this, we will need to change `output` to `Option<Arc<Path>>`.

libwild/src/args.rs
```diff
@@ -32,6 +32,8 @@ use std::path::PathBuf;
 
 pub mod elf;
 pub mod macho;
+#[cfg(feature = "wasm")]
```
marxin: Similarly to `Mach-O` port - I would not guard it by a feature flag.

libwild/src/file_kind.rs
```diff
@@ -72,6 +73,10 @@ impl FileKind {
                 "Expected object file"
             );
             Ok(FileKind::MachOObject)
+        } else if bytes.starts_with(b"\0asm") {
+            // Wasm binary magic number is `\0asm` followed by a 4-byte version.
+            ensure!(bytes.len() >= 8, "Invalid wasm file (too short)");
```
marxin: I think we should consistently use the upper `W`: `Wasm` in the PR. Can you please check it and unify?

libwild/src/file_kind.rs
```diff
@@ -119,6 +124,7 @@ impl std::fmt::Display for FileKind {
             FileKind::ElfObject => "ELF object",
             FileKind::ElfDynamic => "ELF dynamic",
             FileKind::MachOObject => "MachO object",
+            FileKind::WasmObject => "wasm object",
```
marxin: Similarly here.

marxin (changes requested):
Great you've starting working on the port - looking forward to the results.

libwild/src/args.rs
lapla-cogito: I'm not sure what the advantage is of not guarding this with a feature flag. Since we're planning to release a new version soon, incomplete features should probably be behind feature flags.

libwild/src/file_kind.rs
lapla-cogito: Thanks, done.

libwild/src/file_kind.rs
lapla-cogito: Done.

libwild/src/args/wasm.rs
mati865: Does that even matter though? Callers will override it with explicit value anyway.

libwild/src/args.rs
davidlattimore: Another option would be to put `if !cfg!(feature = "wasm") { bail!("Wasm support is still experimental") }` somewhere in the critical path. That way the wasm code is always compiled and is more easily navigated in the IDE, but users still need to opt-in in order to use it.

Cargo.toml
```diff
@@ -101,6 +98,7 @@ tracing-subscriber = { version = "0.3.16", default-features = false, features =
 uuid = { version = "1.0.0", features = ["v4"] }
 wait-timeout = "0.2.0"
 wait4 = "0.1.3"
+wasmparser = "0.248.0"
```
davidlattimore, line 104: Any reason not to use the object crate for parsing wasm files?

libwild/src/args/wasm.rs
lapla-cogito: Yes, in actual practice I don't think this would be a problem. This comment is merely a casual note.

libwild/src/args.rs
lapla-cogito: > Another option would be to put `if !cfg!(feature = "wasm") { bail!("Wasm support is still experimental") }` somewhere in the critical path. That way the wasm code is always compiled and is more easily navigated in the IDE, but users still need to opt-in in order to use it.

That does sound better indeed. Thanks!

Cargo.toml
lapla-cogito: I previously conducted experiments to test how much of the required elements for WebAssembly linking could be read using `object::read::wasm`. While this internally uses `wasmparser`, I confirmed that it actually cannot read the `reloc.*` sections or certain subsections like `SegmentInfo` under `linking`. On the other hand, I verified that using `wasmparser` directly can successfully read these elements. Therefore, this PR currently uses `wasmparser` directly. For these missing features, I'm considering submitting patches to the `object` side, but for now I believe this is sufficient.

davidlattimore (approved):

libwild/src/args/wasm.rs
```diff
@@ -0,0 +1,164 @@
...
+use crate::ensure;
+use crate::error::Result;
+use crate::platform;
+use crate::save_dir::SaveDir;
+use jobserver::Client;
+use std::path::Path;
+use std::sync::Arc;
+
+/// Default page size (in bytes) for a wasm linear memory page.
+pub(crate) const WASM_PAGE_SIZE: u64 = 64 * 1024;
```
marxin: Maybe this const can be defined by using `WASM_PAGE_ALIGNMENT` constant?

libwild/src/wasm.rs
```diff
@@ -0,0 +1,1069 @@
...
+
+use crate::args::wasm::WasmArgs;
+use crate::ensure;
+use crate::platform;
+
+#[derive(Debug, Copy, Clone, Default)]
+pub(crate) struct Wasm;
+
+/// Magic bytes at the start of every Wasm module.
+pub(crate) const WASM_MAGIC: [u8; 4] = [0x00, b'a', b's', b'm'];
```
marxin, line 14: Theoretically this can be moved to `object` crate, right?

libwild/src/wasm.rs
```diff
@@ -0,0 +1,1069 @@
...
+    pub(crate) version: u32,
+}
+
+impl<'data> platform::ObjectFile<'data> for File<'data> {
+    type Platform = Wasm;
+
+    fn parse_bytes(input: &'data [u8], _is_dynamic: bool) -> crate::error::Result<Self> {
+        ensure!(input.len() >= 8, "Wasm module too short");
+        ensure!(input[..4] == WASM_MAGIC, "missing Wasm magic header");
+        let version = u32::from_le_bytes([input[4], input[5], input[6], input[7]]);
```
marxin: Can we use our `u32_from_slice` helper method?

marxin (approved):

libwild/src/args/wasm.rs
lapla-cogito: Thanks, updated.

libwild/src/wasm.rs
lapla-cogito: If the concerns about the object crate side mentioned [here](https://github.com/wild-linker/wild/pull/1912#discussion_r3214129133) can be addressed, using object crate with the `wasm` feature enabled should become feasible, so I believe this is possible.

libwild/src/wasm.rs
lapla-cogito: Done.

Cargo.toml
mati865: I think fixing `object` will be crucial for the maintenance of Wasm platform. If you are too busy for that, let us know how to reproduce those issues, so somebody can take over.

mati865 (approved):

Cargo.toml
philipc: Wasm parsing improvements in `object` are very welcome, but it mostly only exists there for the unified API. Maybe that will be useful here for the testing framework side of things, but I expect that you probably want to use `wasmparser` directly for the linker itself. `object`'s unified API is likely to be hiding Wasm-specific details that the linker needs.  `object` doesn't have a low-level API for Wasm like it does for ELF/Mach-O etc.

Issue #1903, opened 2026-05-08
linker-plugin LTO doesn't work with --wrap


marxin:
The package itself requires a couple of dependencies: `xorg-util-macros font-util meson xtrans` and reproduction steps are the following:

```
wget https://www.x.org/pub/individual/xserver/xorg-server-21.1.22.tar.xz
tar xvJf xorg-server-21.1.22.tar.xz
cd xorg*
mkdir build
cd build
CFLAGS="-B /home/marxin/Programming/wild" meson setup -Db_lto=true ..
ninja
...
[...]

PR #1917, opened 2026-05-11
refactor: Change internal representation of --sym-def symbols

davidlattimore:


davidlattimore:
This is aimed at making it possible to use this mechanism to help change how --wrap is implemented in order to fix #1903. I think the new structure of the code is slightly nicer anyway. In particular, we avoid `defsym_defined_via_cli_option`, which I didn't really like.

mati865 (approved):

PR #1918, opened 2026-05-11
port(wasm): Parse `linking` and `reloc.*` custom sections

lapla-cogito:
This PR adds a layer to parse Wasm object files using `wasmparser`. It produces the following:

- An ordered table of all top-level sections (`WasmSection`), preserving the original byte ranges and the indices used by `reloc.*` sections.
- The symbol table from the `linking` custom section, exposed as `WasmSymbol` with convenient accessors over `wasmparser::SymbolFlags`.
- Per-data-segment metadata from the `WASM_SEGMENT_INFO` subsection (`WasmSegmentInfo`).
- Relocation entries from `reloc.*` custom sections, grouped by target section (`WasmRelocSection`).
- The raw bytes of the `target_features` custom section, retained for future processing.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2,10 +2,21 @@
 #![allow(unused_variables)]
 #![allow(unused)]
 
+use crate::alignment::Alignment;
 use crate::args::wasm::WasmArgs;
 use crate::ensure;
+use crate::error::Context as _;
+use crate::error::Result;
 use crate::platform;
 use linker_utils::utils::u32_from_slice;
+use std::ops::Range;
+use wasmparser::KnownCustom;
+use wasmparser::Linking;
+use wasmparser::Parser;
+use wasmparser::Payload;
+use wasmparser::RelocationEntry;
+use wasmparser::SegmentFlags;
+use wasmparser::SymbolInfo;
 
 #[derive(Debug, Copy, Clone, Default)]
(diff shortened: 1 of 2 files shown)


libwild/src/wasm.rs
```diff
@@ -1068,3 +1181,121 @@ impl platform::Platform for Wasm {
...
+                                id,
+                                name,
+                                range,
+                                payload: payload_bytes,
+                            });
+                            continue;
+                        }
+                    };
+                    linking_version = Some(linking.version());
+                    parse_linking_subsections(linking, &mut symbols, &mut segments)?;
```
davidlattimore: If you moved these two lines up into the `KnownCustom::Linking` branch, then it looks like you eliminate the default branch, changing it to `_ => {}`, since it's otherwise a repeat of the push at the end of the loop.  Something like this:

```rust
if section_name == LINKING_SECTION_NAME {
    match reader.as_known() {
        KnownCustom::Linking(linking) => {
            linking_version = Some(linking.version());
            parse_linking_subsections(&linking, &mut symbols, &mut segments)?;
        }
        _ => {}
    };
}
```

As far as I can tell, that would be equivalent and simpler.

lapla-cogito: Thanks, done.

davidlattimore (approved):

PR #1919, opened 2026-05-11
port(MachO): emit code signature

marxin:
I’m happy to announce that we’ve reached the point where Wild can emit a fully runnable simple binary by providing the code signature.

Issue: #757

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1694,8 +1694,8 @@ impl platform::Platform for Elf {
         }
     }
 
-    fn default_layout_rules() -> &'static [SectionRule<'static>] {
-        DEFAULT_SECTION_RULES
+    fn default_layout_rules(_args: &Self::Args) -> Vec<SectionRule<'static>> {
+        DEFAULT_SECTION_RULES.to_vec()
     }
 
     fn linker_script_rules_pre_build(rule_builder: &mut crate::layout_rules::LayoutRulesBuilder) {
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1978,13 +1978,6 @@ impl platform::Platform for Elf {
     fn default_symtab_entry() -> Self::SymtabEntry {
         Default::default()
     }
-
-    fn extend_last_part_of_file(
-        _record: &OutputRecordLayout,
-        _last_part_id: part_id::PartId,
-    ) -> Result<usize> {
-        Ok(0)
-    }
 }
 
 /// Marks the symbol version associated with the dynamic symbol `GLIBC_ABI_DT_RELR` as needed.
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1978,6 +1978,13 @@ impl platform::Platform for Elf {
     fn default_symtab_entry() -> Self::SymtabEntry {
         Default::default()
     }
+
+    fn extend_last_part_of_file(
+        _file_size: usize,
+        _last_part_id: part_id::PartId,
+    ) -> Result<usize> {
+        Ok(0)
+    }
 }
 
 /// Marks the symbol version associated with the dynamic symbol `GLIBC_ABI_DT_RELR` as needed.
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1980,7 +1980,7 @@ impl platform::Platform for Elf {
     }
 
     fn extend_last_part_of_file(
-        _file_size: usize,
+        _record: &OutputRecordLayout,
         _last_part_id: part_id::PartId,
     ) -> Result<usize> {
         Ok(0)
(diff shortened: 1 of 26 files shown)


libwild/src/elf.rs
```diff
@@ -1978,6 +1978,13 @@ impl platform::Platform for Elf {
     fn default_symtab_entry() -> Self::SymtabEntry {
         Default::default()
     }
+
+    fn extend_last_part_of_file(
```
davidlattimore: If it's common for platforms to just return `Ok(0)`, then we could consider making that the default implementation.

libwild/src/platform.rs
```diff
@@ -534,6 +534,11 @@ pub(crate) trait Platform:
         args: &Self::Args,
     ) -> Result;
 
+    /// Gives the platform a chance to append data to the last part_id given
+    /// the layout knowledge of the last part.
+    fn extend_last_part_of_file(record: &OutputRecordLayout, last_part_id: PartId)
```
davidlattimore: I'm not sure about the name, since calling this function doesn't do the extension, it returns how much to extend by. Perhaps something like `last_part_size_to_extend`, `size_to_extend_last_part` or `last_part_extension_size`?

libwild/src/macho.rs
```diff
@@ -133,6 +136,137 @@ unsafe impl<E: Endian + 'static> object::Pod for DyldChainedFixupsHeader<E> {}
...
+    // number of index entries following
+    pub(crate) count: U32<BigEndian>,
+    // (count) entries
+    // CodeSignatureBlobIndex index[];
+    // followed by Blobs in no particular order as indicated by offsets in index
+}
+
+// Safety:
+// Is repr(C) and contains only `UX<E>` fields, and has no padding.
+unsafe impl object::Pod for CodeSignatureSuperBlob {}
```
davidlattimore: Would it work to use zerocopy's abstractions for these rather than object's? That's what we do in `elf.rs`, since it allows us to avoid the unsafe.

libwild/src/macho.rs
```diff
@@ -1405,6 +1541,18 @@ impl platform::Platform for MachO {
             n_value: Default::default(),
         }
     }
+
+    fn extend_last_part_of_file(
+        record: &OutputRecordLayout,
+        last_part_id: part_id::PartId,
+    ) -> Result<usize> {
+        ensure!(
```
davidlattimore, line 1536: Since the implementer of this method knows which PartId it wants to extend, an alternative API that comes to mind is:

```rust
fn last_part_extension(part_layouts: &OutputSectionPartMap<OutputRecordLayout>) -> Option<(PartId, usize)>;
```

But I don't feel strongly either way, so whichever you prefer.

libwild/src/macho.rs
```diff
@@ -1405,6 +1541,18 @@ impl platform::Platform for MachO {
...
+            last_part_id == part_id::CODE_SIGNATURE,
+            "code signature must be last part_id"
+        );
+        // The SHA hashes encompass the entire file excluding the CODE_SIGNATURE part ID.
+        Ok(record.file_offset.div_ceil(CS_BLOCK_SIZE) * CS_HASH_SIZE as usize)
```
davidlattimore, line 1543: I assume there's some reason why we can't say have the epilogue allocate `CS_HASH_SIZE` bytes in `part_id::CODE_SIGNATURE`, possibly with an alignment of `CS_BLOCK_SIZE`? It would be good to document that reason, probably here.

davidlattimore (commented):
Great!

libwild/src/macho.rs
marxin: Yes, the main reason is we need to allocate dynamically based on the file size (except the `part_id::CODE_SIGNATURE`). Will include a comment.

libwild/src/platform.rs
marxin: I like the suggest name.

libwild/src/macho.rs
marxin: Well, I intentionally wanted to narrow the interface to just the part that's at the very end of the layout. By doing that, we won't have to adjust and shift a part somewhere in the middle of the layout. Hope it makes sense.

libwild/src/macho.rs
```diff
@@ -1542,15 +1529,17 @@ impl platform::Platform for MachO {
...
     ) -> Result<usize> {
         ensure!(
             last_part_id == part_id::CODE_SIGNATURE,
             "code signature must be last part_id"
         );
-        // The SHA hashes encompass the entire file excluding the CODE_SIGNATURE part ID.
+        // The CODE_SIGNATURE size depends on the final file size, excluding the
+        // signature itself. Compute it after layout because there is one SHA hash
+        // per file block (4 KiB) covered by the signature.
         Ok(record.file_offset.div_ceil(CS_BLOCK_SIZE) * CS_HASH_SIZE as usize)
```
davidlattimore, line 1543: Thanks for the explanation. That makes sense now. My brain assumed that there was an `+` in this line, when it's actually a `*`.

davidlattimore (approved):

PR #1916, opened 2026-05-10
fix: Do not output SFrame section unless asked

mati865:
With this change Wild will discard `.sframe` sections unless `--wild-experimental-sframe` flag is explicitly added.

PR #1922, opened 2026-05-11
refactor: Change default_layout_rules to take Args and return a Vec

davidlattimore:


davidlattimore:
This is to help with #1916. But having the layout rules depend on arguments seems like a pretty reasonable thing to do, so I suspect we'd want to keep it even if we didn't have the use-case of conditionally discarding sframe sections.

mati865 (approved):

PR #1923, opened 2026-05-12
fix: More robust code to get line numbers from debug info

davidlattimore:
Also fixed 64 bit relocation sizes for riscv64 and loongarch64, which are used in debug info.


linker-utils/src/riscv64.rs
```diff
@@ -145,7 +145,7 @@ pub const fn relocation_type_from_raw(r_type: u32) -> Option<RelocationKindInfo>
         ),
         object::elf::R_RISCV_64 => (
             RelocationKind::Absolute,
-            RelocationSize::ByteSize(4),
+            RelocationSize::ByteSize(8),
```
marxin, line 148: Whoops! Good finding.

lapla-cogito (approved):

PR #1925, opened 2026-05-12
feat: evaluate expressions within PROVIDE with `evaluate_expression`

vishruth-thimmaiah:
This PR adds support for the `PROVIDE`/`PROVIDE_HIDDEN` commands to parse and evaluate expressions using the `evaluate_expression` function. This PR also lets us evaluate complex expressions within `--defsym` arguements.

Currently, `SEGMENT_START` isn't yet handled by `evaluate_expression`.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -15,6 +15,7 @@ use crate::linker_script::MemoryRegion;
 use crate::output_section_id::OutputSections;
 use crate::output_section_id::SectionName;
 use crate::output_section_map::OutputSectionMap;
+use crate::parsing::SymbolLoc;
 use crate::platform::Platform;
 
 /// Compute 1-based line number by counting newlines before `remainder` in `file_bytes`.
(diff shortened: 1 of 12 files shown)


libwild/src/expression_eval.rs
```diff
@@ -67,10 +68,24 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
...
                     output_sections,
-                    warning_callback,
                     &parsed.memory_regions,
+                    &|name| {
+                        // Symbol resolution in ASSERT is not yet implemented. Rather than failing the
+                        // link (which would be our fault, not the user's), emit a warning and skip the
+                        // assertion by returning 1 (true).
+                        warning_callback(Warning::new(format!(
+                            "ASSERT: symbol references not yet supported ('{}'), skipping assertion",
+                            str::from_utf8(name).unwrap_or_else(|_| {
```
davidlattimore: For messages, using `String::from_utf8_lossy` is reasonable and what we use elsewhere

libwild/src/parsing.rs
```diff
@@ -282,16 +220,13 @@ impl<'data, P: Platform> Prelude<'data, P> {
...
-                DefsymValue::SymbolWithOffset(target, offset) => {
-                    SymbolPlacement::Redirect(Redirect {
-                        kind: RedirectKind::DefSym,
-                        target_name: target.as_bytes(),
-                        offset: *offset,
-                    })
-                }
-            };
+            let mut value = winnow::BStr::new(value);
+            let expr = crate::linker_script::parse_expression(&mut value).unwrap();
```
davidlattimore: It's fine to panic when a condition isn't met that indicates a bug in our code, but ideally, we'd like to not panic on invalid user input. Can you propagate the error up instead of using unwrap here?

davidlattimore (commented):
Sorry for the merge conflicts arising from my PR yesterday.

libwild/src/parsing.rs
vishruth-thimmaiah: Done.

davidlattimore (approved):

PR #1928, opened 2026-05-13
style: Change C to 100 columns

davidlattimore:


davidlattimore:
It has been bothering me for a while that our C code has a column width of 80, while all our Rust code has a column width of 100.

The first commit changes the clang-format config then reruns the formatter. In the second commit, I manually went to files with C-style comments and reflowed the comments. The automatic formatter would have been happy to leave the comments wrapped at 80, but I wasn't.

mati865 (approved):

PR #1930, opened 2026-05-13
chore: format TOML files and use taplo for checking

marxin:
Taplo is a library that powers a very nice VSCode extension called [Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml). I see it pretty handy having a unified format also for the TOML files in the project.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
         if: ${{ contains(matrix.container, 'opensuse') }}
       - run: apk add build-base lld clang clang-extra-tools bash curl tar
         if: ${{ contains(matrix.container, 'alpine') }}
+      - run: curl -fsSL https://github.com/tamasfe/taplo/releases/latest/download/taplo-linux-x86_64.gz | gzip -d - | install -m 755 /dev/stdin /usr/local/bin/taplo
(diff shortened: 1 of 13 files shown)


mati865:
I thought we had TOML formatter already.

Any insights about Tombi vs Taplo?

I see that Zed editor lists Tombi as their LSP for TOML and when testing about a year ago in Helix, Tombi was more helpful as the LSP.

marxin:
I am fine with any of the options as long as we've got one.

davidlattimore:
Auto-formatting toml sounds good to me.

I haven't used any of the options, so don't have any thoughts as to which we should prefer. In terms of people discovering that they need to reformat before they create a PR, I do wonder if we should have a tidy test like we do for C formatting. From a brief look, it appears that taplo at least has a library, so possibly the tidy test could even depend on that and not require a binary to be installed.

> I thought we had TOML formatter already.

Perhaps you were thinking of the yaml format check?

mati865:
I guess the fact that we already have `wild/tests/external_tests/taplo.toml` just threw me off.

marxin:
I've just included it as a new tidy test - using the binary, very similarly to `clang-format`.

davidlattimore (approved):

PR #1931, opened 2026-05-14
feat: implement symbol resolution within the ASSERT command

vishruth-thimmaiah:
Adds support for evaluating ASSERT statements that reference symbols.


davidlattimore (commented):
Nice! It'd be good to add one ExpectError test as well

davidlattimore (approved):

PR #1936, opened 2026-05-14
port(wasm): Implement object file section accessors

lapla-cogito:
Fill in `section_iter` / `section` / `section_by_name` / `section_name` / `raw_section_data` and friends so that the layout pipeline can enumerate and read the input file's sections.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -214,7 +214,7 @@ impl<'data> platform::ObjectFile<'data> for File<'data> {
         0
     }
 
-    fn section_iter(&self) -> <Self::Platform as platform::Platform>::SectionIterator<'data> {
+    fn section_iter<'a>(&'a self) -> <Self::Platform as platform::Platform>::SectionIterator<'a> {
         [].iter()
     }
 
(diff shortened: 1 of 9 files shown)


lapla-cogito:
The first commit is some refactoring. It decouples `section_iter` / `enumerate_sections` / `section` / `section_by_name` from `'data` so that platforms whose section metadata is not zero-copy parseable from the input bytes can satisfy the trait without `unsafe`.

libwild/src/wasm.rs
```diff
@@ -64,26 +64,46 @@ pub(crate) struct File<'data> {
...
     pub(crate) fn is_custom(&self) -> bool {
         self.id == 0
     }
+
+    pub(crate) fn payload_range_usize(&self) -> Range<usize> {
+        self.payload_range.start as usize..self.payload_range.end as usize
+    }
+}
+
+fn standard_section_name(id: u8) -> Option<&'static [u8]> {
```
marxin, line 91: I would rather use a static array that will be indexed `SECTION_NAMES.get(id)`

lapla-cogito: I haven't identified any advantages to that approach over the current implementation 🤔 . The main reasons I believe the current implementation is more appropriate are as follows:
- The number-to-section name mapping in the Wasm spec is more straightforward
- Custom sections (i.e., sections with ID 0) don’t have canonical names, requiring ad-hoc workarounds like making `b""` the first element when using a static array (or creating `[Option<&[u8]>]` and putting `None` in the first element)

marxin: Great, sure, understood :)

davidlattimore (approved):

PR #1943, opened 2026-05-15
feat: Enable linker plugin support by default

davidlattimore:


davidlattimore:
Also contains a 1 line line change I've moved out to #1962.

PR #1950, opened 2026-05-19
test: Validate that loadable segments don't overlap

davidlattimore:


davidlattimore:
I also changed it so that validation and assertions run before linker-diff, since I think that in general is less confusing.

wild/tests/integration_tests.rs
```diff
@@ -3800,6 +3801,50 @@ fn verify_no_overlapping_sections(obj: &object::File) -> Result {
...
+                return None;
+            }
+            Some((i, start, start + size))
+        })
+        .collect();
+
+    segments.sort_by_key(|seg| seg.1);
+
+    for i in 0..segments.len() {
+        for j in (i + 1)..segments.len() {
```
mati865: I'm not sure if checking overlap with all further segments is necessary. If a segment overlaps two segments, then it would also overlap just a one segment.

Not that it matters much given how few segments there are.

mati865 (approved):

wild/tests/integration_tests.rs
```diff
@@ -3800,6 +3801,50 @@ fn verify_no_overlapping_sections(obj: &object::File) -> Result {
...
+        .filter_map(|(i, seg)| {
+            let start = seg.p_vaddr(elf_obj.endian());
+            let size = seg.p_memsz(elf_obj.endian());
+            let p_type = seg.p_type(elf_obj.endian());
+            if p_type != object::elf::PT_LOAD {
+                return None;
+            }
+            Some((i, start, start + size))
+        })
+        .collect();
```
marxin: could be replaced with `sorted_by_key(...).collect_vec()` from Itertools.

wild/tests/integration_tests.rs
marxin: I have the same feeling, maybe simply using - `for ((idx1, start1, end1), (idx2, start2, end2)) in  segments.enumerate().tuple_windows()`?

davidlattimore: Nice. I hadn't seen tuple_windows before. Done

wild/tests/integration_tests.rs
davidlattimore: Good idea. I didn't need the collect_vec, since we can feed the iterator directly into the `tuple_windows` call

PR #1951, opened 2026-05-19
tools: Use WILD_LOG environment variable rather than RUST_LOG

davidlattimore:
Using RUST_LOG didn't work so well with cargo run because any change to RUST_LOG causes a rebuild.


mati865 (approved):
I have considered that but for other reason: when all programs use the same variable enabling logging for just one of them is annoying.

PR #1954, opened 2026-05-19
test: Don't register plugin tests when plugins feature is disabled

davidlattimore:


davidlattimore:
This mostly matters for cargo nextest, since cargo test shows that the test was ignored, whereas cargo nextest was previously showing it as passing. For example if you run `cargo nextest run elf/x86_64/trivial-lto/default`, you'll now get an error because no tests will match, whereas before it showed 1 test passing.

Issue #1948, opened 2026-05-17
LTO: cannot build `mariadb-connector-c` package[/-] [+]


marxin:
The package build fails with the following error message:
```
wget https://dlm.mariadb.com/4441391/Connectors/c/connector-c-3.4.7/mariadb-connector-c-3.4.7-src.tar.gz
mkdir mariadb
cd mariadb
tar xvzf ../mariadb-connector-c-3.4.7-src.tar.gz
mkdir build
cd build
cmake .. -DENABLE_BOOST=OFF -D CMAKE_C_FLAGS="-flto=auto -B/home/marxin/Programming/wild -Wno-error=discarded-qualifiers"
make
...
[ 72%] Linking C shared library libmariadb.so
wild: error: Failed copying from /tmp/ccA4H9nl.ltrans0.ltrans.o (9728 (38/0)) to output file
  Caused by:
    Failed to apply relocations in section `.text` of /tmp/ccA4H9nl.ltrans0.ltrans.o
    Failed to apply relocation of type R_X86_64_PLT32 to symbol `mysql_errno` (16625 local=369) in file #9728 (38/0) (/tmp/ccA4H9nl.ltrans0.ltrans.o) defined as 15637 in file #9494 (37/22) (LTO input `CMakeFiles/mariadb_obj.dir/mariadb_stmt.c.o`) (ABSOLUTE) at offset 0x6fd8
    Missing resolution for: symbol `mysql_errno` (16625 local=369) in file #9728 (38/0) (/tmp/ccA4H9nl.ltrans0.ltrans.o) defined as 15637 in file #9494 (37/22) (LTO input `CMakeFiles/mariadb_obj.dir/mariadb_stmt.c.o`) (ABSOLUTE)
```

davidlattimore:
The cause of the link failure was that during resolution, in `resolve_group`, we weren't consuming from `symbol_definitions_slice` for linker script symbols or `SyntheticSymbols`. This didn't previously matter because these only contained definitions, not references and there were no references after them. However, linker scripts changed that because we started having references to symbols after linker scripts. By not consuming an appropriate amount from `symbol_definitions_slice`, subsequent objects (linker plugin related objects) wrote their resolutions into the wrong places in the symbol definitions vec.

[...]

PR #1956, opened 2026-05-20
fix: Prevent GC of linker script symbol aliases

davidlattimore:
Related to #1948. Note, that dealing with archive activation is left to a subsequent PR.

The newly added test has two configs. Config defsym-obj was already passing. Config script-obj was failing and is what's being fixed by this PR.


wild/tests/sources/elf/versioned-script-symbol/versioned-script-symbol.c
```diff
@@ -1,7 +1,13 @@
...
 //#RunEnabled:false
+// GNU ld discards foo. Wild, like lld doesn't.
+//#EnableLinker:lld
+//#SkipLinker:ld
 //#LinkArgs:--shared -znow ./versioned-script-symbol.map
 //#ExpectSym:mysql_affected_rows@libmysqlclient_18
 //#DiffIgnore:section.got
+//#DiffIgnore:section.gnu.version_d.alignment
+//#DiffIgnore:version_d.verdef_1
+//#DiffIgnore:dynsym.mysql_affected_rows.section
```
mati865, line 12: Neither LD nor LLD export this symbol. Is it an issue with version script parser or the linker?

mati865 (approved):

wild/tests/sources/elf/versioned-script-symbol/versioned-script-symbol.c
davidlattimore: Not sure where the bug is, but I've added a TODO.

PR #1964, opened 2026-05-22
port(wasm): Support relocation encoding

lapla-cogito:
Also wired `write_output_file()` through `wasm_writer`.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -479,12 +479,18 @@ impl<'data> platform::ObjectFile<'data> for File<'data> {
 
     fn relocations(
         &self,
-        _index: object::SectionIndex,
+        index: object::SectionIndex,
         _relocations: &<Self::Platform as platform::Platform>::RelocationSections,
     ) -> crate::error::Result<<Self::Platform as platform::Platform>::RelocationList<'data>> {
-        // Relocation entries are parsed and stored on `File::reloc_sections` but not yet plumbed
-        // through the layout pipeline.
+        let target = u32::try_from(index.0).unwrap_or(u32::MAX);
+        let entries = self
+            .reloc_sections
+            .iter()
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -49,9 +49,26 @@ pub(crate) mod section_id {
 /// Size of a `[Option<u32>; _]` lookup that can be indexed by any standard section id.
 pub(crate) const STANDARD_SECTION_LOOKUP_LEN: usize = section_id::MAX as usize + 1;
 
+pub(crate) mod reloc_type {
```
lapla-cogito, line 52: The relocation types listed here represent only a subset of those documented in [the tool convention](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#relocation-sections). This is primarily because I've excluded types that appear in 64-bit Wasm or for features we haven't planned to support initially, such as PIC.

davidlattimore (approved):

libwild/src/wasm.rs
```diff
@@ -233,10 +250,94 @@ impl WasmRelocation {
...
+            | reloc_type::GLOBAL_INDEX_I32
+            | reloc_type::FUNCTION_INDEX_I32 => 4,
+            _ => 0,
+        }
+    }
+}
+
+/// Write `value` as a 5-byte fixed-width unsigned LEB128. Used for wasm reloc slots that reserve
+/// exactly 5 bytes regardless of the encoded value.
+pub(crate) fn write_uleb128_5(buf: &mut [u8; 5], value: u32) {
```
marxin, line 281: Might be something for the future (will need a testing), but can't we fill up the buffer with:
```
buf.fill(0);
let mut writer = buf.as_mut_slice();
leb128::write::unsigned(&mut writer, u64::from(value)).unwrap();
```

?

PR #1970, opened 2026-05-23
test: update Mold external tests

marxin:
There are a couple of new tests that fail with Wild - mostly related to the LTO.
@davidlattimore Do you want me to create an issue for them?

### DIFF
--- a/.github/workflows/external_tests.yml
+++ b/.github/workflows/external_tests.yml
@@ -20,10 +20,10 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     container:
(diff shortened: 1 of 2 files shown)


lapla-cogito (approved):

davidlattimore:
I'd say file bugs once we've identified what it is that we're doing wrong and have decided it's something we want fixed.

davidlattimore (approved):

marxin:
We might want to improve the story about LTO FAT objects (have both assembly and IR), but it's pretty legacy compatibility option I would say.

PR #1973, opened 2026-05-23
port(MachO): parse more options and identify stub libs

marxin:
The change adds parsing support for all options required when linking with `clang++ main.cc` and treats the stub library as a distinct input type.

Issue: #757


davidlattimore (approved):
I'll be interested to see how the plugin API here compares to the GNU one

marxin:
Well, the Mach-O C/C++ toolchain is LLVM-based, so it's going to be hopefully identical to what you implemented for LLVM on Linux.

PR #1976, opened 2026-05-25
feat: recognise ppc64 (PowerPC64, ELFv2) ELF objects

runlevel5:
Initial scaffold for PowerPC64 little-endian (ELFv2). It wires the target in but doesn't process relocations yet, so it can't link — wild just recognises ppc64 inputs and fails cleanly.

NOTES: IBM does offer free ppc64le GitHub Runner, please check https://github.com/IBM/actionspz which would come in handy for CI build/release process. Alternatively cross-compilation could be an option (though slower than native silicon).

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ The following platforms / architectures are currently supported:
 * ARM64 on Linux
 * RISC-V (riscv64gc) on Linux
 * LoongArch64 on Linux (initial support)
+* PPC64LE on Linux (initial support)
 
 The following is working with the caveat that there may be bugs:
 
--- a/libwild/src/arch.rs
+++ b/libwild/src/arch.rs
@@ -2,6 +2,7 @@ use crate::bail;
 use crate::error::Result;
 use object::elf::EM_AARCH64;
 use object::elf::EM_LOONGARCH;
+use object::elf::EM_PPC64;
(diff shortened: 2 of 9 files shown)


davidlattimore:
Cross compilation works pretty well speed-wise for our other architectures. It's only the binary created by the test that gets run under qemu and those binaries only contain a small number of instructions. I guess the other criteria would be whether a suitable cross compiler is available.

I'm curious. What's your motivation for wanting to port to ppc64le? Is it a platform you use or just a platform that you're interested in?

davidlattimore (approved):

runlevel5:
> Cross compilation works pretty well speed-wise for our other architectures. It's only the binary created by the test that gets run under qemu and those binaries only contain a small number of instructions. I guess the other criteria would be whether a suitable cross compiler is available.

I highly recommend you go with the native runner instead because it would save much efforts / time. 

> I'm curious. What's your motivation for wanting to port to ppc64le? Is it a platform you use or just a platform that you're interested in?

Personally I am interested in the project out of frustration of waiting for so long for the linking process for many projects that I am working on (for example Firefox JIT port). I am too working on contract for enterprise clients that are still running POWER9 and now POWER10 system though do not believe those enterprise projects would ever use wild (or even mold)

PR #1978, opened 2026-05-25
feat: Add support for the DISCARD section in linker scripts

### LINKED ISSUE #1829: Add linker-script support for /DISCARD/ pseudo-section

vishruth-thimmaiah:
Support discarding sections that are marked as discardable with the `/DISCARD/` command in linker scripts.

Closes: https://github.com/wild-linker/wild/issues/1829

### DIFF
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -181,6 +181,26 @@ impl<'data> LayoutRulesBuilder<'data> {
                 for sec_cmd in &sections.commands {
                     match sec_cmd {
                         SectionCommand::Section(sec) => {
+                            if sec.output_section_name == b"/DISCARD/" {
+                                for contents_cmd in &sec.commands {
+                                    match contents_cmd {
+                                        ContentsCommand::Matcher(matcher) => {
+                                            for pattern in &matcher.input_section_name_patterns {
+                                                self.add_section_rule(SectionRule::new(
+                                                    pattern,
+                                                    matcher.input_file_pattern,
+                                                    crate::layout_rules::SectionRuleOutcome::Discard,
+                                                )?);
+                                            }
+                                        }
+                                        _ => {
+                                            return Err(crate::error!(
+                                                "Illegal use of /DISCARD/ section"
+                                            ));
+                                        }
+                                    }
+                                }
+                            }
                             let min_alignment = sec
                                 .alignment
                                 .unwrap_or(alignment::MIN)
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -193,11 +193,7 @@ impl<'data> LayoutRulesBuilder<'data> {
                                                 )?);
                                             }
                                         }
-                                        _ => {
-                                            return Err(crate::error!(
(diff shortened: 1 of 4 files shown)


libwild/src/layout_rules.rs
```diff
@@ -192,9 +176,31 @@ impl<'data> LayoutRulesBuilder<'data> {
...
+                                            for pattern in &matcher.input_section_name_patterns {
+                                                self.add_section_rule(SectionRule::new(
+                                                    pattern,
+                                                    matcher.input_file_pattern,
+                                                    crate::layout_rules::SectionRuleOutcome::Discard,
+                                                )?);
+                                            }
+                                        }
+                                        _ => {
+                                            return Err(crate::error!(
```
davidlattimore: Simpler to just use the `bail!` macro

wild/tests/sources/elf/linker-script-discard/linker-script-discard.c
```diff
@@ -0,0 +1,14 @@
...
+//#LinkArgs:-shared -z now -T ./linker-script-discard.ld
+//#DiffIgnore:section.got
+//#DiffIgnore:section.riscv.attributes
+//#DiffIgnore:segment.RISCV_ATTRIBUTES.*
+// GNU ld emits `.riscv.attributes`, but Wild does not
+//#DiffIgnore:riscv_attributes.*
+//#DiffIgnore:segment.LOAD.RX.alignment
+//#DiffIgnore:segment.LOAD.RWX.alignment
+//#DoesNotContain:DISCARD
+//#DoesNotContain:.text
```
davidlattimore, line 17: I wonder if we want to test whether we emit the symbol `foo` or not. From a quick experiment, it looks like wild and lld don't emit `foo`, while GNU ld weirdly does. So if we went that route ("NoSym:foo"), then we'd likely need to disable GNU ld and enable lld for the test.

vishruth-thimmaiah: Making `foo` static stops ld from emitting it in `.dynsym`, so we should be able to keep ld as well.

davidlattimore: A static that isn't referenced could be eliminated by the compiler, so might make the test pass even without `/DISCARD/` support

vishruth-thimmaiah: I see, I've updated the test to only use lld.

davidlattimore (approved):

PR #1979, opened 2026-05-25
fix: Keep `--defsym`/linker-script redirect targets alive through LTO

### LINKED ISSUE #1960: Symbols referenced by aliases can get eliminated by LTO
See the LTO variant of the test `linker-script-symbols`. Although note that that test only tests aliases from linker scripts. We should also test aliases from --defsym.

lapla-cogito:
When `--defsym=alias=target` (or an equivalent linker-script symbol assignment) referenced a symbol defined in an LTO input, the linker plugin reported the target as `PrevailingDefIronly` because nothing inside the IR referenced it. GCC's LTO was then free to DCE or internalize the symbol.

Mark redirect targets with `HAS_NON_IR_REF` during symbol resolution so that the plugin sees them as `PrevailingDef` and the LTO compiler keeps them in the post-LTO output.

Fix #1960

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -4413,7 +4413,8 @@ fn get_defsym_attributes(
                 ));
 
         if let Some(target_id) = target_symbol_id {
-            get_symbol_attributes(layout, target_id)
+            let canonical_id = layout.symbol_db.definition(target_id);
+            get_symbol_attributes(layout, canonical_id)
         } else {
             Err(redirect.missing_target(target_name))
         }
--- a/libwild/src/parsing.rs
+++ b/libwild/src/parsing.rs
@@ -320,7 +320,7 @@ impl Redirect<'_> {
 
     pub(crate) fn missing_resolution(&self, target_name: &[u8]) -> crate::error::Error {
         crate::error!(
-            "Symbol '{name}' referenced by {kind} has no resolution. Might be issue #1960",
+            "Symbol '{name}' referenced by {kind} has no resolution.",
             name = String::from_utf8_lossy(target_name),
             kind = self.kind.message_text(),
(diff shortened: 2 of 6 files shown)


libwild/src/elf_writer.rs
```diff
@@ -4413,7 +4413,8 @@ fn get_defsym_attributes(
                 ));
 
         if let Some(target_id) = target_symbol_id {
-            get_symbol_attributes(layout, target_id)
+            let canonical_id = layout.symbol_db.definition(target_id);
+            get_symbol_attributes(layout, canonical_id)
```
lapla-cogito, line 4417: This is a bonus fix. The `symbol_alias_lto` test would cause the following error without this:

```
---- elf/x86_64/defsym/symbol_alias_lto ----
Output binary assertions failed. Binary `/home/lapla/repos/wild/wild/tests/build/elf/x86_64/defsym/symbol_alias_lto/defsym.c.wild`. Relink with:
WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/home/lapla/repos/wild/wild/tests/build/elf/x86_64/defsym/symbol_alias_lto/defsym.c.wild /home/lapla/repos/wild/wild/tests/build/elf/x86_64/defsym/symbol_alias_lto/defsym.c.save/run-with cargo run --bin wild --
  Caused by:
    Expected symbol `bar` to be in section `.text`, but it was Absolute
```

like https://github.com/wild-linker/wild/pull/1402/changes/2fb5e252a1de5f731a7aa98e270297dcd1a4baeb

davidlattimore (approved):
Thanks for the fix!

lapla-cogito:
Just removed the `Might be issue #1960` note from the error message.

PR #1980, opened 2026-05-25
chore: Ignore formatting of toml files under target dir

davidlattimore:


davidlattimore:
`cargo package` and `cargo publish` write toml files to the target directory and they use different formatting

mati865 (approved):

PR #1983, opened 2026-05-25
test(ppc64): add ppc64le test infrastructure

runlevel5:
Wires ppc64le into the test framework so we can exercise the ppc64
  backend — no linker capability here, the relocation work lands
  separately.

  - register ppc64le in the integration test harness (target triple,
    `-m elf64lppc`, the `ld64.so.2` dynamic linker, native-host detection)
  - a ppc64 linker-diff module (objdump-based decoding via
    `-m powerpc:common64`, forced little-endian)
  - let libwild build natively on ppc64le (powerpc64 was missing from the
    `perf_unsupported` cfg list)
  - `exit_syscall` + `memcpy` for the test runtime on ppc64le. ppc64le gcc
    lowers large aggregate initialisers to a memcpy call even at `-O0`, where
    the other targets inline it, so the test programs don't link without one.

### DIFF
--- a/linker-diff/src/arch.rs
+++ b/linker-diff/src/arch.rs
@@ -19,6 +19,7 @@ pub(crate) enum ArchKind {
     Aarch64,
     RISCV64,
     LoongArch64,
+    Ppc64,
 }
 
 /// Provides architecture-specific functionality needed by linker-diff.
@@ -281,6 +282,7 @@ impl ArchKind {
             object::elf::EM_AARCH64 => Ok(ArchKind::Aarch64),
             object::elf::EM_RISCV => Ok(ArchKind::RISCV64),
             object::elf::EM_LOONGARCH => Ok(ArchKind::LoongArch64),
+            object::elf::EM_PPC64 => Ok(ArchKind::Ppc64),
             other => bail!("Unsupported object architecture {other}",),
         }
     }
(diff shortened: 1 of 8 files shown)


runlevel5:
I need to look into failed CI build first

linker-diff/src/arch.rs
```diff
@@ -19,6 +19,7 @@ pub(crate) enum ArchKind {
     Aarch64,
     RISCV64,
     LoongArch64,
+    Ppc64,
```
marxin, line 22: I think the proper acronym should be `PPC64`. I’m also wondering whether we should include `LE`, since we technically only support a subset of the ABI.

davidlattimore: Most of the Wild codebase, when using CamelCase, only has the first letter of each acronym uppercased, which would make Ppc64 correct. I think it's `RISCV64` that should be renamed as `RiscV64` (but not in this PR)

davidlattimore:
Most likely you need to adjust the CI setup so that `gcc-powerpc64le-linux-gnu` gets installed

linker-diff/src/arch.rs
marxin: Gotcha, let's make it happen in #1988.

runlevel5:
I am going to close this PR and break it into smaller PRs

PR #1984, opened 2026-05-25
port(Macho): plug stub symbols into symbol DB

marxin:
The change correctly wires in symbols loaded from the stub libraries. Given the issues around `Cow::Owned` and the `&'data` lifetime requirements in the symbols DB, I chose not to support escape sequences in `.tbd` files. Based on my testing, all system stub libraries can be parsed successfully with our implementation.

Issue: #757

### DIFF
--- a/libwild/src/diagnostics.rs
+++ b/libwild/src/diagnostics.rs
@@ -56,6 +56,7 @@ impl SymbolInfoPrinter {
                     ResolvedFile::Prelude(_) => Some(PRELUDE_FILE_ID),
                     ResolvedFile::Object(obj) => Some(obj.common.file_id),
                     ResolvedFile::Dynamic(obj) => Some(obj.common.file_id),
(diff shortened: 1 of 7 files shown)


davidlattimore (approved):
If you want to handle escaped names at some point, there are two options that come to mind. The first would be to store the symbol names in their escaped form. e.g.

```rust
struct StubSymbolName<'data> {
   /// Will be shorter than `bytes.len()` if escape sequences are present.
   len: usize,
   bytes: &'data [u8],
}
```

Although I guess that gets tricky when we want to build a hashmap keyed by symbol name, so maybe that's not a good idea.

The other option would be to store the unescaped names in the herd arena.

marxin:
> The other option would be to store the unescaped names in the herd arena.

Yes - that was the original implementation I had, but didn't like it much. But yes, we can add it later if needed, however I believe the escape sequences for the TBD are super-rare.

PR #1987, opened 2026-05-26
port(wasm): Add section emitters for type / import / function / global / export

lapla-cogito:
Part of #1431

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,6 +108,7 @@ uuid = { version = "1.3.0", features = ["v4"] }
 wait-timeout = "0.2.0"
 wait4 = "0.1.3"
 wasmparser = "0.251.0"
+wasm-encoder = "0.251.0"
 which = "8.0.0"
 winnow = { version = "1.0.0", features = ["simd"] }
 zerocopy = { version = "0.8.27", features = ["derive"] }
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -12,6 +12,15 @@ use crate::output_section_id::SectionName;
 use crate::platform;
 use linker_utils::utils::u32_from_slice;
 use std::ops::Range;
+use wasmparser::BinaryReader;
+use wasmparser::ConstExpr;
+use wasmparser::DataKind;
+use wasmparser::DataSectionReader;
+use wasmparser::ExportSectionReader;
+use wasmparser::FunctionSectionReader;
+use wasmparser::GlobalSectionReader;
+use wasmparser::GlobalType;
+use wasmparser::ImportSectionReader;
 use wasmparser::KnownCustom;
 use wasmparser::Linking;
 use wasmparser::Parser;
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -106,6 +106,7 @@ tracing-subscriber = { version = "0.3.16", default-features = false, features =
 uuid = { version = "1.3.0", features = ["v4"] }
 wait-timeout = "0.2.0"
 wait4 = "0.1.3"
+wasm-encoder = "0.248.0"
```
davidlattimore: I still have reservations about using wasm-encoder to do writing. Mostly because I don't see how we'll be able to make writing multi-threaded. If you're sure you want to go that route, it's fine, but I expect we won't be able to reach the same level of performance as we have for ELF.

davidlattimore (approved):

libwild/src/wasm.rs
```diff
@@ -336,10 +347,231 @@ pub(crate) fn apply_relocation(
...
+                    type_index,
+                });
+            }
+        }
+
+        Ok(out)
+    }
+
+    /// Imported globals in declaration order. Imports of other kinds are skipped.
+    pub(crate) fn global_imports(&self) -> Result<Vec<WasmGlobalImport<'data>>> {
```
davidlattimore, line 461: It seems that there are lots of functions in this PR that aren't called

lapla-cogito: This is intentional (the same pattern as `wasm.rs`, etc.). While much of the layout pipeline remains in TODO state, I'll use these functions after completing that.

Cargo.toml
lapla-cogito: I think that, with the current wasm-encoder API, what is actually feasible is to generate the metadata on a single thread, generate raw instruction streams for the other sections in parallel across multiple threads, and then merge everything back on a single thread at the end to reconcile the indices. I'm not sure how much performance this would sacrifice, though.

davidlattimore: Is generating instruction streams computationally expensive? At least for ELF, most of the work is pretty much using memcpy to copy data around then applying a few relocations. With ELF, the destination of those copies is the actual output file and multiple threads are copying at once, with the destinations being non-overlapping areas of memory. If Wasm has to do a bunch of expensive computation to produce the stream, then that computation might dominate and it might not matter too much. But if like ELF, it's mostly copying data and applying a few relocations, then the part that merges everything back together is likely to be pretty slow, since it will need to perform an additional copy of pretty much the entire output file from a single thread.

libwild/src/wasm.rs
davidlattimore: OK. How close are we to getting something that can run at least part of the way through the linking process and removing the `#[allow(unused)]`?

Cargo.toml
lapla-cogito: Generating streams shouldn't be particularly expensive. 

Because Wasm references are purely index-based and all final indices are already known after layout, we can pre-compute sizes and split the final output buffer into non-overlapping regions per group (using the same `split_output_by_group` + `split_buffers_by_alignment` pattern as ELF). Parallel workers then emit directly into their assigned slices of the real output.

libwild/src/wasm.rs
lapla-cogito: As discussed during our recent meeting, these will be used after we've resolved most `todo!()` in the layout phase.

PR #1990, opened 2026-05-27
feat: Support GDB index

### LINKED ISSUE #811: Add support for --gdb-index

lapla-cogito:
This patch adds support for `--gdb-index` and `--no-gdb-index`. Although we support version 9 here, the test environment may produce a version 7 GDB index when using lld. This patch also adds a test that checks whether specific symbols are included in the symbol table of the `.gdb_index` section. Because of this requirement, the integration test includes logic to support versions earlier than version 9 as well.

Closes #811

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -2078,7 +2078,7 @@ impl platform::Args for ElfArgs {
     }
 
     fn should_write_gdb_index(&self) -> bool {
-        self.gdb_index
+        self.gdb_index && !self.should_strip_debug()
     }
 }
 
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -127,6 +127,7 @@ pub struct ElfArgs {
 
     pub(crate) nmagic: bool,
     pub(crate) rosegment: bool,
+    pub(crate) gdb_index: bool,
 
     rpath_set: IndexSet<String>,
 
@@ -233,7 +234,6 @@ const SILENTLY_IGNORED_SHORT_FLAGS: &[&str] = &[
 ];
 
 const IGNORED_FLAGS: &[&str] = &[
-    "gdb-index",
     "fix-cortex-a53-835769",
     "fix-cortex-a53-843419",
     "discard-all",
@@ -332,6 +332,7 @@ impl Default for ElfArgs {
 
             experimental_sframe: false,
             debug_compression_kind: None,
+            gdb_index: false,
         }
     }
 }
@@ -1114,6 +1115,24 @@ fn setup_argument_parser() -> ArgumentParser<ElfArgs> {
                 "zstd" => args.debug_compression_kind = Some(CompressionKind::Zstd),
                 value => bail!("--compress-debug-sections={value}"),
             }
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("gdb-index")
+        .help("Generate GDB index")
+        .execute(|args, _modifier_stack| {
+            args.gdb_index = true;
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("no-gdb-index")
+        .help("Disable GDB index generation")
+        .execute(|args, _modifier_stack| {
+            args.gdb_index = false;
             Ok(())
         });
 
@@ -2056,6 +2075,10 @@ impl platform::Args for ElfArgs {
 
     fn should_output_partial_object(&self) -> bool {
         self.should_output_partial_object
+    }
+
+    fn should_write_gdb_index(&self) -> bool {
+        self.gdb_index
     }
 }
 
--- a/libwild/src/gdb_index.rs
+++ b/libwild/src/gdb_index.rs
@@ -261,7 +261,9 @@ pub(crate) fn compute_gdb_index_size(groups: &[GroupState<'_, Elf>]) -> u64 {
             }
             cu_index_base += boundaries.len() as u32;
 
-            collect_pubnames_symbols(object, &offset_to_idx, base_idx, &mut symbol_map);
+            for_each_pubname_entry(object, &offset_to_idx, base_idx, |name, entry| {
+                symbol_map.entry(name).or_default().push(entry);
+            });
         }
     }
 
@@ -470,25 +472,13 @@ fn build_address_and_symbol_tables<'data>(
             }
             cu_offset += boundaries.len() as u32;
 
-            for section_name in [".debug_gnu_pubnames", ".debug_gnu_pubtypes"] {
-                let Some(data) = raw_section_by_name(object, section_name) else {
-                    continue;
-                };
-                for set in parse_pubnames_sets(data) {
-                    let cu_idx = offset_to_idx
-                        .get(&set.debug_info_offset)
-                        .copied()
-                        .unwrap_or(base);
-                    for (name, attrs) in set.entries {
-                        let entry = encode_cu_vector_entry(cu_idx, attrs);
-                        let sd = sym_map.entry(name).or_insert_with(|| SymData {
-                            cv_entries: Vec::new(),
-                            hash: gdb_hash(name),
-                        });
-                        sd.cv_entries.push(entry);
-                    }
-                }
-            }
+            for_each_pubname_entry(object, &offset_to_idx, base, |name, entry| {
+                let sd = sym_map.entry(name).or_insert_with(|| SymData {
+                    cv_entries: Vec::new(),
+                    hash: gdb_hash(name),
+                });
+                sd.cv_entries.push(entry);
+            });
         }
     }
 
@@ -507,12 +497,13 @@ fn build_address_and_symbol_tables<'data>(
     }
 }
 
-/// Collect pubnames/pubtypes symbols from an object into the global map.
-fn collect_pubnames_symbols<'data>(
+/// Iterate over `.debug_gnu_pubnames` and `.debug_gnu_pubtypes` entries in an object,
+/// calling `on_entry(name, encoded_entry)` for each symbol.
+fn for_each_pubname_entry<'data>(
     object: &crate::elf::File<'data>,
     offset_to_idx: &HashMap<u64, u32>,
     fallback_cu: u32,
-    symbol_map: &mut HashMap<&'data [u8], Vec<u32>>,
+    mut on_entry: impl FnMut(&'data [u8], u32),
 ) {
     for section_name in [".debug_gnu_pubnames", ".debug_gnu_pubtypes"] {
         let Some(data) = raw_section_by_name(object, section_name) else {
@@ -524,8 +515,7 @@ fn collect_pubnames_symbols<'data>(
                 .copied()
                 .unwrap_or(fallback_cu);
             for (name, attrs) in set.entries {
-                let entry = encode_cu_vector_entry(cu_idx, attrs);
-                symbol_map.entry(name).or_default().push(entry);
+                on_entry(name, encode_cu_vector_entry(cu_idx, attrs));
             }
         }
     }
--- a/libwild/src/gdb_index.rs
+++ b/libwild/src/gdb_index.rs
@@ -228,10 +228,10 @@ pub(crate) fn compute_gdb_index_size(groups: &[GroupState<'_, Elf>]) -> u64 {
             };
             let object = obj.object;
 
-            let obj_cu_count = raw_section_by_name(object, ".debug_info")
-                .map_or(0, |data| parse_cu_boundaries(data).len());
-
-            if obj_cu_count == 0 {
+            let boundaries = raw_section_by_name(object, ".debug_info")
+                .map(parse_cu_boundaries)
+                .unwrap_or_default();
+            if boundaries.is_empty() {
                 continue;
             }
 
@@ -251,17 +251,16 @@ pub(crate) fn compute_gdb_index_size(groups: &[GroupState<'_, Elf>]) -> u64 {
                 }
             }
 
-            total_cus += obj_cu_count;
+            total_cus += boundaries.len();
             total_addr_entries += obj_addr_count;
 
             let base_idx = cu_index_base;
-            let mut offset_to_idx: HashMap<u64, u32> = HashMap::new();
-            if let Some(di_data) = raw_section_by_name(object, ".debug_info") {
-                for (i, cu) in parse_cu_boundaries(di_data).iter().enumerate() {
-                    offset_to_idx.insert(cu.offset, base_idx + i as u32);
-                }
+            let mut offset_to_idx: HashMap<u64, u32> =
+                HashMap::with_capacity(boundaries.len());
+            for (i, cu) in boundaries.iter().enumerate() {
+                offset_to_idx.insert(cu.offset, base_idx + i as u32);
             }
-            cu_index_base += obj_cu_count as u32;
+            cu_index_base += boundaries.len() as u32;
 
             collect_pubnames_symbols(object, &offset_to_idx, base_idx, &mut symbol_map);
         }
--- a/libwild/src/gdb_index.rs
+++ b/libwild/src/gdb_index.rs
@@ -326,29 +326,26 @@ pub(crate) fn write_gdb_index(buf: &mut [u8], output_buf: &[u8], layout: &Layout
     let short_off = sym_off + (ht_slots * HASH_SLOT_SIZE) as u32;
     let cp_off = short_off + SHORTCUT_TABLE_SIZE as u32;
 
-    // Build constant pool: CU vectors first, then name strings.
-    let mut cv_data = Vec::new();
-    let mut str_data = Vec::new();
+    // Write constant pool: CU vectors first, then name strings.
     let mut cv_offsets = Vec::with_capacity(sorted.len());
-    let mut name_offsets = Vec::with_capacity(sorted.len());
-
+    let mut off = cp_off as usize;
     for (_, sd) in &sorted {
-        cv_offsets.push(cv_data.len() as u32);
-        cv_data.extend_from_slice(&(sd.cv_entries.len() as u32).to_le_bytes());
+        cv_offsets.push((off - cp_off as usize) as u32);
+        buf[off..off + 4].copy_from_slice(&(sd.cv_entries.len() as u32).to_le_bytes());
+        off += 4;
         for &e in &sd.cv_entries {
-            cv_data.extend_from_slice(&e.to_le_bytes());
+            buf[off..off + 4].copy_from_slice(&e.to_le_bytes());
+            off += 4;
         }
     }
+    let mut name_offsets = Vec::with_capacity(sorted.len());
     for (name, _) in &sorted {
-        name_offsets.push((cv_data.len() + str_data.len()) as u32);
-        str_data.extend_from_slice(name);
-        str_data.push(0);
+        name_offsets.push((off - cp_off as usize) as u32);
+        buf[off..off + name.len()].copy_from_slice(name);
+        off += name.len();
+        buf[off] = 0;
+        off += 1;
     }
-
-    // Emit into the output buffer.
-    let total = cp_off as usize + cv_data.len() + str_data.len();
-    let len = buf.len().min(total);
-    let buf = &mut buf[..len];
 
     let hdr = GdbIndexHeader {
         version: GDB_INDEX_VERSION,
@@ -388,10 +385,6 @@ pub(crate) fn write_gdb_index(buf: &mut [u8], output_buf: &[u8], layout: &Layout
         name_of_main_offset: 0,
     };
     buf[so..so + SHORTCUT_TABLE_SIZE].copy_from_slice(sc.as_bytes());
-
-    let cpo = cp_off as usize;
-    buf[cpo..cpo + cv_data.len()].copy_from_slice(&cv_data);
-    buf[cpo + cv_data.len()..cpo + cv_data.len() + str_data.len()].copy_from_slice(&str_data);
 }
 
 /// Build the CU list from the already-written output `.debug_info`.
--- a/libwild/src/gdb_index.rs
+++ b/libwild/src/gdb_index.rs
@@ -431,13 +431,13 @@ fn scan_objects_for_gdb_index<'data>(
         }
         cu_offset += boundaries.len() as u32;
 
-        for_each_pubname_entry(object, &offset_to_idx, base, |name, entry| {
+        for (name, entry) in collect_pubname_entries(object, &offset_to_idx, base) {
             let sd = sym_map.entry(name).or_insert_with(|| SymData {
                 cv_entries: Vec::new(),
                 hash: gdb_hash(name),
             });
             sd.cv_entries.push(entry);
-        });
+        }
     }
 
     for sd in sym_map.values_mut() {
@@ -507,14 +507,14 @@ fn build_address_entries(layout: &Layout<'_, Elf>) -> Vec<GdbIndexAddressEntry> 
     entries
 }
 
-/// Iterate over `.debug_gnu_pubnames` and `.debug_gnu_pubtypes` entries in an object,
-/// calling `on_entry(name, encoded_entry)` for each symbol.
-fn for_each_pubname_entry<'data>(
+/// Collect encoded pubname/pubtype entries from an object's `.debug_gnu_pubnames`
+/// and `.debug_gnu_pubtypes` sections, returning `(name, encoded_cu_vector_entry)` pairs.
+fn collect_pubname_entries<'data>(
     object: &crate::elf::File<'data>,
     offset_to_idx: &HashMap<u64, u32>,
     fallback_cu: u32,
-    mut on_entry: impl FnMut(&'data [u8], u32),
-) {
+) -> Vec<(&'data [u8], u32)> {
+    let mut entries = Vec::new();
     for section_name in [".debug_gnu_pubnames", ".debug_gnu_pubtypes"] {
         let Some(data) = raw_section_by_name(object, section_name) else {
             continue;
@@ -525,10 +525,11 @@ fn for_each_pubname_entry<'data>(
                 .copied()
                 .unwrap_or(fallback_cu);
             for (name, attrs) in set.entries {
-                on_entry(name, encode_cu_vector_entry(cu_idx, attrs));
+                entries.push((name, encode_cu_vector_entry(cu_idx, attrs)));
             }
         }
     }
+    entries
 }
 
 /// Insert symbols into the open-addressing hash table region of `buf`.
--- a/libwild/src/gdb_index.rs
+++ b/libwild/src/gdb_index.rs
@@ -116,32 +116,30 @@ struct CuBoundary {
 }
 
 /// Walk `.debug_info` bytes and return `(offset, total_length)` for each CU.
-fn parse_cu_boundaries(data: &[u8]) -> Result<Vec<CuBoundary>> {
+fn parse_cu_boundaries(data: &[u8]) -> Vec<CuBoundary> {
     let mut cus = Vec::new();
     let mut offset = 0usize;
     while offset + 4 <= data.len() {
         let init_len = u32_from_slice(&data[offset..]);
         let total = if init_len == 0xFFFF_FFFF {
-            crate::ensure!(
-                offset + 12 <= data.len(),
-                "truncated DWARF64 initial length in .debug_info at offset {offset:#x}"
-            );
+            if offset + 12 > data.len() {
+                break;
+            }
             let len = u64_from_slice(&data[offset + 4..]);
             12 + len as usize
         } else {
             4 + init_len as usize
         };
-        crate::ensure!(
-            total > 0 && offset + total <= data.len(),
-            "invalid CU length {total} in .debug_info at offset {offset:#x}"
-        );
+        if total == 0 || offset + total > data.len() {
+            break;
+        }
         cus.push(CuBoundary {
             offset: offset as u64,
             length: total as u64,
         });
         offset += total;
     }
-    Ok(cus)
+    cus
 }
 
 struct PubnamesSet<'data> {
@@ -153,7 +151,7 @@ struct PubnamesSet<'data> {
 ///
 /// Each set has a header pointing to a CU in `.debug_info`, followed by
 /// (die_offset, attrs_byte, NUL-terminated name) entries.
-fn parse_pubnames_sets(data: &[u8]) -> Result<Vec<PubnamesSet<'_>>> {
+fn parse_pubnames_sets(data: &[u8]) -> Vec<PubnamesSet<'_>> {
     let mut sets = Vec::new();
     let mut pos = 0;
     while pos + 4 <= data.len() {
@@ -161,19 +159,17 @@ fn parse_pubnames_sets(data: &[u8]) -> Result<Vec<PubnamesSet<'_>>> {
 
         let (header_size, set_end, debug_info_offset) = if init_len == 0xFFFF_FFFF {
             // DWARF64: 4 + 8(len) + 2(ver) + 8(offset) + 8(size) = 30
-            crate::ensure!(
-                pos + 30 <= data.len(),
-                "truncated DWARF64 pubnames header at offset {pos:#x}"
-            );
+            if pos + 30 > data.len() {
+                break;
+            }
             let len = u64_from_slice(&data[pos + 4..]);
             let dio = u64_from_slice(&data[pos + 14..]);
             (30, pos + 12 + len as usize, dio)
         } else {
             // DWARF32: 4(len) + 2(ver) + 4(offset) + 4(size) = 14
-            crate::ensure!(
-                pos + 14 <= data.len(),
-                "truncated DWARF32 pubnames header at offset {pos:#x}"
-            );
+            if pos + 14 > data.len() {
+                break;
+            }
             let dio = u64::from(u32_from_slice(&data[pos + 6..]));
             (14, pos + 4 + init_len as usize, dio)
         };
@@ -185,18 +181,16 @@ fn parse_pubnames_sets(data: &[u8]) -> Result<Vec<PubnamesSet<'_>>> {
 
         while ep < set_end {
             let die_offset = if is_64 {
-                crate::ensure!(
-                    ep + 8 <= set_end,
-                    "truncated DWARF64 pubnames entry at offset {ep:#x}"
-                );
+                if ep + 8 > set_end {
+                    break;
+                }
                 let v = u64_from_slice(&data[ep..]);
                 ep += 8;
                 v
             } else {
-                crate::ensure!(
-                    ep + 4 <= set_end,
-                    "truncated DWARF32 pubnames entry at offset {ep:#x}"
-                );
+                if ep + 4 > set_end {
+                    break;
+                }
                 let v = u64::from(u32_from_slice(&data[ep..]));
                 ep += 4;
                 v
@@ -204,20 +198,18 @@ fn parse_pubnames_sets(data: &[u8]) -> Result<Vec<PubnamesSet<'_>>> {
             if die_offset == 0 {
                 break;
             }
-            crate::ensure!(
-                ep < set_end,
-                "truncated pubnames entry (missing attrs/name) at offset {ep:#x}"
-            );
+            if ep >= set_end {
+                break;
+            }
             let attrs = data[ep];
             ep += 1;
             let name_start = ep;
             while ep < set_end && data[ep] != 0 {
                 ep += 1;
             }
-            crate::ensure!(
-                ep < set_end,
-                "unterminated pubnames name string at offset {name_start:#x}"
-            );
+            if ep >= set_end {
+                break;
+            }
             entries.push((&data[name_start..ep], attrs));
             ep += 1;
         }
@@ -228,7 +220,7 @@ fn parse_pubnames_sets(data: &[u8]) -> Result<Vec<PubnamesSet<'_>>> {
         });
         pos = set_end;
     }
-    Ok(sets)
+    sets
 }
 
 /// Read raw section data from an input object by name.
@@ -398,7 +390,7 @@ fn build_cu_list(output_buf: &[u8], layout: &Layout<'_, Elf>) -> Result<Vec<GdbI
         ".debug_info layout extends beyond output buffer ({end} > {})",
         output_buf.len()
     );
-    Ok(parse_cu_boundaries(&output_buf[start..end])?
+    Ok(parse_cu_boundaries(&output_buf[start..end])
         .into_iter()
         .map(|cu| GdbIndexCuEntry {
             cu_offset: cu.offset,
@@ -430,7 +422,7 @@ fn scan_objects_for_gdb_index<'data>(
 
     for (object, sections) in objects {
         let boundaries = match raw_section_by_name(object, DEBUG_INFO_SECTION_NAME_STR)? {
-            Some(data) => parse_cu_boundaries(data)?,
+            Some(data) => parse_cu_boundaries(data),
             None => continue,
         };
         if boundaries.is_empty() {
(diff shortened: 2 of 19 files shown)


libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+
+/// Iterate over `.debug_gnu_pubnames` and `.debug_gnu_pubtypes` entries in an object,
+/// calling `on_entry(name, encoded_entry)` for each symbol.
+fn for_each_pubname_entry<'data>(
+    object: &crate::elf::File<'data>,
+    offset_to_idx: &HashMap<u64, u32>,
+    fallback_cu: u32,
+    mut on_entry: impl FnMut(&'data [u8], u32),
+) {
+    for section_name in [".debug_gnu_pubnames", ".debug_gnu_pubtypes"] {
```
lapla-cogito: The symbol table construction assumes the presence of these two sections by adding the compiler flag `-ggnu-pubnames` (see the added integration test). Without these sections, the symbol table will not be built and only the CU list and address table will be constructed. This effectively means you won't benefit from GDB index, but since lld and mold behave this way, I've aligned the implementation.

lapla-cogito:
I did `git commit --amend --allow-empty` because the **release** workflow was triggered by the previous commit (which I canceled)...
https://github.com/wild-linker/wild/actions/runs/26509302406

I have no idea what happened... 🤷

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    offset: u64,
+    length: u64,
+}
+
+/// Walk `.debug_info` bytes and return `(offset, total_length)` for each CU.
+fn parse_cu_boundaries(data: &[u8]) -> Vec<CuBoundary> {
+    let mut cus = Vec::new();
+    let mut offset = 0usize;
+    while offset + 4 <= data.len() {
+        let init_len = u32::from_le_bytes(data[offset..offset + 4].try_into().unwrap());
```
marxin: I would use `u32_from_slice` and similarly `u64_from_slice` function we have.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+struct CuBoundary {
+    offset: u64,
+    length: u64,
+}
+
+/// Walk `.debug_info` bytes and return `(offset, total_length)` for each CU.
+fn parse_cu_boundaries(data: &[u8]) -> Vec<CuBoundary> {
+    let mut cus = Vec::new();
+    let mut offset = 0usize;
+    while offset + 4 <= data.len() {
```
marxin, line 131: Might be good linking a place where the CU encoding is actually defined.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+            if offset + 12 > data.len() {
+                break;
+            }
+            let len = u64::from_le_bytes(data[offset + 4..offset + 12].try_into().unwrap());
+            12 + len as usize
+        } else {
+            4 + init_len as usize
+        };
+        if total == 0 || offset + total > data.len() {
+            break;
```
marxin: Shouldn't we rather report a warning as the emitted GDB index will be likely incomeplete/corrupted?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    (num_symbols * 4 / 3 + 1).next_power_of_two()
+}
+
+struct CuBoundary {
+    offset: u64,
+    length: u64,
+}
+
+/// Walk `.debug_info` bytes and return `(offset, total_length)` for each CU.
+fn parse_cu_boundaries(data: &[u8]) -> Vec<CuBoundary> {
```
marxin: As I am more thinking about it - in general, can't we just the `gimli` crate for the parsing of the various DWARF-related data structures? Like here, using https://docs.rs/gimli/latest/gimli/read/struct.DebugInfo.html#method.units. I know there might be some performance penalty, but we should definitely measure it.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+struct PubnamesSet<'data> {
+    debug_info_offset: u64,
+    entries: Vec<(&'data [u8], u8)>,
+}
+
+/// Parse `.debug_gnu_pubnames` / `.debug_gnu_pubtypes` section data.
+///
+/// Each set has a header pointing to a CU in `.debug_info`, followed by
+/// (die_offset, attrs_byte, NUL-terminated name) entries.
+fn parse_pubnames_sets(data: &[u8]) -> Vec<PubnamesSet<'_>> {
```
marxin: Similarly here: https://docs.rs/gimli/latest/gimli/read/struct.DebugPubNames.html#method.items

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+        });
+        pos = set_end;
+    }
+    sets
+}
+
+/// Read raw section data from an input object by name.
+fn raw_section_by_name<'data>(object: &crate::elf::File<'data>, name: &str) -> Option<&'data [u8]> {
+    let (_index, header) = object.section_by_name(name)?;
+    object.raw_section_data(header).ok()
```
marxin: Should we really swallow the error with `ok`? I would rather return `Result<Option<&[u8]>>`.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+
+            let mut obj_addr_count = 0usize;
+            for (si, slot) in obj.sections.iter().enumerate() {
+                let SectionSlot::Loaded(section) = slot else {
+                    continue;
+                };
+                if section.size == 0 {
+                    continue;
+                }
+                let Ok(header) = object.section(object::SectionIndex(si)) else {
```
marxin: Maybe same here, shouldn't we rather propagate an error?

libwild/src/gdb_index.rs
marxin: Maybe declaring section name constants?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    AddressAndSymbolData {
+        addr_entries,
+        sorted_symbols: sorted,
+        ht_slots,
+    }
+}
+
+/// Iterate over `.debug_gnu_pubnames` and `.debug_gnu_pubtypes` entries in an object,
+/// calling `on_entry(name, encoded_entry)` for each symbol.
+fn for_each_pubname_entry<'data>(
```
marxin: Given the fact, the `parse_pubnames_sets` allocates an vector, shouldn't we just concatenate the 2 vectors and return it as a return value (and don't pass a `on_entry` callback). And have a processing of the data in a loop in the caller? What do you think about it?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+        return;
+    }
+    let mask = (ht_slots - 1) as u32;
+    for (i, (_, sd)) in sorted.iter().enumerate() {
+        let h = sd.hash;
+        let step = ((h >> 3) & mask) | 1;
+        let mut slot = h & mask;
+        loop {
+            let so = ht_start + slot as usize * HASH_SLOT_SIZE;
+            let existing_name = u32::from_le_bytes(buf[so..so + 4].try_into().unwrap());
```
marxin: Maybe using a zerocopy for reading and writing of the pair of the values?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+
+    if total_cus == 0 {
+        return 0;
+    }
+
+    let mut cv_bytes = 0usize;
+    let mut str_bytes = 0usize;
+    for (name, entries) in &mut symbol_map {
+        entries.sort_unstable();
```
marxin: What about using `BTreeSet` as the value type of the `symbol_map`?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+        entries.dedup();
+        cv_bytes += 4 + entries.len() * 4;
```
marxin: Can you explain the `4 + ` part here?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+/// and re-scans input objects for address ranges and pubnames/pubtypes symbols.
+pub(crate) fn write_gdb_index(buf: &mut [u8], output_buf: &[u8], layout: &Layout<'_, Elf>) {
+    if buf.is_empty() {
+        return;
+    }
+
+    let cu_entries = build_cu_list(output_buf, layout);
+    let AddressAndSymbolData {
+        addr_entries,
+        sorted_symbols: sorted,
```
marxin: Can't we rather use `sorted_names`, it's more descriptive than `sorted`?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    let cpo = cp_off as usize;
+    buf[cpo..cpo + cv_data.len()].copy_from_slice(&cv_data);
+    buf[cpo + cv_data.len()..cpo + cv_data.len() + str_data.len()].copy_from_slice(&str_data);
+}
+
+/// Build the CU list from the already-written output `.debug_info`.
+fn build_cu_list(output_buf: &[u8], layout: &Layout<'_, Elf>) -> Vec<GdbIndexCuEntry> {
+    let Some(id) = layout
+        .output_sections
+        .section_id_by_name(SectionName(b".debug_info"))
```
marxin: I would add a new section name constant for `".debug_info"`.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    else {
+        return Vec::new();
+    };
+    let sl = layout.section_layouts.get(id);
+    let start = sl.file_offset;
+    let end = start + sl.file_size;
+    if end > output_buf.len() {
```
marxin: Again, maybe let's propagate an Error?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+        buf,
+        ht_slots,
+        sym_off as usize,
+        &sorted,
+        &name_offsets,
+        &cv_offsets,
+    );
+
+    let so = short_off as usize;
+    let sc = GdbIndexShortcutTable {
```
marxin, line 377: Should we add a code comment that explains why we currently skip the information? Or even better - can we fill up the information?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+
+            // Address entries: map each executable section to its resolved address.
+            for (si, slot) in obj.sections.iter().enumerate() {
+                let SectionSlot::Loaded(section) = slot else {
+                    continue;
+                };
+                if section.size == 0 {
+                    continue;
+                }
+                let Ok(header) = object.section(object::SectionIndex(si)) else {
```
marxin: Again, probably report an Error here?

libwild/src/gdb_index.rs
marxin: Can we emit a warning to the user telling the option is leading to a GDB index result?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+            });
+        }
+    }
+
+    for sd in sym_map.values_mut() {
+        sd.cv_entries.sort_unstable();
+        sd.cv_entries.dedup();
+    }
+
+    let mut sorted: Vec<(&[u8], SymData)> = sym_map.into_iter().collect();
```
marxin: You can use directly `sorted_by_key_unstable` on the Iterator from Itertools, no need for a mutable vector that is then sorted.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+}
+
+struct AddressAndSymbolData<'data> {
+    addr_entries: Vec<GdbIndexAddressEntry>,
+    sorted_symbols: Vec<(&'data [u8], SymData)>,
+    ht_slots: usize,
+}
+
+/// Build address entries and symbol table in a single pass over input objects.
+fn build_address_and_symbol_tables<'data>(
```
marxin: This function seems pretty close to `compute_gdb_index_size` - can't we just call the function twice and use it first for the size computation and later on for the actual encoding?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+    r
+}
+
+/// Encode a CU vector entry: bits 0-23 = CU index, bits 28-30 = kind, bit 31 = is_static.
+///
+/// The attrs byte from `.debug_gnu_pubnames`/`.debug_gnu_pubtypes` packs kind in bits 4-6
+/// and is_static in bit 7.
+fn encode_cu_vector_entry(cu_index: u32, attrs: u8) -> u32 {
+    let kind = u32::from((attrs >> 4) & 0x7);
```
marxin: What about using `extract_bit_range`?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+    for (name, _) in &sorted {
+        name_offsets.push((cv_data.len() + str_data.len()) as u32);
+        str_data.extend_from_slice(name);
+        str_data.push(0);
+    }
+
+    // Emit into the output buffer.
+    let total = cp_off as usize + cv_data.len() + str_data.len();
+    let len = buf.len().min(total);
```
marxin: This should be a hard error if the allocated buffer does not match the expected written data.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    let short_off = sym_off + (ht_slots * HASH_SLOT_SIZE) as u32;
+    let cp_off = short_off + SHORTCUT_TABLE_SIZE as u32;
+
+    // Build constant pool: CU vectors first, then name strings.
+    let mut cv_data = Vec::new();
+    let mut str_data = Vec::new();
+    let mut cv_offsets = Vec::with_capacity(sorted.len());
+    let mut name_offsets = Vec::with_capacity(sorted.len());
+
+    for (_, sd) in &sorted {
```
marxin: Can we directly encode the `cv_data` and `str_data` into the output buffer?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,612 @@
...
+    let ht_end = ht_start + ht_slots * HASH_SLOT_SIZE;
+    buf[ht_start..ht_end].fill(0);
+
+    if ht_slots == 0 {
+        return;
+    }
+    let mask = (ht_slots - 1) as u32;
+    for (i, (_, sd)) in sorted.iter().enumerate() {
+        let h = sd.hash;
+        let step = ((h >> 3) & mask) | 1;
```
marxin: Similarly to the `gdb_hash`, what about making it a function?

marxin: Btw. looking at the documentation, the formula is `((hash * 17) & (size - 1)) | 1` - which does not correspond to `hash >> 3` part - or do I overlook something?

marxin (changes requested):
Given the numerous findings I created - can you explain what was the genesis of the PR and how much was a LLM involved? I know we had a PR for the very same functionality that was closed right after it was opened. Is it an incarnation of the changes, or was it created completely independently?

lapla-cogito:
@marxin (I haven't seen the review details yet)
> we had a PR for the very same functionality that was closed right after it was opened

I think you mean #1605. I used that as a reference when implementing this, since I had also reviewed it at the time and therefore already had some familiarity with the surrounding context and implementation details. However, I did not reuse it directly for licensing concerns (and in any case, it does not support version 9 as-is). I mainly used LLMs to help refine the implementation through review and iterative revisions.

marxin:
> I think you mean #1605. I used that as a reference when implementing this, since I had also reviewed it at the time and therefore already had some familiarity with the surrounding context and implementation details.

Yes - fine, that's a good approach you chose ;)

> However, I did not reuse it directly for licensing concerns (and in any case, it does not support version 9 as-is). I mainly used LLMs to help refine the implementation through review and iterative revisions.

All good! Let's iterate on the PR, given you knowledge, I guess you can address the findings pretty fast. Happy to review and help you with the feature.

libwild/src/gdb_index.rs
lapla-cogito: Since these section names aren't referenced anywhere else, I think there isn't much benefit to making them constants.

libwild/src/gdb_index.rs
lapla-cogito: Is this a performance-related comment? If so, generally speaking, the number of CU vec entries is typically just a few. Therefore, I believe the current implementation is more efficient than using `BTreeSet` in this case.

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/ac5bbc46973c82150ad39a097c6a25bc1ec98aad

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/e3cc626208cf06971dd06c0b6ed0657e9cafb5a1

libwild/src/gdb_index.rs
lapla-cogito: Thanks, done https://github.com/wild-linker/wild/pull/1990/commits/e3cc626208cf06971dd06c0b6ed0657e9cafb5a1

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/e3cc626208cf06971dd06c0b6ed0657e9cafb5a1

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/31ff30173de96d9fdc7a082f468cfb46a973f64f

libwild/src/gdb_index.rs
lapla-cogito: I also considered a similar one during implementation, but I don't think it's trivial. The symbol data constructed in `compute_gdb_index_size()` includes `&'data [u8]`, so when merging these 2 functions, we need to store this in the `Layout` struct and retain it until the write phase.

libwild/src/gdb_index.rs
marxin: Ok.

libwild/src/gdb_index.rs
marxin: No, also for convenience reasons - no need for:
```
sd.cv_entries.sort_unstable();
sd.cv_entries.dedup();
```

libwild/src/gdb_index.rs
marxin: No, I mean calling the function 2x is fine to me - first time for the size estimation and second time for the actual writting of the data. But in both cases relying on the same function. Does it make sense to you?

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/121674b7c63f177acd9be561401500fd6fe0e1c3

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/aef860f29257815d2aaf3b83e224e135a6cfe933

libwild/src/gdb_index.rs
lapla-cogito: Done https://github.com/wild-linker/wild/pull/1990/commits/b8f4da76f5dc38a6e58efa37911ba818c62e1076

libwild/src/gdb_index.rs
lapla-cogito: Ah makes sense, thanks. https://github.com/wild-linker/wild/pull/1990/commits/d9abf65a55344e5cf1d64401c8f03043d4c96753

libwild/src/gdb_index.rs
lapla-cogito: Ah right, done in https://github.com/wild-linker/wild/pull/1990/commits/21b9dca539f24fbd72beea9a9d19c127d396de7b

libwild/src/gdb_index.rs
lapla-cogito: I think it's neat. https://github.com/wild-linker/wild/pull/1990/commits/9d099f323fd8e11b7f1121c29a2a64dfc873a670

libwild/src/gdb_index.rs
marxin: Great!

libwild/src/gdb_index.rs
lapla-cogito: gimli doesn't appear to support GNU attributes bytes, so I think we can't use it as-is.
https://github.com/gimli-rs/gimli/blob/45dde2309bda8f385acdf0c945ebc4f275ffe124/src/read/lookup.rs#L186-L199

libwild/src/gdb_index.rs
lapla-cogito: From [this](https://github.com/wild-linker/wild/pull/1990#discussion_r3329635809), the only function that can currently be written using gimli is this one. It feels somewhat excessive to bring in gimli just for this purpose (since constructing indices requires only `offset` and `length`)... 🤔

libwild/src/gdb_index.rs
lapla-cogito: Maybe I'm bad at naming things. https://github.com/wild-linker/wild/pull/1990/commits/9de1cc65d65fd1e035fddfe91c58cfb7e485ef80

libwild/src/gdb_index.rs
lapla-cogito: Done in https://github.com/wild-linker/wild/pull/1990/commits/55610bc8cbf96f570b33911031d8c9e17a3e6dba. This commit also resolves other issues related to error handling.

libwild/src/gdb_index.rs
marxin: All right. I thought the crate we be more beneficial for reading, but as the e.g. `.debug_gnu_pubnames` is unsupported, I would go without the crate.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,632 @@
...
+fn parse_pubnames_sets(data: &[u8]) -> Vec<PubnamesSet<'_>> {
+    let mut sets = Vec::new();
+    let mut pos = 0;
+    while pos + 4 <= data.len() {
+        let init_len = u32_from_slice(&data[pos..]);
+
+        let (header_size, set_end, debug_info_offset) = if init_len == 0xFFFF_FFFF {
+            // DWARF64: 4 + 8(len) + 2(ver) + 8(offset) + 8(size) = 30
+            if pos + 30 > data.len() {
+                break;
```
marxin: I believe most of the `break`s in the function should be actually replaced with an error handling instead.

libwild/src/gdb_index.rs
marxin: Ah, you are right. I filled an issue for the future: https://github.com/gimli-rs/gimli/issues/879.

libwild/src/gdb_index.rs
marxin: Thanks for the comment.

libwild/src/gdb_index.rs
lapla-cogito: As shown by [this test failure](https://github.com/wild-linker/wild/actions/runs/26735903714/job/78788993654), some toolchains don't generate strictly correct `.debug_info`. Therefore, both lld and mold are designed to be tolerant of such structures, so the original `break` approach should work well I think.

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,644 @@
...
+                .context("Failed to read .gdb_index hash table slot")?;
+            if existing.name_offset == 0 && existing.cu_vector_offset == 0 {
+                let new_slot = GdbIndexHashSlot {
+                    name_offset: name_offsets[i],
+                    cu_vector_offset: cv_offsets[i],
+                };
+                buf[so..so + HASH_SLOT_SIZE].copy_from_slice(new_slot.as_bytes());
+                break;
+            }
+            slot = (slot + step) & mask;
```
davidlattimore, line 727: Do you think it'd be worthwhile asserting that we haven't gone back to the starting slot? In case we somehow allocated too few slots - it'd be better to crash that infinitely loop

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,644 @@
...
+    name: &str,
+) -> Result<Option<&'data [u8]>> {
+    let Some((_index, header)) = object.section_by_name(name) else {
+        return Ok(None);
+    };
+    Ok(Some(object.raw_section_data(header)?))
+}
+
+/// Pre-scan all input objects to compute the `.gdb_index` section size.
+pub(crate) fn compute_gdb_index_size(groups: &[GroupState<'_, Elf>]) -> Result<u64> {
```
davidlattimore: Could do with a timing_phase here

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,644 @@
...
+    total_addr_entries: usize,
+    sorted_symbols: Vec<(&'data [u8], SymData)>,
+    ht_slots: usize,
+}
+
+/// Scan input objects to build the symbol table and count CUs / address entries.
+fn scan_objects_for_gdb_index<'data>(
+    objects: impl Iterator<Item = (&'data crate::elf::File<'data>, &'data [SectionSlot])>,
+) -> Result<GdbIndexScanResult<'data>> {
+    let mut total_cus = 0usize;
```
davidlattimore: Could do with a timing_phase here

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,644 @@
...
+            let FileLayout::Object(obj) = f else {
+                return None;
+            };
+            Some((obj.object, obj.sections.as_slice()))
+        });
+    let GdbIndexScanResult {
+        sorted_symbols: sorted_names,
+        ht_slots,
+        ..
+    } = scan_objects_for_gdb_index(objects)?;
```
davidlattimore: I notice that this function is called twice. Is it expensive? If so, would it make sense to store the result from the first call?

davidlattimore (commented):
I ran a quick benchmark to see what effect adding the flag `--gdb-index` has on the performance of the different linkers. With lld and mold, the performance difference between `--gdb-index` and `--no-gdb-index` was negligible. With wild however, it slows down linking by about 2x. Benchmark was with 32 threads linking zed.

I expect that the main problem is that there's lots of work being done single-threaded. e.g. iterating through all the objects. Also, looking up relevant sections by name is likely to be slow - as opposed to our usual approach where SectionRules determines what we should do with each input section.

libwild/src/gdb_index.rs
marxin: Well, I am not aligned with that: apparently it seems we're decoding a compress ZLIB stream, so our parsing logic is effectively hiding a real bug:
```
failures:
        elf/x86_64/gdb-index/enabled

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

  stderr ───
    [libwild/src/gdb_index.rs:127:9] init_len = 1
    [libwild/src/gdb_index.rs:137:9] (total, offset) = (
        5,
        0,
    )
    [libwild/src/gdb_index.rs:127:9] init_len = 3053453312
    [libwild/src/gdb_index.rs:137:9] (total, offset) = (
        3053453316,
        5,
    )
...
❯ eu-readelf -zS ./wild/tests/build/elf/x86_64/gdb-index/enabled/gdb-index.c.o
...
[ 5] .debug_info          PROGBITS     0000000000000000 000000a0 0000007f  0 C      0   0  8
     [ELF ZLIB (1) 000000b6  1]
...
./wild/tests/build/elf/x86_64/gdb-index/enabled/gdb-index.c.o:  file format elf64-x86-64

.debug_info contents:
0x00000000: Compile Unit: length = 0x000000b2, format = DWARF32, version = 0x0005, unit_type = DW_UT_compile, abbr_offset = 0x0000, addr_size = 0x08 (next unit at 0x000000b6)
```

so the expected CU size is `0x000000b2`. So please, let's start with a strict mode and we can then make an intensive testing if that's really the case. If so, then we have basically 2 options: not emitting the `.gdbindex`, or bail-out and provide an error message.

marxin:
After spending a little bit of time researching the area, I noticed there's a systematic DWARF extension **Lookup by Name** aka `.debug_names` (DWARF 5 6.1.1): https://dwarfstd.org/doc/DWARF5.pdf. It's pretty similar to `.gdbindex`, but supported by both `lldb` and `gdb`. Clang can emit the format with `-g -gpubnames` (can be dumped with `llvm-dwarfdump --debug-names`) and `lld` has an option that builds one global cache (`--debug-names`). Moreover, what's interesting, `gdb` itself can generate (and consume) the index format: `gdb-add-index --dwarf-5`. If I'm not mistaken, it's only GCC that lacks the support.

With that said, shouldn't the long-term, future-focused approach be to build out the Debug Names format rather than introducing an ad hoc, GDB-specific index file format?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,644 @@
...
+    while pos + 4 <= data.len() {
+        let init_len = u32_from_slice(&data[pos..]);
+
+        let (header_size, set_end, debug_info_offset) = if init_len == 0xFFFF_FFFF {
+            // DWARF64: 4 + 8(len) + 2(ver) + 8(offset) + 8(size) = 30
+            if pos + 30 > data.len() {
+                break;
+            }
+            let len = u64_from_slice(&data[pos + 4..]);
+            let dio = u64_from_slice(&data[pos + 14..]);
```
philipc, line 181: Is there a reason we don't parse the cu size field here, and use it instead of the size from `parse_cu_boundaries`? And if we do use the cu size from here, is there any reason to do `parse_cu_boundaries` at all? I think the cu offset can be adjusted with a simple section base addition.

libwild/src/gdb_index.rs
marxin: Now the `gimli` support got added: https://github.com/gimli-rs/gimli/pull/883.

libwild/src/gdb_index.rs
lapla-cogito: While an actual infinite loop is impossible to occur (since `compute_hash_table_slots()` calculates the number of slots using `(num_symbols * 4 / 3 + 1).next_power_of_two()`, ensuring more slots are allocated than there are symbols), I've added the `ensure!()` call as a precaution.

libwild/src/gdb_index.rs
lapla-cogito: Indeed this call is expensive (and naturally contributes to performance degradation). Therefore, I also believe it should be cached in the future, but achieving this would require layout refactoring (as we discussed during our previous meeting, parallelizing object traversal (mentioned in above) also requires such changes). I'm unsure whether we should include these in this PR.

edit: After actually trying it out, I realized the change weren't as extensive as I thought, so I added it: https://github.com/wild-linker/wild/pull/1990/commits/0c0b335592d13e1b74a4342728510f68bbc3062e

libwild/src/gdb_index.rs
lapla-cogito: pubnames/pubtypes sets only cover CUs that have public names. CUs with no exported symbols won't have a set. The gdb index CU list have to include all CUs (the address table indexes into it), so we need `parse_cu_boundaries()` to discover the complete set and assign correct global indices.

lapla-cogito:
> After spending a little bit of time researching the area, I noticed there's a systematic DWARF extension **Lookup by Name** aka `.debug_names` (DWARF 5 6.1.1): https://dwarfstd.org/doc/DWARF5.pdf. It's pretty similar to `.gdbindex`, but supported by both `lldb` and `gdb`. Clang can emit the format with `-g -gpubnames` (can be dumped with `llvm-dwarfdump --debug-names`) and `lld` has an option that builds one global cache (`--debug-names`). Moreover, what's interesting, `gdb` itself can generate (and consume) the index format: `gdb-add-index --dwarf-5`. If I'm not mistaken, it's only GCC that lacks the support.
> 
> With that said, shouldn't the long-term, future-focused approach be to build out the Debug Names format rather than introducing an ad hoc, GDB-specific index file format?

Interesting. While we could indeed consider adding `.debug_names` support, I don't think it would be mutually exclusive with `.gdb_index` support. Moreover, considering implementation complexity and usage frequency, I don't think this would serve as a blocking factor for supporting GDB index.

libwild/src/gdb_index.rs
philipc: My understanding was that .debug_gnu_pubnames does include local symbols, so would the problem instead be that some files were not compiled with -ggnu-pubnames?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,687 @@
...
+                if !header.is_alloc() || !header.is_executable() {
+                    continue;
+                }
+                if let Some(addr) = obj.section_resolutions[si].address()
+                    && addr != 0
+                {
+                    entries.push(GdbIndexAddressEntry {
+                        low_address: addr,
+                        high_address: addr + section.size,
+                        cu_index: base_cu,
```
philipc: If a file contains multiple CUs, does this assign all addresses to the first CU?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,687 @@
...
+    buf: &mut [u8],
+    output_buf: &[u8],
+    layout: &Layout<'_, Elf>,
+    scan: &GdbIndexScanResult<'_>,
+) -> Result {
+    if buf.is_empty() {
+        return Ok(());
+    }
+
+    timing_phase!("Write GDB index");
```
davidlattimore: Performance has improved. This time I benchmarked linking of zed-release and zed-debug. zed-release slowed down 12% when `--gdb-index` was passed, while zed-debug slowed down by 52%. The slowest phase is definitely "Write GDB index". It looks like `write_gdb_index` does a few different things. `write_hash_table` is already a separate function. Would any of the other things in here make sense to separate out to separate functions? Could you also add timing phases to any bits that seem like they might be significant?

libwild/src/args/elf.rs
```diff
@@ -2057,6 +2076,10 @@ impl platform::Args for ElfArgs {
     fn should_output_partial_object(&self) -> bool {
         self.should_output_partial_object
     }
+
+    fn should_write_gdb_index(&self) -> bool {
```
davidlattimore, line 2080: It looks like lld and mold, if --strip-debug is supplied (or presumably --strip, but I didn't try that) won't write the gdb-index. Could we do the same (and have a test for it)? As a separate PR is fine, but also fine in this PR.

libwild/src/gdb_index.rs
lapla-cogito: Ah, you're right that the lack of support for compressed debug sections is indeed the issue. I've added this support and replaced most `break` statements with `ensure!`.

libwild/src/gdb_index.rs
lapla-cogito: I see. Is this something that can happen in scenarios like `--relocatable`? Anyay, I was able to reproduce it with a test case and have fixed it. Thanks.

libwild/src/gdb_index.rs
lapla-cogito: > would the problem instead be that some files were not compiled with -ggnu-pubnames?

Ah, right. Since we need to store all CUs in the CU list, we can't rely on `.debug_gnu_pubnames` to enumerate them. `parse_cu_boundaries()` walks `.debug_info` directly, which is always present with `-g`, so it handles this correctly.

libwild/src/gdb_index.rs
lapla-cogito: Actually I was already looking into optimizing the write phase, so while I was at it, I added a few more profiling points as well. https://github.com/wild-linker/wild/pull/1990/changes/fb3ccdf20ff3359b795bc9e041046a1cd616ea3f

libwild/src/args/elf.rs
lapla-cogito: Testing locally, neither lld nor mold emits a GDB index when either `--strip-all` or `--strip-debug` is passed, so I changed the implementation to match that behavior. https://github.com/wild-linker/wild/pull/1990/commits/aa4f1c368abc6840d29a595b113c84d1281d7ca5

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,788 @@
...
+
+struct SymData {
+    cv_entries: BTreeSet<u32>,
+    hash: u32,
+}
+
+pub(crate) struct GdbIndexScanResult {
+    total_cus: usize,
+    total_addr_entries: usize,
+    sorted_symbols: Vec<(Vec<u8>, SymData)>,
```
marxin, line 418: Not sure if possible, but can't we rather use `&'data [u8]` slices for the names?

libwild/src/gdb_index.rs
```diff
@@ -0,0 +1,788 @@
...
+                cv_entries: BTreeSet::new(),
+                hash: gdb_hash(name),
+            });
+            sd.cv_entries.insert(entry);
+        }
+    }
+
+    let sorted: Vec<(Vec<u8>, SymData)> = sym_map
+        .into_iter()
+        .sorted_unstable_by(|(a, _), (b, _)| a.cmp(b))
```
marxin, line 521: Probably can be simplified with `sorted_unstable_by_key`.

libwild/src/gdb_index.rs
marxin: We can always change it if we want - am fine with the current approach.

libwild/src/gdb_index.rs
marxin: Great, thanks.

marxin (approved):
The reworked implementation is much better 👍 one thing I would consider: given the close connection to gdb, I would consider the integration of `gdb` into our testing harness.
There  might be some debugging output that will show if the index was read and then we can inspect it? Just something for a consideration.

libwild/src/gdb_index.rs
lapla-cogito: However, this would likely require cloning the `Vec`, so keeping it as is might be preferable I think.

libwild/src/gdb_index.rs
lapla-cogito: `Vec` is used here because `section_data_cow()` returns a Cow. If we didn't use Cow, we would have to introduce a lifetime param, which would make the layout more cumbersome to manage. (See [this commit](https://github.com/wild-linker/wild/pull/1990/changes/1011cec0a31ec6e5c2c127812a4ec0099eadbb48), where `&'data [u8]` was replaced with `Vec`.)

marxin: Well, `Cow` seems fine to me, but leaving that for now.

libwild/src/gdb_index.rs
marxin: You are right ;)

lapla-cogito:
Thank you for all the reviews. I'm going to merge this. For test integration, I'm also in favor of extending the tests to use gdb, but at least at this stage, I believe the current integration tests and mold tests (which already use `gdb` internally) are sufficient.

PR #1994, opened 2026-05-28
feat: Implement SORT() directive for linker scripts

### LINKED ISSUE #1661: Implement linker-script SORT(...) for input sections

Abh10705:
Transitioned from hashmaps to flat vectors (Vec<ScriptSortedSectionDetail>) for harvested sections. This ensures continuous memory layout and significantly improves L1 cache locality during the final layout binary search.
Fixes https://github.com/wild-linker/wild/issues/1661


libwild/src/elf_writer.rs
```diff
@@ -1462,11 +1462,36 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
 
     let _span = debug_span!("write_file", filename = %object.input).entered();
     let _file_span = layout.args().common().trace_span_for_file(object.file_id);
+
+    // Fast O(1) lookup to skip harvested sections
```
davidlattimore: Anything that involves a loop is unlikely to be O(1).

libwild/src/elf_writer.rs
```diff
@@ -1462,11 +1462,36 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
+    let epilogue = layout
```
davidlattimore: I feel like there's an easier way to get the epilogue. Pretty sure we store the FileId of the epilogue somewhere

libwild/src/elf_writer.rs
```diff
@@ -3968,11 +3994,30 @@ fn write_epilogue<A: Arch<Platform = Elf>>(
     // The actual build-id will be filled in later once all writing has completed. It's important
     // that we fill it with zeros now however, since if we're overwriting an existing file, there
     // might be other data there and we don't zero it, then the build ID will be hashing that data.
+
```
davidlattimore: This extra blank line seems unnecessary and separates the above comment from the code it's referring to

libwild/src/layout.rs
```diff
@@ -290,14 +297,39 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+        // Ensure the memory address correctly aligns with CPU instruction requirements
+        *offset = sec.alignment.align_up(*offset);
+        sec.mem_offset = *offset;
+        *offset += sec.size;
+
+        // Push directly into our flat Vector—no hashing math required!
+        harvested_sections_registry.push(sec.clone());
+    }
+
+    // Crucial: We MUST sort by the IDs here, otherwise the Binary Search in Step 3 will fail.
```
davidlattimore: To what extent was this PR written by an LLM and to what extent do you understand the code that was produced? I ask here because referring to "step 3" seems like the kind of thing an LLM would do. There doesn't seem to be numbered "steps" defined anywhere.

libwild/src/layout.rs
```diff
@@ -5408,3 +5480,65 @@ impl OutputRecordLayout {
...
+
+#[derive(Clone, Debug)]
+pub(crate) struct HarvestedSortedSection<'data> {
+    pub(crate) file_id: FileId,
+    pub(crate) section_index: object::SectionIndex,
+    pub(crate) part_id: PartId,
+    pub(crate) size: u64,
+    pub(crate) alignment: Alignment,
+    pub(crate) mem_offset: u64,
+    pub(crate) _name: &'data [u8],
```
davidlattimore: Why is this here?

libwild/src/layout.rs
```diff
@@ -5408,3 +5480,65 @@ impl OutputRecordLayout {
...
+                                _name: note.name, // pointer
+                            },
+                        ));
+                    }
+                }
+            }
+        }
+    }
+
+    temp.sort_by(|a, b| a.0.cmp(b.0));
```
davidlattimore: sort_by_key would be simpler

libwild/src/resolution.rs
```diff
@@ -764,6 +769,9 @@ pub(crate) struct ResolvedObject<'data, P: Platform> {
     custom_sections: Vec<CustomSectionDetails<'data>>,
 
     init_fini_sections: Vec<InitFiniSectionDetail>,
+    //// Stored as a flat Vec to ensure L1 cache locality during the final layout binary search
```
davidlattimore: This comment seems unnecessary. Also, the way things are stored affects all caches, not just L1

wild/tests/sources/elf/script-sort/script-sort.c
```diff
@@ -0,0 +1,17 @@
...
+__attribute__((used, section(".text.func_c"))) int func_c() { return 3; }
+
+__attribute__((used, section(".text.func_a"))) int func_a() { return 1; }
+
+__attribute__((used, section(".text.func_b"))) int func_b() { return 2; }
+
+void _start(void) {
+  runtime_init();
+
+  // Safely exit with the standard success code
```
davidlattimore: This comment is unnecessary. All our tests exit with this exit code

wild/tests/sources/elf/script-sort/script-sort.c
```diff
@@ -0,0 +1,17 @@
+//#LinkArgs:-T ./script-sort.ld
+//#Object:../common/runtime.c
+
+#include "../common/runtime.h"
+
+__attribute__((used, section(".text.func_c"))) int func_c() { return 3; }
+
+__attribute__((used, section(".text.func_a"))) int func_a() { return 1; }
+
+__attribute__((used, section(".text.func_b"))) int func_b() { return 2; }
```
davidlattimore: This test doesn't seem  to actually be testing that the sections are sorted. If I disable sorting, e.g. by always setting SectionPattern.sorted to false, the test still passes. It'd also be good to test that sorted sections from multiple input objects get correctly sorted together.

libwild/src/linker_script.rs
```diff
@@ -1111,10 +1115,33 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
+        // Inside SORT(...), accept a single wildcard pattern up to ')'
+        let name = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+        // Closing ')'
+        ')'.parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+
+        Ok(SectionPattern { name, sorted: true })
+    } else {
+        // Original behavior: bare pattern
```
davidlattimore: Documenting "original behaviour" makes sense from the perspective of reading a PR, but doesn't make sense for someone reading the code later, after the PR is merged. Comments should always be written with the future reader in mind. This comment is probably unnecessary.

libwild/src/linker_script.rs
```diff
@@ -1111,10 +1115,33 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
 }
 
-fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
-    let pattern = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
-    skip_comments_and_whitespace(input)?;
-    Ok(pattern)
+fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<SectionPattern<'input>> {
+    // Handling SORT(...) wrapper: SORT is an alias for SORT_BY_NAME in GNU ld.
+    if input.starts_with(b"SORT") {
+        // Consume "SORT"
```
davidlattimore: Comments like this don't really add anything. It's pretty much just repeating what the next line says.

libwild/src/linker_script.rs
```diff
@@ -1111,10 +1115,33 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
     })
 }
 
-fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<&'input [u8]> {
-    let pattern = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
-    skip_comments_and_whitespace(input)?;
-    Ok(pattern)
+fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<SectionPattern<'input>> {
+    // Handling SORT(...) wrapper: SORT is an alias for SORT_BY_NAME in GNU ld.
+    if input.starts_with(b"SORT") {
```
davidlattimore: I'm pretty sure there is a helper that checks `starts_with` and then consumes

wild/tests/sources/elf/script-sort/script-sort.ld
```diff
@@ -0,0 +1,16 @@
...
+SECTIONS
+{
+    /* Start at base + 4KB to safely leave room for ELF headers without using math */
+    . = 0x401000;
+
+    .text : {
+        *(.text)
+        
+        /* Protect our harvested sections from the GC! */
+        KEEP(*(SORT(.text.*)))
```
davidlattimore: It'd be good to test both with and without KEEP and make sure that GC works as expected for sorted sections that aren't referenced.

davidlattimore (commented):
I've only partially reviewed, but I just noticed that this is a new PR. Was there a reason to open a new PR rather than continuing with 1857? Also, I just noticed that some of the comments I made on that PR haven't yet been addressed in this PR, so it's probably worthwhile looking back at the comments on that PR

Abh10705:
> I've only partially reviewed, but I just noticed that this is a new PR. Was there a reason to open a new PR rather than continuing with 1857? Also, I just noticed that some of the comments I made on that PR haven't yet been addressed in this PR, so it's probably worthwhile looking back at the comments on that PR

yeh that just skipped my mind, i will close this one and go back to the previous one

libwild/src/layout.rs
Abh10705: yeh that was during the merge conflict i coudn't figure it out, so i pasted my file and the file from the wild repo and asked it to make a new merged file, it asked for some context as to what I was doing/ changing to the code and added these comments I should have paid more attention in removing them

Abh10705: if you go over our conversation on zulip, you should get the idea if I understand what Im doing, respectfully

davidlattimore: Thanks for the explanation. Hope I didn't offend, it's just that I know this PR is a difficult undertaking for someone without previous experience in the codebase and I worry when I see signs of LLM use in those sorts of circumstances. That makes sense that you used it for fixing merge conflicts and that it introduced changes that were perhaps unintended.

Abh10705: wasn't offended at all, it didn't even merge those files properly I had to restart and fix those myself line by line, while i asked it to merge the file I also asked it to review my work and that is where the comments are from, I removed them as i was fixing the merge conflicts but ig this one got left

libwild/src/elf_writer.rs
```diff
@@ -3964,15 +3984,32 @@ fn write_epilogue<A: Arch<Platform = Elf>>(
             &epilogue_offsets,
         )?;
     }
-
```
davidlattimore, line 4028: Can you restore this blank line?

libwild/src/layout.rs
```diff
@@ -290,14 +297,38 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+    let starting_mem_offsets_by_group =
+        compute_start_offsets_by_group(&group_states, mem_offsets.clone());
 
     let merged_string_start_addresses = MergedStringStartAddresses::compute(
         &output_sections,
         &starting_mem_offsets_by_group,
         &merged_strings,
     );
 
+    // --- SECTION SORTING & ADDRESS ASSIGNMENT ---
```
davidlattimore: We don't usually use comments of this form in the linker. If it's worth having a "section" delimiter, then it's probably worth trying to factor out the code. Perhaps the code below can be extracted out into a separate function with a meaningful name? Then there won't be a "section", there will just be a call to some function.

libwild/src/layout.rs
```diff
@@ -290,14 +297,38 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+    let mut epilogue_offsets = starting_mem_offsets_by_group.last().unwrap().clone();
+
+    for sec in &mut script_sorted_sections {
+        let offset = epilogue_offsets.get_mut(sec.part_id);
+        // Ensure the memory address correctly aligns with CPU instruction requirements
+        *offset = sec.alignment.align_up(*offset);
+        sec.mem_offset = *offset;
+        *offset += sec.size;
+
+        // Push directly into our flat Vector—no hashing math required!
```
davidlattimore: I'm not sure that this comment adds anything useful. Probably best to remove it.

libwild/src/layout.rs
```diff
@@ -290,14 +297,38 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+    // addresses. We perform a two-step finalization:
+    // Sort the harvested sections according to the requested criteria.
+    // Linearize them in memory, starting from the base offset of their respective output section
+    // part, and advancing by the size of each section.
+    let mut harvested_sections_registry = Vec::with_capacity(script_sorted_sections.len());
+    let mut epilogue_offsets = starting_mem_offsets_by_group.last().unwrap().clone();
+
+    for sec in &mut script_sorted_sections {
+        let offset = epilogue_offsets.get_mut(sec.part_id);
+        // Ensure the memory address correctly aligns with CPU instruction requirements
```
davidlattimore: I'm not sure the comment is accurate. We're not checking anything related to the CPU, we're just aligning based on whatever the input file. I'd suggest just removing the comment.

libwild/src/layout.rs
```diff
@@ -5408,3 +5471,67 @@ impl OutputRecordLayout {
...
+
+#[derive(Clone, Debug)]
+pub(crate) struct HarvestedSortedSection {
+    pub(crate) file_id: FileId,
+    pub(crate) section_index: object::SectionIndex,
+    pub(crate) part_id: PartId,
+    pub(crate) size: u64,
+    pub(crate) alignment: Alignment,
+    pub(crate) mem_offset: u64,
+    //pub(crate) _name: &'data [u8],
```
davidlattimore: Why just comment it out? If we don't need it, it should be removed

libwild/src/sframe.rs
```diff
@@ -279,8 +279,15 @@ pub(crate) fn sort_sframe_section(
     let total_fre_size: usize = entries.iter().map(|e| e.fre_bytes.len()).sum();
     let total_size = header_end_offset + fde_size + total_fre_size;
 
-    if total_size > section.len() {
+    /*   if total_size > section.len() {
         bail!("Merged SFrame section too large");
+    } */
+    if total_size > section.len() {
```
davidlattimore, line 282: This change doesn't look to be related to sorted sections. Can it be reverted?

libwild/src/layout.rs
```diff
@@ -5408,3 +5471,67 @@ impl OutputRecordLayout {
...
+                                mem_offset: 0,
+                            },
+                        ));
+                    }
+                }
+            }
+        }
+    }
+
+    temp.sort_by_key(|a| a.0);
```
davidlattimore: If I comment out this line, the tests still pass. That suggests that the test aren't properly testing that section sorting is occurring. It should be possible to write some C code that runs as part of the test that verifies that the functions being sorted are in the expected order.

libwild/src/layout.rs
```diff
@@ -5408,3 +5471,67 @@ impl OutputRecordLayout {
...
+    if !has_any_sorting {
+        return Vec::new();
+    }
+
+    let mut temp = Vec::new();
+    for group in group_states.iter_mut() {
+        for file in &mut group.files {
+            if let FileLayoutState::Object(obj) = file {
+                for section_req in &obj.script_sorted_sections {
+                    //for section_req in &obj.script_sorted_sections {
```
davidlattimore: Commented out code should generally be cleaned up

libwild/src/layout.rs
```diff
@@ -155,12 +156,18 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
+        &output_sections,
+        &symbol_db.section_part_ids,
+    );
 
     group_states.push(GroupState {
         files: vec![FileLayoutState::Epilogue(EpilogueLayoutState::new(
             symbol_db.args,
             symbol_db.output_kind,
             &mut dynamic_symbol_definitions,
+            script_sorted_sections.clone(),
```
davidlattimore: Cloning here seems error-prone. It means that we have two copies of this vec, one that gets modified below and one that isn't. i.e. the two copies get out of sync. Why do we need two copies of the vec and doesn't it matter that they're not the same?

libwild/src/layout.rs
```diff
@@ -290,14 +297,38 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     )?;
 
     let mem_offsets: OutputSectionPartMap<u64> = starting_memory_offsets(&section_part_layouts);
-    let starting_mem_offsets_by_group = compute_start_offsets_by_group(&group_states, mem_offsets);
+    let starting_mem_offsets_by_group =
+        compute_start_offsets_by_group(&group_states, mem_offsets.clone());
```
davidlattimore: Why is this call to `clone` here? It seems unnecessary

libwild/src/layout.rs
```diff
@@ -3977,22 +4026,36 @@ impl<'data, P: Platform> ObjectLayoutState<'data, P> {
...
                         sframe_ranges.push(offset..offset + len);
                     }
+
+                    if let Ok(idx) = resources
+                        .harvested_sections_registry
+                        .binary_search_by_key(&(self.file_id, sec_idx), |s| {
+                            (s.file_id, s.section_index.0)
+                        })
+                    {
+                        address = resources.harvested_sections_registry[idx].mem_offset;
```
davidlattimore, line 4237: Commenting out this line doesn't result in any test failures. That suggests that the test isn't checking that we're correctly assigning addresses to symbols in sorted sections. Likely, the simplest way to test this would be to call the functions in the sorted section and verify that the functions return the expected values. The functions and the calls to the functions should probably be in separate compilation units, otherwise there's the risk that the compiler could inline the function.

libwild/src/layout.rs
Abh10705: ownership issues, i'm trying to find a better way to fix it now

wild/tests/sources/elf/script-sort/script-sort.ld
```diff
@@ -0,0 +1,28 @@
...
+    .text : {
+        *(.text)
+        
+        /* Our custom sorted sections */
+        *(SORT(.text.sort.*))
+        KEEP(*(SORT(.text.kept.*)))
+        *(SORT(.text.drop.*))
+    }
+
+    /* * using a quick workaround to push the metadata onto a separate 64KB page.
```
davidlattimore: What's the cause of the segfault that you're seeing? This doesn't look like something we'd be able to merge, so I think we need to find a better approach.

davidlattimore: To clarify what I'm asking, you mention alignment being the problem, but why are things not aligned? Is this a problem with executing a binary where a linker script was used, or is this something specific to sorting?

Abh10705: No, it has nothing to do with sorting. It's just a side effect of using a custom linker script with LLD

from my research i found out that  It happens because the minimal test script doesn't have page alignment directives

lld  puts .text (executable) and .eh_frame (read-only) on the exact same memory page and the OS  applies the read only protection to the whole page and that messes up with execute permissions which causes a crash.

I mean using ALIGN is the standard way of to do things but we dont support that yet as I got syantax errors when i tried to do that.

I only hardcoded stuff to bypass this the sorting logic isnt the issue

libwild/src/layout.rs
```diff
@@ -5408,3 +5459,91 @@ impl OutputRecordLayout {
...
+
+    if !has_any_sorting {
+        return Vec::new();
+    }
+
+    let mut temp = Vec::new();
+    for group in group_states.iter_mut() {
+        for file in &mut group.files {
+            if let FileLayoutState::Object(obj) = file {
+                for section_req in &obj.script_sorted_sections {
```
davidlattimore: Why the name `section_req`? To me that implies "section request", but I'm not sure where the "request" part is coming from.

libwild/src/layout.rs
```diff
@@ -5408,3 +5459,91 @@ impl OutputRecordLayout {
...
+                false
+            }
+        })
+    });
+
+    if !has_any_sorting {
+        return Vec::new();
+    }
+
+    let mut temp = Vec::new();
```
davidlattimore: Can probably have a better name than `temp`. How about `sections_out`?

wild/tests/sources/elf/script-sort/script-sort.ld
davidlattimore: OK, thanks for the explanation. The comment here in the file seems like too much and makes it sound like we're working around some bug. Probably fine to remove it or just put something simple like "Ensure valid alignment for subsequent sections".

libwild/src/elf_writer.rs
```diff
@@ -1462,11 +1462,30 @@ fn write_object<'data, A: Arch<Platform = Elf>>(
...
     let _span = debug_span!("write_file", filename = %object.input).entered();
     let _file_span = layout.args().common().trace_span_for_file(object.file_id);
+
+    let Some(FileLayout::Epilogue(epilogue)) =
+        layout.group_layouts.last().and_then(|g| g.files.last())
+    else {
+        unreachable!("Epilogue is broken and must be the last file in the final layout group");
+    };
+
+    let mut is_harvested = vec![false; object.sections.len()];
```
davidlattimore: Seems unfortunate to create this Vec, especially for the common case when nothing is sorted. Would it work to add a new kind of SectionSlot instead?

libwild/src/resolution.rs
```diff
@@ -764,6 +768,8 @@ pub(crate) struct ResolvedObject<'data, P: Platform> {
     custom_sections: Vec<CustomSectionDetails<'data>>,
 
     init_fini_sections: Vec<InitFiniSectionDetail>,
+    //// same as layout file
```
davidlattimore: It's not obvious to me what this comment means. Just removing it would be an option.

libwild/src/linker_script.rs
```diff
@@ -1111,10 +1115,32 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
...
+    use winnow::combinator::opt;
+    use winnow::token::literal;
+    if opt(literal(b"SORT")).parse_next(input)?.is_some() {
+        skip_comments_and_whitespace(input)?;
+
+        '('.parse_next(input)?;
+        skip_comments_and_whitespace(input)?;
+
+        // Inside SORT(...), accept a single wildcard pattern up to ')'
+        let name = take_while(1.., |b| !b" \n\t)".contains(&b)).parse_next(input)?;
```
davidlattimore: It looks like we're looking for a very exact sequence of whitespace and will fail to parse if it's not found. Why?

Abh10705:
I might have messed up the merge issue and messed up some of your code, just take a look at the epilogue layout plz

davidlattimore:
> I might have messed up the merge issue and messed up some of your code, just take a look at the epilogue layout plz

The best way to verify if the merge did anything it shouldn't have is to read through the diff (e.g. here in github) and see if there's any changes that you don't recognise as your changes. Let us know when you've addressed the previous comments and I'll take another look.

libwild/src/elf_writer.rs
Abh10705: so the issue i had at the start was with borrow checkers, when we start the script sorting phase, the core object structures and section slot are already being shared across all of our threads. i dont think we can get mutable access to section slot to mark it with sort.
This may be wrong but to my understanding even if I added a tag inside section slot enum how would the elf_writter know where the sections are in memory and we would have to go over everything 1 by 1 which doesnt sound optimized
so before i had a boolean vec which was taking up space on the heap, now what we do is we collect sections, sort them into a vec . Then when the elf_writer is goign through those sections we use binary search whhich finds a match in O(logN)

davidlattimore (approved):
There's still a few things that I'd like done differently, including my previous suggestions of using a new variant of SectionSlot and avoiding having two separate copies of the sorted sections state. As this is, the feature works and doesn't regress performance, which is great. So I think to avoid further merge conflicts and dragging this out even longer, it's best if I merge this as-is and I'll follow up with the other changes I wanted. Thanks for all your work on this!

PR #1996, opened 2026-05-29
fix(LoongArch64/RV64): allow empty sequence for merge_eflags

marxin:
Noticed that while debugging the Mold external tests on LoongArch64 (and RV64) - for an empty shared library, we can have none eflags to be merged and so we should allow empty sequence and build `object::elf::FileFlags::(0)`.


marxin:
Good, I've just verified we behave the same as Mold and LLD on the LoongArch64 platform - 2 Mold external tests are fixed with that.

davidlattimore (approved):
This doesn't necessitate an update to the mold skip tests list? Otherwise, looks good.

marxin:
These tests were not excluded in the `mold_skip_tests.toml` because we support them on platforms like `x86_64`, where we don't merge the eflags.

PR #1999, opened 2026-05-30
feat(ppc64): REL24 local entry plus REL32/ADDR32/REL64

runlevel5:
Extends the ppc64le static-relocation support added in #1977 with the
  remaining pieces needed for a static link to come out correct on ELFv2:

  - **REL24 local entry.** An ELFv2 callee that sets up r2 has a separate
    local entry point. A direct call that already shares the callee's TOC
    should branch there, skipping the global-entry prologue. The offset is
    encoded in the callee's `st_other` (`STO_PPC64_LOCAL`); functions with
    no separate local entry (offset 0) are unaffected.

  - **REL32 / ADDR32.** 32-bit PC-relative and absolute. REL32 is what
    `.eh_frame` is built out of, so without it nothing with exception
    tables links cleanly.

  - **REL64.** 64-bit PC-relative — same shape as REL32 but 8 bytes.

  A small generic hook `Arch::local_entry_offset(st_other)` is added so the
  REL24 path can ask the target which entry to bias the branch by; default
  is 0 (no other architecture cares).

### DIFF
--- a/libwild/src/elf_ppc64.rs
+++ b/libwild/src/elf_ppc64.rs
@@ -3,6 +3,7 @@ use crate::elf::Elf;
 use crate::error;
 use crate::error::Result;
 use crate::platform::Platform;
+use linker_utils::bit_misc::BitExtraction;
 use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::RelocationKindInfo;
 use linker_utils::elf::ppc64_rel_type_to_string;
@@ -67,6 +68,14 @@ impl crate::platform::Arch for ElfPpc64 {
 
     fn high_part_relocations() -> &'static [u32] {
         &[]
+    }
+
+    fn local_entry_offset(st_other: u8) -> u64 {
+        // ELFv2 encodes the distance from a function's global entry point to its local entry point
+        // in st_other bits [7:5]: offset = ((1 << k) >> 2) << 2 (k=0,1 -> 0; 2 -> 4; 3 -> 8; ...).
+        let bit = u32::from(object::elf::STO_PPC64_LOCAL_BIT);
+        let k = u64::from(st_other).extract_bit_range(bit..bit + 3);
+        u64::from(((1u32 << k) >> 2) << 2)
     }
 
     #[allow(unused_variables)]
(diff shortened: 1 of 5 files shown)


linker-utils/src/ppc64.rs
```diff
@@ -39,14 +39,37 @@ impl RelaxationKind {
...
             RelocationSize::ByteSize(8),
             AllowedRange::no_check(),
             1,
             0,
         ),
+        object::elf::R_PPC64_ADDR32 => (
+            RelocationKind::Absolute,
+            RelocationSize::ByteSize(4),
+            // Can represent either a signed or an unsigned 32-bit value.
+            AllowedRange::new(-(2i64.pow(31)), 2i64.pow(32)),
```
davidlattimore, line 58: Does a negative absolute address make sense? What does it mean?

libwild/src/elf_ppc64.rs
```diff
@@ -69,6 +69,13 @@ impl crate::platform::Arch for ElfPpc64 {
         &[]
     }
 
+    fn local_entry_offset(st_other: u8) -> u64 {
+        // ELFv2 encodes the distance from a function's global entry point to its local entry point
+        // in st_other bits [7:5]: offset = ((1 << k) >> 2) << 2 (k=0,1 -> 0; 2 -> 4; 3 -> 8; ...).
+        let k = (st_other & object::elf::STO_PPC64_LOCAL_MASK) >> object::elf::STO_PPC64_LOCAL_BIT;
+        u64::from(((1u32 << k) >> 2) << 2)
```
davidlattimore, line 78: Consider using `extract_bit_range` (grep for it to see other uses)

linker-utils/src/ppc64.rs
runlevel5: I merely match the binutils' `complain_overflow_bitfield`. The `R_AARCH64_ABS32` in `linker-utils/src/aarch64.rs` is also the same. Hence my decision to keep it that way for consistency. Having said that I am more than happy to make it stricter as you suggested.

davidlattimore (approved):

PR #2003, opened 2026-05-31
test(ppc64): enable ppc64le in the test harness and CI

runlevel5:
This PR wires ppc64le into the integration-test harness, linker-diff, and CI so the port is actually covered, and installs the cross toolchain so it runs in GitHub Actions under qemu.


lapla-cogito:
Please create a new section for ppc64 in [`tests/external_tests/mold_skip_tests.toml`](https://github.com/wild-linker/wild/blob/main/wild/tests/external_tests/mold_skip_tests.toml) and put `arch-ppc64le-save-restore-gprs.sh` into it to make the external tests CI pass.

davidlattimore (approved):

PR #2004, opened 2026-05-31
chore: Simplify taplo install in CI

lapla-cogito:


.github/workflows/ci.yml
```diff
@@ -256,22 +249,6 @@ jobs:
...
-      - uses: actions/checkout@v6
-        with:
-          persist-credentials: false
-      - name: Install Taplo (x86_64)
-        if: ${{ runner.arch == 'X64' }}
-        run: curl -fsSL https://github.com/tamasfe/taplo/releases/latest/download/taplo-linux-x86_64.gz | gzip -d - | install -m 755 /dev/stdin /usr/local/bin/taplo
-      - name: Install Taplo (AArch64)
-        if: ${{ runner.arch == 'ARM64' }}
-        run: curl -fsSL https://github.com/tamasfe/taplo/releases/latest/download/taplo-linux-aarch64.gz | gzip -d - | install -m 755 /dev/stdin /usr/local/bin/taplo
-      - run: taplo format --check --diff
```
lapla-cogito, line 273: TBH I don't understand the purpose behind this job that only displays diffs (which is why I removed it). Seeing other tidy tests like `clang-format`, this job is likely unnecessary.

taplo.toml
```diff
@@ -1,5 +1 @@
-exclude = [
-  ".cargo-home/**",
-  "external_test_suites/mold/third-party/blake3/Cargo.toml",
```
lapla-cogito, line 3: Considering future updates to external tests, all files under `external_test_suites` should be included as ignored-files.

.github/workflows/ci.yml
marxin: Good point!

marxin (approved):
Thank you for the clean-up.

PR #2005, opened 2026-05-31
feat: support using an expression when setting the location counter

vishruth-thimmaiah:
This PR adds support for using expressions when setting the current location within `SECTIONS`.

This PR doesn't add support for using symbols when setting the location pointer.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -490,7 +490,7 @@ fn update_defsym_symbol_resolution<'data, P: Platform>(
     if let SymbolPlacement::Redirect(redirect) = &def_info.placement {
         let value = crate::expression_eval::evaluate_expression(
             &redirect.expression,
-            redirect.loc,
+            &redirect.loc,
             section_layouts,
             output_sections,
             memory_regions,
@@ -4858,7 +4858,7 @@ fn layout_section_parts<'data, P: Platform>(
                         // The OrderEvent::SetLocation is ELF-specific only.
                         mem_offset = crate::expression_eval::evaluate_expression(
                             &expr,
-                            crate::parsing::SymbolLoc::None,
+                            &crate::parsing::SymbolLoc::None,
                             &empty_section_layouts,
                             output_sections,
                             memory_regions,
@@ -4894,7 +4894,7 @@ fn layout_section_parts<'data, P: Platform>(
                 if let Some(ref expr) = section_info.location {
                     mem_offset = crate::expression_eval::evaluate_expression(
                         expr,
-                        crate::parsing::SymbolLoc::None,
+                        &crate::parsing::SymbolLoc::None,
                         &empty_section_layouts,
                         output_sections,
                         memory_regions,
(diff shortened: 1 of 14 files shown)


wild/tests/sources/elf/linker-script-executable/linker-script-executable.ld
```diff
@@ -2,9 +2,9 @@ ENTRY(begin_here)
 
 SECTIONS
 {
-    . = 0x600000;
+    . = 0x300000 * 2;
+	start_of_sections = .;
```
vishruth-thimmaiah, line 6: Setting the location before defining any sections sets the location for the first section in the section table, not the first section in the linker script. Hence the need for this change, as the first section might not be `.text`.

davidlattimore: I guess possibly another way of looking at what you said is that when the linker reads the linker script, when it encounters the first section, it emits certain linker-generated sections before that first section. This alternative way of looking at it might suggest a slightly different implementation. e.g. perhaps when reading the linker script, those implicit sections should be inserted, almost as if they were in the file. Not sure if that would have any implications or how it would compare in terms of complexity. Don't feel that you need to change anything in this regard... just something to think about.

Also related... I assume the location counter can be set either before the first section or between sections. Are both tested?

libwild/src/layout_rules.rs
```diff
@@ -302,37 +289,40 @@ impl<'data> LayoutRulesBuilder<'data> {
...
+                                SymbolLoc::Expression(_, section_id) => section_id,
+                                SymbolLoc::SectionStart(section_id)
+                                | SymbolLoc::SectionEnd(section_id) => Some(section_id),
+                                _ => None,
+                            };
+                            loc = SymbolLoc::Expression(new_location.address.clone(), section_id);
+                            if current_section_id.is_none() {
+                                output_sections.set_base_address(new_location.address.clone());
+                                continue;
+                            }
```
davidlattimore, line 303: If I comment out this if-block, the tests still seem to still all pass. Is it possible to make the test exercise and verify the behaviour here?

libwild/src/layout.rs
```diff
@@ -232,13 +233,28 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
         &symbol_db,
     );
 
+    let memory_regions: Vec<crate::linker_script::MemoryRegion<'_>> = symbol_db
```
davidlattimore, line 236: If I replace this Vec with and empty one, the tests still pass. Is it practical to add a test that catches it?

vishruth-thimmaiah: We don't fully support the syntax for defining `MEMORY` regions that ld requires for an executable, so writing a test for this might not be possible here. `memory_regions` here is used for evaluating `ORIGIN` and `LENGTH` functions in an expression.

wild/tests/sources/elf/linker-script-executable/linker-script-executable.ld
vishruth-thimmaiah: > Also related... I assume the location counter can be set either before the first section or between sections. Are both tested?

Yes, this is already tested in line 10.

libwild/src/layout_rules.rs
vishruth-thimmaiah: I suppose we could verify .text is not at the base address, since it is not the first section outputed.

davidlattimore (approved):

PR #2009, opened 2026-06-02
refactor: Remove unused Cie::referenced symbols

davidlattimore:


libwild/src/elf.rs
```diff
@@ -2615,7 +2610,7 @@ fn process_eh_frame_relocations<'data, 'scope, A: Arch<Platform = Elf>, R: Reloc
 
         if prefix.cie_id == 0 {
             // This is a CIE
```
lapla-cogito, line 2612: Isn't this line unnecessary?

lapla-cogito (approved):

libwild/src/elf.rs
davidlattimore: You mean the comment? I find the comment helpful. There's an equivalent line below that says "This is an FDE". They let me quickly see what the two branches of the if-else are for

lapla-cogito: > There's an equivalent line below that says "This is an FDE". 

Ah I hadn't noticed that. Then this is helpful.

PR #2012, opened 2026-06-02
chore: Remove redundant trim calls in integration test

davidlattimore:


davidlattimore:
We already call `trim()` once at the start of the method

PR #2013, opened 2026-06-03
feat: support defining custom phdrs in linker scripts

vishruth-thimmaiah:
This PR adds support for defining program headers in linker scripts, using the `PHDRS` command.

This PR doesn't add support for defining multiple phdrs in the same section.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1897,7 +1897,7 @@ impl platform::Platform for Elf {
         output_kind: OutputKind,
         output_sections: &OutputSections<'data, Self>,
         secondary: &OutputSectionMap<Vec<OutputSectionId>>,
-        linker_script: &[&SequencedLinkerScript<'data, Self>],
+        linker_scripts: &[&SequencedLinkerScript<'data, Self>],
         phdr_map: &mut hashbrown::HashMap<&[u8], Vec<OutputSectionId>>,
     ) -> Result<(OutputOrder<'data>, ProgramSegments<Self::ProgramSegmentDef>)> {
         let mut builder =
@@ -1908,7 +1908,7 @@ impl platform::Platform for Elf {
         let mut insert_r = false;
         let mut first_load = false;
 
-        for script in linker_script {
+        for script in linker_scripts {
             for phdr in &script.parsed.program_headers {
                 let ptype = expression_eval::evaluate_const(&phdr.ptype)?;
                 let flags = phdr
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -13,9 +13,11 @@ use crate::ensure;
 use crate::error;
 use crate::error::Context as _;
 use crate::error::Result;
+use crate::expression_eval;
 use crate::file_kind::FileKind;
 use crate::file_writer::copy_section_data;
 use crate::grouping::Group;
+use crate::grouping::SequencedLinkerScript;
 use crate::input_data::InputBytes;
 use crate::input_data::InputRef;
 use crate::layout;
@@ -967,6 +969,7 @@ impl platform::Platform for Elf {
                 min_alignment: d.min_alignment,
                 location: None,
                 secondary_order: None,
+                phdr_name: None,
             })
             .collect()
     }
@@ -1827,7 +1830,8 @@ impl platform::Platform for Elf {
         output_sections: &OutputSections<'data, Self>,
         secondary: &OutputSectionMap<Vec<OutputSectionId>>,
     ) -> (OutputOrder<'data>, ProgramSegments<Self::ProgramSegmentDef>) {
-        let mut builder = OutputOrderBuilder::<Self>::new(output_kind, output_sections, secondary);
+        let mut builder =
+            OutputOrderBuilder::<Self>::new(output_kind, output_sections, secondary, false);
 
         builder.add_section(output_section_id::FILE_HEADER);
         builder.add_section(output_section_id::PROGRAM_HEADERS);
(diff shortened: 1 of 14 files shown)


libwild/src/expression_eval.rs
```diff
@@ -204,6 +204,50 @@ pub(crate) fn evaluate_expression<'data, P: Platform>(
...
+        Expression::RightShift(l, r) => {
+            Ok(evaluate_const(l)?.wrapping_shr(evaluate_const(r)? as u32))
+        }
+        Expression::LogicalAnd(l, r) => Ok(u64::from(
+            evaluate_const(l)? != 0 && evaluate_const(r)? != 0,
+        )),
+        Expression::LogicalOr(l, r) => Ok(u64::from(
+            evaluate_const(l)? != 0 || evaluate_const(r)? != 0,
+        )),
+        Expression::LogicalNot(expression) => Ok(u64::from(!evaluate_const(expression)? != 0)),
```
davidlattimore: This doesn't seem right to me. It's doing a bitwise not, then comparing the result with 0. I'd have thought that a logical not should output 0 for every input other than 0, but this looks like it will output 0 only if the input is u64::MAX.

libwild/src/elf.rs
```diff
@@ -1888,6 +1892,98 @@ impl platform::Platform for Elf {
...
+                            let mut pflags = SegmentFlags(0);
+                            if flags.contains(shf::ALLOC) {
+                                pflags |= pf::READABLE;
+                            }
+                            if flags.contains(shf::WRITE) {
+                                pflags |= pf::WRITABLE;
+                            }
+                            if flags.contains(shf::EXECINSTR) {
+                                pflags |= pf::EXECUTABLE;
+                            }
```
davidlattimore, line 1934: If I delete these three if statements, the test still passes. Do you think we need some way for the test to assert or otherwise verify the segment properties? I could imagine having an "ExpectSection" command in the tests, similar to "ExpectSym" - then have a range of properties that can be asserted about the section - e.g. what permissions the LOAD segment that the section is contained in has.

libwild/src/output_section_id.rs
```diff
@@ -741,6 +783,67 @@ impl<'data, P: Platform> OutputSections<'data, P> {
...
+                if info.section_attributes.is_alloc() {
+                    custom.ro.push(id);
+                } else {
+                    custom.nonalloc.push(id);
+                }
+            } else if info.section_attributes.is_no_bits() {
+                custom.bss.push(id);
+            } else {
+                custom.data.push(id);
+            }
```
davidlattimore, line 805: This block of code looks to be identical to what's in `output_order`. Do you think it'd make sense to extract it out to a function to avoid the duplication? Alternatively, we could consider merging `output_order_from_linker_script` with `output_order`, since there's other code that the two share. I'm pretty sure the function isn't very performance sensitive.

libwild/src/output_section_id.rs
```diff
@@ -741,6 +783,67 @@ impl<'data, P: Platform> OutputSections<'data, P> {
         P::build_output_order_and_program_segments(&custom, output_kind, self, &secondary)
     }
 
+    pub(crate) fn output_order_from_linker_script(
+        &self,
+        output_kind: OutputKind,
+        linker_script: &[&SequencedLinkerScript<'data, P>],
```
davidlattimore: linker_script -> linker_scripts

libwild/src/output_section_id.rs
```diff
@@ -741,6 +783,67 @@ impl<'data, P: Platform> OutputSections<'data, P> {
...
+                return;
+            }
+
+            if let Some(phdr_name) = info.phdr_name {
+                let sections = phdr_map.get_mut(phdr_name);
+                if let Some(sections) = sections {
+                    sections.push(id);
+                } else {
+                    phdr_map.insert(phdr_name, vec![id]);
+                }
```
davidlattimore: This can probably be written more simply as `phdr_map.entry(phdr_name).or_default().push(id)`

libwild/src/output_section_id.rs
```diff
@@ -192,13 +194,15 @@ pub(crate) struct OutputOrderBuilder<'scope, 'data, P: Platform> {
     output_sections: &'scope OutputSections<'data, P>,
     secondary: &'scope OutputSectionMap<Vec<OutputSectionId>>,
     output_kind: OutputKind,
+    is_custom_phdrs: bool,
```
davidlattimore: I think the name would be clearer if we replaced "is" with "has"

libwild/src/expression_eval.rs
vishruth-thimmaiah: Fixed, the negation was unnecessary.

libwild/src/output_section_id.rs
vishruth-thimmaiah: Fixed, I've merged `output_order_from_linker_script` into `output_order`.

libwild/src/output_section_id.rs
vishruth-thimmaiah: Fixed

libwild/src/output_section_id.rs
vishruth-thimmaiah: Fixed.

libwild/src/output_section_id.rs
vishruth-thimmaiah: Fixed.

libwild/src/elf.rs
vishruth-thimmaiah: As we already have a `ExpectProgramHeader` check, would it make sense to add the flag checks here instead, along with a `sections` property, which takes a list of sections, to check which sections come under which segment? That would allow us to verify both the segment flags and that each section is placed in the correct program header segment.

davidlattimore: So you mean something like `ExpectProgramHeader:LOAD flags=RX sections=.text`?

That could work. If there were sections in the segment that weren't listed, would that be an error?

vishruth-thimmaiah: > So you mean something like `ExpectProgramHeader:LOAD flags=RX sections=.text`?

Yes.

> That could work. If there were sections in the segment that weren't listed, would that be an error?

Perhaps we can support adding wildcards in the section list, to make this toggleable?

vishruth-thimmaiah: Also, since this PR is already quite big, would you prefer having this implemented in another PR?

davidlattimore: Either is fine, but I'm somewhat inclined to say that we'd want to land the PR with the extended testing support first, that way we can more thoroughly verify the correctness of this PR before we merge it. But I don't feel strongly.

vishruth-thimmaiah: I've opened #2017 implementing this, once that is merged the tests here should also pass.

davidlattimore (approved):

PR #2014, opened 2026-06-03
test: On FreeBSD, skip tests where GCC is the driver.

daniel-levin:
I'm back on my mission to get Wild working on non-Linux Unix and Unix-like operating systems. Unfortunately, the GCC that one can install on FreeBSD refuses to delegate to another linker. Indeed, it also resists the `-B` symlink trick. This change skips the tests that are consequently non-runnable on FreeBSD. On a positive note, there are far fewer non-passing tests than the last time I hacked on Wild (in November some time).

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -593,7 +593,13 @@ impl Architecture {
 
     fn default_target_triple(&self, platform: PlatformKind) -> String {
         match platform {
-            PlatformKind::Elf => format!("{}-unknown-linux-gnu", self.triple_arch()),
+            PlatformKind::Elf => {
+                if cfg!(target_os = "freebsd") {
+                    format!("{}-unknown-freebsd", self.triple_arch())
+                } else {
+                    format!("{}-unknown-linux-gnu", self.triple_arch())
+                }
+            }
             PlatformKind::MachO => format!("{}-apple-darwin", self.darwin_arch_name()),
         }
     }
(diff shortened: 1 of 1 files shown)


mati865:
How about using `-fuse-ld=lld` and adding `-B` to directory where Wild is symlinked as `ld.lld`. For example fakes-debug from the checkout?

daniel-levin:
@mati865 Good suggestion. I tried it and unfortunately, no luck.

```
❯ gcc -fuse-ld=lld -B /workplace/wild/fakes-debug/ prog.c && readelf -p .comment a.out

String dump of section '.comment':
[     0] FreeBSD clang version 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd708029e0b2)
[    67] GCC: (FreeBSD Ports Collection) 14.2.0
```

I DTrace'd the system in another tmux pane to double check. I also tried making additional fakes, with names like 'ld.bfd'. Still no luck.

```
❯ cat script.d
#!/usr/sbin/dtrace

syscall:freebsd:execve:entry {
        printf("pid=%d user=%s cmd=%s\n", pid, execname, copyinstr(arg0));
}
```

```
  4  83144                     execve:entry pid=21259 user=gcc14 cmd=/usr/local/libexec/gcc14/gcc/x86_64-portbld-freebsd15.0/14.2.0/cc1

  7  83144                     execve:entry pid=21260 user=gcc14 cmd=/usr/local/bin/as

  5  83144                     execve:entry pid=21261 user=gcc14 cmd=/usr/local/libexec/gcc14/gcc/x86_64-portbld-freebsd15.0/14.2.0/collect2

  0  83144                     execve:entry pid=21262 user=collect2 cmd=/usr/local/bin/ld
```

daniel-levin:
Force pushed an updated message which should make it a bit clearer that GCC on FreeBSD can only drive linkers it knows about, at fixed locations.

mati865 (approved):
Oh, that's unfortunate, but also beyond our reach.

davidlattimore (approved):

PR #2015, opened 2026-06-04
linker-diff: Don't use global mutable state for colour output

davidlattimore:


marxin:
Can you please explain the motivation behind the change? Is it causing a contention in the integration tests?

davidlattimore:
Sure. I'd like to add some tests for linker-diff. Those tests would be snapshot tests that take the output of a diff report and compare it with the expected output stored in a file. But for those tests to work reliably, we don't want the output to depend on whether stdout is a tty - we just want colour to be off. Having global mutable state that affects the output is problematic for those tests. Especially when other tests might be running concurrently in other threads and setting the colour mode to always.

davidlattimore:
But I think it's also just nicer to not have a library changing global mutable state that might also be used by other things. See comment on `Report::from_config` (the comment that is now deleted).

marxin:
Ah, I see, makes sense.

PR #2017, opened 2026-06-04
tests: add support for defining flags and sections in a phdr

vishruth-thimmaiah:
This PR adds support for verifying a phdr's flags and the sections included in a segment.

Enables the `serde` feature of the bitflags crate.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3896,10 +3896,10 @@ impl Assertions {
                         continue;
                     }
 
-                    if let Some(expected_flags) = flags {
-                        if header.p_flags(endian).0 != expected_flags.bits() {
-                            continue;
-                        }
+                    if let Some(expected_flags) = flags
+                        && header.p_flags(endian).0 != expected_flags.bits()
+                    {
+                        continue;
                     }
 
                     if !expected_sections.is_empty() {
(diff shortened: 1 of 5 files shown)


vishruth-thimmaiah:
CI failures are unrelated to the changes made.

wild/tests/integration_tests.rs
```diff
@@ -3790,15 +3876,68 @@ impl Assertions {
         }
 
         let endian = obj.endian();
-        let mut header_types = HashSet::new();
 
-        for header in obj.elf_program_headers() {
+        let headers = obj.elf_program_headers();
+        let mut header_sections = Vec::with_capacity(headers.len());
+        let mut header_types = HashSet::new();
```
davidlattimore, line 3882: It looks like `header_types` is no longer used, so should probably be removed.

wild/tests/integration_tests.rs
```diff
@@ -3790,15 +3876,68 @@ impl Assertions {
...
+            header_sections.push(self.get_sections_in_segment(obj, header));
             header_types.insert(header.p_type(endian).0);
         }
 
-        for header in &self.expected_program_headers {
-            if !header_types.contains(&(*header as u32)) {
-                bail!("Expected program header `{header}' not found.");
+        for expected in &self.expected_program_headers {
+            let mut found = false;
+            for (i, header) in headers.iter().enumerate() {
```
davidlattimore: If `i` is just used in order to look up a corresponding entry in `header_sections`, then perhaps you could just use `zip` intead?

davidlattimore (commented):
It's possible that at some point, we might need to deal with linkers that emit multiple program segments with the same permissions. But if we don't need that yet, then it's fine to leave it for later.

wild/tests/integration_tests.rs
vishruth-thimmaiah: `header_types` are still used to track `absent_program_headers`, at line 3944.

wild/tests/integration_tests.rs
vishruth-thimmaiah: Done.

davidlattimore (approved):

PR #2018, opened 2026-06-05
test: Add some snapshot tests for linker-diff

davidlattimore:


marxin:
@davidlattimore have you consider using the `insta` crate for the actual comparison against the existing snapshots? I've had some experience with that crate and enjoyed the diffs and ability to quickly update the expected snapshots once an expected change lands.

davidlattimore:
I did look briefly at insta when I started this, but it seems to be aimed at rust tests where each snapshot file corresponds to a different macro invocation in the test. Our scenario is somewhat different in that our tests are generated dynamically at runtime by finding files on the filesystem. It's not obvious to me how we'd hook into `insta`. Even if we could, it's not clear what we'd get from doing so. The code to update the expected outputs is only a few lines. It sounds like they've got some neat tools for viewing the diffs of what changed, but git/jj are also really good at viewing diffs of what changed. Updating expected outputs is currently just a matter of running the tests with `WILD_SNAPSHOT=update`. I can definitely see the benefit of `insta` if you've got lots of `#[test]`-style rust tests with assertions in them that are subject to change.

marxin (approved):
Gotcha, thanks for the explanation.

mati865 (approved):
> I did look briefly at insta when I started this, but it seems to be aimed at rust tests where each snapshot file corresponds to a different macro invocation in the test. Our scenario is somewhat different in that our tests are generated dynamically at runtime by finding files on the filesystem. It's not obvious to me how we'd hook into `insta`. Even if we could, it's not clear what we'd get from doing so. The code to update the expected outputs is only a few lines. It sounds like they've got some neat tools for viewing the diffs of what changed, but git/jj are also really good at viewing diffs of what changed. Updating expected outputs is currently just a matter of running the tests with `WILD_SNAPSHOT=update`. I can definitely see the benefit of `insta` if you've got lots of `#[test]`-style rust tests with assertions in them that are subject to change.

This would be more like a case for https://docs.rs/trycmd/latest/trycmd/ but that would require some changes to how we test. Given the cost of creating the snapshotting infra is already paid (this PR) and it integrates nicely with existing tests, current approach looks good.

> git/jj are also really good at viewing diffs of what changed

I'm not sure about git. Last time I used it without any third-party tools for viewing diffs of snapshots it wasn't that good. On the other hand, jj is really nice.
Perhaps git experience can be improved with highlighters like https://github.com/dandavison/delta but that's a matter of personal preference.

marxin:
> Perhaps git experience can be improved with highlighters like https://github.com/dandavison/delta but that's a matter of personal preference.

Oh, what an interesting crate - have you been using it on a daily basis @mati865?

mati865:
> Oh, what an interesting crate - have you been using it on a daily basis @mati865?

I had used to, but like two years ago I've made the switch to jj. Since then, I didn't feel the need to use tools like that.

I found delta diffs better for reading but annoying for copy pasting. Arguably I should have made an alias to easily disable it, but it wasn't annoying enough to bother.

PR #2019, opened 2026-06-05
ci: Don't install SDE in external tests workflow

davidlattimore:


davidlattimore:
CI has been failing today because the download of SDE has been failing. It seems that if we just don't install SDE, that the tests pass, so perhaps we no longer need it? If we do need it, then we should probably find a more reliable way to obtain the binary. One option might be to check it in to our https://github.com/wild-linker/test-files repo.

davidlattimore:
I'm going to merge this now just to get our CI working again, but feel free to comment post-merge

.github/workflows/external_tests.yml
```diff
@@ -26,8 +26,7 @@ jobs:
           apt-get update && apt-get -y install gcc g++ clang llvm lld curl bubblewrap git zstd qemu-user gdb dwarfdump perl wget xz-utils \
               binutils-aarch64-linux-gnu binutils-riscv64-linux-gnu binutils-loongarch64-linux-gnu \
               gcc-aarch64-linux-gnu gcc-riscv64-linux-gnu gcc-loongarch64-linux-gnu \
-              g++-aarch64-linux-gnu g++-riscv64-linux-gnu g++-loongarch64-linux-gnu \
-              && mkdir /sde && wget -O- -q https://downloadmirror.intel.com/859732/sde-external-9.58.0-2025-06-16-lin.tar.xz | tar -C /sde --strip-components=1 --xz -xf - && ln -s /sde/sde64 /usr/bin
```
lapla-cogito, line 30: I added this since SDE is required for `arch-x86_64-apx-*.sh` tests. This is based on [mold's setup](https://github.com/rui314/mold/blob/e9c3d1629fe92f178f6204295779c27fa40bdafc/.github/workflows/install-extras.sh#L74). Without this, APX-related tests will be [skipped](https://github.com/rui314/mold/blob/e9c3d1629fe92f178f6204295779c27fa40bdafc/test/arch-x86_64-apx-gottpoff.sh#L5).

> One option might be to check it in to our https://github.com/wild-linker/test-files repo.

It might be a good idea I think.

PR #2021, opened 2026-06-05
feat: always emit .riscv.attribute when using linker scripts

vishruth-thimmaiah:
Both ld and lld always emit a `RISCV_ATTRIBUTE` segment along with a `.riscv.attribute` section, regardless of whether or not it is defined in the linker script.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3987,7 +3987,7 @@ impl Assertions {
                 _ => section.flags() != object::SectionFlags::None,
             };
 
-            if !is_alloc && header.p_type(endian).0 != object::elf::PT_RISCV_ATTRIBUTES.0 {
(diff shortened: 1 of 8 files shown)


wild/tests/integration_tests.rs
```diff
@@ -3979,12 +3980,14 @@ impl Assertions {
                 _ => section.flags() != object::SectionFlags::None,
             };
 
-            if !is_alloc {
+            if !is_alloc && header.p_type(endian).0 != object::elf::PT_RISCV_ATTRIBUTES.0 {
```
davidlattimore: It looks like we could remove the two `.0` in the line and directly compare the outer values

vishruth-thimmaiah: Fixed.

davidlattimore (approved):

davidlattimore:
Just merged #2029. That should fix the tests. Can you sync to head and reupload?

davidlattimore (approved):

PR #2022, opened 2026-06-05
feat: Support `--rosegment` and `--no-rosegment`

### LINKED ISSUE #1993: Support `--[no-]rosegment`
# What
It would be nice if `wild` could support the following options (docs taken from `lld.ld`):

```text
--no-rosegment          Do not put read-only non-executable sections in their own segment   
--rosegment             Put read-only non-executable sections in their own segment (default)
```

# Why
This option fixes some issues with `perf`, see https://issues.chromium.org/issues/41434218 . Note that my use case is NOT for Android or Chrome (or any Google project), but a Rust + C project.

# Alternatives
One could also just swallow the option for now. This won't create the "correct" binary, but at least this allows `wild` to be used in compilation pipelines that set this option.

lapla-cogito:
`--rosegment` places read-only and non-executable sections into the `R` segment, while `--no-rosegment` merges them into executable segments.

Closes #1993

### DIFF
--- a/libwild/src/platform.rs
+++ b/libwild/src/platform.rs
@@ -1141,6 +1141,7 @@ pub(crate) trait ProgramSegmentDef: Copy + Send + Sync + Display + 'static {
         self,
         section_info: &crate::output_section_id::SectionOutputInfo<Self::Platform>,
         section_id: OutputSectionId,
+        rosegment: bool,
     ) -> bool;
 
     /// Returns whether the current RW segment should end when this segment ends.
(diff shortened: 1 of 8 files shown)


libwild/src/platform.rs
```diff
@@ -1158,16 +1141,13 @@ pub(crate) trait ProgramSegmentDef: Copy + Send + Sync + Display + 'static {
         self,
         section_info: &crate::output_section_id::SectionOutputInfo<Self::Platform>,
         section_id: OutputSectionId,
+        rosegment: bool,
```
davidlattimore, line 1144: If we add any more arg-derived parameters, then I'd say we'd want to gather them up into a struct. But it's OK to wait until that happens.

davidlattimore (approved):

PR #2027, opened 2026-06-06
fix: Don't bail when a linker plugin doesn't claim an IR archive member

lapla-cogito:
Archive members are only selected if their symbols are referenced. If the active LTO plugin refuses to claim one, we used to error unconditionally. Skip it instead and let later symbol resolution fail if the member turns out to be needed.

### DIFF
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -1043,14 +1043,18 @@ impl<'data, P: Platform> LoadedInputs<'data, P> {
                     file,
                     kind,
                 } = *obj;
-                let result = plugin.as_mut()
+                let plugin_result = plugin.as_mut()
                     .with_context(|| {
                         format!(
                             "Input file {input_ref} contains {kind}, but linker plugin was not supplied"
                         )
                     })
                     .and_then(|plugin| plugin.process_input(input_ref, &file, kind));
-                self.lto_objects.push(result);
+                match plugin_result {
+                    Ok(Some(info)) => self.lto_objects.push(Ok(info)),
+                    Ok(None) => {} // Skipped, e.g. unclaimed IR member inside an archive
+                    Err(e) => self.lto_objects.push(Err(e)),
+                }
             }
         }
     }
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -174,20 +174,23 @@ impl<'data> LinkerPlugin<'data> {
         input_ref: InputRef<'data>,
         file: &File,
         kind: FileKind,
-    ) -> Result<Box<LtoInputInfo<'data>>> {
+    ) -> Result<Option<Box<LtoInputInfo<'data>>>> {
         verbose_timing_phase!("Linker plugin process input");
 
         let fd = file.as_raw_fd();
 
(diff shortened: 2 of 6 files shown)


lapla-cogito:
Adding a regression test for this is a bit tricky because we need to use different compilers (clang and gcc) for archive and other builds 🤔 
While it's possible to commit an object file compiled with gcc, this would be somewhat hacky. A more ideal approach would be to extend test directives to support entries like `//#Archive:unclaimed.c:gcc:-flto -fno-fat-lto-objects`. However, implementing this should extend support to other similar directives rather than just limiting it to `Archive`, suggesting it would require significant refactoring.

davidlattimore:
Does it work to use gcc as the linker driver, clang as the compiler, then specify -flto only for the one archive?

davidlattimore (commented):
Does this mean that now we'll now, in general, report an undefined symbol error instead of reporting that the archive contains the wrong kind of IR?

lapla-cogito:
> Does it work to use gcc as the linker driver, clang as the compiler, then specify -flto only for the one archive?

Ah right. There was a simple solution.

> Does this mean that now we'll now, in general, report an undefined symbol error instead of reporting that the archive contains the wrong kind of IR?

While this isn't true in general (it's limited to the archive case), yes. I think users naturally expect an "undefined symbol" error if a symbol was meant to be provided via an archive but the IR member wasn't claimed by the plugin.

davidlattimore:
> I think users naturally expect an "undefined symbol" error if a symbol was meant to be provided via an archive but the IR member wasn't claimed by the plugin.

I'm not sure what users expect, but an error telling them that they're trying to link LLVM IR using the GCC plugin would almost certainly be more useful than an undefined symbol error. We have tests for that, but I guess the tests only cover the case of an object, not an archive. I just tried adding the following test:

```
//#Config:clang-link-gcc-archive:error
//#Compiler:clang
//#CompArgs:-flto
//#LinkerDriver:gcc
//#SkipLinker:ld
//#LinkArgs:-Wl,-znow -flto -nostdlib
//#Object:runtime.c
//#Archive:linker-plugin-lto-2.c
//#ExpectError:(contains LLVM-IR, but the linker plugin|Wild was compiled without linker-plugin support)
```

Strangely it passes even with your PR. Perhaps you'd like to add it to your PR to ensure we don't regress that case. Although I'd be interested to learn why it passes.

lapla-cogito:
> I'd be interested to learn why it passes

I think it's because `//#CompArgs:-flto` makes  `runtime.c` an LLVM IR input as well. Direct (non-archive) IR files still produce the "... contains LLVM IR, but the linker plugin ..." error, so the test ends up passing.

lapla-cogito:
> I'm not sure what users expect, but an error telling them that they're trying to link LLVM IR using the GCC plugin would almost certainly be more useful than an undefined symbol error.

I generally (edit: I mean the general case other than archives) agree, but in the case of archives, the situation differs because archived components may contain members with different IR that aren't referenced. In fact, GNU ld and lld exhibit the same behavior (they pass the mold tests that no longer need to be skipped in this PR).

davidlattimore (approved):
That makes sense

PR #2030, opened 2026-06-06
feat: implement support for the SIZEOF_HEADERS builtin function

vishruth-thimmaiah:
This PR adds support for the `SIZEOF_HEADERS` builtin, which is used to get the combined size of the elf header and program headers.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -34,6 +34,7 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
     section_layouts: &OutputSectionMap<OutputRecordLayout>,
     output_sections: &OutputSections<'data, P>,
     resolutions: &[Option<Resolution<P>>],
+    sizeof_headers: u64,
 ) -> Result {
     for group in &symbol_db.groups {
         let Group::LinkerScripts(scripts) = group else {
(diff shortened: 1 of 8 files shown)


libwild/src/linker_script.rs
```diff
@@ -894,8 +896,12 @@ fn parse_identifier_or_function<'a>(input: &mut &'a BStr) -> winnow::Result<Expr
             _ => Err(ContextError::default()),
         }
     } else {
-        // It's a symbol
-        Ok(Expression::Symbol(ident))
+        if ident == b"SIZEOF_HEADERS" {
```
davidlattimore: Can this be collapsed as an else-if on the line above?

libwild/src/expression_eval.rs
```diff
@@ -34,6 +34,7 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
     section_layouts: &OutputSectionMap<OutputRecordLayout>,
     output_sections: &OutputSections<'data, P>,
     resolutions: &[Option<Resolution<P>>],
+    sizeof_headers: u64,
```
davidlattimore, line 37: I feel like at some point soon, we might want to gather up the input data needed to evaluate expressions into a struct. Probably makes sense to do it separately though.

libwild/src/linker_script.rs
vishruth-thimmaiah: Fixed.

davidlattimore (approved):

PR #2031, opened 2026-06-06
port(MachO): support for linking against stub libraries

marxin:
This PR is a nice milestone: we can now properly link against stub libraries such as `libSystem` and `libc++`. As a result, a typical “Hello, world!” program can be linked with Wild and run successfully.

```
❯ cat hw.c
#include <stdio.h>

int main() {
  printf("hello world\n");
  return 0;
}
❯ clang hw.c -fuse-ld=/users/apple/Programming/wild/fakes-debug/ld64.ld && ./a.out
wild: warning: Fat object file is not supported yet: /opt/homebrew/Cellar/llvm/22.1.3/lib/clang/22/lib/darwin/libclang_rt.osx.a
hello world
```

The change adds support for emitting `__got`, `__stubs` - the Mach-O equivalent of ELF’s `.plt` section - and the chained fixups section.

Known limitations:
- currently only a single stub library supported - the load command allocation needs to be reworked so that we can allocate multiple `LC_LOAD_DYLIB` commands (one for each library)
- linking against classical `.dylib` not supported yet
- the allocation and emission of `CHAINED_FIXUPS` is pretty complex: depends on number of active segments, number of symbols from libraries and their names -> that's why the streaming happens in Epilogue and why I introduced the `ImportedSymbol` type - maybe we can come up with a better approach @davidlattimore
- the implicitly used archive (`libclang_rt.osx.a`) is skipped right now as we don't support FAT objects yet

Anyway, once the PR gets merged, I am planning to spend some time on refactoring and filling up of MachO-related issues. I feel it's a good time for others to join the initiative as we'll have the foundation established.

Issue: #757

### DIFF
--- a/libwild/src/args/macho.rs
+++ b/libwild/src/args/macho.rs
@@ -297,6 +297,14 @@ fn setup_argument_parser() -> ArgumentParser<MachOArgs> {
             args.common_mut().file_write_mode = Some(FileWriteMode::UpdateInPlace);
             Ok(())
         });
+    parser
+        .declare_with_param()
+        .long("sym-info")
+        .help("Show symbol information. Accepts symbol name or ID.")
+        .execute(|args, _modifier_stack, value| {
+            args.common_mut().sym_info = Some(value.to_owned());
+            Ok(())
+        });
 
     super::declare_common_args(&mut parser);
 
--- a/libwild/src/args/macho.rs
+++ b/libwild/src/args/macho.rs
@@ -220,6 +220,15 @@ fn setup_argument_parser() -> ArgumentParser<MachOArgs> {
         });
     parser
         .declare_with_param()
+        .prefix("L")
+        .help("Add directory to library search path")
+        .execute(|args, _modifier_stack, value| {
+            args.common_mut().save_dir.handle_file(value);
+            args.lib_search_path.push(Box::from(Path::new(value)));
+            Ok(())
+        });
+    parser
+        .declare_with_param()
         .prefix("l")
         .help("Link with library")
         .sub_option_with_value(
@@ -348,6 +357,9 @@ mod tests {
         "-enable-linkonceodr-outlining",
         "-o",
         "a.out",
+        "-L/foo/lib",
+        "-L",
+        "/bar/lib",
         "main.o",
         "-lc++",
     ];
@@ -371,6 +383,16 @@ mod tests {
             InputSpec::Lib(f) => f.as_ref() == "c++",
             InputSpec::File(_) | InputSpec::Search(_) => false,
         }));
+        assert!(
+            args.lib_search_path
+                .iter()
+                .any(|p| p.as_ref() == Path::new("/foo/lib"))
+        );
+        assert!(
+            args.lib_search_path
+                .iter()
+                .any(|p| p.as_ref() == Path::new("/bar/lib"))
+        );
         assert_eq!(args.plugin_path, Some("/foo/bar/libLTO.dylib".to_owned()));
     }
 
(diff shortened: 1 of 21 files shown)


libwild/src/args/macho.rs
```diff
@@ -297,6 +306,14 @@ fn setup_argument_parser() -> ArgumentParser<MachOArgs> {
...
             Ok(())
         });
+    parser
+        .declare_with_param()
+        .long("sym-info")
+        .help("Show symbol information. Accepts symbol name or ID.")
+        .execute(|args, _modifier_stack, value| {
+            args.common_mut().sym_info = Some(value.to_owned());
+            Ok(())
+        });
```
davidlattimore, line 316: Can this be moved into `declare_common_args` (called below)?

libwild/src/input_data.rs
```diff
@@ -739,10 +741,16 @@ impl<'data, P: Platform> TemporaryState<'data, P> {
         if input_ref.is_archive_entry() && kind != FileKind::ElfObject {
             bail!("Unexpected archive member of kind {kind:?}: {input_ref}");
         }
-        ensure!(
-            kind != FileKind::FatMachOObject,
-            "Fat object file is not supported yet: {input_ref}"
-        );
+        // ensure!(
```
davidlattimore: Are you likely to want to restore the error? If not, we could delete this commented-out code.

libwild/src/layout.rs
```diff
@@ -662,6 +677,8 @@ pub struct Layout<'data, P: Platform> {
     pub(crate) has_variant_pcs: bool,
     pub(crate) per_symbol_flags: PerSymbolFlags,
     pub(crate) dynamic_symbol_definitions: Vec<DynamicSymbolDefinition<'data, P>>,
+    /// A list of imported STUB library symbols (Mach-O specific).
```
davidlattimore: If `imported_symbols` is just for Mach-O, could it be added to `LayoutExt` instead? At some point we should perhaps rename `properties_and_attributes` (below) to `format_specific` for more consistency.

libwild/src/macho_writer.rs
```diff
@@ -205,32 +230,116 @@ fn write_prelude<'data, A: Arch<Platform = MachO>>(
...
+}
+
+#[derive(derive_more::Debug, Clone, Copy)]
+struct ImportedSymbolWithResolution<'data> {
+    symbol: ImportedSymbol<'data>,
+    got_address: NonZeroU64,
+    plt_address: Option<NonZeroU64>,
+}
+
+fn get_imported_sorted_by_got<'data>(
```
davidlattimore: This function gets called 3 times. Should we do it once during layout and store the result?

libwild/src/layout.rs
marxin: Yep, I'm going to move it to `LayoutExt`. Feel free to rename it.

libwild/src/layout.rs
```diff
@@ -189,6 +190,9 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
         &finalise_sizes_resources,
     )?;
 
+    let imported_symbols = collect_imported_symbols(&group_states);
```
davidlattimore, line 193: If you added `group_states` as an argument to the `P::create_layout_properties` call above, could it be computed there? If so, that would allow removal of `imported_symbols` from `FinaliseSizesResources` and would also remove the need for `set_imported_symbols` on `Platform`.

davidlattimore (approved):
I'll preemptively approve. No need to wait for another round of review on this.

libwild/src/layout.rs
marxin: Yes, I could be able to construct `Vec<ImportedSymbol> in `create_layout_properties`, but the function is called before the resolution is done and thus `ImportedSymbolWithResolution` instances can't be built. So I guess not so easy to adjust?

PR #2032, opened 2026-06-06
chore(clippy): allow pendantic with exceptions

marxin:
When experimenting with clippy, I noticed the vast majority of `pedantic` lints are reasonable to me and I would like to enable them. Thoughts on that?


davidlattimore (approved):

davidlattimore:
Look reasonable to me. We can always add more exclusions if any of the rules turn out to be annoying.

PR #2037, opened 2026-06-08
port(wasm): Build initial output module layout

lapla-cogito:
Add inert implementations for Wasm layout hooks that don't map to the current output model, and introduce `WasmLayout` as the layout state for output module data.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -1378,6 +1378,23 @@ impl<'data> platform::RelocationList<'data> for RelocationList<'data> {
     }
 }
 
+#[derive(Debug, Default)]
+pub(crate) struct WasmLayout<'data> {
+    pub(crate) output_types: Vec<wasmparser::FuncType>,
+    pub(crate) imports: Vec<crate::wasm_writer::OutputImport<'data>>,
+    pub(crate) function_type_indices: Vec<u32>,
+    pub(crate) globals: Vec<crate::wasm_writer::OutputGlobal<'data>>,
+    pub(crate) exports: Vec<crate::wasm_writer::OutputExport<'data>>,
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct WasmObjectLayout<'data> {
+    pub(crate) type_index_base: u32,
+    pub(crate) function_index_base: u32,
+    pub(crate) global_index_base: u32,
+    _phantom: std::marker::PhantomData<&'data ()>,
+}
+
 impl platform::Platform for Wasm {
     type File<'data> = File<'data>;
     type FileFlags = u32;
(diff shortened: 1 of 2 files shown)


lapla-cogito:
Therefore, the necessary next step (unsure it will be done in a single PR) is to call `wasm_writer::build_*_section` from `WasmLayout` to generate section bytes, update the layout with their encoded sizes, and then split and write the output buffer by section/group.

libwild/src/wasm.rs
```diff
@@ -1378,6 +1382,174 @@ impl<'data> platform::RelocationList<'data> for RelocationList<'data> {
...
+) -> Result<WasmLayout<'data>>
+where
+    'data: 'files,
+{
+    let objects = objects.collect::<Vec<_>>();
+    let mut layout = WasmLayout::default();
+    let mut next_function_import_index = 0u32;
+    let mut next_global_import_index = 0u32;
+
+    for object in &objects {
```
davidlattimore: It looks like everything is being done from the main thread. Is it possible to do layout the same way as we do with ELF? i.e. have a sizing phase where the input objects figure out (in parallel) how much space they need in different sections, then allocate base addresses to each object from a single thread (doesn't take long), then figure out everything else relative to those base addresses (in parallel).

lapla-cogito: It matches what I was about to do next (see the comment above).

davidlattimore: So the plan is to do layout in parallel? I got the impression from your comment above that you were mostly referring to the write phase?

lapla-cogito: > So the plan is to do layout in parallel?

Originally that was the plan, but as of now not all parts of the pipeline are complete, so I had planned to postpone it. However, I realized it would be easier to follow the ELF patterns, so I've decided to handle part of this parallelization in this PR, thanks.

davidlattimore (commented):
This feels very separate to the pre-existing linker implementation. That will likely make it hard to reuse things like the existing GC algorithm. Are there plans to try to unify things a bit more?

davidlattimore (approved):

lapla-cogito:
> This feels very separate to the pre-existing linker implementation. That will likely make it hard to reuse things like the existing GC algorithm. Are there plans to try to unify things a bit more?

Note about the above comment:

David and I discussed the matter and agreed that adapting the existing ELF GC pipeline for Wasm would require some refactoring. Wasm sections are fixed-order containers; the actual link/GC units consist of functions, globals, data segments, and element segments. If we force this structure into the ELF input-section model, the GC would either retain excessive data (e.g., the entire code section) or necessitate numerous special cases. Therefore, when integrating the GC pipeline (which I expect to occur after basic linking functionality is complete), I plan to implement a future refactoring that creates an enum to represent GC units and switches between them using a trait.

PR #2040, opened 2026-06-08
docs: clarify wild output imposes no licensing requirements

### LINKED ISSUE #1856: Attribution when distributing binaries linked with wild?
The MIT license and Apache 2.0 license require attribution. If I link a C executable with wild, do I have to include the wild license files when distributing the binaries? (if the C standard libraries are not included)

The LLVM linker uses the Apache 2.0 license with the [LLVM exceptions](https://releases.llvm.org/14.0.0/LICENSE.TXT), which says that attribution isn't needed:

> As an exception, if, as a result of your compiling your source code, portions
> of this Software are embedded into an Object form of such source code, you
> may redistribute such embedded portions in such Object form without complying
> with the conditions of Sections 4(a), 4(b) and 4(d) of the License.

The wild licenses don't have such exceptions.

mvanhorn:
## Summary

Adds a short clarification to the README `## License` section stating that wild's dual MIT/Apache-2.0 license applies to wild's own source and places no licensing or attribution requirements on the binaries you link with it.

## Why this matters

A user asked whether distributing a binary linked with wild requires shipping wild's license files, noting LLD uses the Apache-2.0 LLVM exception that waives attribution for embedded portions (#1856). @davidlattimore clarified that wild does not embed its own copyrighted material into output (it emits only short, psABI-defined PLT stubs) and that relicensing the whole repo with an LLVM-style exception is impractical, but proposed documenting the situation in the README instead. This adds that paragraph in the maintainer's stated wording: the output has no licensing requirements imposed by the linker, and any future embedded content would be very permissively licensed.

The license terms themselves are unchanged; this is a clarification, not legal advice.

## Testing

Documentation only; no code paths affected.

Closes #1856

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,12 @@ queries, please email wild-mod@googlegroups.com.
 Licensed under either of [Apache License, Version 2.0](LICENSE-APACHE) or [MIT license](LICENSE-MIT)
 at your option.
 
+For clarity, this is a documentation clarification and not legal advice: the licensing above applies
+to Wild's own source code and does not place any licensing or attribution requirements on binaries
+you produce with Wild. Wild does not embed its own copyrighted material in output files; it emits
+only short PLT stubs defined by the platform psABI. If Wild ever embeds substantial material in
+output files, that material will be under a very permissive license.
+
 Unless you explicitly state otherwise, any contribution intentionally submitted for inclusion in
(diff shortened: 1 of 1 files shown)


davidlattimore:
Thanks, but we haven't entirely decided how we want to handle this. Also, please see `CONTRIBUTING.md`. In particular, the bit about communication using AI.

Issue #2028, opened 2026-06-06
Use custom containers for CI


davidlattimore:
I've been meaning to do this for a while. Currently we use stock os containers then install required packages. This is simple but has some downsides. Firstly, it's unreliable. Outages of whatever we're downloading packages from cause CI failures as happened most of yesterday with openSUSE. It also puts load on those package repository servers, which seems a bit antisocial. It may also be slower than it needs to be. 

My plan is to have a workflow or workflows for building the containers. Then another workflow to create a PR to update to the new version of the containers. That way we can test the container before we commit to using it.

marxin:
Really welcome the initiative! By having that we'll get super-fast CI runs, which is a nice prospect.

davidlattimore:
Based on my experiments, I don't think we'll get super-fast CI runs. We no longer need to apt-get install packages (or equivalent for other distros) and we also no longer need to use a github action to install rust. However, the docker image is now much larger, so takes considerably longer to start. From what I can see, I think most of the jobs that I've converted end up slightly faster than before, but not much. I haven't touched the Nix stuff and usually it's the slowest.

mati865:
Nice, that could make adding Arch Linux job feasible. Currently the tests are stable but every now and then the packages see a bigger update, and the tests break. Having pinned image would avoid breaking CI in unrealated PRs.

davidlattimore:
Completed in https://github.com/wild-linker/wild/pull/2049

PR #2046, opened 2026-06-10
ci: Build docker images for use by CI

davidlattimore:


davidlattimore:
Issue #2028

mati865:
1. Should we do a sanity check that Wild still passes the tests before publishing new image? Rolling releases like OpenSUSE Tumbleweed can perform "breaking" changes on typical upgrade. For example bumping GCC or Binutils version can cause our tests to start failing.

2. Is it worth building each merged commit? I guess the weekly builds would be sufficient.

davidlattimore:
> Should we do a sanity check that Wild still passes the tests before publishing new image? Rolling releases like OpenSUSE Tumbleweed can perform "breaking" changes on typical upgrade. For example bumping GCC or Binutils version can cause our tests to start failing.

My current plan is that once the build completes, we wouldn't be using the new images straight away, since CI would be pinned to a specific version of the images. We'll then have a PR that will update to the new images. That PR would only be merged if CI passed (with the new images).

> Is it worth building each merged commit? I guess the weekly builds would be sufficient.

Do you mean change the trigger so that runs only weekly instead of every time one of the dockerfiles gets changed?

davidlattimore:
https://github.com/wild-linker/wild/pull/2049 is roughly what I had in mind for the PR to use the images

mati865:
> We'll then have a PR that will update to the new images. That PR would only be merged if CI passed (with the new images).

Do you mean bot/action that creates update PR or manual one? Sounds fine either way.

> Do you mean change the trigger so that runs only weekly instead of every time one of the dockerfiles gets changed?

Ah, there is `path` filter in the job. So weekly or biweekly trigger would be a good addition to that IMO. Otherwise you can end up with a really old Tumbleweed image that passes the tests, but on real (up-to-date) Tumbleweed the tests might be failing a long time.

davidlattimore:
> Do you mean bot/action that creates update PR or manual one? Sounds fine either way.

At the moment what I'm looking at is having the action push a change to a branch. We'd then need to create a PR from that branch. Actually creating a PR seems like it would be more complicated, but could probably be done.

> So weekly or biweekly trigger would be a good addition to that IMO. Otherwise you can end up with a really old Tumbleweed image that passes the tests, but on real (up-to-date) Tumbleweed the tests might be failing a long time.

I imagine we'll want to update the images from time to time anyway, e.g. if we want to use a new rust version. I'll add a manual trigger for the build workflow. I'd probably only want a cron trigger if we were automatically creating the PR to update, otherwise we'd likely end up with a bunch of automatically built packages that we never transitioned to, which could be bad when combined with a cron job that automatically deletes old images.

mati865:
Looks like there are handful solutions for cron based PR creation: https://stackoverflow.com/questions/68057744/create-pull-request-with-github-action

So, adding logic there shouldn't prove too hard, I suppose. Anyway this is not something for the first iteration.

davidlattimore:
I changed the trigger to manual trigger only. At least until such time as we have PR creation automated, I figured it best not to create new images when the files are changed. I intend to also have a job that automatically deletes all but the few most recent image versions.

I also added a prefix "sha-" to the version strings. This is mostly to make automatic updates easier, since we can search for a regex like ":sha-[0-9a-f]+$" and be sure that we don't match anything else in our CI workflow.

davidlattimore:
> So, adding logic there shouldn't prove too hard, I suppose. Anyway this is not something for the first iteration.

I had read that github wouldn't allow PRs to be created from a workflow unless you had an appropriate token set up. I got the impression that this was to avoid workflows creating PRs that trigger workflows and that going in an infinite loop. I'm not sure why having a token would prevent that. I guess we can try it and see if it works.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,115 @@
...
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        include:
+          - file: docker/ci/ubuntu-base.Dockerfile
+            container: ci-ubuntu-base-amd64
+            platform: linux/amd64
+          - file: docker/ci/ubuntu-base.Dockerfile
+            container: ci-ubuntu-base-arm64
+            platform: linux/arm64
```
lapla-cogito, line 26: Is there any reason to build an arm64 image with the x86_64 runner?

docker/ci/alpine.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM rust:1.94-alpine
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN apk add build-base lld clang clang-extra-tools bash curl tar \
```
lapla-cogito:
```suggestion
RUN apk add --no-cache build-base lld clang clang-extra-tools bash curl tar \
```

docker/ci/opensuse.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM opensuse/tumbleweed:latest
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN zypper refresh \
+    && zypper in -y gcc gcc-c++ glibc-devel-static clang lld curl rustup \
```
lapla-cogito, line 8: Shouldn't `zypper clean --all` be added here too, similar to Ubuntu?

lapla-cogito:
Sorry for the late review

docker/ci/alpine.Dockerfile
davidlattimore: Done

docker/ci/opensuse.Dockerfile
davidlattimore: Done

.github/workflows/build-images.yml
davidlattimore: Good catch! I hadn't realised I was doing that. Fixed

davidlattimore (commented):
Thanks for the thorough review!

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,122 @@
...
+          tags: |
+            ghcr.io/${{ github.repository }}/${{ matrix.container }}:latest
+            ghcr.io/${{ github.repository }}/${{ matrix.container }}:sha-${{ github.sha }}
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/${{ matrix.container }}:buildcache
+          cache-to: type=registry,ref=ghcr.io/${{ github.repository }}/${{ matrix.container }}:buildcache,mode=max
+
+  # Build images that depend on other images.
+  build-dependent:
+    name: ${{ matrix.container }}
+    runs-on: ubuntu-24.04
```
lapla-cogito: Same as https://github.com/wild-linker/wild/pull/2046#discussion_r3407579030

lapla-cogito (approved):
Thanks!

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,127 @@
...
+            base_image: ghcr.io/${{ github.repository }}/ci-ubuntu-base-amd64:sha-${{ github.sha }}
+          - file: docker/ci/ubuntu.Dockerfile
+            container: ci-ubuntu-arm64
+            runs-on: ubuntu-24.04-arm
+            platform: linux/arm64
+            base_image: ghcr.io/${{ github.repository }}/ci-ubuntu-base-arm64:sha-${{ github.sha }}
+          - file: docker/ci/ubuntu-cross.Dockerfile
+            container: ci-ubuntu-cross-amd64
+            platform: linux/amd64
+            runs-on: ubuntu-24.04-arm
```
mati865, line 92: @davidlattimore sorry for pinging in the old PR, but do you recall why this job runs amd64 container on aarch64 machine?
IIUC it will cause to run entire amd64 userspace via QEMU with `docker-container` driver, and fail with `docker` driver as shown by the errors in this WIP job: https://github.com/wild-linker/wild/actions/runs/30292425078/job/90065924045?pr=2277

Eatmydata or apt fails there with `0.096 exec /bin/sh: exec format error` because `docker` driver doesn't handle binfmt OOTB, so it cannot run amd64 binaries on aarch64 without explicit QEMU.

Was the intention really to test whether we can build amd64 image on aarch64 system? I suppose the goal was to build it natively on amd64.

davidlattimore: Definitely just a mistake on my part. I have trouble with how similar amd64 and arm64 look.

PR #2042, opened 2026-06-08
port(wasm): Build and emit metadata section

lapla-cogito:
Build the output module metadata during layout, encode the type, import, function, global and export sections, and account for their encoded sizes in the layout.

The writer now splits the output buffer by section, writes the Wasm preamble, and copies the pre-encoded metadata sections into their allocated slices. It still stops before code and data emission because those sections require function/data body layout and relocation application.

Part of #1431

PR #2047, opened 2026-06-10
port(wasm): Build and emit code and memory sections

lapla-cogito:
As with #2042, this patch enables building and emitting code and memory sections.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -11,9 +11,7 @@ use crate::layout::ImportedSymbol;
 use crate::layout_rules::SectionKind;
 use crate::output_section_id::SectionName;
 use crate::platform;
-use crate::wasm_writer::OutputDataSegment;
 use crate::wasm_writer::OutputExport;
-use crate::wasm_writer::OutputFunctionBody;
 use crate::wasm_writer::OutputGlobal;
 use crate::wasm_writer::OutputImport;
 use crate::wasm_writer::OutputImportEntity;
@@ -1438,9 +1436,8 @@ pub(crate) struct WasmLayout<'data> {
     pub(crate) function_type_indices: Vec<u32>,
     pub(crate) globals: Vec<OutputGlobal<'data>>,
     pub(crate) exports: Vec<OutputExport<'data>>,
-    pub(crate) function_bodies: Vec<OutputFunctionBody<'data>>,
+    pub(crate) function_bodies: Vec<WasmFunctionBody<'data>>,
     pub(crate) memories: Vec<MemoryType>,
-    pub(crate) data_segments: Vec<OutputDataSegment<'data>>,
     pub(crate) unsupported_output: Vec<&'static str>,
     pub(crate) object_index_maps: Vec<WasmObjectIndexMap>,
     pub(crate) encoded_sections: WasmEncodedSections,
@@ -1455,7 +1452,6 @@ pub(crate) struct WasmEncodedSections {
     pub(crate) export: Option<Vec<u8>>,
     pub(crate) code: Option<Vec<u8>>,
     pub(crate) memory: Option<Vec<u8>>,
-    pub(crate) data: Option<Vec<u8>>,
 }
 
 impl WasmEncodedSections {
@@ -1467,7 +1463,6 @@ impl WasmEncodedSections {
         add_encoded_section_size(sizes, crate::part_id::WASM_EXPORT, self.export.as_ref());
         add_encoded_section_size(sizes, crate::part_id::WASM_CODE, self.code.as_ref());
         add_encoded_section_size(sizes, crate::part_id::WASM_MEMORY, self.memory.as_ref());
-        add_encoded_section_size(sizes, crate::part_id::WASM_DATA, self.data.as_ref());
     }
 }
 
@@ -1520,14 +1515,11 @@ impl<'data> WasmLayout<'data> {
             self.encoded_sections.memory = Some(encode_wasm_section(&memory_section));
         }
 
-        let code_section = crate::wasm_writer::build_code_section(&self.function_bodies);
+        let code_section = crate::wasm_writer::build_code_section(
+            self.function_bodies.iter().map(|body| body.bytes),
+        );
         if !code_section.is_empty() {
             self.encoded_sections.code = Some(encode_wasm_section(&code_section));
-        }
-
-        let data_section = crate::wasm_writer::build_data_section(&self.data_segments)?;
-        if !data_section.is_empty() {
-            self.encoded_sections.data = Some(encode_wasm_section(&data_section));
         }
 
         Ok(())
@@ -1558,9 +1550,8 @@ struct WasmObjectLayoutInput<'data> {
     module_functions: Vec<WasmModuleFunction>,
     globals: Vec<OutputGlobal<'data>>,
     exports: Vec<OutputExport<'data>>,
-    function_bodies: Vec<OutputFunctionBody<'data>>,
+    function_bodies: Vec<WasmFunctionBody<'data>>,
     memories: Vec<MemoryType>,
-    data_segments: Vec<OutputDataSegment<'data>>,
     unsupported_output: Vec<&'static str>,
 }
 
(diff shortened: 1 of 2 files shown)


libwild/src/wasm_writer.rs
```diff
@@ -149,6 +168,42 @@ pub(crate) fn build_global_section(globals: &[OutputGlobal<'_>]) -> Result<Globa
...
+        section.memory(convert_memory_type(memory));
+    }
+    section
+}
+
+pub(crate) fn build_code_section(function_bodies: &[OutputFunctionBody<'_>]) -> CodeSection {
+    let mut section = CodeSection::new();
+    for body in function_bodies {
+        section.raw(body.bytes);
+    }
```
lapla-cogito, line 180: When actually parallelizing this, size calculations must account for relocations. Therefore, these kind of parallelization should be implemented during relocation handling (since attempting to parallelize it prematurely would likely result in needing to rewrite the implementation anyway).

libwild/src/wasm.rs
```diff
@@ -1390,6 +1438,10 @@ pub(crate) struct WasmLayout<'data> {
     pub(crate) function_type_indices: Vec<u32>,
     pub(crate) globals: Vec<OutputGlobal<'data>>,
     pub(crate) exports: Vec<OutputExport<'data>>,
+    pub(crate) function_bodies: Vec<OutputFunctionBody<'data>>,
+    pub(crate) memories: Vec<MemoryType>,
+    pub(crate) data_segments: Vec<OutputDataSegment<'data>>,
+    pub(crate) unsupported_output: Vec<&'static str>,
```
lapla-cogito, line 1441: Previously, we always bailed during the write phase, but now we only do so when encountering unsupported items like `relocations` or `data_count`. This change allows builds to succeed within the currently supported scope without generating errors.

lapla-cogito:
We can verify:

(code-only module)
```
（▰╹◡╹）❯  cat min.wat
(module
  (func (result i32) (i32.const 42))
  (export "f" (func 0)))

（▰╹◡╹）❯  wat2wasm min.wat -o min.wasm

（▰╹◡╹）❯  wasm-objdump -h min.wasm

min.wasm:       file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000f (size=0x00000005) count: 1
 Function start=0x00000011 end=0x00000013 (size=0x00000002) count: 1
   Export start=0x00000015 end=0x0000001a (size=0x00000005) count: 1
     Code start=0x0000001c end=0x00000022 (size=0x00000006) count: 1

（▰╹◡╹）❯  ~/repos/wild/target/debug/wild -flavor wasm -o min.wild.wasm min.wasm

（▰╹◡╹）❯  wasm-objdump -h min.wild.wasm

min.wild.wasm:  file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000f (size=0x00000005) count: 1
 Function start=0x00000011 end=0x00000013 (size=0x00000002) count: 1
   Export start=0x00000015 end=0x0000001a (size=0x00000005) count: 1
     Code start=0x0000001c end=0x00000022 (size=0x00000006) count: 1

（▰╹◡╹）❯  wasm2wat min.wild.wasm
(module
  (type (;0;) (func (result i32)))
  (func (;0;) (type 0) (result i32)
    i32.const 42)
  (export "f" (func 0)))
```

(memory module)
```
（▰╹◡╹）❯  cat min_mem.wat
(module
  (memory 1)
  (func (result i32) (i32.const 0))
  (export "memory" (memory 0))
  (export "f" (func 0)))

（▰╹◡╹）❯  wat2wasm min_mem.wat -o min_mem.wasm

（▰╹◡╹）❯  wasm-objdump -h min_mem.wasm

min_mem.wasm:   file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000f (size=0x00000005) count: 1
 Function start=0x00000011 end=0x00000013 (size=0x00000002) count: 1
   Memory start=0x00000015 end=0x00000018 (size=0x00000003) count: 1
   Export start=0x0000001a end=0x00000028 (size=0x0000000e) count: 2
     Code start=0x0000002a end=0x00000030 (size=0x00000006) count: 1

（▰╹◡╹）❯  ~/repos/wild/target/debug/wild -flavor wasm -o min_mem.wild.wasm min_mem.wasm

（▰╹◡╹）❯  wasm-objdump -h min_mem.wild.wasm

min_mem.wild.wasm:      file format wasm 0x1

Sections:

     Type start=0x0000000a end=0x0000000f (size=0x00000005) count: 1
 Function start=0x00000011 end=0x00000013 (size=0x00000002) count: 1
   Memory start=0x00000015 end=0x00000018 (size=0x00000003) count: 1
   Export start=0x0000001a end=0x00000028 (size=0x0000000e) count: 2
     Code start=0x0000002a end=0x00000030 (size=0x00000006) count: 1

（▰╹◡╹）❯  wasm2wat min_mem.wild.wasm
(module
  (type (;0;) (func (result i32)))
  (func (;0;) (type 0) (result i32)
    i32.const 0)
  (memory (;0;) 1)
  (export "memory" (memory 0))
  (export "f" (func 0)))
```

libwild/src/wasm.rs
```diff
@@ -1652,6 +1775,20 @@ impl<'data> WasmObjectLayoutInput<'data> {
             index_map.global_indices.push(output_global_index);
         }
 
+        for i in 0..self.memories.len() {
```
davidlattimore: It looks like this loop is just doing the equivalent of `index_map.extend(u32::try_from(index_bases.memory_base.index_bases.memory_base as u64 + self.memories.len() as u64).context("Wasm memory index overflow")?);`. Do we need the explicit loop? Also, do we need to do two separate overflow checks with two separate messages?

libwild/src/wasm_writer.rs
```diff
@@ -178,6 +233,17 @@ pub(crate) struct OutputGlobal<'a> {
     pub(crate) init_expr_body: &'a [u8],
 }
 
+#[derive(Debug, Copy, Clone)]
+pub(crate) struct OutputFunctionBody<'a> {
```
davidlattimore: Do we need a separate type for this when we also have `WasmFunctionBody`? Are they likely to end up different?

libwild/src/wasm.rs
```diff
@@ -1546,7 +1629,39 @@ impl<'data> WasmObjectLayoutInput<'data> {
...
+        }
+
         let module_functions = file.module_functions()?;
+        let function_bodies = file.function_bodies()?;
+        ensure!(
+            module_functions.len() == function_bodies.len(),
+            "Wasm function and code section counts differ"
+        );
+        let memories = file.memories()?;
+        let data_segments = file
```
davidlattimore: Why do we convert things from the input form into our own representation? Is it so that we can merge them into one large `data_segments` Vec that contains info from all the input objects? Wherever possible, I've tried to avoid copying data out of the input objects. The exceptions are when we're deriving something from that input data and it would be too expensive to rederive it later. I guess the follow-up question is why do we need the large `data_segments` Vec on `WasmLayout`?

libwild/src/wasm.rs
lapla-cogito: Since similar processing would be required when adding table indices, data segment indices, or element segment indices, I've extracted this as a generic helper function for reuse.

libwild/src/wasm_writer.rs
lapla-cogito: I kept this separate because it is intended to become the output-side body record once relocation-aware emission lands. At that point it will likely carry output offsets / rewritten bytes / relocations, while `WasmFunctionBody` remains a view of the input body. However this should be added in relocation implementation so I removed that.

libwild/src/wasm.rs
lapla-cogito: The reason for doing this is as you guessed but I started to consider that data emission should probably remain per-object until layout has assigned output ranges, especially once data relocations are applied. Therefore flattening here is too early and this PR doesn't support emitting data sections.

davidlattimore (approved):

PR #2050, opened 2026-06-10
Fix: Support R_X86_64_PC32 relocations to absolute addresses

### LINKED ISSUE #1420: Failed to apply relocation of type R_X86_64_PC32 to absolute
Assembling the following program with `GNU assembler (GNU Binutils) 2.45.1`

```asm
.text
.globl main
.type main, @function
main:
    callq 0
    ret
.size main, .-main

.text
.globl _start
.type _start, @function
_start:
    call main
    ud2
.size _start, .-_start

.section .note.GNU-stack,"",@progbits
```

and then attempting to link the object file results in an error:

```
wild: error: Failed copying from ./minasm.o (256 (1/0)) to output file
  Caused by:
    Failed to apply relocations in section `.text` of ./minasm.o
    Failed to apply relocation of type R_X86_64_PC32 to absolute at offset 0x1
    Unsupported absolute relocation
```

`lld`, `ld`, `mold` all seem to be OK with this relocation (no warning or error)

deepakshirkem:
Fixes #1420
Previously, Wild would error with "Unsupported absolute relocation" when  encountering a PC-relative relocation to an absolute address (e.g. `callq 0`). lld and GNU ld both handle this for non-PIE executables by computing the PC-relative offset to the absolute target address. For PIE/shared objects, we now emit a clearer error message similar to mold.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -2854,9 +2854,22 @@ fn apply_relocation<
 
             return Ok(RelocationModifier::Normal);
         }
+        RelocationKind::Relative if rel.symbol().is_none() => {
+            if layout.symbol_db.output_kind.is_relocatable() {
+                bail!(
+                    "relocation of type {} to absolute address cannot be used in \
+                    position-independent output; recompile with -fPIC",
+                    rel.raw_type()
+                );
+            }
+            let place = section_info.section_address + offset_in_section;
+            let value = (rel.addend() as u64).wrapping_sub(place);
+            A::relocation_from_raw(r_type)?
+                .write_to_buffer(value, &mut out[offset_in_section as usize..])?;
+            return Ok(RelocationModifier::Normal);
+        }
         _ => {}
     }
-
     let (resolution, symbol_index, local_symbol_id) = get_resolution(rel, object_layout, layout)?;
     let flags = layout.flags_for_symbol(local_symbol_id);
     let mut next_modifier = RelocationModifier::Normal;
--- a/wild/tests/sources/elf/abs-reloc/abs-reloc.s
+++ b/wild/tests/sources/elf/abs-reloc/abs-reloc.s
@@ -1,0 +1,17 @@
+// Test that R_X86_64_PC32 to absolute address is handled for non-PIE executables.
+// `call 0` generates R_X86_64_PC32 with no symbol and addend=-4.
+// lld and GNU ld handle this by computing a PC-relative offset to address 0.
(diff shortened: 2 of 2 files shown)


libwild/src/elf_writer.rs
```diff
@@ -2857,6 +2857,20 @@ fn apply_relocation<
         _ => {}
     }
 
+    if rel.symbol().is_none() && rel.raw_type() == object::elf::R_X86_64_PC32 {
```
davidlattimore: There's overlap in the numeric values for relocation types for different architectures, so using the raw X86_64 relocation type from generic ELF code means this will likely do the wrong thing when linking for non-x86-64 CPUs. It'd also be better if we could fold this into the match above.

deepakshirkem: You're right. I've updated the fix to use `RelocationKind::Relative` instead of the raw relocation type check.

While testing, I also noticed that `RISC-V` `jal ra, 0` triggers the same code path, but it fails with a range overflow because Wild's default load address of `0x400000` is outside the 21-bit JAL range. GNU ld appears to use `0x10000` for RISC-V, which keeps it within range.

I'm not sure whether this is an actual issue or just a difference in defaults. Do you think it's worth creating an issue for this.

libwild/src/elf_writer.rs
```diff
@@ -2857,6 +2857,23 @@ fn apply_relocation<
         _ => {}
     }
 
+    if rel.symbol().is_none() {
+        let rel_info = A::relocation_from_raw(rel.raw_type())?;
+        if matches!(rel_info.kind, RelocationKind::Relative) {
```
davidlattimore: I think the code would be simpler if this were moved into the match above. Possibly something like `RelocationKind::Relative if rel.symbol().is_none() => {...}`

libwild/src/elf_writer.rs
davidlattimore: > While testing, I also noticed that `RISC-V` `jal ra, 0` triggers the same code path, but it fails with a range overflow because Wild's default load address of `0x400000` is outside the 21-bit JAL range. GNU ld appears to use `0x10000` for RISC-V, which keeps it within range.

Sure, sounds worthwhile investigating. I think I used a default load address of 0x400000 because that's what I observed GNU ld doing (at least on x86_64), but I think possibly lld uses a different load address, so it'd probably be fine for us to do so. Or, we could make it arch-dependent.

davidlattimore (approved):
Thanks for working on this :)

PR #2054, opened 2026-06-13
fix: Don't output -ggnu-pubnames sections when --gdb-index is active

davidlattimore:


davidlattimore:
I noticed this because we were producing a substantially larger binary than the other linkers when using --gdb-index. This reduced the size of the binary in question from 690M to 525M (24% reduction).

lapla-cogito (approved):
Thanks!

PR #2055, opened 2026-06-13
fix: Use 0x10000 as default load address for RISC-V

### LINKED ISSUE #2053: Default load address is not architecture-dependent
While working on #1420, I noticed that Wild uses `0x400000` as the default  load address for all ELF architectures. However GNU ld and lld use different addresses per architecture.

## Comparison

| Linker      | x86_64     | RISC-V     | AArch64    | LoongArch64    |
|-------------|------------|------------|------------|----------------|
| GNU ld      | 0x400000   | 0x10000    | 0x400000   | 0x120000000    |
| lld         | 0x400000   | ~0x11000   | 0x200000   | 0x10000        |
| Wild        | 0x400000  | 0x400000  | 0x400000  | 0x400000     |

## Impact

For RISC-V, Wild's `0x400000` load address causes range overflow for JAL  relocations since JAL has a 21-bit signed range (~1MB). Both GNU ld and lld use `~0x10000` for RISC-V which keeps it within range.

## Reproducer

```bash
cat > test.s <<\ASM
.text
.globl _start
_start:
    jal ra, 0
    ret
ASM

riscv64-linux-gnu-as -o test.o test.s
wild -o test.out test.o -nostdlib -m elf64lriscv
# Error: Relocation -4199376 outside of bounds [-1048576, 1048575)
```

## Notes
Discovered while working on #1420.

deepakshirkem:
Fixes #2053

Wild previously used 0x400000 as the default load address for all ELF architectures. Both GNU ld and lld use 0x10000 for RISC-V, which is required for JAL relocations to reach addresses within the 21-bit signed range (~1MB). Added DEFAULT_LOAD_ADDRESS const to the Arch trait so architectures can override the default load address independently of the Platform trait.

Note: There's a warning about `start_memory_address` on the Platform trait being unused since we moved the logic to the Arch trait. Happy to remove it if that's the preferred approach.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -136,12 +136,6 @@ use zerocopy::FromBytes;
 use zerocopy::IntoBytes;
 use zerocopy::KnownLayout;
 
-/// Our starting address in memory when linking non-relocatable executables. We can start memory
-/// addresses wherever we like, even from 0. We pick 400k because it's the same as what ld does and
-/// because picking a distinctive non-zero values makes it more obvious what's happening if we mix
-/// up file and memory offsets.
-pub const NON_PIE_START_MEM_ADDRESS: u64 = 0x400_000;
-
 pub(crate) const GLOBAL_POINTER_SYMBOL_NAME: &str = "__global_pointer$";
 
 /// The ppc64 TOC base symbol. Defined to point at the start of the GOT.
@@ -2064,14 +2058,6 @@ impl platform::Platform for Elf {
         }
 
         SectionRuleOutcome::Custom
-    }
-
-    fn start_memory_address(output_kind: OutputKind) -> u64 {
-        if output_kind.is_relocatable() {
-            0
-        } else {
-            crate::elf::NON_PIE_START_MEM_ADDRESS
-        }
     }
 
     fn requires_symtab_shndx(num_sections: usize) -> bool {
--- a/libwild/src/platform.rs
+++ b/libwild/src/platform.rs
@@ -75,6 +75,10 @@ pub(crate) trait Arch: Send + Sync + 'static {
     type Relaxation: Relaxation;
     type Platform: Platform;
 
+    /// Default load address for non-PIE output.
+    /// Override this for architectures that need a different default.
+    const DEFAULT_LOAD_ADDRESS: u64 = 0x400_000;
+
     /// Returns the identifier to be written into the output file that identifies the file as
     /// belonging to this architecture. e.g. for ELF, this is the header magic for the architecture.
     fn arch_identifier() -> <Self::Platform as Platform>::ArchIdentifier;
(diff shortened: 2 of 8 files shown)


deepakshirkem:
> Note: There's a warning about `start_memory_address` on the Platform trait being unused since we moved the logic to the Arch trait. Happy to remove it if that's the preferred approach.

@davidlattimore

wild/tests/sources/elf/riscv64-load-address/riscv64-load-address.s
```diff
@@ -0,0 +1,16 @@
+/*
+ * Test that RISC-V non-PIE executables use 0x10000 as default load address.
```
davidlattimore: I'd probably be inclined to say that it tests that we can handle a relative relocation to address 0 from in a non-relocatable executable -  we don't actually check the default load address

libwild/src/platform.rs
```diff
@@ -75,6 +75,10 @@ pub(crate) trait Arch: Send + Sync + 'static {
     type Relaxation: Relaxation;
     type Platform: Platform;
 
+    /// Default load address for non-PIE output.
+    /// Override this for architectures that need a different default.
+    const DEFAULT_LOAD_ADDRESS: u64 = crate::elf::NON_PIE_START_MEM_ADDRESS;
```
davidlattimore: It feels like we shouldn't reference the `elf` module from here, or if we do, then we should rename `DEFAULT_LOAD_ADDRESS` to something like `DEFAULT_ELF_LOAD_ADDRESS`. Perhaps delete `NON_PIE_START_MEM_ADDRESS` and make the default here be 0x400000.

davidlattimore (commented):
I haven't looked at the CI failures, but given that you mentioned there was a warning, that's likely what it is. Our CI verifies that the code has no warnings. Deleting dead code is usually the best option.

libwild/src/elf.rs
davidlattimore: You left behind the doc comment for the constant that you deleted.

libwild/src/platform.rs
```diff
@@ -193,6 +197,15 @@ pub(crate) trait Arch: Send + Sync + 'static {
         // overridden.
         unimplemented!();
     }
+
+    /// Return the starting load address for non-PIE output.
+    fn start_memory_address(output_kind: OutputKind) -> u64 {
+        if output_kind.is_relocatable() {
```
davidlattimore, line 203: I don't know if it matters, it but it looks like you've changed the behaviour for the Mach-O implementation. Previously, it would always return 0x1_000_000. Now, it'll return 0 if `output_kind` is relocatable.

deepakshirkem: Hi @davidlattimore, you're right. I've added a `start_memory_address` override in `MachOAArch64` to preserve the original behavior.

davidlattimore (approved):

PR #2057, opened 2026-06-14
perf: Make gdb-index building do less heap allocation

davidlattimore:


davidlattimore:
This change (together with #2056) gives about a 10% speedup on a benchmark with --gdb-index.

marxin (approved):

PR #2058, opened 2026-06-14
port(wasm): Apply code section index relocations

lapla-cogito:
This patch enables the application of code index relocations (`R_WASM_TYPE_INDEX_LEB`, `R_WASM_FUNCTION_INDEX_LEB`, `R_WASM_FUNCTION_INDEX_I32`, `R_WASM_GLOBAL_INDEX_LEB`, `R_WASM_GLOBAL_INDEX_I32`). The byte-level encoding primitives (`apply_relocation()`) were already implemented in a prior patch and this patch adds the symbol-to-output-index resolution, per-body relocation grouping, and replaces the `wasm_encoder::CodeSection` builder with direct emission into the output buffer.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -16,6 +16,7 @@ use crate::wasm_writer::OutputGlobal;
 use crate::wasm_writer::OutputImport;
 use crate::wasm_writer::OutputImportEntity;
 use linker_utils::utils::u32_from_slice;
+use linker_utils::utils::uleb128_size;
 use rayon::prelude::*;
 use std::ops::Range;
 use wasmparser::BinaryReader;
@@ -296,31 +297,10 @@ impl WasmRelocation {
     }
 }
 
-/// Number of bytes needed to encode `value` as an unsigned LEB128.
-pub(crate) fn uleb128_size(mut value: u32) -> usize {
-    let mut size = 1;
-    while value >= 0x80 {
-        value >>= 7;
-        size += 1;
-    }
-    size
-}
-
 /// Write `value` as an unsigned LEB128 into `buf`, returning the number of bytes written.
-pub(crate) fn write_uleb128(buf: &mut [u8], mut value: u32) -> usize {
-    let mut i = 0;
-    loop {
-        let mut byte = (value & 0x7f) as u8;
-        value >>= 7;
-        if value != 0 {
-            byte |= 0x80;
-        }
-        buf[i] = byte;
-        i += 1;
-        if value == 0 {
-            return i;
-        }
-    }
+pub(crate) fn write_uleb128(buf: &mut [u8], value: u64) -> usize {
+    let mut writable = &mut *buf;
+    leb128::write::unsigned(&mut writable, value).unwrap()
 }
 
 /// Write `value` as a 5-byte fixed-width unsigned LEB128. Used for wasm reloc slots that reserve
@@ -1583,16 +1563,16 @@ fn compute_code_section_size(bodies: &[WasmFunctionBody<'_>]) -> u64 {
         return 0;
     }
     let count = bodies.len() as u32;
-    let count_leb_size = uleb128_size(count) as u64;
+    let count_leb_size = uleb128_size(u64::from(count)) as u64;
     let bodies_with_prefix_total: u64 = bodies
         .iter()
         .map(|b| {
-            let body_len = b.bytes.len() as u32;
-            uleb128_size(body_len) as u64 + b.bytes.len() as u64
+            let body_len = b.bytes.len() as u64;
+            uleb128_size(body_len) as u64 + body_len
         })
         .sum();
     let payload_size = count_leb_size + bodies_with_prefix_total;
-    let payload_size_leb_size = uleb128_size(payload_size as u32) as u64;
+    let payload_size_leb_size = uleb128_size(payload_size) as u64;
 
     // section id (1 byte) + payload size LEB + payload
     1 + payload_size_leb_size + payload_size
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -400,6 +400,8 @@ pub(crate) struct WasmDataSegment<'data> {
 #[derive(Debug, Clone)]
 pub(crate) struct WasmFunctionBody<'data> {
     pub(crate) bytes: &'data [u8],
+    /// Byte offset of this body (starting at its size prefix) within the code section payload.
+    pub(crate) code_offset: u32,
 }
 
 impl<'data> File<'data> {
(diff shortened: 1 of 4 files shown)


lapla-cogito:
With more practical testing cases becoming available, I plan to begin establishing a basic testing infrastructure once this gets merged.

```
（▰╹◡╹）❯  cat reloc_test.wat

(module
  (type $t0 (func (param i32) (result i32)))
  (type $t1 (func (param i32 i32) (result i32)))

  (func $add (type $t1) (param $a i32) (param $b i32) (result i32)
    local.get $a
    local.get $b
    i32.add
  )

  (func $double (type $t0) (param $x i32) (result i32)
    local.get $x
    local.get $x
    call $add
  )

  (export "add" (func $add))
  (export "double" (func $double))
)

（▰╹◡╹）❯  wat2wasm reloc_test.wat -o reloc_test.wasm --relocatable

（▰╹◡╹）❯  wasm-objdump -rd reloc_test.wasm

reloc_test.wasm:        file format wasm 0x1

Code Disassembly:

000031 func[0] <add>:
 000032: 20 00                      | local.get 0
 000034: 20 01                      | local.get 1
 000036: 6a                         | i32.add
 000037: 0b                         | end
000039 func[1] <double>:
 00003a: 20 00                      | local.get 0
 00003c: 20 00                      | local.get 0
 00003e: 10 80 80 80 80 00          | call 0 <add>
           00003f: R_WASM_FUNCTION_INDEX_LEB 0 <add>
 000044: 0b                         | end

（▰╹◡╹）❯  ~/repos/wild/target/debug/wild -flavor wasm reloc_test.wasm -o reloc_test.wild.wasm

（▰╹◡╹）❯  wasmtime run --invoke add reloc_test.wild.wasm 3 4
warning: using `--invoke` with a function that takes arguments is experimental and may break in the future
warning: using `--invoke` with a function that returns values is experimental and may break in the future
7

（▰╹◡╹）❯  wasmtime run --invoke double reloc_test.wild.wasm 5
warning: using `--invoke` with a function that takes arguments is experimental and may break in the future
warning: using `--invoke` with a function that returns values is experimental and may break in the future
10
```

libwild/src/wasm.rs
```diff
@@ -399,7 +426,11 @@ pub(crate) struct WasmDataSegment<'data> {
 
 #[derive(Debug, Clone)]
 pub(crate) struct WasmFunctionBody<'data> {
-    pub(crate) bytes: &'data [u8],
+    /// Raw body bytes (locals + operators) without the LEB128 size prefix. Borrowed when no
+    /// relocations are applied; owned when the body has been patched.
+    pub(crate) bytes: std::borrow::Cow<'data, [u8]>,
```
davidlattimore: Why do we need to store the function body with relocations applied rather than applying the relocations directly after copying into the output file like we do for ELF?

libwild/src/wasm.rs
```diff
@@ -1888,6 +2037,55 @@ fn allocate_wasm_object_index_bases(
...
+    }
+
+    // Group relocations by body. Both relocs (by offset) and bodies (by code_offset) are ordered.
+    let mut reloc_iter = relocs.iter().peekable();
+    for body in &mut bodies {
+        let body_start = body.code_offset;
+        let body_end = body_start + body.bytes.len() as u32;
+
+        // Collect relocations that fall within this body's range.
+        let mut body_relocs: Vec<WasmRelocation> = Vec::new();
```
davidlattimore: Is this Vec needed? It looks like the loop that uses it below could just consume directly from `reloc_iter`

libwild/src/wasm.rs
```diff
@@ -296,6 +296,33 @@ impl WasmRelocation {
     }
 }
 
+/// Number of bytes needed to encode `value` as an unsigned LEB128.
+pub(crate) fn uleb128_size(mut value: u32) -> usize {
```
davidlattimore: I notice that we have a closure in elf.rs that's called `size_of_uleb_encoded`. Should this function be put somewhere common so that we can deduplicate at some stage?

libwild/src/wasm.rs
```diff
@@ -296,6 +296,33 @@ impl WasmRelocation {
...
+    let mut size = 1;
+    while value >= 0x80 {
+        value >>= 7;
+        size += 1;
+    }
+    size
+}
+
+/// Write `value` as an unsigned LEB128 into `buf`, returning the number of bytes written.
+pub(crate) fn write_uleb128(buf: &mut [u8], mut value: u32) -> usize {
```
davidlattimore: We have other code that calls `leb128::write`. Possibly we should use that unless there's a reason to do something different here?

davidlattimore (commented):
Nice! Looking forward to seeing some tests in a later PR :)

libwild/src/wasm.rs
marxin: Yes, please. Btw. I made `leb128` crate PR that adds the functionality - https://github.com/gimli-rs/leb128/pull/32.

libwild/src/wasm.rs
marxin: Yes, please use `leb128` crate.

libwild/src/wasm_writer.rs
```diff
@@ -102,6 +103,62 @@ fn copy_encoded_section(encoded: Option<&Vec<u8>>, out: &mut [u8]) -> Result<()>
...
+    // Function bodies: size prefix + body bytes.
+    for body in bodies {
+        let body_len = body.bytes.len() as u32;
+        pos += write_uleb128(&mut out[pos..], body_len);
+        let len = body.bytes.len();
+        out[pos..pos + len].copy_from_slice(&body.bytes);
+        pos += len;
+    }
+
+    if pos != out.len() {
```
marxin: `ensure!`

libwild/src/wasm_writer.rs
```diff
@@ -102,6 +103,62 @@ fn copy_encoded_section(encoded: Option<&Vec<u8>>, out: &mut [u8]) -> Result<()>
     Ok(())
 }
 
+// Each `WasmFunctionBody.bytes` is the raw body content (locals + operators) without a size prefix.
+// This function writes the LEB128 size prefix for each body.
+fn write_code_section(bodies: &[WasmFunctionBody<'_>], out: &mut [u8]) -> Result<()> {
+    if bodies.is_empty() {
+        if !out.is_empty() {
```
marxin: `ensure!`

libwild/src/wasm.rs
lapla-cogito: Has been resolved along with resolving https://github.com/wild-linker/wild/pull/2058#discussion_r3409050655

libwild/src/wasm.rs
lapla-cogito: I've left a TODO comment to note the helper should be updated once the PR merged.

libwild/src/wasm.rs
```diff
@@ -397,9 +404,23 @@ pub(crate) struct WasmDataSegment<'data> {
...
+    pub(crate) value: u32,
+}
+
 #[derive(Debug, Clone)]
 pub(crate) struct WasmFunctionBody<'data> {
+    /// Raw body bytes (locals + operators) without the LEB128 size prefix.
     pub(crate) bytes: &'data [u8],
+    /// Byte offset of this body (starting at its size prefix) within the code section payload.
+    pub(crate) code_offset: u32,
+    pub(crate) relocations: Vec<ResolvedCodeReloc>,
```
davidlattimore, line 423: It looks to me like it might be possible to avoid this Vec and store just enough information to find the relocations later on. e.g. a &'data [WasmRelocation] or even just a u32 that represents the index of the first relocation for this function. But it's perfectly reasonable to leave this optimisation until later.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: If we do that, we'll have to propagate relocation value resolution to the writer side, which would require each object to access its own index map and symbols reference, so the current implementation should be kept I think. But of course we can revisit this if allocation overhead becomes a concern.

PR #2061, opened 2026-06-14
chore: Don't call .context(format!(...))

davidlattimore:


davidlattimore:
`.context`, because it doesn't take a closure, results in `format!` always being evaluated, even on the non-error path, which means we end up heap allocating a string even when there's no error. I've changed `.context` so that it now only accepts a `&'static str`, so we can't make this mistake going forward.

marxin:
Welcome the change - to be honest I probably mixed these 2 multiple times in the part.

PR #2062, opened 2026-06-14
port(MachO): simplify LOAD_COMMANDS and support more LC_LOAD_DYLIB

marxin:
The change makes a simplification where all the load commands live under a single `part_id::LOAD_COMMANDS`. By having that, we can load multiple stub libraries (and allocate an arbitrary number of `LC_LOAD_DYLIB` commands). Right now, we can run a C++ programs :fireworks:

There's an opportunity for a refactoring when it comes to `imported_symbols` and `imported_sections`, where I'm planning to fill up an issue, hoping @davidlattimore will help me with that.

Issue: #757


libwild/src/layout.rs
```diff
@@ -1188,6 +1205,7 @@ pub(crate) struct CommonGroupState<'data, P: Platform> {
 
     /// A list of imported STUB library symbols (Mach-O specific).
     imported_symbols: Vec<ImportedSymbol<'data>>,
+    imported_libraries: HashSet<ImportedLibrary<'data>>,
```
davidlattimore, line 1208: If these are Mach-O specific, can they go into `CommonGroupStateExt` instead? Same for the pre-existing `imported_symbols`. But I'm guessing this is what you meant when you mentioned refactoring in the PR description. Fine to do it separately. I'm also happy to do it if you like.

davidlattimore (approved):

libwild/src/layout.rs
marxin: Yes, that's going to be the subject of the mentioned refactoring and I would like help. Will create an issue shortly.

marxin: #2075

PR #2073, opened 2026-06-15
perf: Avoid b-tree and parallelise symbol sort in gdb-index

davidlattimore:


davidlattimore:
3 commits - all pretty small. Happy to squash or merge separately.

First commit switches from a BTreeSet to a Vec. We then sort the Vec. We don't deduplicate since it seems like we never have duplicate values anyway.

Middle commit adds some tracing and time points plus moves some code around in order to make that easier.

Last commit switches to using a parallel sort when sorting the list of symbols.

First commit reduces time by about 8%. Everything together reduces by about 18%.

```
Benchmark 1 (24 runs): /home/david/save/mold-gdb-index/run-with env-rand /home/d/wild-builds/2026-06-15.opt
  wall_time           833ms ± 4.83ms     824ms …  843ms          0 ( 0%)        0%
  peak_rss           7.53MB ±  101KB    7.29MB … 7.76MB          2 ( 8%)        0%
  cpu_cycles         9.22G  ±  104M     8.96G  … 9.33G           1 ( 4%)        0%
  instructions       11.8G  ± 30.0M     11.8G  … 11.9G           3 (13%)        0%
  cache_references    313M  ± 2.34M      310M  …  317M           0 ( 0%)        0%
  cache_misses       89.9M  ± 1.16M     88.0M  … 91.7M           0 ( 0%)        0%
  branch_misses      25.0M  ±  143K     24.8M  … 25.5M           1 ( 4%)        0%
Benchmark 2 (27 runs): /home/david/save/mold-gdb-index/run-with env-rand /home/d/wild-builds/2026-06-15.opt.no-btree
  wall_time           765ms ± 3.63ms     759ms …  775ms          1 ( 4%)        ⚡-  8.1% ±  0.3%
  peak_rss           7.51MB ±  154KB    7.23MB … 7.91MB          1 ( 4%)          -  0.2% ±  1.0%
  cpu_cycles         9.12G  ±  142M     8.87G  … 9.36G           0 ( 0%)          -  1.0% ±  0.8%
  instructions       11.7G  ± 31.9M     11.6G  … 11.7G           0 ( 0%)        ⚡-  1.4% ±  0.1%
  cache_references    308M  ± 3.10M      303M  …  315M           0 ( 0%)        ⚡-  1.8% ±  0.5%
  cache_misses       88.8M  ± 1.30M     86.6M  … 91.1M           0 ( 0%)          -  1.2% ±  0.8%
  branch_misses      24.3M  ±  139K     24.0M  … 24.6M           0 ( 0%)        ⚡-  2.9% ±  0.3%
Benchmark 3 (30 runs): /home/david/save/mold-gdb-index/run-with env-rand target/opt/wild
  wall_time           680ms ± 5.13ms     667ms …  692ms          0 ( 0%)        ⚡- 18.4% ±  0.3%
  peak_rss           7.55MB ±  107KB    7.32MB … 7.74MB          0 ( 0%)          +  0.3% ±  0.8%
  cpu_cycles         9.12G  ±  208M     8.71G  … 9.44G           0 ( 0%)          -  1.0% ±  1.0%
  instructions       11.8G  ± 38.2M     11.7G  … 11.9G           0 ( 0%)          -  0.4% ±  0.2%
  cache_references    312M  ± 3.06M      307M  …  319M           0 ( 0%)          -  0.5% ±  0.5%
  cache_misses       89.7M  ± 1.64M     86.6M  … 92.5M           0 ( 0%)          -  0.3% ±  0.9%
  branch_misses      24.8M  ±  111K     24.5M  … 25.0M           0 ( 0%)          -  1.0% ±  0.3%
```

libwild/src/gdb_index.rs
```diff
@@ -454,19 +456,26 @@ fn merge_gdb_index_scans(per_object: Vec<Option<PerObjectGdbScan>>) -> GdbIndexS
         for (name, local_cu_idx, attrs) in scan.symbol_entries {
             let entry = encode_cu_vector_entry(base + local_cu_idx, attrs);
             let sd = sym_map.entry(name).or_insert_with_key(|name| SymData {
-                cv_entries: BTreeSet::new(),
+                cv_entries: Vec::new(),
```
mati865: The difference is pretty small when inserting more elements, but for a just single element you can achieve better assembly by preallocating: https://godbolt.org/z/brG1v9MWT
```suggestion
                cv_entries: Vec::with_capacity(1),
```

mati865:
> First commit switches from a BTreeSet to a Vec. We then sort the Vec. We don't deduplicate since it seems like we never have duplicate values anyway.

You could add check guarded on debug assertions to test if there are no duplicates, if their presence would a problem here.

mati865 (approved):

libwild/src/gdb_index.rs
davidlattimore: Done. I went with a capacity of 4 though, since that's the capacity that we would end up with if did:

```
let mut a = Vec::new();
a.push(...);
```

Coincidentally, 4 is also the average number of entries that we end up with on the benchmark I've been running.

marxin: Wouldn't be `SmallVec<[_; 4]>` a better fit for the `cv_entires?

davidlattimore: Only if a high enough percentage ended up not needing heap allocation. Otherwise the extra complexity of SmallVec could potentially make it slower. I just tried it and any difference in performance wasn't sufficient to be easily measured.

marxin: Good - thanks for the re-measurement.

PR #2077, opened 2026-06-15
chore: use latest `leb128` crate functionality

marxin:


libwild/src/elf.rs
```diff
@@ -3712,40 +3713,38 @@ pub(crate) fn gnu_property_notes_section_size(gnu_property_notes: &[GnuProperty]
...
 fn riscv_attributes_section_size(riscv_attributes: &[RiscVAttribute]) -> u64 {
-    let size_of_uleb_encoded = linker_utils::utils::uleb128_size;
-
     (if riscv_attributes.is_empty() {
         0
     } else {
         1 // 'A'
             + 4 // sizeof(u32)
-            + size_of_uleb_encoded(TAG_RISCV_WHOLE_FILE)
+            + unsigned_len(TAG_RISCV_WHOLE_FILE)
```
davidlattimore: It's unfortunate that this function doesn't include leb128 in its name. I guess we could refer to it by its fully qualified name, although the "write" part seems excessive. Alternatively, we could alias it or keep the `uleb128_size` function, but change the implementation to use this. But I don't feel strongly, if you think `unsigned_len` is sufficient, then feel free to go with that.

davidlattimore (approved):

libwild/src/elf.rs
marxin: You are right - I reworked it and started using `use leb128::write::unsigned_len as uleb128_size;`.

davidlattimore (approved):

PR #2078, opened 2026-06-15
feat: support setting an output section's region

### LINKED ISSUE #2044: >region memory region placement in linker scripts not implemented
Wild does not currently support the `>region` syntax in linker scripts. This directive assigns output sections to named memory regions defined in a `MEMORY` block and is part of the GNU ld linker script specification.

## Reproducer

```bash
cat > 1.c <<\EOF
int data_var = 42;
int bss_var;
int main() { return data_var + bss_var; }
EOF

cat > memory_test.ld <<\EOF
MEMORY
{
    RAM : ORIGIN = 0x20000000, LENGTH = 64K
    ROM : ORIGIN = 0x08000000, LENGTH = 128K
}

SECTIONS
{
    .text : { *(.text*) } > ROM
    .data : { *(.data*) } > RAM
    .bss  : { *(.bss*)  } > RAM
}
EOF

clang -o 1.o 1.c -c
wild -o 1.out 1.o -T memory_test.ld -nostdlib
```

### Current Behavior
```
wild: error: Failed to parse linker script memory_test.ld:
parse error at line 10, column 5
|
10 |     .data : { (.data) } > RAM
|     ^
```

### Expected Behavior
Sections should be placed into the specified memory region respecting ORIGIN and LENGTH constraints, matching GNU ld behavior.

## Notes
- Blocker for `--print-memory-usage` to report accurate used/available sizes per region
- Related to #44

vishruth-thimmaiah:
closes #2044

This PR implements support for setting an output section's region using the `>region` syntax.


libwild/src/output_section_id.rs
```diff
@@ -324,39 +327,54 @@ impl<'scope, 'data, P: Platform> OutputOrderBuilder<'scope, 'data, P> {
...
                 }
             }
         }
 
+        let section_region = section_info.region_name;
         P::program_segment_defs()
             .iter()
-            .zip(self.active_segment_kinds.iter_mut())
-            .for_each(|(segment_def, active_id)| {
+            .enumerate()
```
davidlattimore: It's probably nicer to keep the zip rather than switching to enumerate. You can either call `.zip` on a zip - you'll end up with nested tuples - `((a, b), c)` or you can use [ `itertools::multizip`](https://docs.rs/itertools/latest/itertools/macro.izip.html) which would give you a single tuple `(a, b, c)`.

libwild/src/layout.rs
```diff
@@ -253,10 +254,19 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
...
     );
 
-    let memory_regions: Vec<crate::linker_script::MemoryRegion<'_>> = linker_scripts
-        .iter()
-        .flat_map(|s| s.parsed.memory_regions.iter().cloned())
-        .collect();
+    let mut memory_regions = HashMap::new();
+    for s in &linker_scripts {
+        for region in &s.parsed.memory_regions {
+            memory_regions.insert(
```
davidlattimore: Should it be an error if multiple linker scripts define a memory region with the same name? Or perhaps only one script is allowed to define memory regions? Separate PR is fine if you like. We'll want tests for the error cases.

libwild/src/layout.rs
```diff
@@ -5131,11 +5151,36 @@ fn layout_section_parts<'data, P: Platform>(
...
+                    mem_offset = region.origin + region.used;
+                }
                 if let Some(ref expr) = section_info.location {
-                    mem_offset = expression_eval(expr)?;
+                    let offset = expression_eval(expr, memory_regions)?;
+                    if let Some(region) = region
+                        && (offset < mem_offset || offset > mem_offset + region.length)
+                    {
+                        bail!(
+                            "address 0x{offset:x} of  section '{}' is not within region `{}'",
```
davidlattimore: Extra space between "of" and "section"

libwild/src/layout.rs
```diff
@@ -5131,11 +5151,36 @@ fn layout_section_parts<'data, P: Platform>(
                 let section_info = output_sections.output_info(section_id);
                 let part_id_range = section_id.part_id_range();
                 let max_alignment = sizes.max_alignment(part_id_range.clone(), output_sections);
+                let region = section_info
+                    .region_name
+                    .map(|region_name| {
+                        memory_regions.get(region_name).with_context(|| {
+                            format!(
+                                "Memory region '{}' not found for section '{}'",
```
davidlattimore: Could be good to add an ExpectError test for this

libwild/src/layout.rs
```diff
@@ -5226,6 +5271,20 @@ fn layout_section_parts<'data, P: Platform>(
...
                         }
                     });
+
+                if let Some(region_name) = section_info.region_name
+                    && let Some(region) = memory_regions.get_mut(region_name)
+                {
+                    let max_offset = region.origin + region.length;
+                    if mem_offset > max_offset {
+                        bail!(
+                            "region '{}' overflowed by {} bytes",
```
davidlattimore, line 5287: Could be good to add two tests for this. One that fills the region exactly and one that goes a single byte over. That'll allow us to verify that we have the same behaviour as the other linkers and don't have any off-by-one errors

libwild/src/layout.rs
vishruth-thimmaiah: GNU ld and lld differ on this, ld just emits a warning and ignores any repeated definitions, whereas lld emits an error. I've made wild return an error here, similar to lld.

libwild/src/layout.rs
vishruth-thimmaiah: I've added a new region in `linker-script-region.ld` that is exactly the size of the section it contains, and `linker-script-region-overflow.ld` that is one byte smaller.

libwild/src/output_section_id.rs
vishruth-thimmaiah: Fixed.

davidlattimore (approved):

PR #2081, opened 2026-06-16
fix: Support `--wrap` combined with linker plugin LTO

### LINKED ISSUE #1903: linker-plugin LTO doesn't work with --wrap

lapla-cogito:
When `--wrap` and `-flto` were used together, the LTO plugin would internalize/remove `__wrap_*` symbols because it didn't know they were needed. This caused link failures.

Three changes fix this:

- In `get_symbol_resolution` (`get_symbols_v3`), report wrapped symbols as resolved to a regular (non-IR) object rather than letting the name-table override mislead the plugin into thinking they resolved to IR.
- Before calling the plugin's `all_symbols_read`, mark the wrapped symbol, `__wrap_*`, and `__real_*` variants with `HAS_NON_IR_REF` so the plugin emits them in its output instead of internalizing them.
- After loading the plugin's output objects, re-apply wrap overrides so the name table points to the new definitions rather than the LTO input objects.

Fixes #1903

### DIFF
--- a/libwild/src/linker_plugins.rs
+++ b/libwild/src/linker_plugins.rs
@@ -25,6 +25,7 @@ use crate::input_data::FileLoader;
 use crate::input_data::InputRef;
 use crate::layout_rules::LayoutRulesBuilder;
 use crate::output_section_id::OutputSections;
+use crate::platform::Args as _;
 use crate::platform::Platform;
 use crate::platform::RawSymbolName as _;
 use crate::resolution::ResolutionResources;
@@ -211,6 +212,11 @@ impl<'data> LinkerPlugin<'data> {
 
         timing_phase!("Linker plugin codegen");
 
+        // Mark wrapped symbol names and their __wrap_/__real_ variants as referenced by non-IR
+        // code. This ensures the plugin keeps them in the LTO output rather than
+        // internalising/removing them.
+        mark_wrap_symbols_as_non_ir_ref(symbol_db, per_symbol_flags);
+
         let plugin_outputs = self.store.loaded()?.with_callbacks(|callbacks| {
             if let Some(cb) = callbacks.all_symbols_read {
                 let ctx = AllSymbolsReadContext {
@@ -237,6 +243,10 @@ impl<'data> LinkerPlugin<'data> {
             layout_rules_builder,
             plugin_loaded,
         )?;
+
+        // Re-apply --wrap overrides so that wrapped names now map to definitions from the LTO
+        // output rather than the LTO input objects.
+        symbol_db.apply_wrapped_symbol_overrides();
 
         resolver.resolve_symbols_and_select_archive_entries(symbol_db, per_symbol_flags)?;
 
@@ -922,6 +932,20 @@ fn get_symbol_resolution<'data>(
     if !sym.version.is_null() {
         raw_name.version_name = Some(unsafe { CStr::from_ptr(sym.version) }.to_bytes());
     }
+
+    // If the symbol was wrapped via --wrap, the name table has been modified to map the original
+    // name to __wrap_<name>. We must not report the wrapped resolution to the plugin, since the
+    // plugin needs to know the pre-wrap state.
+    let wrap_names = symbol_db.args.symbol_names_to_wrap();
+    let is_wrapped = wrap_names.iter().any(|w| w.as_bytes() == raw_name.name);
+    if is_wrapped {
+        return if sym.is_undefined() {
+            PluginSymbolResolution::ResolvedExec
+        } else {
+            PluginSymbolResolution::PreemptedReg
+        };
+    }
+
     let symbol_id = symbol_db
         .get(&PreHashedSymbolName::from_raw(&raw_name), true)
         .map(|id| symbol_db.definition(id));
@@ -1417,4 +1441,25 @@ pub(crate) fn resolve_lto_symbols<'data, 'scope>(
                 Ok(())
             },
         )
+}
+
+/// Marks symbols related to --wrap as having non-IR references. This ensures that the linker
+/// plugin preserves these symbols in its output rather than internalising them.
+fn mark_wrap_symbols_as_non_ir_ref<'data, P: Platform>(
+    symbol_db: &SymbolDb<'data, P>,
+    per_symbol_flags: &mut PerSymbolFlags,
+) {
+    for name in symbol_db.args.symbol_names_to_wrap() {
+        for lookup_name in [
+            name.clone(),
+            format!("__wrap_{name}"),
+            format!("__real_{name}"),
+        ] {
+            if let Some(symbol_id) =
+                symbol_db.get_unversioned(&UnversionedSymbolName::prehashed(lookup_name.as_bytes()))
+            {
+                per_symbol_flags.set_flag(symbol_id, ValueFlags::HAS_NON_IR_REF);
+            }
+        }
+    }
 }
--- a/wild/tests/sources/elf/wrap-lto/wrap-lto.c
+++ b/wild/tests/sources/elf/wrap-lto/wrap-lto.c
@@ -1,0 +1,30 @@
+//#AbstractConfig:default
+//#RequiresLinkerPlugin:true
+//#Object:runtime.c
+//#Object:wrap-lto-2.c:-fno-lto
+//#SkipLinker:ld
+//#CompArgs:-flto
+//#LinkArgs:-flto -nostdlib -z now -Wl,-wrap,foo
+
+//#Config:gcc:default
+//#LinkerDriver:gcc
+
+//#Config:clang:default
+//#Compiler:clang
+//#LinkerDriver:clang
+//#EnableLinker:lld
+
+#include "../common/runtime.h"
+
+int foo(void);
+int __real_foo(void);
+
+int __wrap_foo(void) { return __real_foo() + 32; }
+
+void _start(void) {
+  runtime_init();
+  if (foo() != 42) {
+    exit_syscall(100);
+  }
+  exit_syscall(42);
+}
--- a/libwild/src/lib.rs
+++ b/libwild/src/lib.rs
@@ -303,8 +303,6 @@ impl Linker {
             loaded,
         )?;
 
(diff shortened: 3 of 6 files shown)


libwild/src/linker_plugins.rs
```diff
@@ -1418,3 +1443,29 @@ pub(crate) fn resolve_lto_symbols<'data, 'scope>(
...
 }
+
+/// Marks symbols related to --wrap as having non-IR references. This ensures that the linker
+/// plugin preserves these symbols in its output rather than internalising them.
+fn mark_wrap_symbols_as_non_ir_ref<'data, P: Platform>(
+    symbol_db: &SymbolDb<'data, P>,
+    per_symbol_flags: &mut PerSymbolFlags,
+) {
+    let wrap = symbol_db.args.symbol_names_to_wrap();
+    if wrap.is_empty() {
```
davidlattimore: This early return seems unnecessary. The only thing it skips is a loop that would otherwise have zero iterations.

lapla-cogito:
With this patch, I've confirmed that xorg-server (which was mentioned in the original issue) can also be built with LTO enabled.

libwild/src/linker_plugins.rs
```diff
@@ -1418,3 +1443,29 @@ pub(crate) fn resolve_lto_symbols<'data, 'scope>(
...
+        return;
+    }
+
+    for name in wrap {
+        for lookup_name in [
+            name.clone(),
+            format!("__wrap_{name}"),
```
davidlattimore, line 1455: We previously only had the format string "__wrap_{name}" once, but now we have it in two places. Can we avoid the duplication? One possibility would be to define a function that takes `name` and returns the "__wrap_" string. Alternatively, it might be possible to store the symbol IDs somewhere, but I haven't thought too hard about that option. Same comment applies to "__real_{name}".

wild/tests/sources/elf/wrap-lto/wrap-lto.c
```diff
@@ -0,0 +1,30 @@
+//#AbstractConfig:default
+//#RequiresLinkerPlugin:true
+//#Object:runtime.c
+//#Object:wrap-lto-2.c:-fno-lto
+//#SkipLinker:ld
+//#CompArgs:-flto
+//#LinkArgs:-flto -nostdlib -z now -Wl,-wrap,foo
+
+//#Config:gcc:default
```
davidlattimore, line 9: Could be good to add a comment here saying why we don't run this test with GNU ld

libwild/src/linker_plugins.rs
```diff
@@ -922,6 +932,21 @@ fn get_symbol_resolution<'data>(
     if !sym.version.is_null() {
         raw_name.version_name = Some(unsafe { CStr::from_ptr(sym.version) }.to_bytes());
     }
+
+    // If the symbol was wrapped via --wrap, the name table has been modified to map the original
+    // name to __wrap_<name>. We must not report the wrapped resolution to the plugin, since the
+    // plugin needs to know the pre-wrap state.
+    let wrap_names = symbol_db.args.symbol_names_to_wrap();
+    let is_wrapped =
+        !wrap_names.is_empty() && wrap_names.iter().any(|w| w.as_bytes() == raw_name.name);
```
davidlattimore: `!wrap_names.is_empty() &&` seems unnecessary

libwild/src/linker_plugins.rs
lapla-cogito: I think a helper function like `fn wrap_name(name: &str) -> String` would add more indirection than it saves 🤔. The body is essentially just `format!("__wrap_{name}")` (or `__real_`), which is barely shorter than the signature itself. The two call sites also use the formatted strings differently: `symbol_db.rs` immediately copies into an arena allocator, while `linker_plugins.rs` uses them as temporary values for hash lookups. A shared function returning String would work for both, but it wouldn't simplify either caller much.
For the latter one, `apply_wrapped_symbol_overrides` is called twice (before and after LTO), and symbol IDs can change between calls as LTO output objects are added, so caching IDs would require invalidation logic that adds more complexity than it removes.

wild/tests/sources/elf/wrap-lto/wrap-lto.c
lapla-cogito: Are you expecting comments like "GNU ld requires a compiler driver to load the plugin"? If so, I believe they should be added to all existing LTO tests with similar configs.

libwild/src/linker_plugins.rs
lapla-cogito: Similar to `apply_wrapped_symbol_overrides`, I initially considered adding `(verbose_)timing_phase!` and included an early return, but when I later removed `timing_phase!`, I forgot to remove the early return. Thanks for pointing that out.

libwild/src/linker_plugins.rs
davidlattimore: It's not about being shorter - it'll certainly be longer - it's more about having a thing only be defined once in the code. It's why we create constants for things - so that we can give a name to a concept and quickly see everything that uses that concept. That said, I don't feel strongly if you really don't want to extract the format(s) out.

davidlattimore (approved):

PR #2083, opened 2026-06-16
fix: error on sub-pointer-size absolute relocations in shared objects

### LINKED ISSUE #2074: Wild emits wrong dynamic relocation type for absolute relocations in shared objects with --no-gc-sections
## Reproducer

```bash
cat > shared.s <<\EOF
.data
.long _shared
EOF

gcc -c -o shared.o shared.s
wild -shared shared.o -o shared.so --no-gc-sections
readelf -r shared.so
```

## Current behavior
With `--no-gc-sections`, Wild silently succeeds and emits the wrong  dynamic relocation type:

```
Relocation section '.rela.dyn':
R_X86_64_64   _shared + 0   ← wrong, should be R_X86_64_32
```

## Expected behavior
Both GNU ld and lld error with:
`relocation R_X86_64_32 cannot be used when making a shared object; recompile with -fPIC`

Wild should do the same — emit an error instead of silently producing a binary with the wrong relocation type.

## Note
With `--gc-sections` (Wild's default) all three linkers behave the same since the unreferenced `.data` section gets garbage collected. The bug only manifests with `--no-gc-sections`.

deepakshirkem:
R_X86_64_32 and R_X86_64_32S cannot be used when making a shared object because the dynamic linker has no way to patch 32-bit absolute relocations at runtime. Wild was silently emitting R_X86_64_64 instead of erroring, producing a broken shared object.

Add an early check in apply_relocation to error when these relocation types are encountered while building a shared object, matching the behavior of GNU ld and lld.

Fixes #2074

### DIFF
--- a/wild/tests/sources/elf/pic-required/pic-required.s
+++ b/wild/tests/sources/elf/pic-required/pic-required.s
@@ -1,0 +1,10 @@
+// Test that R_X86_64_32 cannot be used when making a shared object.
+// .long _shared generates R_X86_64_32 which is illegal in shared objects.
+// Both GNU ld and lld error: recompile with -fPIC
+//#Arch:x86_64
+//#LinkArgs:--no-gc-sections -shared
+//#ExpectError:recompile with -fPIC
+//#ExpectErrorWild:R_X86_64_32.*cannot be used when making a shared object
+
+.data
+.long _shared
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -5018,6 +5018,13 @@ fn process_relocation<'data, 'scope, A: Arch<Platform = Elf>, R: Relocation>(
                     .store(true, atomic::Ordering::Relaxed);
             }
         } else if flags_to_add.needs_direct() && flags.is_interposable() {
+            if symbol_db.output_kind.is_shared_object() && A::is_illegal_in_shared_object(r_type) {
+                bail!(
+                    "relocation {} cannot be used when making a shared object; \
+                    recompile with -fPIC",
+                    A::rel_type_to_string(r_type),
+                );
+            }
             if section_is_writable {
                 common.allocate(part_id::RELA_DYN_GENERAL, elf::RELA_ENTRY_SIZE);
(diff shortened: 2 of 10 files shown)


wild/tests/sources/elf/pic-required/pic-required.s
```diff
@@ -0,0 +1,10 @@
+// Test that R_X86_64_32 cannot be used when making a shared object.
+// .long _shared generates R_X86_64_32 which is illegal in shared objects.
+// Both GNU ld and lld error: recompile with -fPIC
+//#Arch:x86_64
+//#LinkArgs:--no-gc-sections -shared
+//#SkipLinker:ld
```
davidlattimore: Could be good to enable ld (and possibly lld) for the test. If all linkers have "recompile with -fPIC" then you could have that as the error matcher. If you want to verify that Wild says the message below, then you could add "ExpectErrorWild" (implementing it should be pretty trivial - there's already an "ExpectWarningWild")

deepakshirkem:
@davidlattimore, One question: should we extend this PR to also report an error for sub-pointer-size absolute relocations on other 64-bit architectures such as AArch64, RISC-V64, PPC64, and LoongArch64?
They seem to have the same fundamental issue with `ByteSize(4)` absolute relocations in shared objects.
I could either implement a generic check based on the architecture trait, or add explicit relocation-type checks in each architecture file, similar to what we did for x86_64. Which approach do you think would be preferable?

davidlattimore:
That'd be great if you can make the check generic. It could also be good to look to see if the check can be done in the layout phase rather than the write phase. That said, it's OK if it stays in the write phase.

deepakshirkem:
> That'd be great if you can make the check generic. It could also be good to look to see if the check can be done in the layout phase rather than the write phase. That said, it's OK if it stays in the write phase.

The check now lives in the layout phase (process_relocation in elf.rs) rather than the write phase, catching the error earlier. Each arch overrides the method with its specific illegal relocation types.

davidlattimore (approved):
Thanks :)

PR #2085, opened 2026-06-17
test: Add basic Wasm integration test infrastructure

lapla-cogito:
This patch adds an integration test infrastructure for WebAssembly. Since Wild currently lacks support for features required to link C code targeting Wasm, such as memory imports, data section emission, and symbols like `__stack_pointer`, the added tests are based on `.wat` files instead.

For each test case, the input is first linked with wasm-ld as a reference. If linking succeeds, the same input is linked with Wild, and the resulting output is validated using `wasm-tools validate` to verify that it is a well-formed Wasm binary. The binary produced by Wild is then executed with wasmtime to verify its runtime behavior.

Part of #1431

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,27 @@ jobs:
       # spawning processes, which isn't supported on wasi.
       - run: cargo test --profile ci -p libwild --target wasm32-wasip1
 
+  wasm-test:
+    name: Test Wasm
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@stable
+        id: rust-toolchain
+      - uses: Swatinem/rust-cache@v2
+      - run: sudo apt-get update && sudo apt-get install -y wabt lld
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: wasm-tools
+      - name: Setup Wasmtime
+        uses: bytecodealliance/actions/wasmtime/setup@v1
+        with:
+          version: '43.0.0'
+      - run: cargo test --features wasm -- wasm
+
   macho-build:
     name: Test Mach-O (AArch64)
     runs-on: macos-latest
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -330,6 +330,8 @@ fn collect_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
 
     let host_arch = get_host_architecture();
 
+    collect_wasm_tests(tests, filter)?;
+
     for platform in [PlatformKind::Elf, PlatformKind::MachO] {
         // Right now, the Mach-O provided Clang and the ld linker do not support the ELF format.
         if platform == PlatformKind::Elf && cfg!(target_os = "macos") {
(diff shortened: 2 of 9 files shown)


wild/tests/sources/wasm/multi-object/multi-object.wat
```diff
@@ -0,0 +1,9 @@
+;;#Object:multi-object2.wat
+;;#SkipRun
```
lapla-cogito: Since cross-module symbol resolution has not yet been implemented, we are currently unable to run test cases that involve multiple object files.

.github/workflows/ci.yml
```diff
@@ -142,6 +142,28 @@ jobs:
       # spawning processes, which isn't supported on wasi.
       - run: cargo test --profile ci -p libwild --target wasm32-wasip1
 
+  wasm-test:
```
lapla-cogito, line 145: As the test suite grows, we may consider creating a dedicated container image for Wasm. For now, however, that would be premature I think.

wild/tests/integration_tests.rs
```diff
@@ -420,6 +422,210 @@ fn collect_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
...
+    let mut extra_objects = Vec::new();
+    let mut expect_error = false;
+    let mut skip_run = false;
+    for line in source.lines() {
+        if let Some(rest) = line.trim().strip_prefix(";;#") {
+            if let Some(obj_name) = rest.strip_prefix("Object:") {
+                extra_objects.push(obj_name.trim().to_owned());
+            } else if rest.starts_with("ExpectError") {
+                expect_error = true;
+            } else if rest.starts_with("SkipRun") {
```
davidlattimore: Can we make the test directives consistent with the tests for other platforms? ("RunEnabled")

wild/tests/integration_tests.rs
```diff
@@ -420,6 +422,210 @@ fn collect_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
...
+
+    let root = src_path(platform_name);
+    if !root.exists() {
+        return Ok(());
+    }
+
+    let dir = std::fs::read_dir(&root)
+        .with_context(|| format!("Failed to read directory {}", root.display()))?;
+
+    for entry in dir {
```
davidlattimore, line 428: This is duplicating a lot of code in `collect_tests`. Any reason we can't reuse it with a little refactoring?

wild/tests/integration_tests.rs
```diff
@@ -420,6 +422,210 @@ fn collect_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
     Ok(())
 }
 
+fn collect_wasm_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
```
lapla-cogito: It is true that the Wasm-specific logic currently feels somewhat isolated. However, since the tests are currently compiled using `wat2wasm`, I don't think there is much benefit in forcing additional abstraction or code sharing at this stage. The amount of Wasm-specific code is still relatively small, and I think it would make more sense to refactor and unify the infrastructure once we are able to replace all of the tests with C, C++, or Rust sources.

wild/tests/integration_tests.rs
```diff
@@ -420,6 +422,210 @@ fn collect_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
...
+    std::fs::create_dir_all(&build_dir)?;
+
+    // Parse directives from .wat file (;;# prefix)
+    let source = std::fs::read_to_string(wat_file)?;
+    let mut extra_objects = Vec::new();
+    let mut expect_error = false;
+    let mut skip_run = false;
+    for line in source.lines() {
+        if let Some(rest) = line.trim().strip_prefix(";;#") {
+            if let Some(obj_name) = rest.strip_prefix("Object:") {
```
davidlattimore, line 494: I know at the moment wasm only supports a few of the test directives that we support for ELF, but as the list grows, I think it'd be good if we could share the code as much as possible to avoid inconsistencies.

davidlattimore (commented):
Great to see some wasm integration tests :)

wild/tests/integration_tests.rs

wild/tests/integration_tests.rs
lapla-cogito: Extracted directory scanning portion into helper.

wild/tests/integration_tests.rs
lapla-cogito: As noted [here](https://github.com/wild-linker/wild/pull/2085#discussion_r3425193622), I believe we'll eventually need to do a refactor to eliminate Wasm-specific logic. Since we expect few additional test directives before this becomes possible, I think it makes sense to standardize them during the refactoring process.

davidlattimore (approved):

marxin:
Very nice progress :rocket:

Issue #2075, opened 2026-06-15
Mach-O: move `imported_symbols/imported_libraries` to `P::CommonGroupStateExt`


marxin:
The following 2 fields can be moved to Mach-O specific `P::CommonGroupStateExt`
https://github.com/wild-linker/wild/blob/a9536f43bbe960552996b9ffc56535a898d583bf/libwild/src/layout.rs#L1206-L1208

PR #2089, opened 2026-06-17
refactor: Make late-stage layout phase platform interface more flexible

davidlattimore:
Three commits. One creates FinaliseSizesExt, where previously we just uses LayoutExt. For now all `Platform` implementations use the same type for both `FinaliseSizesExt` and `LayoutExt`, but that will likely change later. The second commit passes Groups states directly to `create_finalise_sizes_ext`, where previously we passed an iterator over objects. The last commit is just a rename to a more general and more consistent name.


davidlattimore:
Issue #2075

PR #2090, opened 2026-06-18
chore: Fix test filtering following #2085

davidlattimore:


wild/tests/integration_tests.rs
```diff
@@ -439,10 +435,6 @@ fn for_each_test_dir(
             .context("Non-UTF-8 path")?
             .to_owned();
 
-        if filter.excludes(&format!("{platform_name}/{name}")) {
```
davidlattimore, line 442: The bug was in this line. The filter format is `{platform_name}/{arch}/{name}`. When running tests with `cargo nextest`, which uses full test names for filtering for running all tests, this meant that all the tests were getting excluded, but nextest didn't know this and was reporting them as passing even though they hadn't actually been run.

PR #2093, opened 2026-06-18
fix(loongarch64): Treat R_LARCH_CALL36 as PLT-generating

wszqkzqk:
## Problem

Run wild with `-mcmodel=medium`:

```
Direct relocation (R_LARCH_CALL36) to dynamic symbol from non-writable
section, but copy relocations are disabled because output is a shared
object.
```

`R_LARCH_CALL36` was mapped to `RelocationKind::Relative`, so the linker tried to resolve it directly. For interposable symbols in non-writable sections, the only fallback is a copy relocation, which is disallowed for shared objects.

## Fix

Treats CALL36 like R_LARCH_B26 and call relocations on other architectures.

This matches the behavior of ld.lld, which routes R_LARCH_CALL36 through processR_PLT_PC.

## Test

Ran the integration tests on Arch Linux for Loong64 (this distro defaults to `-mcmodel=medium` so it uses R_LARCH_CALL36 instead of R_LARCH_B26):

- Before: 13 failures in `elf/loongarch64/*` with the CALL36 error.
- After: 275 passed, 0 failed, 859 ignored.

### DIFF
--- a/libwild/src/elf_loongarch64.rs
+++ b/libwild/src/elf_loongarch64.rs
@@ -128,6 +128,15 @@ impl crate::platform::Arch for ElfLoongArch64 {
         let offset = offset_in_section as usize;
 
         match relocation_kind {
+            object::elf::R_LARCH_CALL36 if !interposable => {
+                relocation.kind = RelocationKind::Relative;
+                return Some(Relaxation {
+                    kind: RelaxationKind::NoOp,
+                    rel_info: relocation,
+                    mandatory: true,
+                });
+            }
+
             object::elf::R_LARCH_B26 if !interposable => {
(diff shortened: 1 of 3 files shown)


davidlattimore (commented):
Thanks for the fix. Can you add an integration test? Also, your PR description reads like it was written by AI. Please see `CONTRIBUTING.md`.

wszqkzqk:
> Also, your PR description reads like it was written by AI. Please see `CONTRIBUTING.md`.

Sorry, I used it to translate my Chinese discription to English. Let me organize it manually.

wszqkzqk:
> Thanks for the fix. Can you add an integration test?

Done.

davidlattimore (approved):

PR #2103, opened 2026-06-19
feat: calculate section layouts and part layouts at the same time

vishruth-thimmaiah:
By computing both at the same time, linker scripts can use the values of previously computed sections within expressions defined for the current output section.


davidlattimore (approved):
Great work! It'd be good to follow up with a change that detects forward references and gives an appropriate error.

PR #2104, opened 2026-06-20
fix: Don't panic on linker script section names shorter than 4 bytes

### LINKED ISSUE #2102: Wild panics on linker script section names shorter than 4 characters
## Reproducer

```bash
cat > short.s <<\EOF
.section foo
.long 0
EOF

cat > short.ld <<\EOF
SECTIONS {
    foo 0x1000 : { *(foo) }
}
EOF

gcc -c -o short.o short.s
wild -T short.ld short.o -o out -nostdlib
```

## Current behavior
hread 'main' panicked at libwild/src/layout_rules.rs:514:18:
Prefixes of length less than 4 not yet supported

## Expected behavior
Both GNU ld and lld link successfully with no errors.

deepakshirkem:
Section names with fewer than 4 bytes (e.g. 'foo') caused Wild to panic when building the section rule hash table. The hash function required exactly 4 bytes from the section name prefix.

Fix by zero-padding short names to 4 bytes before hashing, preserving the existing 4-byte prefix optimization while supporting short names. GNU ld and lld both handle short section names without error.

Fixes #2102

### DIFF
--- a/libwild/src/layout_rules.rs
+++ b/libwild/src/layout_rules.rs
@@ -513,8 +513,7 @@ impl<'data> SectionRules<'data> {
             rules: HashTable::with_capacity(rules.len() * RULE_TABLE_CAPACITY_MULTIPLIER),
         };
         for rule in rules {
-            let hash = section_name_prefix_hash(rule.name_matcher.prefix_bytes())
-                .expect("Prefixes of length less than 4 not yet supported");
+            let hash = section_name_prefix_hash(rule.name_matcher.prefix_bytes()).unwrap_or(0);
 
             map.rules.insert_unique(hash, rule.clone(), |existing| {
                 section_name_prefix_hash(existing.name_matcher.prefix_bytes()).unwrap_or(0)
@@ -551,11 +550,16 @@ impl<'data> SectionRules<'data> {
     }
 }
 
-/// Returns a hash of the first four bytes of the supplied name or `None` if the name is shorter
-/// than 4 bytes.
-#[inline(always)]
+/// Returns a hash of the first four bytes of the supplied name, zero-padding if shorter than 4
+/// bytes. Returns `None` if the name is empty.
 fn section_name_prefix_hash(name: &[u8]) -> Option<u64> {
-    Some(hash_bytes(name.get(..4)?))
+    if name.is_empty() {
+        return None;
+    }
+    let mut buf = [0u8; 4];
+    let len = name.len().min(4);
+    buf[..len].copy_from_slice(&name[..len]);
+    Some(hash_bytes(&buf))
 }
 
 /// Determines, where if anywhere, we should place an input section with no name.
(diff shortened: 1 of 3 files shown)


wild/tests/sources/elf/linker-script-short-section-name/linker-script-short-section-name.s
```diff
@@ -0,0 +1,13 @@
+//#Arch:x86_64
+//#LinkerScript:linker-script-short-section-name.ld
+//#Mode:static
+//#RunEnabled:false
+//#SkipLinker:ld
```
lapla-cogito, line 5: This test could ideally assert a bit more, but that's not a blocker for me since the main point of this patch is fixing the panic, and it does that.

lapla-cogito (approved):
Looks good, thanks!

libwild/src/layout_rules.rs
```diff
@@ -551,11 +550,16 @@ impl<'data> SectionRules<'data> {
     }
 }
 
-/// Returns a hash of the first four bytes of the supplied name or `None` if the name is shorter
-/// than 4 bytes.
-#[inline(always)]
```
davidlattimore, line 556: Why remove the inline directive? Most of the inline directives in wild were added after profiling showed that they provided a performance gain.

wild/tests/sources/elf/linker-script-short-section-name/linker-script-short-section-name.s
davidlattimore: Not crashing is good, but I think we should really be checking that the section rule is actually being applied. At the moment, as far as I can tell the section `foo` isn't in the output binary, probably because we default to `--gc-sections` and the section isn't referenced.

libwild/src/layout_rules.rs
```diff
@@ -551,11 +550,16 @@ impl<'data> SectionRules<'data> {
...
+/// Returns a hash of the first four bytes of the supplied name, zero-padding if shorter than 4
+/// bytes. Returns `None` if the name is empty.
 fn section_name_prefix_hash(name: &[u8]) -> Option<u64> {
-    Some(hash_bytes(name.get(..4)?))
+    if name.is_empty() {
```
davidlattimore, line 556: If all the callers just do `.unwrap_or(0)`, then there doesn't seem to be any reason to special-case the empty name. I'd be inclined to change this function to not return an Option.

davidlattimore (commented):
This approach is fine for matchers like `foo`, but won't work for prefix matchers - e.g. `foo*`. I'm fine with not supporting patterns like `foo*`, but we should error if we encounter such a pattern rather than silently doing the wrong thing.

deepakshirkem:
> This approach is fine for matchers like `foo`, but won't work for prefix matchers - e.g. `foo*`. I'm fine with not supporting patterns like `foo*`, but we should error if we encounter such a pattern rather than silently doing the wrong thing.

@davidlattimore, Would you prefer that I revert these changes first and then submit the updated fix as a separate patch?

davidlattimore:
It's fine to leave them provided we're going to add an error for the prefix case before the next release. Note, if you intend to try and actually support short prefixes, then just be careful of performance - mapping input sections to output sections is something we do a lot of, so we don't want to regress performance for the common case of having no linker script and thus no short perfix matchers.

PR #2105, opened 2026-06-20
fix: Accept full byte range for R_X86_64_8 and R_X86_64_16 relocations

### LINKED ISSUE #2101: R_X86_64_8 and R_X86_64_16 have incorrect range checks
Wild rejects valid values for `R_X86_64_8` and `R_X86_64_16` relocations due to incorrect range checks. Wild uses a signed-only range `[-128, 128)` for `R_X86_64_8`, but both GNU ld and lld accept the full mixed range `[-128, 255]` (signed minimum, unsigned maximum). Same issue applies to `R_X86_64_16`.

## Reproducer

```bash
cat > test.s <<\EOF
.globl _start
_start:
.data
.byte foo - 1
EOF

cat > foo.s <<\EOF
.global foo
foo = 256
EOF

llvm-mc -filetype=obj -triple=x86_64 foo.s -o foo.o
wild --no-gc-sections test.o foo.o -o out
```

## Current behavior
Relocation 255 outside of bounds [-128, 128)

## Expected behavior
Both GNU ld and lld accept value 255 for `R_X86_64_8` with range `[-128, 255]`.

## Priority
Low  `R_X86_64_8` and `R_X86_64_16` are rarely used in modern code.

deepakshirkem:
R_X86_64_8 and R_X86_64_16 are absolute relocations that should accept the full mixed range (signed minimum, unsigned maximum): [-128, 255] for 8-bit and [-32768, 65535] for 16-bit. Wild was incorrectly using signed-only ranges, rejecting valid values like 255 for R_X86_64_8.

GNU ld and lld both accept the full mixed range.

Fixes #2101

### DIFF
--- a/wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
+++ b/wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
@@ -1,0 +1,37 @@
+// Test that R_X86_64_8 and R_X86_64_16 accept the full mixed range [-128, 255] and
+// [-32768, 65535] respectively, not just the signed range.
+//#Arch:x86_64
+//#LinkArgs:-nostdlib --no-gc-sections
+//#Mode:static
+//#RunEnabled:false
+
+.globl _start
+_start:
+    ret
+
+.data
+.byte foo8_max - 1
+.byte foo8_mid - 1
+.byte foo8_base - 129
+
+.short foo16_max - 1
+.short foo16_mid - 1
+.short foo16_base - 32769
+
+.globl foo8_max
+foo8_max = 256
+
+.globl foo8_mid
+foo8_mid = 128
+
+.globl foo8_base
+foo8_base = 1
+
+.globl foo16_max
+foo16_max = 65536
+
+.globl foo16_mid
+foo16_mid = 32768
+
+.globl foo16_base
+foo16_base = 1
--- a/linker-utils/src/x86_64.rs
+++ b/linker-utils/src/x86_64.rs
@@ -306,6 +306,14 @@ const RELOC_1_BYTE_SIGNED: RelocSizeAndRange = (
     RelocationSize::ByteSize(1),
     AllowedRange::from_byte_size(1, Sign::Signed),
 );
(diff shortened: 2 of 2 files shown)


wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
```diff
@@ -0,0 +1,30 @@
+// Test that R_X86_64_8 and R_X86_64_16 accept the full mixed range [-128, 255] and
+// [-32768, 65535] respectively, not just the signed range.
+//#Arch:x86_64
+//#LinkArgs:-nostdlib --no-gc-sections
+//#Mode:static
+//#RunEnabled:false
+//#SkipLinker:ld
```
davidlattimore: Why disable GNU ld? Generally we want most of our tests to run against at least one of GNU ld or lld so that we can verify that we're testing behaviour common to other linkers. If we do need to disable other linkers, then it's usually good to document why.

wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
```diff
@@ -0,0 +1,30 @@
...
+
+.globl _start
+_start:
+    ret
+
+.data
+.byte foo8_max - 1
```
davidlattimore, line 13: Should we test that the minimum signed value works too?

wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
deepakshirkem: @davidlattimore Thank-you. Removed SkipLinker:ld  both GNU ld and lld accept this test. I will make sure to follow this guideline when writing future tests.

wild/tests/sources/elf/reloc-8-16-range/reloc-8-16-range.s
deepakshirkem: Added signed minimum test cases

davidlattimore (approved):
Thanks :)

PR #2117, opened 2026-06-22
port(wasm): Add per-object data segment layout and emission

lapla-cogito:
This PR implements the data segment layout and emission. Features like applying data relocations (such as `R_WASM_MEMORY_ADDR_*`) and handling passive segments are not included in this PR.

Part of #1431

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -486,6 +486,7 @@ fn run_wasm_test(wat_file: &Path) -> Result {
     let mut extra_objects = Vec::new();
     let mut expect_error = false;
     let mut should_run = true;
+    let mut expected_sections = Vec::new();
     for line in source.lines() {
         if let Some(rest) = line.trim().strip_prefix(";;#") {
             if let Some(obj_name) = rest.strip_prefix("Object:") {
@@ -494,6 +495,8 @@ fn run_wasm_test(wat_file: &Path) -> Result {
                 expect_error = true;
             } else if let Some(val) = rest.strip_prefix("RunEnabled:") {
                 should_run = val.trim().parse().context("Invalid bool for RunEnabled")?;
+            } else if let Some(section) = rest.strip_prefix("ExpectSection:") {
+                expected_sections.push(section.trim().to_owned());
             }
         }
     }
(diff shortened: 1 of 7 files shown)


wild/tests/integration_tests.rs
```diff
@@ -584,13 +587,47 @@ fn run_wasm_test(wat_file: &Path) -> Result {
...
+    }
     if should_run {
         run_wasm_with_wasmtime(&wasm_wild_output, "wild")?;
     }
 
     Ok(())
 }
 
+fn ensure_wasm_section(wasm_file: &Path, section_name: &str, linker_name: &str) -> Result {
+    let output = Command::new("wasm-objdump")
```
davidlattimore: I feel like running an external command to dump stuff might be a little fragile. Why not parse the file and validate the properties we need to that way? Also, we're dumping the file once for each section that we want to check, which seems a little wasteful in terms of test execution time.

wild/tests/integration_tests.rs
```diff
@@ -584,13 +587,47 @@ fn run_wasm_test(wat_file: &Path) -> Result {
     );
 
     validate_wasm(&wasm_wild_output, "wild")?;
+    for section in &expected_sections {
+        ensure_wasm_section(&wasm_wild_output, section, "wild")?;
```
davidlattimore: It'd be good if we could perform the same checks on both wild's output and the reference linkers output. Doing that helps validate that the test is correct and that we're not asserting wild-specific behaviour without knowing it.

wild/tests/integration_tests.rs
lapla-cogito: Changed to use wasmparser.

wild/tests/integration_tests.rs
lapla-cogito: Done

wild/tests/sources/wasm/data-segment/data-segment.wat
```diff
@@ -0,0 +1,11 @@
+;;#RunEnabled:false
+;;#LinkArgs:--no-gc-sections
```
lapla-cogito, line 2: Without this, wasm-ld drops the data segment.

davidlattimore (approved):

PR #2118, opened 2026-06-22
fix: --compress-debug-sections drops merged-string offset map

### LINKED ISSUE #2113: `--compress-debug-sections` breaks `.debug_str_offsets` relocations on large objects: "Failed to find merge-string"
With `--compress-debug-sections=zlib`, wild fails to apply a relocation into a merged
string section when linking a sufficiently large `-g` object:

```
wild: error: Failed copying from out/big.o (512 (2/0)) to output file
  Caused by:
    Failed to apply relocations in section `.debug_str_offsets` of out/big.o
    Failed to apply relocation of type R_X86_64_32 to section `.debug_str` (61 local=5) in file #512 (2/0) (out/big.o) (NON_INTERPOSABLE) at offset 0x8
    Failed to find merge-string at offset 0
```

The same link succeeds with `ld.lld`, and with wild itself once the compress flag is dropped.

## Reproduce

```Makefile
#   make                # 20000-function object, wild + compress=zlib -> FAILS (the bug)
#   make ok-lld         # ld.lld + same flag                          -> links OK
#   make ok-nocompress  # wild, no compress flag                      -> links OK
#   make repro N=5000   # smaller object                              -> links OK (below size threshold)

CLANG ?= clang
WILD  ?= wild
LLD   ?= ld.lld
N     ?= 10000

.DEFAULT_GOAL := repro
.PHONY: gen repro ok-lld ok-nocompress clean

gen: | out
	@awk -v n=$(N) 'BEGIN{for(i=0;i<n;i++)printf "int f%d(int x){return x+%d;}\n",i,i; print "int main(void){return 0;}"}' > out/big.c
	$(CLANG) -g -c out/big.c -o out/big.o
	@echo "generated out/big.c (N=$(N)) and compiled out/big.o"

repro: gen
	$(CLANG) out/big.o -o out/big --ld-path=$(WILD) -Wl,--compress-debug-sections=zlib

ok-lld: gen
	$(CLANG) out/big.o -o out/big --ld-path=$(LLD) -Wl,--compress-debug-sections=zlib

ok-nocompress: gen
	$(CLANG) out/big.o -o out/big --ld-path=$(WILD)

out:
	@mkdir -p out

clean:
	rm -rf out
```
(repro minimized by claude, I ran into this in a regular project trying to compile `gpui-component`)

## Environment

- wild: `Wild 0.9.0 non-git-build (compatible with GNU linkers)`
- rustc (only for the related issue): `1.98.0-nightly (61d7280f3 2026-06-06)`
- clang driver: `clang version 22.1.6`; reference linker that works: `LLD 22.1.6`
- OS: Arch Linux, kernel `7.0.12-arch1-1`, x86_64

kimjune01:
I've been developing some agent tooling to improve solution generalization, and Wild just happened to be welcoming of going out to look for these types of bugs. To be up front, I actually don't understand the specifics of the code being proposed here. However, I did read your contributing.md and made sure to look over past PRs for what counts as a general fix. Here, I directed my agent to use a custom tool to go beyond passing tests. I can provide reasoning traces, demos, and further explanation on how this tool works if you're curious. I hope that better tooling will push the project forward faster. Thanks for review.

---

*Generated change summary (agent):*

Fixes #2113 ("Failed to find merge-string at offset 0" with `--compress-debug-sections` on large `-g` objects).

- **Cause:** after compressing a debug section, `update_allocation_sizes` reset the whole `MergedStringsSection`. For a compressed `.debug_str`, that dropped `string_offsets`, still needed to resolve `.debug_str_offsets` → `.debug_str` relocations applied later. Only hits objects where `.debug_str` is large enough to compress (and so gets cleared).
- **Fix:** clear only `buckets`, not the whole struct. `len()` still goes to 0 (so the compressed section isn't re-emitted) while the offset maps survive for relocation resolution.
- **Test:** `compressed-debug-merge` (clang + DWARF5) — fails on `main`, passes here.

### DIFF
--- a/wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
+++ b/wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
@@ -1,22 +1,23 @@
 //#AbstractConfig:default
-// clang + DWARF5 so the compiler emits `.debug_str_offsets` (gcc keeps string offsets inline);
-// the bug is a relocation from `.debug_str_offsets` into a compressed `.debug_str`.
+// clang + DWARF5 emits `.debug_str_offsets`; gcc keeps the string offsets inline.
 //#Compiler:clang
 //#CompArgs:-g -gdwarf-5
 //#Object:runtime.c
 //#DiffIgnore:section.debug_*
-// `debug_info`: at this scale wild and GNU ld compress `.debug_info` to different sizes; that
-// cross-linker difference is unrelated to the merge-string bug under test.
+// wild and GNU ld compress `.debug_info` to different sizes; same underlying cause as #2119.
 //#DiffIgnore:debug_info
-// wild types `.eh_frame` as SHT_PROGBITS where GNU ld uses SHT_X86_64_UNWIND; a pre-existing
-// cosmetic difference, unrelated to this bug.
+// wild types `.eh_frame` as SHT_PROGBITS, GNU ld as SHT_X86_64_UNWIND.
 //#DiffIgnore:section.eh_frame.type
-// Link-only: this reproduces a link-time failure (#2113), and a statically linked test binary
-// can't be executed in every CI environment.
-//#RunEnabled:false
 
 //#Config:zlib:default
-//#LinkArgs:--no-gc-sections --compress-debug-sections=zlib
+//#LinkArgs:--compress-debug-sections=zlib
+
+//#Config:zstd:default
+//#RequiresLinkerFlags:--compress-debug-sections=zstd
+//#LinkArgs:--compress-debug-sections=zstd
+
+//#Config:none:default
+//#LinkArgs:--compress-debug-sections=none
 
 #include "../common/runtime.h"
 
--- a/wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
+++ b/wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
@@ -1,0 +1,98 @@
+//#AbstractConfig:default
+// clang + DWARF5 so the compiler emits `.debug_str_offsets` (gcc keeps string offsets inline);
+// the bug is a relocation from `.debug_str_offsets` into a compressed `.debug_str`.
+//#Compiler:clang
+//#CompArgs:-g -gdwarf-5
+//#Object:runtime.c
+//#DiffIgnore:section.debug_*
+// `debug_info`: at this scale wild and GNU ld compress `.debug_info` to different sizes; that
+// cross-linker difference is unrelated to the merge-string bug under test.
+//#DiffIgnore:debug_info
+// wild types `.eh_frame` as SHT_PROGBITS where GNU ld uses SHT_X86_64_UNWIND; a pre-existing
+// cosmetic difference, unrelated to this bug.
+//#DiffIgnore:section.eh_frame.type
+// Link-only: this reproduces a link-time failure (#2113), and a statically linked test binary
+// can't be executed in every CI environment.
+//#RunEnabled:false
+
+//#Config:zlib:default
+//#LinkArgs:--no-gc-sections --compress-debug-sections=zlib
+
+#include "../common/runtime.h"
+
+// Regression test for #2113. With `--compress-debug-sections`, a merge-string debug section
+// (`.debug_str`) that actually compresses had its merged-strings map freed after compression,
+// before `.debug_str_offsets` relocations into it were resolved, causing "Failed to find
+// merge-string at offset 0". We need *enough* distinct, compressible debug strings that
+// `.debug_str` compresses (and so is freed); a narrow program doesn't trigger it. Each instance
+// contributes distinct type/member/function names to `.debug_str`.
+#define GENERATE_DEBUG_STUFF(id)                          \
+  struct data_blob_##id {                                 \
+    int field_a_##id;                                     \
+    char field_b_##id[32];                                \
+    double field_c_##id;                                  \
+    float field_d_##id;                                   \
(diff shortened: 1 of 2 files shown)


wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
+//#AbstractConfig:default
+// clang + DWARF5 so the compiler emits `.debug_str_offsets` (gcc keeps string offsets inline);
+// the bug is a relocation from `.debug_str_offsets` into a compressed `.debug_str`.
+//#Compiler:clang
+//#CompArgs:-g -gdwarf-5
+//#Object:runtime.c
+//#DiffIgnore:section.debug_*
+// `debug_info`: at this scale wild and GNU ld compress `.debug_info` to different sizes; that
+// cross-linker difference is unrelated to the merge-string bug under test.
```
lapla-cogito: I think statements like "This difference is unrelated to this bug" shouldn't be included in tests. If any test directive requires a special excuse, explaining why it's necessary should be sufficient for future readers.

wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
...
+//#DiffIgnore:debug_info
+// wild types `.eh_frame` as SHT_PROGBITS where GNU ld uses SHT_X86_64_UNWIND; a pre-existing
+// cosmetic difference, unrelated to this bug.
+//#DiffIgnore:section.eh_frame.type
+// Link-only: this reproduces a link-time failure (#2113), and a statically linked test binary
+// can't be executed in every CI environment.
+//#RunEnabled:false
```
lapla-cogito: I don't think it's correct to say that it cannot be run in a CI environment.

wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
...
+
+//#Config:zlib:default
+//#LinkArgs:--no-gc-sections --compress-debug-sections=zlib
```
lapla-cogito: Why `--no-gc-sections` here?

wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
...
+
```
lapla-cogito, line 21: It would be good to also test zstd cases.

lapla-cogito (changes requested):

wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
...
+// the bug is a relocation from `.debug_str_offsets` into a compressed `.debug_str`.
+//#Compiler:clang
+//#CompArgs:-g -gdwarf-5
+//#Object:runtime.c
+//#DiffIgnore:section.debug_*
+// `debug_info`: at this scale wild and GNU ld compress `.debug_info` to different sizes; that
+// cross-linker difference is unrelated to the merge-string bug under test.
+//#DiffIgnore:debug_info
+// wild types `.eh_frame` as SHT_PROGBITS where GNU ld uses SHT_X86_64_UNWIND; a pre-existing
+// cosmetic difference, unrelated to this bug.
```
davidlattimore: Can probably remove everything from the ';' onwards. Talking about "this bug" makes sense in the context of a code review, but less so afterwards.

wild/tests/sources/elf/compressed-debug-merge/compressed-debug-merge.c
```diff
@@ -0,0 +1,98 @@
+//#AbstractConfig:default
+// clang + DWARF5 so the compiler emits `.debug_str_offsets` (gcc keeps string offsets inline);
+// the bug is a relocation from `.debug_str_offsets` into a compressed `.debug_str`.
+//#Compiler:clang
+//#CompArgs:-g -gdwarf-5
+//#Object:runtime.c
+//#DiffIgnore:section.debug_*
+// `debug_info`: at this scale wild and GNU ld compress `.debug_info` to different sizes; that
```
davidlattimore: I don't think the issue is different sizes. This is the same problem as noted in #2119. A diffignore is reasonable for the time being.

davidlattimore:
Oh, and thanks for being up-front an explicit about where stuff came from :)

kimjune01:
Thanks both, all addressed in f48267d:

- **\`--no-gc-sections\`**: removed. It had no effect on the repro: \`--gc-sections\` only collects allocatable sections, and the debug sections involved are non-alloc, so they survive regardless. Confirmed the test still fails on \`main\` and passes here without it.
- **\`RunEnabled:false\`**: removed; the test now runs like the sibling \`compressed-debug\`. You're right it runs in CI: I'd over-generalized from a local Rosetta \`bss_size overflow\` that turns out to hit every static run-test here, not just this one.
- **zstd**: added \`zstd\` and \`none\` configs mirroring \`compressed-debug.c\`. Only \`zlib\` independently reproduces #2113 on \`main\`; the others are compress-path coverage plus an uncompressed control.
- **DiffIgnore comments**: dropped the "unrelated to this bug" framing on all three; each now just states what differs, and the \`.debug_info\` one points at #2119 as the shared cause rather than claiming a size difference.

davidlattimore (approved):

lapla-cogito (approved):

PR #2119, opened 2026-06-22
fix: Don't compress `SHF_ALLOC` debug sections

### LINKED ISSUE #2112: `--compress-debug-sections` fails on gdb debug script due to "Insufficient space allocated to section `.debug_gdb_scripts`"
Linking a Rust binary that embeds a GDB pretty-printer via
`#![debugger_visualizer(gdb_script_file = ...)]` fails under wild when
`--compress-debug-sections=zlib` (or `=zstd`) is passed. 

```
wild: error: Failed copying from viz.viz.8b01e95c87876ea7-cgu.0.rcgu.o (260 (1/4)) to output file
  Caused by:
    Insufficient space allocated to section `.debug_gdb_scripts`. Tried to take 59 bytes, but only 0 remain
```

The same link succeeds with `ld.lld`, and with wild itself once the compress flag is dropped.

## Repro
**viz.rs**
```rs
#![debugger_visualizer(gdb_script_file = "viz.gdb")]

fn main() {}
```
**viz.gdb**
```gdb
x
```

**Makefile**
```Makefile
#   make                # wild + --compress-debug-sections=zlib  -> FAILS (the bug)
#   make ok-lld         # ld.lld + same flag                     -> links OK
#   make ok-nocompress  # wild, no compress flag             -> links OK
#   make clean

RUSTC ?= rustc
CLANG ?= clang
WILD  ?= wild
LLD   ?= ld.lld

# rustc emits .debug_gdb_scripts only when debug info is on, hence -g.
BASE = $(RUSTC) --edition 2021 -g ../viz.rs -o viz -Clinker=$(CLANG)
COMPRESS = -Clink-arg=-Wl,--compress-debug-sections=zlib

.DEFAULT_GOAL := repro
.PHONY: repro ok-lld ok-nocompress check clean

repro: | out
	cd out && $(BASE) -Clink-arg=--ld-path=$(WILD) $(COMPRESS)

ok-lld: | out
	cd out && $(BASE) -Clink-arg=--ld-path=$(LLD) $(COMPRESS)

ok-nocompress: | out
	cd out && $(BASE) -Clink-arg=--ld-path=$(WILD)

out:
	@mkdir -p out

clean:
	rm -rf out
```

lapla-cogito:
Rust's `.debug_gdb_scripts` section is marked `SHF_ALLOC`. Compressing it zeroed the object file buffer while the section was still copied via `write_object_section()` instead of being skipped like other compressed debug sections.

Fixes #2112


wild/tests/sources/elf/compressed-debug-gdb-scripts/compressed-debug-gdb-scripts.rs
```diff
@@ -0,0 +1,19 @@
+//#AbstractConfig:default
+//#ExpectSection:.debug_gdb_scripts
+//#DiffIgnore:section.debug_*
+//#SkipArch: ppc64le
+
+//#Config:zlib:default
+//#CompArgs:-g -Clink-arg=-Wl,--compress-debug-sections=zlib
+// TODO: wild should emit one zlib stream per compressed section
+//#DiffIgnore:debug_info
```
lapla-cogito, line 9: The diff is
```
debug_info
  wild Uncompressed data size does not match compression header
  ld OK
```

Wild should emit one zlib stream per compressed section. Large `.debug_info` here is split into concatenated 64 KiB zlib streams, so standard decompressors only read the first chunk and the size no longer matches `ch_size`.

davidlattimore: Hopefully it's possible to still compress multi-threaded

davidlattimore (approved):

Issue #2065, opened 2026-06-14
Mach-O: support for fat objects archives and shared libraries


marxin:
The Mach-O can contain an object or an archive that contains sub-objects for various different architectures. One example of that is the Clang's standard library used by default when linking a C object:
```
❯ lipo -info /opt/homebrew/Cellar/llvm/22.1.3/lib/clang/22/lib/darwin/libclang_rt.osx.a
Architectures in the fat file: /opt/homebrew/Cellar/llvm/22.1.3/lib/clang/22/lib/darwin/libclang_rt.osx.a are: i386 x86_64 x86_64h arm64 arm64e
```

We should just inspect the objects (or libraries) and consider `aarch64` parts of them. Object's documentation:
https://docs.rs/object/latest/object/macho/struct.FatArch64.html

PR #2122, opened 2026-06-22
port(MachO): Add support for fat objects

davidlattimore:
Issue #2065

### DIFF
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -26,8 +26,10 @@ use crate::verbose_timing_phase;
 use colosseum::sync::Arena;
 use crossbeam_queue::SegQueue;
 use hashbrown::HashMap;
+use itertools::Itertools as _;
 #[cfg(not(target_family = "wasm"))]
 use memmap2::Mmap;
+use object::read::macho::FatArch as _;
 use rayon::Scope;
 use rayon::iter::IntoParallelIterator;
 use rayon::iter::IntoParallelRefIterator;
@@ -138,6 +140,7 @@ impl Deref for FileBytes {
 #[derive(Clone, Copy)]
 pub(crate) struct InputRef<'data> {
     pub(crate) file: &'data InputFile,
+    pub(crate) data: &'data [u8],
     pub(crate) entry: Option<archive::EntryMeta<'data>>,
 }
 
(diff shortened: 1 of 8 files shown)


davidlattimore:
lol. Tests failed for Build Windows. I saw that the failure was in the checkout step and assumed it was a flake. Reran. Same failure. Looked closer and found that it was because one of the filenames I had used for a test was a reserved name on windows. Windows is weird.

libwild/src/input_data.rs
```diff
@@ -591,6 +597,32 @@ fn process_thin_archive<'data, P: Platform>(
...
+        entry: None,
+    };
+
+    match kind {
+        FileKind::Archive => process_archive(&input_ref, file, state),
+        FileKind::MachOObject => {
+            let parsed = state.process_input(input_ref, file, kind)?;
+            Ok(LoadedFileState::Loaded(input_ref.file, parsed))
+        }
+        _ => bail!("Unsupported file type {kind} found in fat object"),
```
marxin: Let's use `FAT object` for consistency with the error message at the start of this function.

libwild/src/input_data.rs
```diff
@@ -773,6 +799,32 @@ impl<'data, P: Platform> TemporaryState<'data, P> {
     }
 }
 
+fn select_fat_entry_for_cpu_type(
+    data: &[u8],
+    target_cpu_type: object::macho::CpuType,
+) -> Result<&[u8]> {
+    let parsed = object::read::macho::MachOFatFile32::parse(data)?;
```
marxin, line 806: What about reporting an error if the file starts with `MachOFatFile64`?

marxin (approved):
Really nice, smooth integration.

libwild/src/input_data.rs
davidlattimore: I just looked into doing this. Turns out we don't actually get here if it's a `MachOFatFile64`. Instead we report that we failed to identify the file type. But in the process of checking that I ended up adding support for having the tests generate the 64 bit variant. Once I'd done that, it then wasn't too much work to actually make them work, so I did that. But it felt like it should probably be a separate PR, so I'll upload that once I merge this.

Issue #2114, opened 2026-06-21
Wild does not error on invalid relocations against DSO symbols and local symbols in PIE/shared objects


deepakshirkem:
## Description

When linking PIE or shared objects, Wild silently succeeds on relocations that should be rejected. Both GNU ld and lld correctly error with "recompile with -fPIC".

## Reproducer

```bash
cat > shared.s <<\EOF
.global zed
zed:
EOF

cat > test.s <<\EOF
.global _start, main, data
_start:
  ret
main:
  ret
.data
data:
.long zed        
.long zed - .    
.rodata
.quad _start     
.quad main       
.quad data       
EOF

llvm-mc -filetype=obj -triple=x86_64 shared.s -o shared.o
llvm-mc -filetype=obj -triple=x86_64 test.s -o test.o
ld.lld shared.o -shared -o shared.so
wild -pie test.o shared.so -o out
```

## Current behavior
Wild silently succeeds producing a broken binary.

## Expected behavior
lld errors with: 
error: relocation R_X86_64_32 cannot be used against symbol 'zed'; recompile with -fPIC
error: relocation R_X86_64_PC32 cannot be used against symbol 'zed'; recompile with -fPIC
error: relocation R_X86_64_64 cannot be used against symbol '_start'; recompile with -fPIC

deepakshirkem:
I would like to work on this

deepakshirkem:
All cases from this issue have been addressed:
#2125 — R_X86_64_32/R_X86_64_32S against DSO symbols in PIE
#2132 — R_X86_64_PC32 against DSO symbols in PIE
#2133 — R_X86_64_64 to local symbols in non-writable sections (already worked, test added)

Happy to close once the remaining PRs merge.

deepakshirkem:
@davidlattimore, We can close this as completed.

PR #2125, opened 2026-06-22
fix: Error on R_X86_64_32/32S against DSO symbols in PIE output

deepakshirkem:
Wild was silently succeeding when R_X86_64_32 or R_X86_64_32S relocations were used against DSO symbols in PIE output. Both GNU ld and lld correctly error with "recompile with -fPIC".

Issue #2114

### DIFF
--- a/libwild/src/output_kind.rs
+++ b/libwild/src/output_kind.rs
@@ -72,7 +72,7 @@ impl OutputKind {
                 | OutputKind::StaticExecutable(RelocationModel::Relocatable)
         )
     }
-    
+
     pub(crate) fn needs_dynsym(self) -> bool {
         matches!(
             self,
--- a/libwild/src/output_kind.rs
+++ b/libwild/src/output_kind.rs
@@ -88,4 +88,13 @@ impl OutputKind {
             OutputKind::DynamicExecutable(_) | OutputKind::SharedObject
         )
     }
+
+    pub(crate) fn is_position_independent(self) -> bool {
+        matches!(
+            self,
+            OutputKind::SharedObject
+                | OutputKind::DynamicExecutable(RelocationModel::Relocatable)
+                | OutputKind::StaticExecutable(RelocationModel::Relocatable)
+        )
+    }
 }
--- a/libwild/src/output_kind.rs
+++ b/libwild/src/output_kind.rs
@@ -64,6 +64,15 @@ impl OutputKind {
         )
     }
 
+    pub(crate) fn is_position_independent(self) -> bool {
+        matches!(
+            self,
+            OutputKind::SharedObject
+                | OutputKind::DynamicExecutable(RelocationModel::Relocatable)
+                | OutputKind::StaticExecutable(RelocationModel::Relocatable)
+        )
+    }
+    
     pub(crate) fn needs_dynsym(self) -> bool {
         matches!(
             self,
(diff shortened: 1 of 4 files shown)


wild/tests/sources/elf/pie-invalid-reloc/pie-invalid-reloc.s
```diff
@@ -0,0 +1,14 @@
+// Test that R_X86_64_32 to a DSO symbol errors in PIE/shared output.
+// Both GNU ld and lld error with "recompile with -fPIC".
+//#Arch:x86_64
+//#Mode:dynamic
+//#Shared:pie-invalid-reloc-dso.s
+//#LinkArgs:-pie --no-gc-sections
+//#ExpectError:R_X86_64_32
+//#RunEnabled:false
```
davidlattimore: Note that ExpectError implies "RunEnabled:False"

davidlattimore:
> * How to distinguish valid vs invalid `R_X86_64_PC32` in PIE context

Possibly whether the target of the relocation has `ValueFlags::FUNCTION` set? But I'd say it's worthwhile experimenting a bit to see in which cases the other linkers accept or error.

> * Whether `R_X86_64_64` to local symbols should error or emit dynamic reloc

If lld errors here, then I'd say it'd be good to do the same. We can always relax it later to accept it like GNU ld if it turns out to be problematic. But it also depends a bit on how hard the implementation is. If it ends up significantly simpler to accept rather than error, then I'd be open to that.

davidlattimore:
It's fine, possibly even preferable to fix each case in a separate PR. So don't feel that you need to fix them all in this PR. Note that github isn't very good at reading intention when issues are referenced, so saying "partial fix for X" is just as likely to close X when the PR is merged. I usually just put "Issue X", since that definitely doesn't close the issue.

wild/tests/sources/elf/pie-invalid-reloc/pie-invalid-reloc.s
deepakshirkem: addressed this comment

deepakshirkem:
@davidlattimore I have updated the PR to only fix R_X86_64_32/32S. I will address R_X86_64_PC32 and R_X86_64_64 cases in separate PRs.

libwild/src/output_kind.rs
```diff
@@ -88,4 +88,13 @@ impl OutputKind {
             OutputKind::DynamicExecutable(_) | OutputKind::SharedObject
         )
     }
+
+    pub(crate) fn is_position_independent(self) -> bool {
```
davidlattimore: This method seems very similar to `is_relocatable` (above). The only difference being the treatment of `OutputKind::Relocatable` (partial linking). Any thoughts on what's best here? If nothing else, we should perhaps put this method next to the other one to make it easier to see that they're similar.

deepakshirkem: @davidlattimore, Yes, you're correct. But we don't want to error on relocations in partial link output since it's an intermediate step. `is_position_independent` excludes that case. I can rename it or move it below `is_relocatable`.

davidlattimore: Yeah, that makes sense. TBH, I wonder how many of the callers of `is_relocatable` should actually be calls to the new `is_position_independent`. Partial linking isn't super thoroughly tested, so it wouldn't be surprising if we were getting some such aspects of it wrong. I just moved the method, so we can merge it once CI passes.

davidlattimore (approved):

PR #2130, opened 2026-06-23
chore: Run tidy tests on all platforms

davidlattimore:


mati865:
Should these checks be limited at all? They look like they should be runnable on all systems.

davidlattimore:
> Should these checks be limited at all? They look like they should be runnable on all systems.

Done

PR #2132, opened 2026-06-23
fix: Error on R_X86_64_PC32 to DSO symbols in non-executable sections in PIE

deepakshirkem:
R_X86_64_PC32 to a DSO symbol in a non-executable section (e.g. .data) is invalid in PIE output. Wild was silently producing a broken binary. R_X86_64_PC32 in executable sections (.text) remains valid as it goes via PLT.

lld errors in this case; GNU ld only warns.

Issue #2114

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -2889,6 +2889,7 @@ fn apply_relocation<
     let flags = layout.flags_for_symbol(local_symbol_id);
     if layout.symbol_db.output_kind.is_position_independent()
         && (flags.is_interposable() || flags.is_dynamic())
+        && !flags.needs_copy_relocation()
         && A::is_illegal_in_shared_object(r_type)
     {
         bail!(
--- a/wild/tests/sources/elf/pie-pc32-dso-code-ref-code/pie-pc32-dso-code-ref-code.s
+++ b/wild/tests/sources/elf/pie-pc32-dso-code-ref-code/pie-pc32-dso-code-ref-code.s
@@ -1,0 +1,17 @@
+// Scenario: code references DSO function via PLT call - valid in PIE
+//#Arch:x86_64
+//#Mode:dynamic
+//#Shared:pie-pc32-dso-shared-fn.s
+//#SoSingleLinker:wild
+//#LinkArgs:-pie
+//#RunEnabled:false
+//#EnableLinker:lld
+//#SkipLinker:ld
+//#DiffIgnore:.dynamic.DT_FLAGS_1.NOW
+//#DiffIgnore:.dynamic.DT_RELA
+//#DiffIgnore:.dynamic.DT_RELAENT
+//#DiffIgnore:section.got.plt.entsize
+.global _start
+_start:
+    call zed_fn
+    ret
--- a/libwild/src/elf_x86_64.rs
+++ b/libwild/src/elf_x86_64.rs
@@ -68,7 +68,10 @@ impl crate::platform::Arch for ElfX86_64 {
     }
 
     fn is_illegal_in_shared_object(r_type: u32) -> bool {
-        matches!(r_type, object::elf::R_X86_64_32 | object::elf::R_X86_64_32S)
+        matches!(
+            r_type,
+            object::elf::R_X86_64_32 | object::elf::R_X86_64_32S | object::elf::R_X86_64_PC32
+        )
     }
 
     fn get_dynamic_relocation_type(relocation: DynamicRelocationKind) -> u32 {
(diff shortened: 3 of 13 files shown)


deepakshirkem:
@davidlattimore Please review when you get a chance.

libwild/src/elf_writer.rs
```diff
@@ -2889,6 +2889,7 @@ fn apply_relocation<
     let flags = layout.flags_for_symbol(local_symbol_id);
     if layout.symbol_db.output_kind.is_position_independent()
         && (flags.is_interposable() || flags.is_dynamic())
+        && !section_info.section_flags.contains(shf::EXECINSTR)
```
davidlattimore: It seems to me that there might be 4 different scenarios:

* code references code
* data references code
* code references data
* data references data

My guess is that the first two are likely OK, while the second two are likely not OK. It seems that you're checking whether the section to which the relocation is applied is executable. But, what matters is probably not where the reference is from, but what it's a reference to. Is it possible to make tests for all 4 combinations?

wild/tests/sources/elf/pie-pc32-dso-code-ref-code/pie-pc32-dso-code-ref-code.s
```diff
@@ -0,0 +1,10 @@
+//#Arch:x86_64
+//#Mode:dynamic
+//#Shared:pie-pc32-dso-shared-fn.s
+//#LinkArgs:-pie
+//#RunEnabled:false
+//#SkipLinker:ld
```
davidlattimore, line 9: I'm guessing that GNU ld just issues a warning. Are we matching lld's behaviour for all these cases? It'd be good to enable lld so that we can verify that we are.

deepakshirkem: @davidlattimore Added EnableLinker:lld for cases 1, 2 and 4. 
For case 3, lld accepts R_X86_64_PC32 to a FUNC DSO symbol via canonical PLT, so it doesn't error  only Wild errors in that case, hence SkipLinker:lld there.

davidlattimore: The comment in the test says:

```
// SkipLinker:lld because the test framework passes -pie to the shared object build,
// which conflicts with lld's -shared flag. Wild correctly errors on this case.
```

It's hard for me to see that this is the same reason. Can you clarify?

Also, I think we have a way to override the linker flags used when linking a shared object. If we don't, then we should. I think you can list them after a ':' on the `#Shared` line.

I'm fuzzy on where we currently stand with canonical PLT entries. I know I don't like them (much like copy relocations), but I thought we did them.

wild/tests/sources/elf/pie-pc32-dso-code-ref-code/pie-pc32-dso-code-ref-code.s
davidlattimore, line 1: If it's not too much trouble, could we rename the tests to indicate which scenario they are - e.g. instead of numbered cases include "data-ref-data", "data-ref-code" etc in the name. Alternatively, just add a one line comment to the start of each test saying which scenario it is.

deepakshirkem: Yeah, that makes more sense. I have updated the test names to be more descriptive and also added a one line comment at the beginning of each test indicating the scenario.

wild/tests/sources/elf/pie-pc32-dso-code-ref-code/pie-pc32-dso-code-ref-code.s
deepakshirkem: >The comment in the test says:

>// SkipLinker:lld because the test framework passes -pie to the shared object build,
// which conflicts with lld's -shared flag. Wild correctly errors on this case.
It's hard for me to see that this is the same reason. Can you clarify?

>Also, I think we have a way to override the linker flags used when linking a shared object. If we don't, then we should. I think you can list them after a ':' on the #Shared line.

The comment was misleading  the real reason for SkipLinker:lld is that lld creates a canonical PLT entry for R_X86_64_PC32 to STT_FUNC DSO symbols and succeeds, while Wild doesn't support canonical PLT and errors instead. Also found LinkSoArgs directive which can override shared object linker flags will use that to clean up the test once canonical PLT behavior is decided.

>I'm fuzzy on where we currently stand with canonical PLT entries. I know I don't like them (much like copy relocations), but I thought we did them.

After checking, Wild does not implement canonical PLT entries. For case 3 (R_X86_64_PC32 to STT_FUNC DSO symbol from .text), lld succeeds via canonical PLT while Wild errors. For now I've kept Wild erroring since it's safer than silently producing a broken binary. If canonical PLT support is needed, I can file a separate issue and fix it in a follow-up PR.

davidlattimore (approved):
Thanks!

PR #2133, opened 2026-06-24
test: Add test for R_X86_64_64 to local symbols in read-only sections in PIE

deepakshirkem:
While investigating #2114, discovered Wild already correctly errors on R_X86_64_64 to local symbols in non-writable sections in PIE. Add a test to document and prevent regression.

Issue #2114

### DIFF
--- a/wild/tests/sources/elf/pie-r64-local-rodata/pie-r64-local-rodata.s
+++ b/wild/tests/sources/elf/pie-r64-local-rodata/pie-r64-local-rodata.s
@@ -1,0 +1,16 @@
+//#Arch:x86_64
+//#Mode:dynamic
+//#LinkArgs:-pie
+//#EnableLinker:lld
+//#ExpectError:R_X86_64_64
+//#SkipLinker:ld
+.global _start, foo, rodata_ref
+_start:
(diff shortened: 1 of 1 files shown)


wild/tests/sources/elf/pie-r64-local-rodata/pie-r64-local-rodata.s
```diff
@@ -0,0 +1,15 @@
+//#Arch:x86_64
+//#Mode:dynamic
+//#LinkArgs:-pie
+//#ExpectError:R_X86_64_64
+//#SkipLinker:ld
```
davidlattimore, line 6: Could be good to enable lld so that we're verifying that we have the same behaviour as at least one other linker

davidlattimore (commented):
Thanks :)

wild/tests/sources/elf/pie-r64-local-rodata/pie-r64-local-rodata.s
deepakshirkem: Sorry,  added EnableLinker:lld to verify behavior matches lld.

davidlattimore (approved):

PR #2135, opened 2026-06-24
port(wasm): Apply data relocations

lapla-cogito:
Resolve `MEMORY_ADDR_*` and `FUNCTION_INDEX_I32` relocations in data segment payloads after per-object layout.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -431,7 +431,6 @@ pub(crate) struct WasmDataSegmentLayout<'data> {
 #[derive(Debug, Default)]
 pub(crate) struct WasmObjectDataLayout<'data> {
     pub(crate) file_id: crate::input_data::FileId,
-    pub(crate) symbol_id_range: crate::symbol_db::SymbolIdRange,
     pub(crate) segments: Vec<WasmDataSegmentLayout<'data>>,
 }
 
@@ -1748,7 +1747,6 @@ fn layout_object_data<'data>(
     }
     Ok(WasmObjectDataLayout {
         file_id: input.file_id,
-        symbol_id_range: input.symbol_id_range,
         segments,
     })
 }
@@ -1800,6 +1798,7 @@ pub(crate) struct WasmObjectIndexMap {
     pub(crate) function_indices: Vec<u32>,
     pub(crate) global_indices: Vec<u32>,
     pub(crate) memory_indices: Vec<u32>,
+    pub(crate) data_addresses: Vec<u32>,
 }
 
 impl WasmObjectIndexMap {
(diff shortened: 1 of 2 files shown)


libwild/src/wasm.rs
```diff
@@ -2462,10 +2475,182 @@ where
...
+        reloc,
+        def_obj_idx,
+        def_sym,
+        layout_inputs,
+        object_data_layouts,
+        object_index_maps,
+    )
+}
+
+fn resolve_data_relocations<'data>(
```
davidlattimore: Why store the resolutions for all the relocations rather than looking up the resolutions for each relocation at write time? For ELF, we store resolutions per symbol (in one large Vec), but we don't store any per-relocation state

libwild/src/wasm.rs
```diff
@@ -2462,10 +2475,182 @@ where
...
+    object_data_layouts: &mut [WasmObjectDataLayout<'data>],
+    layout_inputs: &[WasmObjectLayoutInput<'data>],
+    object_index_maps: &[WasmObjectIndexMap],
+    symbol_db: &SymbolDb<'data, Wasm>,
+) -> Result<()> {
+    let file_id_to_index: HashMap<crate::input_data::FileId, usize> = layout_inputs
```
davidlattimore, line 2533: It looks like this hashmap is effectively providing a way to index files sequentially rather than by group. Grouping ends up being important for keeping overheads low when doing allocation of space in the output file. This ends up meaning that the group is the unit of parallelism for a lot of operations. That said, this only matters when we're doing resolution and space allocation in parallel, which it looks like we're not at the moment. What's your plan regarding parallelism and performance for the wasm port?

davidlattimore (commented):
Are there any new tests that could reasonably be added to test this?

lapla-cogito:
> Are there any new tests that could reasonably be added to test this?

Generating data relocations from WAT using `wat2wasm` is somewhat difficult (see also https://github.com/WebAssembly/wabt/issues/965#issuecomment-443283311). While it's technically possible to inject `reloc.DATA` entries in the integration tests, I believe this would be too ad hoc.
As discussed during the last meeting, the remaining features needed to enable linking of minimal object files generated by clang (roughly memory import) are few in number. Therefore, for integration testing purposes, I think that verifying these features through C code alone would suffice.

libwild/src/wasm.rs
lapla-cogito: Groups should become the unit of parallelism for per-object work, same as ELF. This lookup would then become group-aware (or be replaced by the group indexing we already use). For now the priority is the data model that enables that shape (per-symbol addresses, apply-at-write); the sequential `file_id_to_index` map is only a bridge for today's still-partially-sequential  multi-object layout, not something I expect to keep long-term.

libwild/src/wasm.rs
lapla-cogito: I implemented in this way in anticipation of long-term parallelization plans, but after reconsideration, I realized this was unnecessary and switched to the same processing as ELF. For code sections as well.

davidlattimore (approved):

PR #2136, opened 2026-06-24
feat: use location counters within output sections in linker scripts

vishruth-thimmaiah:
This PR adds support for setting the location counter using the `.` keyword within an output section.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -410,7 +410,7 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>>(
     // Evaluate ASSERT commands from all linker scripts now that layout is complete.
     crate::expression_eval::evaluate_assertions(
         &symbol_db,
-        &section_layouts,
+        &merged_section_layouts,
         &output_sections,
         &symbol_resolutions.resolutions,
         sizeof_headers,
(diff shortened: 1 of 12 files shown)


libwild/src/layout.rs
```diff
@@ -5126,6 +5137,8 @@ fn layout_section<'data, P: Platform>(
                 let section_info = output_sections.output_info(section_id);
                 let part_id_range = section_id.part_id_range();
                 let max_alignment = sizes.max_alignment(part_id_range.clone(), output_sections);
+                let merge_target = output_sections.primary_output_section(section_id);
+                let section_flags = output_sections.section_flags(merge_target);
```
davidlattimore: Given that these two variables are only used in one code path below, can you move them into the scope where they're used?

wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter.ld
```diff
@@ -0,0 +1,21 @@
+ENTRY(begin_here)
+
+SECTIONS
+{
+    . = 0x300000 * 2;
+	start_of_sections = .;
+    .text : {
+        . = 0x10;
```
davidlattimore, line 9: Should we have some assertions regarding the address of .text and the symbols in it?

vishruth-thimmaiah: We cannot assert the exact address of .text as different linkers place it in different locations, I've added assertions for the offsets.

davidlattimore (approved):

PR #2137, opened 2026-06-25
test: Switch from Enable/SkipLinker to ReferenceLinkers

davidlattimore:


davidlattimore:
Interested as to what folks think about this. The idea is to specify all the reference linkers in a single directive rather than having two separate directives to enable/disable them.

I feel that with "SkipLinker" / "EnableLinker" is that it's not always immediately clear what linkers you're running with, especially when inheriting the config. Also, it's a minor thing but the names aren't symetric - the opposite of "Enable" should be "Disable". I figured, while making this change, that I'd switch from calling GNU ld "ld" to calling it "bfd", since "ld" seems a little vague, especially if "ld" can be different things on different platforms.

PR #2141, opened 2026-06-26
chore: Fix build

davidlattimore:


davidlattimore:
#2139 and #1994 both passed independently, but together, it seems they broke compilation

PR #2145, opened 2026-06-27
port(wasm): Link clang objects with memory, stack, and data addresses

lapla-cogito:
Teach the Wasm port enough linker-defined state to run minimal clang `-nostdlib` objects: absorb linear-memory and `__stack_pointer` imports, place static data at the wasm-ld default base (1024), synthesise a stack pointer global, and export `_start` by default.

This allows Wild to link basic C programs.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -11,6 +11,7 @@ use crate::layout;
 use crate::layout_rules::SectionKind;
 use crate::output_section_id::SectionName;
 use crate::platform;
+use crate::symbol::UnversionedSymbolName;
 use crate::symbol_db::SymbolDb;
 use crate::wasm_writer::OutputExport;
 use crate::wasm_writer::OutputGlobal;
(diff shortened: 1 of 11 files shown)


libwild/Cargo.toml
```diff
@@ -64,7 +64,7 @@ perf-event.workspace = true
 nix.workspace = true
 
 [build-dependencies]
-cc = "1.1.12"
+cc = "1.2.0"
```
lapla-cogito: This was necessary to pass the cargo-minimal-versions check (the same applies to updating other dependencies).

davidlattimore (commented):
It looks like this increases our CI time quite a lot. What's the reason for switching to using the wasmtime crate rather than the binary? I recall checking the exit codes was possibly part of the reason? But it looks like you're trapping when the test fails, which I would have assumed would have worked with the wasmtime binary too?

lapla-cogito:
> It looks like this increases our CI time quite a lot.

The actual increase in time seems to be occured only in Nix CI.

edit: Ah, I didn't notice that some Linux jobs also had about ~50s increase (I was only looking at the rough time summary in the bottom of this tab). Hmm, that's a bit difficult...

> What's the reason for switching to using the wasmtime crate rather than the binary? I recall checking the exit codes was possibly part of the reason?

Yes, the primary reason is to check return values when invoking functions, with an additional motivation to make testing easier as we advance WASI support (this point came up after the previous meeting).  In other words, I think using Wasmtime API is eventually preferable at the Wasm port. However, implementing return value validation requires WASI support that can dynamically satisfy imports like `proc_exit`. Nevertheless, one practical reason for adopting Wasmtime APIs at this stage is that certain features added in this PR, such as [verifying the presence of `_start` exports](https://github.com/lapla-cogito/wild/blob/be24bbdf28378dd1e58a7cbf7ffa5ec82b370fc4/wild/tests/integration_tests.rs#L857-L861), work better with this.

davidlattimore:
How practical would it be to separate the change to the test infrastructure changes from the primary stated purpose of this PR? It seems like they should ideally be separate if possible. That would allow us to move forward with this PR, while having a separate PR to look at whether the benefits of a direct dependency on wasmtime are worth the costs.

Verifying the type signature of the "_start" function seems like something a wasm runtime might do anyway, even if it weren't built into our test infrastructure.

If we do end up needing a wasm runtime, I wonder if a lighter weight one might be enough. e.g. I see there's a tinywasm crate.

lapla-cogito:
I've disabled the WasmTime API at least for now (I didn't expect the CI time to increase this much).

wild/tests/integration_tests.rs
```diff
@@ -470,21 +467,162 @@ fn collect_wasm_tests(tests: &mut Vec<Trial>, filter: &Filter) -> Result {
...
+        } else {
+            tests.push(libtest_mimic::Trial::test(full_name, move || {
+                run_wasm_test(&wat_file).map_err(|e| libtest_mimic::Failed::from(e.to_string()))
+            }));
+        }
         Ok(())
     })
 }
 
+fn run_wasm_clang_test(c_file: &Path) -> Result {
```
davidlattimore, line 545: Any thoughts on unifying config parsing? We seem to be going from having two config parsing - one for ELF+Mac and one for Wasm to three - one for ELF+Mac and two for wasm

libwild/src/wasm.rs
```diff
@@ -2411,6 +2443,168 @@ fn resolve_one_import<'data>(
...
+{
+    if export_name_exists(exports, entry_name) {
+        return Ok(());
+    }
+
+    let objects: Vec<_> = layout::objects_iter(groups)
+        .map(|state| (state.file_id, &state.object))
+        .collect();
+
+    for (input, index_map) in layout_inputs.iter().zip(object_index_maps) {
```
davidlattimore: Do we need to scan every symbol of every file in order to find the entry symbol? For ELF, we just do a hashmap lookup in the existing SymbolDb and from that we know which symbol in which file is the entry point.

lapla-cogito: Neat, thanks.

wild/tests/integration_tests.rs
lapla-cogito: I was considering merging directive processing for Wasm and ELF/Mach-O in a separate PR (we previously discussed merging directives once additional ones were added in another PR), so I was wondering how much merging should happen in this PR. However, I've decided to integrate WAT and C directive parsing instead.

libwild/src/wasm.rs
```diff
@@ -2411,6 +2444,170 @@ fn resolve_one_import<'data>(
...
+        return;
+    }
+    exports.push(OutputExport {
+        name: "memory",
+        kind: wasmparser::ExternalKind::Memory,
+        index: 0,
+    });
+}
+
+fn export_name_exists(exports: &[OutputExport<'_>], name: &str) -> bool {
```
davidlattimore, line 2544: It'd be good at some point to get rid of this as it effectively turns what should be O(n) operations into O(n^2) operations. i.e. every time we export a function, we scan the list of all previously exported functions. For ELF, this seems like the kind of thing we'd represent via setting a bit in the ValueFlags of a symbol. Probably the existing `EXPORT_DYNAMIC` flag would be the one that would make sense. But this can wait until later.

davidlattimore (approved):

PR #2146, opened 2026-06-27
fix: Reject R_X86_64_8/16/PC8/PC16 against preemptible symbols in shared objects

### LINKED ISSUE #2115: Wild does not error on sub-pointer-size relocations against preemptible symbols in shared objects
## Reproducer

```bash
cat > test.s <<\EOF
.global foo
.data
.byte foo
.short foo
.byte foo - .
.short foo - .
EOF

llvm-mc -filetype=obj -triple=x86_64-pc-linux test.s -o test.o
wild -shared test.o -o test.so
```

## Current behavior
Wild silently succeeds producing a broken binary.

## Expected behavior
Both GNU ld and lld error:
error: relocation R_X86_64_8 cannot be used against symbol 'foo'; recompile with -fPIC
error: relocation R_X86_64_16 cannot be used against symbol 'foo'; recompile with -fPIC
error: relocation R_X86_64_PC8 cannot be used against symbol 'foo'; recompile with -fPIC
error: relocation R_X86_64_PC16 cannot be used against symbol 'foo'; recompile with -fPIC

deepakshirkem:
R_X86_64_8, R_X86_64_16, R_X86_64_PC8 and R_X86_64_PC16 relocations cannot be used against preemptible symbols in shared objects because the dynamic linker has no way to patch sub-pointer-size values at runtime. Wild was silently succeeding and producing a broken binary.

Both GNU ld and lld error with "recompile with -fPIC".

Fixes #2115

### DIFF
--- a/wild/tests/sources/elf/shared-sub-ptr-reloc/shared-sub-ptr-reloc.s
+++ b/wild/tests/sources/elf/shared-sub-ptr-reloc/shared-sub-ptr-reloc.s
@@ -1,0 +1,10 @@
+//#Arch:x86_64
+//#Mode:dynamic
+//#LinkArgs:-shared --no-gc-sections
+//#ExpectError:R_X86_64_8
+.global foo
+.data
+.byte foo
+.short foo
+.byte foo - .
+.short foo - .
--- a/libwild/src/elf_x86_64.rs
+++ b/libwild/src/elf_x86_64.rs
@@ -70,7 +70,13 @@ impl crate::platform::Arch for ElfX86_64 {
     fn is_illegal_in_shared_object(r_type: u32) -> bool {
         matches!(
             r_type,
-            object::elf::R_X86_64_32 | object::elf::R_X86_64_32S | object::elf::R_X86_64_PC32
+            object::elf::R_X86_64_32
+                | object::elf::R_X86_64_32S
+                | object::elf::R_X86_64_PC32
+                | object::elf::R_X86_64_8
+                | object::elf::R_X86_64_16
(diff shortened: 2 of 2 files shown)


wild/tests/sources/elf/shared-sub-ptr-reloc/shared-sub-ptr-reloc.s
```diff
@@ -0,0 +1,12 @@
+//#Arch:x86_64
+//#Mode:dynamic
+//#LinkArgs:-shared --no-gc-sections
+//#EnableLinker:lld
+//#ExpectError:R_X86_64_8
+//#SkipLinker:ld
```
lapla-cogito: Is there any reason to enable lld instead of using GNU ld? I think comparing against GNU ld should be enough. 🤔

deepakshirkem: @lapla-cogito We are primarily tracking lld compatibility, so I used EnableLinker:lld to verify Wild matches lld's behavior. Happy to switch to GNU ld if you prefer.

davidlattimore: If both linkers have the same behaviour, then it's probably good to use both as reference linkers. We usually only skip GNU ld when it has different behaviour that we don't want to try and match - and in that case we'd usually have a comment indicating why we skipped GNU ld

davidlattimore (approved):

lapla-cogito (approved):
Thanks!

PR #2149, opened 2026-06-27
test: Add runtime test for GOTPCRELX access to hidden absolute symbol in shared object

### LINKED ISSUE #2147: Wild incorrectly relaxes R_X86_64_GOTPCRELX to absolute symbols in shared objects
## Reproducer

```bash
cat > test.s <<\EOF
movq bar@GOTPCREL(%rip), %rax
.data
.global bar
.hidden bar
bar = 42
EOF

wild --no-gc-sections -shared test.o -o test.so
objdump -d test.so | grep mov
```

## Current behavior
Wild relaxes the GOT access to a direct immediate:
mov $0x2a,%rax    ← wrong, puts value 42 instead of address of bar

## Expected behavior
lld keeps the GOT access:
mov 0x1071(%rip),%rax    ← correct, loads address of bar from GOT

deepakshirkem:
When an absolute symbol is accessed via GOTPCRELX or REX_GOTPCRELX in a shared object or PIE, Wild was incorrectly relaxing the GOT-indirect load to a direct immediate (`mov $42, %rax`). This puts the symbol's value instead of its address into the register.

The GOT access must be preserved in position-independent output so that the symbol's address can be loaded at runtime. lld keeps the GOT access correctly. GNU ld has the same bug as the old Wild behavior.

Fixes #2147

### DIFF
--- a/wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
+++ b/wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
@@ -1,0 +1,21 @@
+// Test that GOTPCRELX access to a hidden absolute symbol in a shared object
+// produces correct runtime behavior. Wild relaxes to a direct immediate
+// (matching GNU ld), while lld keeps the GOT access. Both are correct.
+// lld is used as reference linker since GNU ld uses different relaxation.
+//#Arch:x86_64
+//#LinkArgs:--no-gc-sections -shared
+//#ReferenceLinkers:lld
+//#RunDynSym:get_bar
+//#DiffIgnore:.dynamic.DT_FLAGS_1.NOW
+//#DiffIgnore:section.got
+//#DiffIgnore:rel.extra-opt.R_X86_64_REX_GOTPCRELX.RexMovIndirectToAbsolute*
+.text
+.globl get_bar
+.type get_bar, @function
+get_bar:
+    movq bar@GOTPCREL(%rip), %rax
+    ret
+.data
+.global bar
+.hidden bar
+bar = 42


wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
```diff
@@ -0,0 +1,14 @@
+// GNU ld incorrectly relaxes to mov $0x2a,%rax, so we skip it.
```
davidlattimore: I'm not entirely sure that GNU ld's behaviour here is wrong. The symbol is marked as hidden, which means it cannot be provided by another DSO. Absolute symbols are weird in that they don't have an address. Code that requests their "address" is actually getting their value

wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
```diff
@@ -0,0 +1,14 @@
...
+//#Arch:x86_64
+//#LinkArgs:--no-gc-sections -shared
+//#EnableLinker:lld
+//#SkipLinker:ld
+//#RunEnabled:false
```
davidlattimore: I think it's also useful, where possible, to test the runtime behaviour. Out test framework supports running of shared objects via the `RunDynSym` directive. My guess is, that if we ignore the different relaxations being applied, that all the linkers will have the same runtime behaviour for GOTPCREL on an absolute symbol - it'd be good to verify though. Even if we decide that we want to stick doing the same optimisation as GNU ld, it might be that our existing tests aren't sufficient in this area.

wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
```diff
@@ -0,0 +1,14 @@
+// GNU ld incorrectly relaxes to mov $0x2a,%rax, so we skip it.
+//#Arch:x86_64
+//#LinkArgs:--no-gc-sections -shared
+//#EnableLinker:lld
+//#SkipLinker:ld
```
davidlattimore: FYI, yesterday I added a new "ReferenceLinkers" directive that replaces both of these. EnableLinker and SkipLinker will still exist for at least a few weeks.

wild/tests/sources/elf/gotpcrelx-abs-no-relax-pic/gotpcrelx-abs-no-relax-pic.s
deepakshirkem: @davidlattimore I investigated further. Upstream Wild and GNU ld both relax to mov $0x2a,%rax and drop the GOT entry. lld keeps the GOT access and fills it with 42. I verified the runtime result is the same in all cases. So upstream Wild's behavior was actually correct — it matches GNU ld. I opened this issue assuming lld's behavior was the reference, but GNU ld's relaxation here is also valid. Should I close this PR, or would it still be useful to add a RunDynSym runtime test to verify correctness of the existing behavior?

davidlattimore: Thanks for confirming. I think it could be useful to add a test, especially given that it sounds like changing the current behaviour didn't cause any of the existing tests to fail.

deepakshirkem:
@davidlattimore decided to only add a test rather than fixing the relaxation behavior, since I didn't want to add unnecessary complexity to the code. This case is also unlikely to occur in practice. Thank you for your guidance!

davidlattimore (approved):

PR #2151, opened 2026-06-27
feat: define multiple location counters that are relative to each other

vishruth-thimmaiah:
This PR adds support for evaluating every location counter, instead of only the most recently referenced one. This enables assignments where a location counter depends on its previous value, example:
```ld
. = 0x1000;
. = . + 0x100;
```

It also brings `. = ALIGN(x)` in line with GNU ld behavior by aligning the location counter rather than setting the section header directly.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -84,6 +84,34 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
     Ok(())
 }
 
+pub(crate) fn evaluate_location(
+    expr_loc: &SymbolLoc,
+    section_layouts: &OutputSectionMap<OutputRecordLayout>,
+    resolved_location_counters: &[(u64, Option<u64>)],
+) -> Result<u64> {
+    match expr_loc {
+        SymbolLoc::SectionStart(id) => Ok(section_layouts.get(*id).mem_offset),
+        SymbolLoc::SectionEnd(id) => {
+            let layout = section_layouts.get(*id);
+            Ok(layout.mem_offset + layout.mem_size)
+        }
+        SymbolLoc::FirstSection | SymbolLoc::None => Ok(0),
+        SymbolLoc::LocationCounter(idx, section_id) => {
+            let entry = resolved_location_counters.get(*idx).ok_or_else(|| {
+                crate::error!(
+                    "location counter index {idx} out of range (len: {})",
+                    resolved_location_counters.len()
+                )
+            })?;
+            if section_id.is_none() {
+                Ok(entry.1.unwrap_or(entry.0))
+            } else {
+                Ok(entry.0)
+            }
+        }
+    }
+}
+
 pub(crate) fn evaluate_expression<'data, P: Platform>(
     expr: &Expression<'data>,
     expr_loc: &SymbolLoc,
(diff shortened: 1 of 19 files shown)


libwild/src/expression_eval.rs
```diff
@@ -82,14 +84,43 @@ pub(crate) fn evaluate_assertions<'data, P: Platform>(
     Ok(())
 }
 
+pub(crate) fn evaluate_location(
+    expr_loc: &SymbolLoc,
+    section_layouts: &OutputSectionMap<OutputRecordLayout>,
+    resolved_location_counters: &[(u64, Option<u64>)],
```
davidlattimore: Given that this tuple is passed around, can we give it a name? (define a struct) That'll make it clearer what the two values represent

libwild/src/layout_rules.rs
```diff
@@ -377,6 +383,8 @@ impl<'data> LayoutRulesBuilder<'data> {
             }
         }
 
+        self.num_location_counters += location_counters.len();
```
davidlattimore, line 388: Changing this `+=` to just `=` doesn't seem to cause any of the tests to fail. Do you think it'd be practical to extend one of the tests so that it does?

libwild/src/output_section_id.rs
```diff
@@ -1090,3 +1156,11 @@ impl Iterator for PartIdIterator {
         }
     }
 }
+
+#[derive(Debug, Clone)]
+pub(crate) struct SectionLocationInfo<'data> {
+    pub(crate) location_counters: (usize, usize),
```
davidlattimore: Can you add a comment on this saying whether this range is inclusive or exclusive of the end index?

libwild/src/layout_rules.rs
```diff
@@ -306,44 +311,45 @@ impl<'data> LayoutRulesBuilder<'data> {
...
-                                SymbolLoc::Expression(_, section_id) => section_id,
-                                SymbolLoc::SectionStart(section_id)
-                                | SymbolLoc::SectionEnd(section_id)
-                                | SymbolLoc::RelativeExpression(_, section_id) => Some(section_id),
-                                _ => None,
-                            };
-                            loc = SymbolLoc::Expression(new_location.address.clone(), section_id);
+                            location_counters
+                                .push(LocationCounter::Absolute(new_location.address.clone()));
+                            last_lc_idx = location_counters.len();
```
davidlattimore: Location counters are identified currently by a usize. It might make it easier to see what things are referring to location counters if we added newtype - e.g. `LocationCounterIndex` or `LocationCounterId`.

libwild/src/layout.rs
```diff
@@ -5084,7 +5107,8 @@ fn perform_iterative_relaxation<'data, A: Arch>(
     Ok(())
 }
 
-fn layout_section<'data, P: Platform>(
+#[allow(clippy::type_complexity)]
```
davidlattimore: Could be good to use `expect` rather than `allow`. That way if we ever change the code such that the warning no longer triggers, we'll get a warning to remove the directive. I probably need to get better at remembering to do this myself - pretty sure I've probably added allows recently :)

davidlattimore (commented):
Nice :)

libwild/src/layout_rules.rs
vishruth-thimmaiah: We use `num_location_counters` to make sure location counter ids do not overlap when multiple linker scripts are provided. I've added a few additional linker scripts to test this.

libwild/src/layout.rs
```diff
@@ -5142,23 +5172,62 @@ fn layout_section<'data, P: Platform>(
...
     let mut records_out = output_sections.new_part_map();
 
     for event in output_order {
         match event {
-            OrderEvent::SetLocation(expr) => {
-                pending_location = Some(expr);
+            OrderEvent::SetLocation(expr, loc, idx) => {
+                let value =
+                    expression_eval(&expr, &loc, memory_regions, &section_layouts, &resolved_lc)?;
+                if resolved_lc.len() > 1 {
```
davidlattimore: Will this work with a linker script like the following?

```
ENTRY(begin_here)

SECTIONS
{
    .text 0x600000 : {
        *(.text)
    }
    . = 0x800000;
    .data : {
        *(.data)
    }
    .bss : {
        KEEP(*(.bss))
    }
}

ASSERT(ADDR(.data) == 0x800000, "ADDR(.data) should be moved by the location counter");
```

libwild/src/output_section_id.rs
davidlattimore, line 457: Do we need to do anything to deal with location counters for secondaries? If so, can you add a test for that?

vishruth-thimmaiah: Looks like I forgot to handle secondary sections here, I've updated the PR.

libwild/src/layout.rs
vishruth-thimmaiah: Removed the check, we don't need it anymore as the size of `resolved_lc` should be equal to the number of location counters defined.

davidlattimore (approved):

PR #2152, opened 2026-06-28
fix: Create canonical PLT entry for R_X86_64_PC32 to STT_FUNC DSO symbols in PIE

### LINKED ISSUE #2144: Wild errors on R_X86_64_PC32 to STT_FUNC DSO symbols instead of creating canonical PLT entry
Reproducer:
```
cat > shared.s <<\EOF
.global zed_fn
.type zed_fn, @function
zed_fn:
    ret
.size zed_fn, .-zed_fn
EOF

cat > test.s <<\EOF
.global _start
_start:
    mov zed_fn - ., %eax
    ret
EOF

llvm-mc -filetype=obj -triple=x86_64 shared.s -o shared.o
ld.lld shared.o -shared -o shared.so
wild -pie --no-gc-sections test.o shared.so -o out
```

Current behavior:

wild: error: relocation R_X86_64_PC32 cannot be used against symbol; recompile with -fPIC

Expected behavior:

lld links successfully by creating a canonical PLT entry for the STT_FUNC DSO symbol. GNU ld also links successfully.

deepakshirkem:
When a PIE binary uses R_X86_64_PC32 against a STT_FUNC DSO symbol, Wild was erroring with "recompile with -fPIC" instead of creating acanonical PLT entry like lld and GNU ld do.

The layout phase already correctly allocates a PLT entry for STT_FUNC DSO symbols via the `flags.is_function()` branch in `elf.rs`. The error in `elf_writer.rs` was firing before relocation application without checking if a PLT entry had already been allocated for the symbol.

Fixes #2144

### DIFF
--- a/wild/tests/sources/elf/pie-pc32-dso-code-ref-data/pie-pc32-dso-code-ref-data.s
+++ b/wild/tests/sources/elf/pie-pc32-dso-code-ref-data/pie-pc32-dso-code-ref-data.s
@@ -1,15 +1,13 @@
-// Scenario: code references DSO data symbol via R_X86_64_PC32 - invalid in PIE.
-// SkipLinker:lld because lld creates a canonical PLT entry for R_X86_64_PC32 to
-// STT_FUNC DSO symbols and succeeds, while Wild doesn't support canonical PLT and errors.
-// This case will be revisited when canonical PLT support is added to Wild.
+// Scenario: code references DSO STT_FUNC symbol via R_X86_64_PC32 in PIE.
+// Wild and lld create a canonical PLT entry and link successfully.
+// GNU ld errors on R_X86_64_PC32 against DSO symbol in PIE, so skip it.
 //#Arch:x86_64
 //#Mode:dynamic
 //#Shared:pie-pc32-dso-shared.s
 //#SoSingleLinker:wild
 //#LinkArgs:-pie --no-gc-sections
-//#ExpectErrorWild:R_X86_64_PC32
 //#SkipLinker:ld
-//#SkipLinker:lld
+//#RunEnabled:false
 .global _start
 _start:
     mov zed_fn - ., %eax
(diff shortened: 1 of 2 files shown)


wild/tests/sources/elf/pie-pc32-dso-code-ref-data/pie-pc32-dso-code-ref-data.s
```diff
@@ -1,15 +1,13 @@
-// Scenario: code references DSO data symbol via R_X86_64_PC32 - invalid in PIE.
-// SkipLinker:lld because lld creates a canonical PLT entry for R_X86_64_PC32 to
-// STT_FUNC DSO symbols and succeeds, while Wild doesn't support canonical PLT and errors.
-// This case will be revisited when canonical PLT support is added to Wild.
+// Scenario: code references DSO STT_FUNC symbol via R_X86_64_PC32 in PIE.
+// Wild creates a canonical PLT entry and links successfully.
+// SkipLinker:ld because SoSingleLinker:wild is incompatible with GNU ld.
```
davidlattimore: This reason doesn't seem right to me. I'm not aware of any reason why GNU ld couldn't link against a shared object produced by wild. If it couldn't, then we'd definitely want to fix it.

deepakshirkem: You're right, the comment was wrong. I tested it and found that GNU ld actually errors on R_X86_64_PC32 against a DSO symbol in PIE. It doesn't support canonical PLT for PIE binaries; only LLD does. So `SkipLinker:ld` is still correct, but for a different reason. I've updated the comment to reflect 

Image

davidlattimore: I don't see your change. Perhaps something went wrong when pushing it?

deepakshirkem: Sorry, I missed adding the test file. Thank you!

davidlattimore (approved):

PR #2153, opened 2026-06-29
test: Unify Wasm integration tests with ELF/Mach-O directives

lapla-cogito:
The structure of the integration tests for Wasm has started to take shape, so we can prevent the test directives from diverging by unifying their handling with that of the ELF and Mach-O.

Part of #1431

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -139,12 +139,11 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
         });
 
     parser
-        .declare_with_param()
+        .declare()
         .long("no-gc-sections")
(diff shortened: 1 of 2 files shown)


libwild/src/args/wasm.rs
```diff
@@ -139,12 +139,11 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
         });
 
     parser
-        .declare_with_param()
+        .declare()
```
lapla-cogito, line 142: This was incorrect, and the Wasm data-segment test would fail without this fix.

https://github.com/wild-linker/wild/actions/runs/28351479535/job/83985098830

davidlattimore (approved):
Thanks :)

PR #2154, opened 2026-06-29
port(wasm): Allow linking against archives containing Wasm objects

lapla-cogito:
Part of #1431


davidlattimore (approved):
Wow! I love how simple this PR is :)

PR #2157, opened 2026-06-29
CI: do not run Malfunction tests in non-Debug profile

marxin:
Sometimes I run tests in release profile and I noticed the tests are failing.


davidlattimore (approved):
Thanks. For a brief moment I wondered what optimisation might be causing them to fail. Then I remembered that for performance reasons, the malfunction check points are never checked except when debug assertions aren't active.

PR #2158, opened 2026-06-30
port(MachO): Support direct linking against dylibs

### LINKED ISSUE #2067: Mach-O: add support for linking against classical shared libraries
For most of the system libraries, the Mach-O provides only STUB files (YAML file with a symbol listing). The initial support for these files was already added. On top of that, we must support also the normal `.dylib` shared libraries built by users, or provided by a third party. The mechanism should be pretty similar to ELF by using `-lmy_library`.

davidlattimore:
There's quite a lot of similarity between our handling of stub libraries and dylibs. I suspect we might want to do some future refactoring to try and unify them some more. For now, where there's duplicated logic, I've just extracted it out into a function and called it from both places. The primary difference between e.g. `DynamicLayoutState` and `StubLibraryLayoutState` is that the former contains a `platform::File` reference, where the latter cannot, since one does not exist. One option would be to remove this reference from `DynamicLayoutState` and leave it up to the `DynamicLayoutStateExt` whether and how to store the reference. I prototyped this change. The difficulty is that the same difference applies to our other stages, e.g. `ResolvedDynamic` and making the same refactoring there started to look complicated, so I decided to hold off on it for now.

Fixes #2067

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,7 @@ jobs:
     env:
       WILD_TEST_IGNORE_FORMAT: "1" # Checked by other jobs
     steps:
+      - run: brew install lld
       - uses: actions/checkout@v7
         with:
           persist-credentials: false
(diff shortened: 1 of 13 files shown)


davidlattimore:
Note that we don't yet support fat dylibs. That will come in a subsequent PR.

PR #2162, opened 2026-07-01
fix: Emit one zlib stream per compressed debug section

lapla-cogito:
Previously, large compressed debug sections could be emitted incorrectly once a section was split into multiple compression chunks (chunk size is at least 64 KiB). Consumers then saw truncated or corrupt DWARF. (Also see https://github.com/wild-linker/wild/pull/2119#discussion_r3449691345)

For example, with a buggy wild-linked binary:

```
（▰╹◡╹）❯  llvm-dwarfdump --statistics wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.wild|head
warning: DWARF unit at offset 0x000c8b80 has unsupported version 0, supported are 2-5
warning: unable to decode LEB128 at offset 0x0004a0fe: malformed uleb128, extends past end
warning: last sequence in debug line table at offset 0x000003fd is not terminated
{
  "version": 9,
  "file": "wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.wild",
  "format": "elf64-x86-64",
  "#functions": 50,
  "#functions with location": 49,
  "#out-of-line functions": 103,
  "#inlined functions": 1798,
  "#inlined functions with abstract origins": 1798,
  "#unique source variables": 13,
```

whereas GNU ld on the same inputs (note that there is no warning and the number of functions is very different from wild's case):

```
（▰╹◡╹）❯  llvm-dwarfdump --statistics wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.ld|head
{
  "version": 9,
  "file": "wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.ld",
  "format": "elf64-x86-64",
  "#functions": 9370,
  "#functions with location": 9370,
  "#out-of-line functions": 1925,
  "#inlined functions": 44899,
  "#inlined functions with abstract origins": 44899,
```

The root cause was concatenating complete zlib streams from `ZlibEncoder::finish()` per chunk. We now compress raw deflate shards in parallel with `Z_SYNC_FLUSH` boundaries, then emit a single zlib stream (header + shards + empty stored block + combined Adler-32).

### DIFF
--- a/libwild/src/compression.rs
+++ b/libwild/src/compression.rs
@@ -48,8 +48,32 @@ const MIN_CHUNK_SIZE: usize = 64 * 1024;
 /// threads because we want the output to not depend on the number of threads.
 const MAX_CHUNKS: usize = 128;
 
-const ZLIB_COMPRESSION_LEVEL: u32 = 1;
+const ZLIB_COMPRESSION_LEVEL: i32 = 1;
 const ZSTD_COMPRESSION_LEVEL: i32 = 3;
+
+/// zlib `windowBits`.
+const ZLIB_WINDOW_BITS: u8 = 15;
+
+/// Initial Adler-32 value (RFC 1950).
+const ADLER32_INITIAL: u32 = 1;
+
+/// zlib CMF+FLG header for deflate with a 32 KiB window and default FCHECK.
+const ZLIB_CMF_FLG: [u8; 2] = [0x78, 0x9c];
+
+/// Empty final deflate block (`BFINAL=1`, empty stored/fixed block) written after
+/// SyncFlush'd shards so the stream can be closed before the Adler-32 trailer.
+const ZLIB_EMPTY_FINAL_BLOCK: [u8; 2] = [0x03, 0x00];
+
+/// Size of multi-shard zlib trailer. Empty final block + Adler-32 (big-endian u32).
+const ZLIB_MULTI_SHARD_TRAILER_LEN: usize =
+    ZLIB_EMPTY_FINAL_BLOCK.len() + std::mem::size_of::<u32>();
+
+/// Extra bytes reserved beyond `compress_bound` so SyncFlush markers fit without an immediate
+/// realloc on the first flush attempt.
+const ZLIB_SYNC_FLUSH_SLACK: usize = 16;
+
+/// How much to grow the output buffer when a Finish/SyncFlush needs more space.
+const ZLIB_OUTPUT_GROW_BYTES: usize = 64;
 
 pub(crate) fn maybe_compress_debug_sections_elf<A: Arch<Platform = Elf>>(
     layout: &mut crate::layout::Layout<Elf>,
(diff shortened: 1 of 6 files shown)


lapla-cogito:
The output of the `llvm-dwarfdump` command after this change: 

```
（▰╹◡╹）❯  llvm-dwarfdump --statistics wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.wild|head
{
  "version": 9,
  "file": "wild/tests/build/elf/x86_64/compressed-debug-gdb-scripts/zlib/compressed-debug-gdb-scripts.rs.wild",
  "format": "elf64-x86-64",
  "#functions": 9370,
  "#functions with location": 9370,
  "#out-of-line functions": 1925,
  "#inlined functions": 44899,
  "#inlined functions with abstract origins": 44899,
  "#unique source variables": 17,
```

libwild/src/compression.rs
```diff
@@ -110,19 +143,101 @@ impl Compressor for ZlibCompressor {
...
+}
+
+/// Compresses one shard as raw deflate terminated with `SyncFlush`. Also returns Adler-32 of the
+/// uncompressed shard.
+fn zlib_compress_shard(input: &[u8]) -> Result<(Vec<u8>, u32)> {
+    // Compute while `input` is hot for the following deflate pass.
+    let checksum = adler32(1, input);
+
+    let mut encoder = Deflate::new(ZLIB_COMPRESSION_LEVEL as i32, false, 15);
+    let mut output = vec![0u8; zlib_rs::compress_bound(input.len()) + 16];
```
davidlattimore: There's quite a few numeric values that it's not clear what they are. Can you define a few constants for them?

libwild/src/compression.rs
```diff
@@ -110,19 +143,101 @@ impl Compressor for ZlibCompressor {
 
 struct ZstdCompressor;
 
-impl Compressor for ZstdCompressor {
-    fn compress(chunk: &[u8]) -> Result<Vec<u8>> {
-        let mut output = Vec::new();
-        zstd::stream::copy_encode(chunk, &mut output, ZSTD_COMPRESSION_LEVEL)?;
-        Ok(output)
+impl SectionCompressor for ZstdCompressor {
```
davidlattimore, line 171: Are tools that consume zstd OK with multiple, independently compressed chunks? i.e. do we need to (in a separate PR) look to make a similar fix for zstd, or is it already OK?

davidlattimore (commented):
Thanks for fixing this!

Do you think it'd be worth adding a unit test that creates some data, zlib compresses it, concatenates the chunks, decompresses it and then checks that we get back to what we started with?

libwild/src/compression.rs
lapla-cogito: In zstd, a sequence of concatenated frames is a valid format (yes, I should have mentioned this from the beginning 😅):

> Zstandard compressed data is made up of one or more frames. Each frame is independent and can be decompressed independently of other frames. The decompressed content of multiple concatenated frames is the concatenation of each frame's decompressed content.

https://www.rfc-editor.org/rfc/rfc8878.html#section-3.1

libwild/src/compression.rs
lapla-cogito: done

davidlattimore (approved):

PR #2163, opened 2026-07-01
port(wasm): Absorb indirect function tables and apply table relocations

lapla-cogito:
Synthesize a single funcref table and an active element segment, absorb `env.__indirect_function_table` imports, and resolve `TABLE_INDEX` / `TABLE_NUMBER` relocations.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -18,6 +18,7 @@ use crate::wasm_writer::OutputGlobal;
 use crate::wasm_writer::OutputImport;
 use crate::wasm_writer::OutputImportEntity;
 use hashbrown::HashMap;
+use hashbrown::HashSet;
 use leb128::write::unsigned_len as uleb128_size;
 use linker_utils::utils::u32_from_slice;
 use rayon::prelude::*;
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -2685,10 +2744,113 @@ where
...
+}
+
+/// Assign indirect-call table slots and synthesize `table` / `element` sections.
+fn finalize_indirect_function_table(
+    layout: &mut WasmLayout<'_>,
+    layout_inputs: &[WasmObjectLayoutInput<'_>],
+) -> Result {
+    // Collect output function indices that must appear in the table.
+    let mut needed: Vec<u32> = Vec::new();
+    let mut seen = HashSet::new();
```
davidlattimore, line 2770: This seems like the sort of thing that could more efficiently be handled by setting a ValueFlag. No need to change anything now (although a todo could be good), but it's something we should consider for future.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: Yes, also taking the discussion from the previous PR into account, I now think it makes sense to introduce a concept equivalent to `ValueFlags` for Wasm as well.

PR #2165, opened 2026-07-01
fix: Relax -z noplt form of TLS GD to Initial Exec on x86_64

### LINKED ISSUE #2164: Wild missing TLS GD to IE relaxation for -z noplt form (RegularNoPlt)
### Reproducer

```bash
cat > tls_noplt_ie.s <<\EOF
.globl _start
_start:
    .byte 0x66
    leaq tlsvar@tlsgd(%rip), %rdi
    .byte 0x66
    .byte 0x48
    call *__tls_get_addr@GOTPCREL(%rip)
EOF

cat > tlsvar.s <<\EOF
.section .tbss,"awT",@nobits
.globl tlsvar
tlsvar:
    .long 0
EOF

llvm-mc -filetype=obj -triple=x86_64-unknown-linux tls_noplt_ie.s -o tls_noplt_ie.o
llvm-mc -filetype=obj -triple=x86_64-unknown-linux tlsvar.s -o tlsvar.o
ld.lld -shared tlsvar.o -o tlsvar.so
wild tls_noplt_ie.o tlsvar.so -o out
```

### Current behavior
wild: error: Undefined symbol __tls_get_addr

### Expected behavior
lld relaxes to IE form: 
mov %fs:0x0,%rax
add offset(%rip),%rax

deepakshirkem:
When a TLS symbol is in a shared object, Wild needs to relax TLS GD to IE (Initial Exec) rather than LE. The -z noplt variant uses an indirect call through the GOT (0x66 0x48 0xff 0x15) instead of a direct PLT call. This form was grouped with the Large variant and returned None, causing Wild to error with "Undefined symbol __tls_get_addr" instead of relaxing to IE.

Both lld and GNU ld already handle this pattern correctly.

Fixes  #2164

### DIFF
--- a/libwild/src/elf_x86_64.rs
+++ b/libwild/src/elf_x86_64.rs
@@ -371,8 +371,10 @@ impl crate::platform::Arch for ElfX86_64 {
             }
             object::elf::R_X86_64_TLSGD if output_kind.is_executable() => {
                 let kind = match TlsGdForm::identify(section_bytes, offset)? {
-                    TlsGdForm::Regular => RelaxationKind::TlsGdToInitialExec,
-                    TlsGdForm::RegularNoPlt | TlsGdForm::Large => {
+                    TlsGdForm::Regular | TlsGdForm::RegularNoPlt => {
+                        RelaxationKind::TlsGdToInitialExec
+                    }
+                    TlsGdForm::Large => {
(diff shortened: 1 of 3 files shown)


davidlattimore (approved):
Thanks :)

PR #2167, opened 2026-07-02
chore: Improve error message if we allocate to sections with no order

davidlattimore:


libwild/src/layout.rs
```diff
@@ -5343,9 +5344,53 @@ fn layout_section<'data, P: Platform>(
...
+    let mut emitted_sections: OutputSectionMap<bool> =
+        OutputSectionMap::with_size(output_sections.num_sections());
+
+    for event in output_order {
+        if let OrderEvent::Section(output_section_id) = event {
+            *emitted_sections.get_mut(output_section_id) = true;
+        }
+    }
+
+    // TODO: Investigate why wasm allocates space to sections not in its output ordering.
```
lapla-cogito: I submitted https://github.com/wild-linker/wild/pull/2170 to resolve this.

lapla-cogito (approved):

libwild/src/layout.rs
davidlattimore: Thanks :)

PR #2171, opened 2026-07-03
chore: Overhaul update-in-place testing

davidlattimore:
Rather than running the linker once for the main test then twice for the update-in-place verification, we only run the linker once for the main test, then one extra time for the update-in-place verification. We also now set `--threads=3` for the verification run so that we can check that the output doesn't change due to a change in the number of threads.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1353,7 +1353,8 @@ impl Config {
     }
 
     fn can_use_wild_in_process(&self) -> bool {
-        self.expect_stderr.is_empty()
+        !self.test_update_in_place
+            && self.expect_stderr.is_empty()
             && self.expect_stdout.is_empty()
             && self.active_malfunction.is_none()
     }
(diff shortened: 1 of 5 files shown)


wild/tests/integration_tests.rs
```diff
@@ -2181,92 +2186,62 @@ impl ProgramInputs {
...
+        reference_output: &LinkOutput,
+    ) -> Result {
+        let original_content = std::fs::read(&reference_output.binary)
+            .with_context(|| format!("Failed to read: {}", reference_output.binary.display()))?;
 
-        // Overwrite the file with random data with a different length
+        // Create initial file with random data and a different length.
         let random_data = (0..original_content.len() + 1024)
-            .map(|i| (i % 256) as u8)
+            .map(|_| rand::random())
```
mati865: Could be just `rand::random_iter().take(original_content.len() + 1024)`?

mati865 (approved):

wild/tests/integration_tests.rs
davidlattimore: Done

PR #2176, opened 2026-07-03
port(wasm): Add linker-defined memory base, `MEMORY_ADDR_REL_SLEB` relocation, and `call_ctors` stubs

lapla-cogito:
Clang `-fPIC` objects that use `__memory_base`-relative addressing, and inputs that import `__stack_pointer` or the `__wasm_call_ctors` / `__wasm_call_dtors` hooks, can now be linked into a runnable module without leaving those as unresolved env imports (empty ctor/dtor bodies for now). This PR handles PIC compatibility and basic absorption of linker-defined symbols.

Absorb `__memory_base` / `__stack_pointer` / `__wasm_call_ctors` / `__wasm_call_dtors` into synthesized globals and empty functions. Apply `R_WASM_MEMORY_ADDR_REL_SLEB`, derive the stack pointer from data_end plus a 64KiB stack, and size memory pages from static data (and the stack when present).

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2587,6 +2587,9 @@ fn remaining_unresolved_imports(unresolved: u32, absorbed: u32) -> Result<u32> {
 struct LinkerDefinedIndices {
     memory_base_global: Option<u32>,
     stack_pointer_global: Option<u32>,
+    /// Index of `__stack_pointer` among the defined globals prepended by
+    /// `emit_reserved_linker_definitions` (not the Wasm module global index).
+    stack_pointer_defined_slot: Option<u32>,
     call_ctors_func: Option<u32>,
     call_dtors_func: Option<u32>,
     num_defined_globals: u32,
@@ -2638,14 +2641,18 @@ impl LinkerDefinedIndices {
         }
 
         let mut next_global = global_import_count;
+        let mut next_defined_global_slot = 0u32;
         let memory_base_global = needs_memory_base.then(|| {
             let idx = next_global;
             next_global += 1;
+            next_defined_global_slot += 1;
             idx
         });
+        let stack_pointer_defined_slot = needs_stack_pointer.then_some(next_defined_global_slot);
         let stack_pointer_global = needs_stack_pointer.then(|| {
             let idx = next_global;
             next_global += 1;
+            next_defined_global_slot += 1;
             idx
         });
         let num_defined_globals = next_global - global_import_count;
@@ -2666,6 +2673,7 @@ impl LinkerDefinedIndices {
         Ok(Self {
             memory_base_global,
             stack_pointer_global,
+            stack_pointer_defined_slot,
             call_ctors_func,
             call_dtors_func,
             num_defined_globals,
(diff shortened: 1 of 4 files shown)


libwild/src/wasm.rs
```diff
@@ -3231,11 +3420,10 @@ impl platform::Platform for Wasm {
     }
 
     fn create_linker_defined_symbols(
```
lapla-cogito, line 3524: This appears to be intended to allow named symbols to be added to the symbol database early, but it conflicts with Wasm implementations that add actual entities to the module's index space later. Therefore, it has been delegated to `finalize_linker_defined_state`.

libwild/src/wasm.rs
```diff
@@ -2495,27 +2531,192 @@ fn any_object_needs_linker_memory(inputs: &[WasmObjectLayoutInput<'_>]) -> bool
...
+        linker_globals.push(OutputGlobal {
+            ty: GlobalType {
+                content_type: wasmparser::ValType::I32,
+                mutable: true,
+                shared: false,
+            },
+            init_expr_body: Cow::Owned(encode_i32_const_body(sp as i32)),
+        });
+    }
+    let inserted = linker_globals.len() as u32;
```
davidlattimore: From what I understand of this code, we're inserting symbols, then fixing up already assigned indexes. Why can't we decide what symbols are needed before indexes are assigned such that no fixing up is necessary?

libwild/src/wasm.rs
davidlattimore: Can you elaborate on what you mean by "conflicts"?

libwild/src/wasm.rs
```diff
@@ -109,14 +111,25 @@ pub(crate) const TARGET_FEATURES_SECTION_NAME: &str = "target_features";
...
-/// `i32.const` body for [`LINKER_STACK_POINTER`] without the trailing `end` opcode.
-const LINKER_STACK_POINTER_INIT_EXPR: &[u8] = &[0x41, 0x90, 0x88, 0x04];
+/// Empty function body: zero locals + `end`.
+const EMPTY_FUNCTION_BODY: &[u8] = &[0x00, 0x0b];
 
-/// Sentinel in `global_indices` for an absorbed `env.__stack_pointer` import.
+/// `i32.const` body for `LINKER_MEMORY_BASE`.
+const LINKER_MEMORY_BASE_INIT_EXPR: &[u8] = &[0x41, 0x80, 0x08];
+
+/// Sentinels in `global_indices` / `function_indices` for absorbed linker-defined symbols.
```
davidlattimore: Do the symbols associated with these placeholders have SymbolIds? If they don't, why not and could the code possibly be simplified a bit if they did?

libwild/src/wasm.rs
lapla-cogito: Ah, I may have phrased that poorly. I didn't mean a hard conflict. Early `SymbolDb` registration is simply unnecessary for Wasm. Unresolved imports for linker-defined names are absorbed during layout and mapped through `LinkerDefinedIndices`, which is enough to place the corresponding module entities.

libwild/src/wasm.rs
lapla-cogito: The original motivation was to leave the existing per-object index-allocation pipeline mostly unchanged. Linker-defined entities are only known to be needed after scanning imports across objects, so the first approach absorbed those imports with placeholders, inserted the entities at the end of layout, and fixed up already-assigned indices (and exports) to account for the insertion. That worked when there were a few entities, but as we added more linker-defined globals/functions the fixup surface grew. I therefore switched to reserving their final indices as soon as their need is known so references get the real indices immediately and no renumbering is required.

libwild/src/wasm.rs
```diff
@@ -2957,20 +3200,48 @@ fn allocate_wasm_object_index_bases(
         next_type_index = next_type_index
             .checked_add(u32::try_from(input.types.len()).context("too many Wasm types")?)
             .ok_or_else(|| crate::error!("Wasm type index overflow"))?;
+
+        let mut absorbed_functions = 0u32;
```
davidlattimore: Why do we need absorbed_functions (same for globals)? Couldn't we just make these be linker-defined symbols?

davidlattimore (commented):
I wasn't sure if you were ready for another round of review, but I had a quick look. When you are ready for re-review, can you click the request-review button in the github UI?

libwild/src/wasm.rs
lapla-cogito: They already are (as module entities via `LinkerDefinedIndices`). `absorbed_*` is the import-index accounting for that decision. Those names appear as unresolved imports in input objects, but we emit them as defined entities instead of import entries, so they must not advance `next_*_import_index`. Without subtracting them, import bases (and everything after them) would be too large.

davidlattimore: OK. I guess what I'm wondering is whether these symbols can be created in `create_linker_defined_symbols` so that they have SymbolIds, allowing us to avoid scanning all imports looking for known names? But I'll approve this, since it's fine to defer until later.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: Even when using SymbolId, I think we still need to scan all import resolution results for each object file.

Issue #1853, opened 2026-04-19
Add support for bitmap packing of RELR relocations


mati865:
https://github.com/wild-linker/wild/pull/1701 implemented the easy part of RELR relocations: emitting RELR entries instead of relative RELA ones, if feasible. However, it's missing bitmap packing, because I failed to reliably achieve that.
This is hard because during the allocation phase we have very little idea about the relocation addresses in output binary, yet we need to figure out all the details regarding bitmap packing.

Some things to look out for that I remember:
- bitmap packing must be identical regardless of inputs grouping
- unaligned relocations may occur in between aligned ones, forcing an earlier end of current bitmap

deepakshirkem:
Hi @davidlattimore and @mati865, I'd like to work on this. I verified the gap: Wild currently emits 4 individual RELR entries for 4 consecutive pointers, while LLD and GNU ld pack them into 2 entries (an address entry followed by a bitmap entry). I plan to implement this incrementally in small PRs. Do you have any suggestions before I get started?

davidlattimore:
Great! Some thoughts on possible simplifications... I think if you only pack relocations within an input section, then you can figure out what space is required during the relocation scan that we do during layout. i.e. don't worry about packing relocations from input sections that just happen to end up adjacent in the output file. While that simplification might leave a little bit of space-saving that we don't achieve, I expect that we'll still get perhaps 90% of the possible savings just looking within input sections.

Regarding the comment above about ending a bitmap if we encounter an unaligned relocation - it may also be fine to continue with the bitmap, but just handle that unaligned relocation separately.

PR #2179, opened 2026-07-03
fix: Implement RELR bitmap packing for consecutive relocations

deepakshirkem:
Pack consecutive 8-byte-aligned RELR relocations within input sections into bitmap entries, matching lld's output.

Before: 4 consecutive pointers → 4 RELR entries (32 bytes)
After:  4 consecutive pointers → 2 RELR entries (16 bytes)

Only packs within input sections as suggested by David — avoids cross-section ordering issues in Wild's parallel layout phase.

Issue #1853

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2827,6 +2827,7 @@ fn process_eh_frame_relocations<'data, 'scope, A: Arch<Platform = Elf>, R: Reloc
                     queue,
                     false,
                     scope,
+                    &mut RelrRunState::default(), // eh_frame relocations are never RELR-eligible
                 )?;
 
                 if rel.symbol().is_none() {
@@ -2929,6 +2930,7 @@ fn process_section_exception_frames<'data, 'scope, A: Arch<Platform = Elf>, R: R
                 queue,
                 true,
                 scope,
+                &mut RelrRunState::default(), // eh_frame relocations are never RELR-eligible
             )?;
         }
         common.format_specific.exception_frame_relocations +=
@@ -3380,6 +3382,26 @@ impl<'data> DynamicTagValues<'data> {
 impl<'data> platform::DynamicTagValues<'data> for DynamicTagValues<'data> {
     fn lib_name(&self, input: &InputRef<'data>) -> &'data [u8] {
         self.soname.unwrap_or_else(|| input.lib_name())
+    }
+}
+
+/// Tracks consecutive RELR-eligible relocations within a single input section
+/// for bitmap packing during the layout phase.
+/// Only pack within a single input section — gives ~90% of savings and avoids
+/// cross-section address ordering issues in Wild's parallel layout phase.
+struct RelrRunState {
+    /// Offset of the last RELR-eligible relocation seen in the current run.
+    last_offset: Option<u64>,
+    /// Whether a bitmap entry has already been allocated for the current run.
+    bitmap_allocated: bool,
+}
+
+impl Default for RelrRunState {
+    fn default() -> Self {
+        Self {
+            last_offset: None,
+            bitmap_allocated: false,
+        }
     }
 }
 
@@ -4964,6 +4986,7 @@ fn load_section_relocations<'scope, 'data, A: Arch<Platform = Elf>, R: Relocatio
     scope: &Scope<'scope>,
 ) -> Result {
     let mut modifier = RelocationModifier::Normal;
+    let mut relr_run = RelrRunState::default();
     for rel in relocations {
         if modifier == RelocationModifier::SkipNextRelocation {
             modifier = RelocationModifier::Normal;
@@ -4982,6 +5005,7 @@ fn load_section_relocations<'scope, 'data, A: Arch<Platform = Elf>, R: Relocatio
             queue,
             false,
             scope,
+            &mut relr_run,
         )
         .with_context(|| {
             format!(
@@ -5005,6 +5029,7 @@ fn process_relocation<'data, 'scope, A: Arch<Platform = Elf>, R: Relocation>(
     queue: &mut layout::LocalWorkQueue,
     is_debug_section: bool,
     scope: &Scope<'scope>,
+    relr_run: &mut RelrRunState,
 ) -> Result<RelocationModifier> {
     let args = resources.symbol_db.args;
     let mut next_modifier = RelocationModifier::Normal;
@@ -5101,9 +5126,25 @@ fn process_relocation<'data, 'scope, A: Arch<Platform = Elf>, R: Relocation>(
             && flags.is_address()
         {
             if section_is_writable {
-                // Odd offsets mean bitmaps in RELR, so we need to fall back to RELA for them.
+                // Odd offsets can't be encoded as RELR address entries (LSB used as
+                // bitmap marker), so fall back to RELA for them.
                 if resources.symbol_db.args.is_relr_enabled() && rel.offset().is_multiple_of(2) {
-                    common.allocate(part_id::RELR_DYN, elf::RELR_ENTRY_SIZE);
+                    let offset = rel.offset();
+                    if relr_run.last_offset == Some(offset.wrapping_sub(8)) {
+                        // Continues current run — consecutive 8-byte-aligned relocation.
+                        if !relr_run.bitmap_allocated {
+                            // Second member of run: allocate one bitmap entry.
+                            common.allocate(part_id::RELR_DYN, elf::RELR_ENTRY_SIZE);
+                            relr_run.bitmap_allocated = true;
+                        }
+                        // Third+ members fit into existing bitmap — no new entry needed.
+                    } else {
+                        // New run: allocate address entry, reset bitmap state.
+                        common.allocate(part_id::RELR_DYN, elf::RELR_ENTRY_SIZE);
+                        relr_run.last_offset = Some(offset);
+                        relr_run.bitmap_allocated = false;
+                    }
+                    relr_run.last_offset = Some(offset);
                 } else {
                     common.allocate(part_id::RELA_DYN_RELATIVE, elf::RELA_ENTRY_SIZE);
                 }
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -1457,6 +1457,10 @@ impl platform::Platform for Elf {
                 mem_sizes.increment(part_id::RELA_DYN_GENERAL, elf::RELA_ENTRY_SIZE);
             } else if flags.is_address() && output_kind.is_relocatable() {
                 if args.is_relr_enabled() {
+                    // TODO: Implement bitmap packing for GOT-based RELR entries.
+                    // Currently counts flat (one entry per GOT slot) to match the
+                    // writer's write_relr_entry_flat. Bitmap packing requires splitting
+                    // the GOT so relative relocations are contiguous.
                     mem_sizes.increment(part_id::RELR_DYN, elf::RELR_ENTRY_SIZE);
                 } else {
                     mem_sizes.increment(part_id::RELA_DYN_RELATIVE, elf::RELA_ENTRY_SIZE);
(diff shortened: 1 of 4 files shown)


libwild/src/elf.rs
```diff
@@ -3383,6 +3385,26 @@ impl<'data> platform::DynamicTagValues<'data> for DynamicTagValues<'data> {
     }
 }
 
+/// Tracks consecutive RELR-eligible relocations within a single input section
+/// for bitmap packing during the layout phase.
+/// Only pack within a single input section — gives ~90% of savings and avoids
+/// cross-section address ordering issues in Wild's parallel layout phase.
+struct RelrRunState {
+    /// Offset of the last RELR-eligible relocation seen in the current run.
+    last_offset: Option<u64>,
```
davidlattimore: The bitmap can contain 1s and 0s, which means that non-consecutive relocations can still be packed into the same bitmap - skipped relocations just encode as a 0. We do however need some additional state, since we need to know how much of the current bitmap remains, after which we'll need to allocate a new bitmap. If ever a relocation address indicates that we would have needed an empty bitmap, that's the point at which we should probably end the previous run and start a new one.

davidlattimore (commented):
90% was a complete guess on my part, so we probably shouldn't put that in the comments

libwild/src/elf.rs
deepakshirkem: Updated both the layout counting and writer to handle gaps within the 63-slot window.

I think we hit a structural issue with the allocate_resolution path (GOT-based  RELR entries). That function is called inside finalise_symbol_sizes  which iterates in symbol-ID order, but the writer packs in 
output-address order. These don't match, causing validate_empty  failures on aarch64/riscv64/loongarch64 (16 of 32 bytes remain layout counts 4 flat entries, writer packs them into 2).

Two options we can see:
1. Skip bitmap packing for GOT-based RELR entries entirely count  and write flat in both phases. Loses some savings but is consistent.
2. Count GOT entries in output-address order during layout — unclear  if this is feasible without deeper changes to the layout pipeline.

What's the right approach here?

libwild/src/elf.rs
```diff
@@ -3383,6 +3385,23 @@ impl<'data> platform::DynamicTagValues<'data> for DynamicTagValues<'data> {
     }
 }
 
+struct RelrRunState {
```
davidlattimore: I wonder if this might be easier to understand if it were an enum instead, with a variant for each state

libwild/src/elf_writer.rs
```diff
@@ -605,6 +605,13 @@ struct TableWriter<'layout, 'out> {
     rela_dyn_relative: &'out mut [crate::elf::Rela],
     rela_dyn_general: &'out mut [crate::elf::Rela],
     relr_dyn: Option<&'out mut [elf::Relr]>,
+    /// Index of next free slot in relr_dyn.
+    relr_dyn_index: usize,
+    /// Index of the current bitmap entry in relr_dyn, if any.
+    relr_bitmap_index: Option<usize>,
+    /// Address of the start of the current RELR run.
+    relr_run_base: Option<u64>,
+    /// Last address written to relr_dyn, to detect consecutive +8 runs.
```
davidlattimore: This comment seems to be referring to a field that doesn't exist

libwild/src/elf_writer.rs
```diff
@@ -605,6 +605,13 @@ struct TableWriter<'layout, 'out> {
     rela_dyn_relative: &'out mut [crate::elf::Rela],
     rela_dyn_general: &'out mut [crate::elf::Rela],
     relr_dyn: Option<&'out mut [elf::Relr]>,
+    /// Index of next free slot in relr_dyn.
+    relr_dyn_index: usize,
+    /// Index of the current bitmap entry in relr_dyn, if any.
```
davidlattimore: Could plausibly have a similar enum-based state machine here - plausibly, the same enum might be able to be used.

davidlattimore (commented):
I think it's best to leave the GOT for a separate PR. One possible strategy for dealing with it might be to split the GOT into two parts with all the relative relocations together in one part. If we know that they're all together and how many there are, then allocation and writing should be pretty straightforward. But definitely a separate PR for something like that.

deepakshirkem:
While verifying against lld on the existing pack-relative-relocs test, I noticed Wild produces 5 RELR entries vs lld's 4 for the test binary.

### Reproducer (relr.c):

```
static int a, b, c, d;
static int *p1 = &a;
static int *p2 = &b;
static int *p3 = &c;
static int *p4 = &d;
int get() { return *p1 + *p2 + *p3 + *p4; }

gcc -fPIC -c relr.c -o relr.o

# lld: 2 entries (1 address + 1 bitmap 0x0f)
ld.lld -shared -z pack-relative-relocs relr.o -o relr.lld.so
readelf -x .relr.dyn relr.lld.so

# Wild: also 2 entries — matches lld for this simple case
./target/release/wild -shared -z pack-relative-relocs relr.o -o relr.wild.so
readelf -x .relr.dyn relr.wild.so
```

For the existing pack-relative-relocs test binary however Wild produces 5 entries vs lld's 4. The difference comes from .data.rel.local which has relocations at offsets 0, 9, and 26. Offset 9 is odd (RELA fallback). Offsets 0 and 26 are both RELR-eligible but delta=26 is not a multiple of 8, so they can't share a bitmap within the section each gets its own address entry. lld packs globally and sees them consecutive in the output.

This is the known cost of within-section-only packing. Is this acceptable for this PR or would you suggest a different approach?

libwild/src/elf_writer.rs
```diff
@@ -1056,14 +1118,60 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
...
+                            }
+                            relr_writer[idx].0.set(LittleEndian, 1 | (1 << bit));
+                            self.relr_dyn_index += 1;
+                            RelrWriterState::WithBitmap {
+                                base,
+                                bitmap_idx: idx,
+                            }
+                        }
+                    } else {
+                        // Out of window — start new run with address entry.
```
davidlattimore: If we run out of bits, we can start a new bitmap

davidlattimore (commented):
I think it's fine that we won't get quite as minimal packing as we would if we packed between sections. If we want to compare RELR bitmap utilisation, I'd suggest a full sized program rather than one of our test binaries, which are too small to really say much.

BTW, is it intentional that the PR is marked as a draft?

libwild/src/elf_writer.rs
deepakshirkem: Yes, intentionally draft  planning to add a test case before marking  ready for review. Will also fix the bitmap chaining (new bitmap entry  instead of new address entry when delta > 63*8) before marking ready.

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
davidlattimore, line 1: The test passes without the rest of your change. I think to actually test that RELR bitmaps are being used, you'd need to extend the test assertion system and add some assertions about the number of relocations performed via RELR bitmap. Given the slight differences in the number of bitmap-based relocations performed by lld vs wild for this test, you'd probably need to specify a range or a minimum rather than an exact number

deepakshirkem: You were right. Added an `ExpectSectionContains` directive to verify that the expected byte sequence exists in the Wild output. The test now checks for the bitmap value `0x07` in `.relr.dyn`, which is mathematically guaranteed for 3 consecutive 8-byte-aligned entries. I verified that the test fails on upstream/main and passes with this fix.

Do you have any other approach that would be more robust for verifying that the `.relr.dyn` bitmap encoding is happening correctly?

davidlattimore: I'd probably be inclined to have the test framework parse the `.relr.dyn` and count the total number of relr relocations. That should be the same for all linkers, but if it's not, then you make the assertion be the minimum number of relr relocations. The `ExpectSection` directive could be extended to accept properties, similar to how ExpectSym does.

```
//#RelrCount:4
//#ExpectSection:.relr.dyn max_size=2
```

It's good, where possible to have assertions apply to all linkers, not just wild. With these two assertions, it should be possible to configure them such that they pass for both wild and lld - possibly max_size for the section might need to be 3 if wild has one more non-bitmap relocation.

wild/tests/integration_tests.rs
```diff
@@ -67,7 +67,12 @@
 //!
 //! Contains:{string} Checks that the output binary does contain the specified string.
 //!
-//! ExpectSection:{section_name} Checks that the specified section exists in the output binary.
+//! ExpectSection:{section_name} [properties] Checks that the specified section exists in the
+//! output binary. Optional properties:
+//!   max_entries=N: Asserts the section has at most N entries (entry size assumed to be 8 bytes).
```
davidlattimore: Using an entity count seems like a good idea. Rather than assuming 8 byte entities though, how about using the entity size property on the section? It looks like we're setting it to 8 for `.relr.dyn`. If it's set to 0, then we should probably treat it as 1.

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
```diff
@@ -0,0 +1,37 @@
...
+// a bitmap entry.
+//#Object:runtime.c
+//#Object:init.c:-fPIC
+//#Mode:dynamic
+//#LinkArgs:-pie -z now -z pack-relative-relocs --no-gc-sections
+//#DiffMatchAny:true
+// GNU ld ignores `-z pack-relative-relocs` on RISC-V.
+//#ReferenceLinkers:lld
+// Verify all 3 constructor pointers are correctly encoded as RELR relocations.
+//#RelrCount:3
```
deepakshirkem, line 17: @davidlattimore  Not sure how to handle arch-specific values. Can you help me here.

davidlattimore: I think if a value like this is arch-specific, then the best bet is to have variants of the test for each arch that inherit from the base config.

wild/tests/integration_tests.rs
```diff
@@ -4278,6 +4318,62 @@ impl Assertions {
...
+        ensure!(
+            count >= expected_count,
+            "RelrCount: expected at least {} RELR relocations, got {}",
+            expected_count,
+            count,
+        );
+        Ok(())
+    }
+
+    fn verify_expected_sections_with_assertions(
```
davidlattimore, line 4351: ExpectSection already existed. Is there some old code that needs cleaning up?

deepakshirkem: You're right. `ExpectSection` now has two separate code paths depending on whether properties are present. The old `expected_sections` is still used for the Wasm handling and program header checks.

I can either merge everything into a single `Vec<ExpectedSection>` and update all the call sites, or leave it as it is for now and do the cleanup in a follow-up PR. I'm happy to do whichever you think is the better approach.

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
deepakshirkem: Thank-You. It looks like we'd need architecture-specific configs, something like:

```
//#Config:x86_64:default
//#SkipArch:aarch64,riscv64,loongarch64,ppc64le
//#ExpectSection:.relr.dyn max_entries=2
```

The only concern is that I can't run the AArch64, RISC-V64, LoongArch64, or PPC64LE tests locally to verify their `max_entries` values. From the CI failures, I observed the following values: AArch64 = 6, LoongArch64 = 6, RISC-V64 = 7, and PPC64LE = 4.

Do you think it's okay to hardcode these values based on the CI results, or is there a better approach?

wild/tests/integration_tests.rs
davidlattimore: It's fine to leave it for a subsequent PR if you prefer

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
davidlattimore: It's fine to base the values from what you see in CI. If you'd like to get the other architectures working locally, let me know and I can provide guidance. On Ubuntu, or Ubuntu in a docker container, it should just be a matter of apt-get installing some packages.

BTW, I wouldn't suggest using "SkipArch" for this - use "Arch" instead, that way each test can just specify the one arch that it runs for.

deepakshirkem: >If you'd like to get the other architectures working locally, let me know and I can provide guidance. On Ubuntu, or Ubuntu in a docker container, it should just be a matter of apt-get installing some packages.

Yes, That would be great.

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
```diff
@@ -0,0 +1,54 @@
+// Verify that consecutive RELR-eligible relocations are packed into bitmap
+// entries rather than emitted as individual address entries.
+//
+// Three consecutive constructors in .init_array produce function pointers at
+// consecutive 8-byte-aligned offsets. Without bitmap packing, each would get
+// its own RELR address entry. With packing, they share an address entry and
+// a bitmap entry.
+//#Object:runtime.c
```
davidlattimore, line 9: The first config should be declared as abstract, otherwise get run - with all architectures - `#AbstractConfig:default`

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
```diff
@@ -0,0 +1,54 @@
...
+// GNU ld ignores `-z pack-relative-relocs` on RISC-V.
+//#ReferenceLinkers:lld
+// Verify all 3 constructor pointers are correctly encoded as RELR relocations.
+//#RelrCount:3
+// Arch-specific configs to verify bitmap packing reduced the entry count.
+// Values determined from CI output.
+//#Config:x86_64:default
+//#Arch:x86_64
+//#ExpectSection:.relr.dyn max_entries=2
+//#Config:aarch64:default
```
davidlattimore, line 23: Can you put a blank line before each `#Config` directive? Otherwise it's pretty hard to see where they start and end.

wild/tests/sources/elf/relr-bitmap/relr-bitmap.c
davidlattimore: Have a look at `docker/ci/ubuntu-cross.Dockerfile`. That should show which packages will be needed. There's also instructions in `CONTRIBUTING.md`, but they're quite possibly incomplete, so the dockerfile is likely what should be referred to. If you run into any trouble, let me know and I'll see what I can figure out. If there is anything that could be better documented, then feel free to send a PR to update the docs.

davidlattimore (approved):

PR #2181, opened 2026-07-06
feat: support the DEFINED command in linker scripts

vishruth-thimmaiah:
This PR adds support for the `DEFINED` command in linker scripts, which is used to check if a symbol has been declared or not.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -213,6 +213,9 @@ pub(crate) fn evaluate_expression<'data, P: Platform>(
                 eval!(if_false)
             }
         }
+        Expression::Defined(name) => Ok(symbol_db
+            .get_unversioned(&UnversionedSymbolName::prehashed(name))
+            .map_or(0, |_| 1)),
     }
 }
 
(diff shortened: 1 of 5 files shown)


libwild/src/expression_eval.rs
```diff
@@ -213,6 +213,9 @@ pub(crate) fn evaluate_expression<'data, P: Platform>(
                 eval!(if_false)
             }
         }
+        Expression::Defined(name) => Ok(symbol_db
+            .get_unversioned(&UnversionedSymbolName::prehashed(name))
```
davidlattimore, line 217: I'm not sure that this will do the correct thing if there are objects that reference, but don't define the symbol. Can you add a test for that? i.e. have an object that has a weak reference to the symbol.

wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.c
```diff
@@ -4,3 +4,7 @@
 void _start() {}
 
 int symbol4 = 0x4000;
+int symbol6 __attribute__((weak));
```
davidlattimore, line 11: Note that declaring a variable doesn't create a reference to it - you need to actually have some code that uses the variable for it to end up in the symbol table.

davidlattimore: You can verify this by running `readelf -Ws` on the object file to see what undefined symbols are listed.

davidlattimore (approved):

PR #2182, opened 2026-07-06
refactor: Unify ExpectSection assertion handling

deepakshirkem:
Follow-up cleanup from #2179.

This change simplifies the implementation of the `ExpectSection` directive by using a single `expected_sections: Vec<ExpectedSection>` instead of maintaining two separate internal fields. Previously, plain section existence checks and section property assertions followed different code paths. With this change, a plain `ExpectSection:.relr.dyn` creates an `ExpectedSection` with default (empty) assertions, preserving the existing behavior while keeping the implementation simpler. As part of the cleanup, `verify_expected_sections_with_assertions` has also been renamed to `verify_section_max_entries` to better reflect its purpose.

This is a refactoring only and does not change any behavior. All existing tests continue to pass.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1503,8 +1503,7 @@ struct Assertions {
     expected_load_alignments: Vec<u64>,
     expected_dynamic_entries: Vec<String>,
     absent_dynamic_entries: Vec<String>,
-    expected_sections: Vec<String>,
-    expected_sections_with_assertions: Vec<ExpectedSection>,
+    expected_sections: Vec<ExpectedSection>,
     absent_sections: Vec<String>,
     expected_section_bytes: Vec<ExpectedSectionBytes>,
     relr_count: Option<u64>,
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -4173,7 +4169,12 @@ impl Assertions {
         );
 
         validate_wasm(path, linker_used.name())?;
-        ensure_wasm_sections(path, &self.expected_sections, linker_used.name())?;
+        let section_names: Vec<String> = self
+            .expected_sections
+            .iter()
+            .map(|s| s.section_name.clone())
+            .collect();
+        ensure_wasm_sections(path, &section_names, linker_used.name())?;
```
lapla-cogito, line 4235: Wouldn't this silently pass if a Wasm test specifies a directive expecting `max_entries`, since it only checks for the presence of the section? I don't think this actually preserves the existing behavior. Since this isn't something we intend to validate, I think it would be better to reject it explicitly. You can write like this:
```rs
ensure!(
    self.expected_sections
        .iter()
        .all(|s| s.assertions == SectionAssertions::default()),
    "Section property assertions are not supported for Wasm"
);
```

(The Mach-O test doesn't show any behavioral change because it uses a blacklist approach, which can easily miss newly added directives. I think that's somewhat problematic, and I might even submit a patch to switch it to a whitelist approach, similar to what Wasm does.)

wild/tests/integration_tests.rs
```diff
@@ -4348,11 +4350,11 @@ impl Assertions {
...
 
-    fn verify_expected_sections_with_assertions(
+    fn verify_section_max_entries(
         &self,
         obj: &object::read::elf::ElfFile64<'_, object::Endianness>,
     ) -> Result {
-        for expected in &self.expected_sections_with_assertions {
+        for expected in &self.expected_sections {
             let section = obj
                 .section_by_name(&expected.section_name)
```
lapla-cogito, line 4421: While not directly related, I think calling `section_by_name` for each `expected_sections` seems redundant since those with `max_entries: None` have already been verified to exist by `verify_expected_sections`. For processing only those with `max_entries`, you could write it like this:
```rs
for expected in &self.expected_sections {
    let Some(max_entries) = expected.assertions.max_entries else {
        continue;
    };
    let size = section.size();
    let entry_size = section.elf_section_header().sh_entsize.get(obj.endian());
    let entry_size = if entry_size == 0 { 1 } else { entry_size };
    let actual_entries = size / entry_size;
    ensure!(
        actual_entries <= max_entries,
        "Section `{}` has {} entries, expected at most {}. \
        Bitmap packing may not be working correctly.",
        expected.section_name,
        actual_entries,
        max_entries,
    );
}
```

lapla-cogito (approved):
Thanks!

Issue #2185, opened 2026-07-06
Mach-O: Off-by-one in calculating __DATA_CONST index


joanna-jo:
When writing the chained fixup table, `starts_in_image` contains `seg_count` at index 0, but when we calculate `starts_in_image[data_const_segment_index + 1]`, it seems we don’t account for `__PAGEZERO` (since this is calculated from `active_segments`). This leads to the following error from dyld_info: `chained fixups, segment_offset does not match vmaddr from LC_SEGMENT in segment #2` since this now writes to `__DATA` and not `__DATA_CONST` in `starts_in_image`.

joanna-jo:
I found this issue while working on #2063 but happy to work on this too!

PR #2186, opened 2026-07-07
port(MachO): Align __LINKEDIT sections

### LINKED ISSUE #2063: Mach-O: properly align `__LINKEDIT` sections
Right now, we emit a couple of "sections" in the `__LINKEDIT` segment (e.g. `output_section_id::CHAINED_FIXUP_TABLE`), but they are not properly aligned:
```
❯ dyld_info a.out
a.out [arm64]:
   mis-aligned LINKEDIT content 'chained fixups'
```

Based on my quick research, each "section" might need a different alignment.

joanna-jo:
Fixes #2063.

According to dyld ([alignments](https://github.com/apple-oss-distributions/dyld/blob/fd8d0c4d52320ebf64db34f3cb280310d905c5ae/common/MachOLayout.cpp#L129-L207), [alignment check](https://github.com/apple-oss-distributions/dyld/blob/fd8d0c4d52320ebf64db34f3cb280310d905c5ae/common/MachOLayout.cpp#L270-L274)) and lld ([alignments](https://github.com/llvm/llvm-project/blob/6f47c92a82cc142fe5b728a4d90987e15876b372/lld/MachO/SyntheticSections.h#L54-L80), [code signature alignment](https://github.com/llvm/llvm-project/blob/6f47c92a82cc142fe5b728a4d90987e15876b372/lld/MachO/SyntheticSections.cpp#L1567-L1590)), the current `__LINKEDIT` blobs/“sections” we have are expected to have a minimum alignment of: 

```
STRTAB: 1
CHAINED_FIXUPS: 8
SYMTAB: 8
CODE_SIGNATURE: 16
```

`STRTAB` and `CODE_SIGNATURE` are already correctly aligned, so the alignments were only changed for `CHAINED_FIXUPS` and `SYMTAB`. `CHAINED_FIXUPS`’ size was also aligned upwards to USIZE since `OffsetVerifier` needs this “section” to have an aligned size (`wild: error: Unexpected memory offsets: Part #36 (section DYLD_CHAINED_FIXUPS_TABLE alignment: 8) has non aligned size: 0x4e`). This follows what [lld does for padding `CHAINED_FIXUPS`](https://github.com/llvm/llvm-project/blob/6f47c92a82cc142fe5b728a4d90987e15876b372/lld/MachO/SyntheticSections.h#L71-L80). 

I manually checked with dyld_info and otool using misaligned files, and while this fixes the errors for misaligned `__LINKEDIT` content, there’s another dyld_info error (`chained fixups, segment_offset does not match vmaddr from LC_SEGMENT in segment #2`) that seems like an off-by-one in how we calculate the index for `__DATA_CONST` when writing the chained fixup table. I’ve opened a separate issue for that here: #2185.

### DIFF
--- a/libwild/src/macho.rs
+++ b/libwild/src/macho.rs
@@ -1318,7 +1318,10 @@ impl platform::Platform for MachO {
         fixup_table_size += CHAINED_FIXUP_PAGE_START_SIZE
             * (state.imported_symbols.len() as u64).div_ceil(MACHO_PAGE_ALIGNMENT.value());
 
-        mem_sizes.increment(part_id::CHAINED_FIXUP_TABLE, fixup_table_size);
+        mem_sizes.increment(
+            part_id::CHAINED_FIXUP_TABLE,
+            alignment::USIZE.align_up(fixup_table_size),
+        );
     }
 
     fn finalise_sizes_all<'data>(
@@ -1709,10 +1712,12 @@ const SECTION_DEFINITIONS: [BuiltInSectionDetails; NUM_BUILT_IN_SECTIONS] = {
     };
     defs[output_section_id::CHAINED_FIXUP_TABLE.as_usize()] = BuiltInSectionDetails {
         kind: SectionKind::Primary(SectionName(b"DYLD_CHAINED_FIXUPS_TABLE")),
+        min_alignment: alignment::USIZE,
         ..DEFAULT_DEFS
     };
     defs[output_section_id::SYMTAB_GLOBAL.as_usize()] = BuiltInSectionDetails {
         kind: SectionKind::Primary(SectionName(b"SYMTAB")),
+        min_alignment: alignment::USIZE,
         ..DEFAULT_DEFS
     };
     defs[output_section_id::CODE_SIGNATURE.as_usize()] = BuiltInSectionDetails {
--- a/linker-diff/src/header_diff.rs
+++ b/linker-diff/src/header_diff.rs
@@ -22,13 +22,19 @@ use object::ObjectSymbol as _;
 use object::Section;
 #[allow(clippy::wildcard_imports)]
 use object::elf::*;
+use object::macho::LC_CODE_SIGNATURE;
+use object::macho::LC_DYLD_CHAINED_FIXUPS;
 use object::read::elf::Dyn;
+use object::read::macho::LoadCommandVariant;
 use object::read::macho::MachHeader;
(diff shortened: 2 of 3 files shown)


davidlattimore (commented):
Thanks! Do you think it's practical to add a test that would catch this bug?

davidlattimore:
One possible way in which this might be tested is to look at making linker-diff detect that the alignment of these sections is different from what lld emits.

joanna-jo:
Thanks for looking this over and for the suggestion! :) I was also wondering what the best way to add a regression test here would be - I briefly looked into assertions but I’ll try supporting this directly with linker-diff then!

marxin:
Thank you @joanna-jo for the very first contribution! Really appreciate your research part where you investigated what's the expected alignment of the various `__LINKEDIT` sections.

joanna-jo:
This is now the linker-diff output before vs. after our fix on a file with misaligned `SYMTAB`, `DYLD_CHAINED_FIXUPS`, and `CODE_SIGNATURE`:

**Before:**
```
cargo run --bin linker-diff -- --wild-defaults --only 'linkedit.*' --ref /tmp/t.lld /tmp/t.wild
   Compiling linker-diff v0.9.0 (/Users/joannajo/wild/linker-diff)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.18s
     Running `target/debug/linker-diff --wild-defaults --colour never --only 'linkedit.*' --ref /tmp/t.lld /tmp/t.wild`
wild: /tmp/t.wild
lld: /tmp/t.lld
linkedit.CODE_SIGNATURE.dataoff.alignment
  wild offset 0xc11f is not 16-byte aligned
  lld OK

linkedit.DYLD_CHAINED_FIXUPS.dataoff.alignment
  wild offset 0xc027 is not 8-byte aligned
  lld OK

linkedit.SYMTAB.symoff.alignment
  wild offset 0xc0cf is not 8-byte aligned
  lld OK
```

**After:**
```
cargo run --bin linker-diff -- --wild-defaults --only 'linkedit.*' --ref /tmp/t.lld /tmp/t.wild
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/linker-diff --wild-defaults --colour never --only 'linkedit.*' --ref /tmp/t.lld /tmp/t.wild`
No differences or validation failures detected
```

davidlattimore (approved):
Thanks :)

PR #2187, opened 2026-07-07
port(wasm): Deduplicate the output type section

lapla-cogito:
wasm-ld deduplicates type sections with the same signature. You can observe this in existing multi-object test:

```
（▰╹◡╹）❯  wasm-tools print wild/tests/build/wasm/wasm32/multi-object/default/multi-object.wat.wild
(module
  (type (;0;) (func (result i32)))
  (type (;1;) (func))
  (type (;2;) (func (result i32)))
  (export "_start" (func 0))
...
)

（▰╹◡╹）❯  wasm-tools print wild/tests/build/wasm/wasm32/multi-object/default/multi-object.wat.wasm-ld
(module $multi-object.wat.wasm-ld
  (type (;0;) (func))
  (type (;1;) (func (result i32)))
  (memory (;0;) 2)
...
)
```

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -32,6 +32,7 @@ use wasmparser::ConstExpr;
 use wasmparser::DataKind;
 use wasmparser::DataSectionReader;
 use wasmparser::ExportSectionReader;
+use wasmparser::FuncType;
 use wasmparser::FunctionSectionReader;
 use wasmparser::GlobalSectionReader;
 use wasmparser::GlobalType;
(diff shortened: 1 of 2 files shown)


lapla-cogito:
With this patch, Wild no longer emits a simple concatenation of each object's type section and identical function types are collapsed and type indices are rewritten:

```
（▰╹◡╹）❯  wasm-tools print wild/tests/build/wasm/wasm32/multi-object/default/multi-object.wat.wild
(module
  (type (;0;) (func (result i32)))
  (type (;1;) (func))
  (export "_start" (func 0))
  (export "helper" (func 1))
  (func (;0;) (type 1)
    call 1
    i32.const 42
    i32.ne
    if ;; label = @1
      unreachable
    end
  )
  (func (;1;) (type 0) (result i32)
    i32.const 42
  )
)
```

davidlattimore (approved):

PR #2188, opened 2026-07-08
port(wasm): Resolve `__data_end` and `__heap_base` as linker data addresses

lapla-cogito:
Clang references these as undefined data symbols with `MEMORY_ADDR` relocations, not as globals. Assign absolute addresses after static data layout (data_end, and data_end plus the reserved stack gap for heap_base) and grow memory when the stack gap is required.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2844,6 +2844,7 @@ fn emit_reserved_linker_definitions(layout: &mut WasmLayout<'_>, indices: &Linke
 fn fill_linker_defined_values(
     layout: &mut WasmLayout<'_>,
     indices: &LinkerDefinedIndices,
+    needs_stack_gap: bool,
 ) -> Result {
     if let Some(defined_slot) = indices.stack_pointer_defined_slot {
         let sp = layout
@@ -2862,7 +2863,7 @@ fn fill_linker_defined_values(
     }
 
     let mut bytes_needed = u64::from(layout.data_end.max(layout.memory_base));
-    if indices.stack_pointer_defined_slot.is_some() {
+    if indices.stack_pointer_defined_slot.is_some() || needs_stack_gap {
         bytes_needed =
             bytes_needed.max(u64::from(layout.data_end.saturating_add(LINKER_STACK_SIZE)));
     }
@@ -3058,14 +3059,15 @@ where
         ensure_memory_export(&mut layout.exports);
     }
     layout.data_end = memory_cursor;
-    compute_data_addresses(
+    let needs_stack_gap = compute_data_addresses(
         &mut layout.object_index_maps,
         &layout.per_object_symbols,
         &layout.object_data_layouts,
         &layout_inputs,
         symbol_db,
+        layout.data_end,
     )?;
-    fill_linker_defined_values(&mut layout, &indices)?;
+    fill_linker_defined_values(&mut layout, &indices, needs_stack_gap)?;
     ensure_entry_export(
         &mut layout.exports,
         &layout_inputs,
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -3234,19 +3236,54 @@ fn data_symbol_memory_address(
...
+    needs_stack_gap: bool,
+}
+
+/// Absolute addresses linker-defined data symbols.
+fn linker_defined_data_symbol_address(
+    name: &[u8],
+    data_end: u32,
+) -> Result<Option<LinkerDefinedDataAddress>> {
+    match name {
+        b"__data_end" => Ok(Some(LinkerDefinedDataAddress {
```
davidlattimore, line 3253: Similar to previous PRs, the matching on symbol names makes me wary. It feels potentially wasteful and like it ideally shouldn't be necessary. Fine for now, but longer-term it'd be good to consider alternatives.

wild/tests/sources/wasm/data-end-heap-base/data-end-heap-base2.c
```diff
@@ -0,0 +1,4 @@
+extern char __data_end;
```
davidlattimore, line 1: Does this actually cause a symbol reference to be emitted? At least from my previous observations when compiling for x86_64, it doesn't - you'd need to actually have code that references the symbol for that.

libwild/src/wasm.rs
```diff
@@ -3234,19 +3236,54 @@ fn data_symbol_memory_address(
...
+            address: data_end,
+            needs_stack_gap: false,
+        })),
+        b"__heap_base" => {
+            let address = data_end
+                .checked_add(LINKER_STACK_SIZE)
+                .ok_or_else(|| crate::error!("Wasm __heap_base address overflow"))?;
+            Ok(Some(LinkerDefinedDataAddress {
+                address,
+                needs_stack_gap: true,
```
davidlattimore, line 3266: What is it about __heap_base that means it needs stack gap? Could be good to document that here.

wild/tests/sources/wasm/data-end-heap-base/data-end-heap-base2.c
lapla-cogito: Ah, yes, that's true. Done, thanks.

libwild/src/wasm.rs
lapla-cogito: done

davidlattimore: OK, I think I get it. I guess in the ELF parts of our linker, symbols like __heap_base would be attached to the start/end of sections, one of those sections would be the stack and memory addresses would just be computed by adding sizes.

libwild/src/wasm.rs
```diff
@@ -3260,6 +3260,9 @@ fn linker_defined_data_symbol_address(
                 .ok_or_else(|| crate::error!("Wasm __heap_base address overflow"))?;
             Ok(Some(LinkerDefinedDataAddress {
                 address,
+                // We place the stack in [data_end, data_end + LINKER_STACK_SIZE) and define
+                // __heap_base as the end of that range (same as initial __stack_pointer). The
+                // address itself therefore requires that memory cover the stack gap.
                 needs_stack_gap: true,
```
davidlattimore, line 3266: I tried changing this line to `needs_stack_gap: false` and the tests still all passed. Is it possible to make the tests verify that this is working correctly?

lapla-cogito: As noted in the comments for the newly added tests, this is a bit difficult to test with an integration test, so I covered it with unit tests instead.

davidlattimore (approved):

PR #2189, opened 2026-07-08
refactor: Use an allowlist for Mach-O test assertions

lapla-cogito:
context: https://github.com/wild-linker/wild/pull/2182#discussion_r3532232142

Using an allowlist is more robust against future changes, such as the addition of new directives.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -4147,6 +4147,11 @@ impl Assertions {
 
         let obj = object::File::parse(bytes.as_slice())?;
 
+        // For Mach-O.
+        if matches!(obj, object::File::MachO64(_)) {
+            return self.check_macho_path(&obj, &bytes);
+        }
+
         self.verify_file_kind(&obj)?;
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -4205,6 +4191,44 @@ impl Assertions {
...
+            obj.dynamic_symbols(),
+            "dynsym",
+        )?;
+        self.verify_symbols_absent(&self.no_sym, obj.symbols(), "symtab")?;
+        self.verify_expected_sections(obj)?;
+        self.verify_absent_sections(obj)?;
+        self.verify_section_bytes(obj)?;
+        self.verify_strings(bytes)?;
+        verify_no_overlapping_sections(obj)?;
+        verify_no_overlapping_segments(obj)?;
```
lapla-cogito, line 4228: This currently only performs an early return for Mach-O, but has been added in anticipation of future support.

davidlattimore (approved):

PR #2192, opened 2026-07-10
port(wasm): Synthesize `__wasm_call_ctors` from InitFuncs and wrap the command entry

lapla-cogito:
Parse linking InitFuncs and emit a priority-ordered `__wasm_call_ctors` body. When objects do not already reference that symbol, export a synthetic entry that runs constructors before user `_start`, matching wasm-ld.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2870,11 +2870,12 @@ fn function_type_for_symbol<'a>(
     input: &'a WasmObjectLayoutInput<'_>,
     sym: &WasmSymbol,
 ) -> Result<&'a wasmparser::FuncType> {
-    let n_imports = u32::try_from(input.function_imports.len()).context("too many Wasm imports")?;
-    let type_index = if sym.index < n_imports {
-        input.function_imports[sym.index as usize].type_index
+    let sym_index = sym.index as usize;
+    let n_imports = input.function_imports.len();
+    let type_index = if sym_index < n_imports {
+        input.function_imports[sym_index].type_index
     } else {
-        let local = (sym.index - n_imports) as usize;
+        let local = (sym_index - n_imports);
         input
             .module_functions
             .get(local)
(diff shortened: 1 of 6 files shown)


libwild/src/wasm.rs
```diff
@@ -2783,14 +2839,98 @@ fn deduplicate_output_types(layout: &mut WasmLayout<'_>) {
...
+    }
+    bytes.push(0x0b); // end
+    bytes
+}
+
+fn function_type_for_symbol<'a>(
+    input: &'a WasmObjectLayoutInput<'_>,
+    sym: &WasmSymbol,
+) -> Result<&'a wasmparser::FuncType> {
+    let n_imports = u32::try_from(input.function_imports.len()).context("too many Wasm imports")?;
```
davidlattimore: I think this might look a bit nicer if you instead did something like `let sym_index = sym.index as usize`. Then this could stay as a usize and you wouldn't need the expect call.

libwild/src/wasm.rs
```diff
@@ -2783,14 +2839,98 @@ fn deduplicate_output_types(layout: &mut WasmLayout<'_>) {
...
+) -> Result<Vec<u32>> {
+    let mut items = Vec::new();
+    for (obj_idx, input) in inputs.iter().enumerate() {
+        let index_map = object_index_maps
+            .get(obj_idx)
+            .context("missing Wasm object index map for init function")?;
+        for init in &input.init_funcs {
+            let sym = input
+                .symbols
+                .get(init.symbol_index as usize)
```
davidlattimore: Usually we just use indexing operators to index unless bad user input could cause an index to be invalid or we think an internal error is likely to result in a bad index. That'd allow the ok_or_else bit to disappear.

wild/tests/sources/wasm/constructors-priority/constructors-priority.c
davidlattimore, line 1: Do you think it'd be worthwhile having a second file containing constructors?  Just to make sure that constructors are properly ordered between files

lapla-cogito: That makes sense, thanks.

davidlattimore (approved):

PR #2193, opened 2026-07-10
chore(wasm): Defer relocations targeting debug sections

lapla-cogito:
When enhancing WASI support, most of the crt and `libc.a` included in the `wasi-libc` package contain debug relocations. Since we haven't yet added DWARF support for Wasm, link failures occur whenever debug relocations are encountered. Therefore, for the time being, we will disable the processing of debug relocations to facilitate WASI support enhancement. I believe implementing DWARF support is appropriate only after the final layout processing for Wasm has been stabilized.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -484,7 +484,25 @@ pub(crate) struct WasmFunctionBody<'data> {
     pub(crate) object_index: usize,
 }
 
+fn is_debug_section_name(name: &[u8]) -> bool {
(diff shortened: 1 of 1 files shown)


lapla-cogito:
I intentionally used "chore(wasm)" instead of "port(wasm)" for the PR title because this PR is not intended to introduce a new feature, but rather to provide a temporary workaround.

davidlattimore (approved):

PR #2196, opened 2026-07-11
feat: GOT-based RELR bitmap packing

### LINKED ISSUE #1853: Add support for bitmap packing of RELR relocations

deepakshirkem:
Follow-up to #2179 which added RELR bitmap packing for direct relocations. This PR implements bitmap packing for GOT-based RELR entries.

Approach 
Split the GOT into two parts:
- GOT: non-RELR entries (dynamic, TLS, ifunc, downgraded-to-local)
- GOT_RELR: RELR-eligible entries (contiguous for bitmap packing)

Fixes  #1853


deepakshirkem:
@davidlattimore, The CI is now green. Before adding the tests, I wanted to get your thoughts on this approach and see if there's anything you think could be improved. Whenever you get a chance.

libwild/src/elf.rs
lapla-cogito, line 1666: This `needs_plt` branch should also contain a conditional branch similar to the one added in the `needs_got` branch, and call `allocate_got_relr` when necessary. Additionally, it would be beneficial to refactor the `is_got_relr` logic in `{allocate, create, process}_resolution` into a single shared function.

libwild/src/elf_writer.rs
```diff
@@ -612,6 +612,8 @@ enum RelrWriterState {
 struct TableWriter<'layout, 'out> {
     output_kind: OutputKind,
     got: &'out mut [u64],
+    got_relr: &'out mut [u64],
+    got_relr_base: Option<u64>,
```
lapla-cogito: Is this field needed? This value doesn't seem to be read from anywhere.

libwild/src/part_id.rs
```diff
@@ -26,54 +26,52 @@ pub(crate) const SECTION_HEADERS: PartId = PartId(3);
...
+pub(crate) const WASM_TABLE: PartId = PartId(41);
+pub(crate) const WASM_MEMORY: PartId = PartId(42);
+pub(crate) const WASM_GLOBAL: PartId = PartId(43);
+pub(crate) const WASM_EXPORT: PartId = PartId(44);
+pub(crate) const WASM_START: PartId = PartId(45);
+pub(crate) const WASM_ELEMENT: PartId = PartId(46);
+pub(crate) const WASM_DATA_COUNT: PartId = PartId(47);
+pub(crate) const WASM_CODE: PartId = PartId(48);
+pub(crate) const WASM_DATA: PartId = PartId(49);
+pub(crate) const NUM_SINGLE_PART_SECTIONS: u32 = 50;
```
lapla-cogito, line 75: It would be better to keep the blank line here.

```suggestion
pub(crate) const WASM_DATA: PartId = PartId(49);

pub(crate) const NUM_SINGLE_PART_SECTIONS: u32 = 50;
```

libwild/src/elf_writer.rs
deepakshirkem: removed. Thank-You

libwild/src/elf.rs
deepakshirkem: Done: 
I extracted the `is_got_relr` logic into a shared `is_got_relr_eligible()` function, which is now used by `allocate_resolution`, `create_resolution`, and `process_resolution`.

For the `needs_plt + needs_ifunc_got_for_address` case, I added a TODO comment for now. Do you think it would be okay to handle that in a follow-up PR?

libwild/src/part_id.rs
deepakshirkem: Done:

libwild/src/elf.rs
lapla-cogito: > For the `needs_plt + needs_ifunc_got_for_address` case, I added a TODO comment for now. Do you think it would be okay to handle that in a follow-up PR?

Hm, I don't really see the rationale for fixing this in a separate PR 🤔 I assumed like:

```rs
let is_got_relr = is_got_relr_eligible(flags, has_dynamic_symbol, args, output_kind);
let num_got_entries = if flags.needs_ifunc_got_for_address() {
    2
} else {
    1
};
resolution.format_specific.got_address = if is_got_relr && num_got_entries == 1 {
    Some(allocate_got_relr(memory_offsets));
} else {
    Some(allocate_got(num_got_entries, memory_offsets));
}
```

This is a relatively small change, but the current implementation has inconsistent assumptions about allocate/create/write operations on the resolution, so addressing this in this PR is worthwhile I think.

deepakshirkem: @lapla-cogito, Thank you! I've updated the fix. Could you please verify it when you get a chance?

libwild/src/value_flags.rs
```diff
@@ -179,6 +183,11 @@ impl ValueFlags {
         self.contains(ValueFlags::DOWNGRADE_TO_LOCAL)
     }
 
+    #[must_use]
+    pub(crate) fn is_section_boundary(self) -> bool {
+        self.contains(ValueFlags::IS_SECTION_BOUNDARY)
```
davidlattimore: If I change this to always just return `false`, the tests still pass, which suggests that either this isn't necessary or the the tests are insufficient.

libwild/src/elf_writer.rs
```diff
@@ -931,12 +943,51 @@ impl<'layout, 'out> TableWriter<'layout, 'out> {
...
+    /// Writes bitmap-packed RELR entries for the entire GOT_RELR block.
+    /// Called after all symbol resolutions are processed.
+    fn write_got_relr_bitmap(&mut self, n: u64, base: u64) -> Result {
+        if base == 0 || n == 0 {
+            return Ok(());
+        }
+        let Some(relr_writer) = &mut self.relr_dyn else {
+            return Ok(());
+        };
+        const BITMAP_SLOTS: u64 = 63;
```
davidlattimore: This constant seems to be repeated elsewhere. Can it be moved into an outer scope so that it can be shared?

libwild/src/elf.rs
```diff
@@ -1331,6 +1337,38 @@ impl platform::Platform for Elf {
...
+    fn post_compute_sizes(section_part_sizes: &mut OutputSectionPartMap<u64>, args: &Self::Args) {
+        if !args.is_relr_enabled() {
+            return;
+        }
+        let got_relr_size = *section_part_sizes.get(part_id::GOT_RELR);
+        if got_relr_size == 0 {
+            return;
+        }
+        let n = got_relr_size / RELR_ENTRY_SIZE;
+        let bitmap_entries = n.saturating_sub(1).div_ceil(63);
```
davidlattimore: Can this 63 refer to a shared constant?

libwild/src/elf.rs
```diff
@@ -1331,6 +1337,38 @@ impl platform::Platform for Elf {
...
+        let relr_entries = 1 + bitmap_entries;
+        section_part_sizes.increment(part_id::RELR_DYN, relr_entries * RELR_ENTRY_SIZE);
+    }
+
+    fn post_compute_sizes_epilogue(
```
davidlattimore: It looks like this function just repeats all the same logic a the function above. Can we find a way to not do that? I wonder if the new post_compute_sizes hooks are needed though. Why can't we use apply_late_size_adjustments_epilogue?

davidlattimore (commented):
Do you think it'd be practical to add a test that verifies that this is working?

libwild/src/elf_writer.rs
deepakshirkem: Done:

libwild/src/elf.rs
deepakshirkem: Done::

deepakshirkem:
@davidlattimore  Addressed all review comments and added a test in  that verifies GOT-based RELR  bitmap packing. Not 100% sure the test approach is optimal. Happy to improve it if you have a better way.

wild/tests/sources/elf/got-relr-bitmap/got-relr-bitmap.s
```diff
@@ -0,0 +1,55 @@
...
+// Four hidden symbols accessed via GOT produce consecutive GOT entries.
+// With bitmap packing, they share one address entry and one bitmap entry
+// instead of four individual RELR address entries.
+//
+// Note: --no-relax is needed to prevent the linker from optimizing away
+// the GOT references (REX_GOTPCRELX relaxation).
+//#AbstractConfig:default
+//#LinkArgs:-shared -z pack-relative-relocs --no-relax
+//#RunEnabled:false
+//#DiffMatchAny:true
```
davidlattimore: DiffMatchAny won't have any effect if there's only one reference linker

wild/tests/sources/elf/got-relr-bitmap/got-relr-bitmap.s
```diff
@@ -0,0 +1,55 @@
+// Verify that consecutive RELR-eligible GOT entries are bitmap-packed.
+//
+// Four hidden symbols accessed via GOT produce consecutive GOT entries.
```
davidlattimore: The symbols are locals, not hidden

wild/tests/sources/elf/got-relr-bitmap/got-relr-bitmap.s
deepakshirkem: Done ::))

davidlattimore (approved):

PR #2197, opened 2026-07-12
docs: fix cross-arch setup instructions for loongarch64 and add ppc64le

deepakshirkem:
- Fix loongarch64 compiler package name on Ubuntu 24.04   (gcc-14-loongarch64-linux-gnu instead of gcc-loongarch64-linux-gnu)
- Add symlink instructions for loongarch64
- Add ppc64le cross-compilation instructions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,9 @@ then use `cargo test` as usual.
 
 Wild supports testing on non-native architectures using QEMU.
 
+> **Note:** Ubuntu 26.04 is recommended for cross-architecture testing.
+
+
 ### Setup
 
 1. Add required Rust targets:
@@ -86,6 +89,9 @@ Wild supports testing on non-native architectures using QEMU.
   
   # For loongarch64
   rustup target add --toolchain nightly loongarch64-unknown-linux-gnu loongarch64-unknown-linux-musl
+
+  # For ppc64le
+  rustup target add --toolchain nightly powerpc64le-unknown-linux-gnu
   ```
 
 2. Install required packages (for apt-based systems):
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -98,7 +101,14 @@ Wild supports testing on non-native architectures using QEMU.
   sudo apt install qemu-user gcc-riscv64-linux-gnu g++-riscv64-linux-gnu binutils-riscv64-linux-gnu
 
   # For loongarch64
-  sudo apt install qemu-user gcc-loongarch64-linux-gnu g++-loongarch64-linux-gnu binutils-loongarch64-linux-gnu
+  sudo apt install qemu-user gcc-14-loongarch64-linux-gnu g++-14-loongarch64-linux-gnu binutils-loongarch64-linux-gnu
+  # Create symlinks so the toolchain can be found without the version suffix:
```
davidlattimore: What OS and version did you need to do this on? I assume that this is only necessary because the above apt install line installed version 14 specifically rather than installing gcc-loongarch64-linux-gnu. I'm running on Ubuntu 26.04 and it doesn't look like I needed to do this. Our CI docker image also doesn't seem to do this.

deepakshirkem: @davidlattimore This is on Ubuntu 24.04 LTS. Should I add a note that this is only needed on Ubuntu 24.04?

davidlattimore: I'd probably add a note that running on Ubuntu 26.04 is recommended. If you wanted to add a section for 24.04, that's fine, but it shouldn't replace the instructions for 26.04.

deepakshirkem: @davidlattimore Your correct . I'll just add a note that Ubuntu 26.04 is recommended.

davidlattimore (approved):

PR #2199, opened 2026-07-14
chore: Disable unused features of all dependencies

davidlattimore:
This removes 11 crates from our Cargo.lock. Some tests needed updating to switch regexes to not require unicode tables.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,12 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "allocator-api2"
-version = "0.2.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -2284,10 +2186,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3eb099dcadcde5be9eef55e3a337128efd4e44b4c93122487e4d2e4e1c6627c"
 dependencies = [
  "bitflags",
- "hashbrown 0.17.1",
- "indexmap",
```
mati865, line 2288: Without `hash-collections` it will use BTree collections which might have perf impact in either way.

davidlattimore: Thanks! I've enabled that feature.

marxin:
Interesting! Have you managed to get a binary reduction after the removal of the default features? Or to any measurable speed-up of the compilation time?

Cargo.lock
```diff
@@ -1929,7 +1853,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.1",
+ "getrandom 0.3.4",
```
mati865, line 1856: Huh? It doesn't matter though, and with https://github.com/davidlattimore/perfetto-recorder/pull/13 we should be able to remove getrandom 0.3 entirely.

davidlattimore:
> Interesting! Have you managed to get a binary reduction after the removal of the default features? Or to any measurable speed-up of the compilation time?

The binary size (for `--profile opt`) reduced by about 2.7%. The cold build speed didn't seem to measurably change.

PR #2200, opened 2026-07-14
fix: use nop padding bytes in executable sections on x86 

### LINKED ISSUE #2191: Use trap/nop instructions instead of zero byte padding within EXEC sections on X86
We currently use zero byte padding when aligning executable sections. I noticed this issue when trying to link linux kernel modules on x86 with wild, the `MODPOST` tool reads zero byte padding as invalid instructions, and fails.

We should probably match lld behavior here, and emit the `int3` instruction on x86.

vishruth-thimmaiah:
Closes #2191

On x86, emit the `int3` instruction, instead of using zero byte padding, in sections with executable instructions.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -3000,7 +3000,8 @@ fn apply_relocation<
             let removed_bytes =
                 relax_deltas.map_or(0u64, |d| u64::from(d.delta_bytes_at(rel.offset())));
             let padding_bytes = addend.saturating_sub(removed_bytes) as usize;
-            A::fill_nop_padding(out, offset_in_section as usize, padding_bytes);
+            let offset_in_section = offset_in_section as usize;
+            A::fill_nop_padding(&mut out[offset_in_section..offset_in_section + padding_bytes]);
 
             return Ok(RelocationModifier::Normal);
         }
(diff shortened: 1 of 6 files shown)


libwild/src/elf_writer.rs
```diff
@@ -2226,7 +2231,11 @@ fn write_section_raw<'out, 'data>(
                         .copy_from_slice(&input_data[input_pos..]);
                     output_pos += remaining;
                 }
-                out[output_pos..].fill(0);
+                if A::requires_nop_padding_for_section(section_flags) {
+                    A::fill_nop_padding(out, output_pos, out.len() - output_pos);
+                } else {
+                    out[output_pos..].fill(0);
+                }
```
davidlattimore: Rather than having two new methods on Platform and repeating the call pattern twice, what about having just one method:

```
A::fill_padding(&mut out[output_pos..], section_flags);
```

The default implementation can just fill with 0, while x86_64 can check `section_flags`, then fill with either 0xcc or with 0.

Also, in the above proposed API, we just pass the slice, not the slice plus offset and length.

vishruth-thimmaiah: > what about having just one method:
> 
> ```
> A::fill_padding(&mut out[output_pos..], section_flags);
> ```

`fill_nop_padding` isn't a new method defined here, it is an existing method that was previously only defined for riscv, we can't replace it with the above api.

Instead I've defined `fill_section_padding`, for filling in section padding. I've also cleaned up `fill_nop_padding` to take in just a slice.

libwild/src/platform.rs
```diff
@@ -180,7 +180,7 @@ pub(crate) trait Arch: Send + Sync + 'static {
     ) -> Option<Self::Relaxation>;
 
     /// Fill `len` bytes of NOP padding at `offset` in `buf`.
```
davidlattimore: Comment is out-of-date

davidlattimore (approved):

PR #2201, opened 2026-07-14
feat: print version and emulation info when the -V flag is used

vishruth-thimmaiah:
This PR adds support for the `-V` flag, which prints version info and supported emulations. It also fixed an issue with `-v`, we now continue linking if input object files are provided.

Also adds powerpcle to the supported target and emulation lists.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -208,7 +208,8 @@ impl Args {
             Args::Elf(_) => {
                 writeln!(
                     stdout,
-                    "supported emulations: elf_x86_64 aarch64elf elf64lriscv elf64loongarch elf64lppc"
+                    "supported emulations: {}",
+                    crate::arch::SUPPORTED_EMULATIONS
                 )?;
             }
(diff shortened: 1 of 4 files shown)


libwild/src/args.rs
```diff
@@ -201,6 +202,19 @@ impl Args {
             Args::Wasm(wasm_args) => &mut wasm_args.common,
         }
     }
+
+    pub(crate) fn print_emulation_info(&self, stdout: &mut dyn Write) -> Result<()> {
+        match self {
+            Args::Elf(_) => {
+                writeln!(
+                    stdout,
+                    "supported emulations: elf_x86_64 aarch64elf elf64lriscv elf64loongarch elf64lppc"
```
davidlattimore: This looks to be duplicating the line from `elf.rs`. Can it be extracted out into a constant or function? If you can keep the list of supported emulations next to the list of supported targets, that'd be good, since it'll reduce the chances of them getting out of sync.

davidlattimore (approved):

PR #2202, opened 2026-07-14
port(wasm): Link static WASI libraries and preserve host imports

lapla-cogito:
Add `-L`/`-l` for Wasm so `libc.a` can be pulled from the search path. Allow table/index relocations in data so wasi-libc objects layout. Absorb unresolved `__tls_base` as a linker-defined global. Unresolved host imports keep their original module and name (`wasi_snapshot_preview1.*`).

Part of #1431

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -202,6 +202,8 @@
 //!
 //! The following input types support an optional template parameter. This should look something
 //! like `Shared:template(--push-state --as-needed $O --pop-state):filename.c`.
+//! `$O` is replaced with the built input path. `$OUT_DIR` is replaced with the test build
+//! directory (same as in `LinkArgs`). At least one of `$O` or `$OUT_DIR` must appear.
 //!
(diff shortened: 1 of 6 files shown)


wild/tests/integration_tests.rs
```diff
@@ -4090,13 +4098,17 @@ fn add_inputs_to_command(config: &Config, inputs: &[LinkerInput], command: &mut
...
             for a in template {
-                if a.contains(TEMPLATE_PLACEHOLDER) {
-                    command.arg(a.replace(TEMPLATE_PLACEHOLDER, path_str));
-                } else {
-                    command.arg(a);
-                }
+                let arg = a
+                    .replace(TEMPLATE_OUT_DIR_PLACEHOLDER, &out_dir)
+                    .replace(TEMPLATE_PLACEHOLDER, path_str);
+                command.arg(arg);
```
lapla-cogito, line 4111: wasm-ld cannot use absolute paths for the `-l` option.

davidlattimore (approved):

PR #2205, opened 2026-07-14
feat(API): add pub ThreadPool::new method

### LINKED ISSUE #2204: `libwild::run` using `activate_thread_pool` overrides global `ThreadPool`
The current public API is somewhat unfortunate: it either requires overriding Rayon’s global thread pool - which can only be initialized once - via `ThreadPoolBuilder::new()...build_global()`, or leaves client code with no way to skip this initialization because `ThreadPool::_jobserver_tokens` is private.

marxin:
The least invasive change that will allow downstream library consumers to skip the global thread pool initialization.

Resolves: #2204

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -486,6 +486,23 @@ pub struct ThreadPool {
     _jobserver_tokens: Vec<Acquired>,
 }
 
+impl ThreadPool {
+    /// Creates a `ThreadPool` marker without configuring the global rayon pool. Use this when the
+    /// global pool is already initialized; otherwise use [`CommonArgs::activate_thread_pool`].
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            _jobserver_tokens: Vec::new(),
+        }
+    }
+}
+
+impl Default for ThreadPool {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 // TODO: remove
 #[allow(clippy::large_enum_variant)]
 pub enum Args {


libwild/src/args.rs
```diff
@@ -486,6 +486,23 @@ pub struct ThreadPool {
     _jobserver_tokens: Vec<Acquired>,
 }
 
+impl ThreadPool {
+    /// Creates a `ThreadPool` marker without configuring the global rayon pool. Use this when the
```
davidlattimore, line 490: The comment on the type `ThreadPool` seems to no longer be true with this PR.

davidlattimore (commented):
Should we perhaps go further and stop using the global thread pool entirely?

marxin:
> Should we perhaps go further and stop using the global thread pool entirely?

Yeah, let me try preparing a change going into that direction.

marxin:
I tried the different approach of doing the `pool.install` approach:
```
    let thread_pool = args.common_mut().build_thread_pool()?;
    let linker = Linker::new();
    linker.run(&args, &thread_pool)?;
...
    pub fn run<'layout_inputs>(
        &'layout_inputs self,
        args: &'layout_inputs Args,
        thread_pool: &crate::args::ThreadPool,
    ) -> error::Result<LinkerOutput<'layout_inputs>> {
        thread_pool.pool.install(|| self.run_in_thread_pool(args))
    }
```

and it seems it messes (or makes it more convoluted) when it comes to the perfetto and the tracing stuff. It might have been easier if
```
    let linker = Linker::new();
    linker.run(&args)?;
```

will be run in a closure of the `thread_pool`, but using this approach, we're gonna break the existing API. What do you think?

davidlattimore:
I expect that there's at most a handful of people using Wild as a library at present, so changing the API should be fine.

I had imagined passing in the thread pool as you showed in the first example, but I wouldn't rule out having the user of the API enter the thread pool if that ends up nicer for some reason.

marxin:
All right, I think I came up with a pretty compact and flexible approach.
Superseded in #2209.

PR #2206, opened 2026-07-15
port(wasm): Align data segments and stack like wasm-ld

lapla-cogito:
Without segment alignment and 16-byte stack alignment, Wild's layout diverged from wasm-ld and WASI binaries could misbehave.

This patch makes wasi-libc-linked C programs actually runnable.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -116,9 +116,6 @@ const LINKER_MEMORY_BASE: u32 = 1024;
 /// Default stack size reserved above `__data_end` for the main stack.
 const LINKER_STACK_SIZE: u32 = 64 * 1024;
 
-/// Stack and post-data gap alignment.
-const LINKER_STACK_ALIGNMENT: Alignment = crate::alignment::STACK_ALIGNMENT;
-
 /// Empty function body: zero locals + `end`.
 const EMPTY_FUNCTION_BODY: &[u8] = &[0x00, 0x0b];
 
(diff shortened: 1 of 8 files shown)


libwild/src/wasm.rs
```diff
@@ -115,6 +116,9 @@ const LINKER_MEMORY_BASE: u32 = 1024;
 /// Default stack size reserved above `__data_end` for the main stack.
 const LINKER_STACK_SIZE: u32 = 64 * 1024;
 
+/// Stack and post-data gap alignment.
+const LINKER_STACK_ALIGNMENT: Alignment = crate::alignment::STACK_ALIGNMENT;
```
davidlattimore: How is this different to STACK_ALIGNMENT which it is defined in terms of? If it is different, perhaps it needs a different name, since the "LINKER_" prefix doesn't do much to distinguish it from `STACK_ALIGNMENT`, which is also part of the linker.

wild/tests/integration_tests.rs
```diff
@@ -633,6 +642,34 @@ type ElfFile64<'data> = object::read::elf::ElfFile64<'data, LittleEndian>;
...
+        .into_iter()
+        .chain(std::iter::once(PathBuf::from("/usr")));
+
+    candidates.find(|root| root.join("lib/wasm32-wasi/libc.a").exists())
+}
+
+fn wasi_sysroot_display() -> String {
+    wasi_sysroot()
+        .unwrap_or_else(|| PathBuf::from("/usr"))
+        .display()
```
davidlattimore: I think `display` should only really be used for display purposes - i.e. in a printed message that we display to the user. Given that this is substituted into a command-line used to invoke a linker, if there's characters that can't be encoded as UTF-8, the command is going to fail in a way that might be hard to debug. Given that this is a test, I think it'd be fine to panic if we encounter a path that we can't convert to a string.

libwild/src/wasm.rs
lapla-cogito: Initially, I kept them separate to explicitly distinguish Wasm's stack alignment from ELF's. However, if future inconsistencies were to arise (though this remains an extremely unlikely scenario) they would still be detectable through unit and integration tests, so I've now unified them.

wild/tests/integration_tests.rs
lapla-cogito: Thanks, done.

davidlattimore (approved):

PR #2207, opened 2026-07-15
chore(wasm): Add timing instrumentation for Wasm layout and write

lapla-cogito:
We've already been able to link several real-world programs for Wasm. For example, I've confirmed that SQLite can be built for WASI (though I've verified that some unsupported features cause incorrect runtime behavior. The build itself succeeds, however). Therefore, introducing instrumentation into each Wasm linking phase at this time seems worthwhile.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -15,6 +15,8 @@ use crate::output_section_id::SectionName;
 use crate::platform;
 use crate::symbol::UnversionedSymbolName;
 use crate::symbol_db::SymbolDb;
+use crate::timing_phase;
+use crate::verbose_timing_phase;
 use crate::wasm_writer::OutputExport;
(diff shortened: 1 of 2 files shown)


lapla-cogito:
For SQLite, the phase currently taking the most time is "Merge Wasm object layouts." This is intuitive given SQLite's object files typically contain several thousand functions, symbols, and data segments and this phase currently involves numerous copy operations, reducing these operations is both realistic and unlikely to conflict with other porting tasks.

There are still several parts that haven't been optimized yet due to prioritization of ensuring proper linking other than this phase, so I'll also incorporate optimizations for those as needed.

image

davidlattimore (approved):
Could also be worth considering if some of these phases that are just scopes within a larger function might be good to extract into separate functions

PR #2209, opened 2026-07-15
feat: do not override the global Rayon thread pool

### LINKED ISSUE #2204: `libwild::run` using `activate_thread_pool` overrides global `ThreadPool`

marxin:
Reorganize the public `libwild` API so downstream clients can control the number of threads used for linking.

Clients can either:
- use `libwild::run`, which derives the thread count from the arguments and jobserver; or
- construct a `Linker` with `Linker::new` and call `run`, allowing the thread count to be controlled externally.

Note I tested the perfetto traces are properly emitted and show reasonable output in the Web UI.

Resolves: #2204

### DIFF
--- a/libwild/src/lib.rs
+++ b/libwild/src/lib.rs
@@ -208,6 +208,12 @@ impl Linker {
     /// Runs the linker. The returned value isn't useful for anything, but is somewhat expensive to
     /// drop, so we leave it up to the caller to decide when to drop it. At the point at which we
     /// return, the output file should be usable.
+    ///
+    /// This method runs in whatever Rayon thread pool is currently installed. If no thread pool is
+    /// installed, then Rayon's default global thread pool will be created and used. In either case,
+    /// the value of `--threads` in `args` has no effect. If you want a thread pool configured from
+    /// the supplied arguments, use the top-level [`run`] function instead, which creates such a
+    /// pool, runs the linker in it, then tears it down together with other associated resources.
     pub fn run<'layout_inputs>(
         &'layout_inputs self,
         args: &'layout_inputs Args,
--- a/libwild/src/lib.rs
+++ b/libwild/src/lib.rs
@@ -125,14 +125,18 @@ use tracing_subscriber::fmt;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 
-/// Runs the linker and cleans up associated resources. Only use this function if you've OK with
+/// Runs the linker in a Rayon thread pool configured from the supplied arguments or the available
+/// jobserver tokens, then cleans up associated resources. Only use this function if you've OK with
 /// waiting for cleanup.
 pub fn run(mut args: Args) -> error::Result {
-    let thread_pool = args.common_mut().activate_thread_pool()?;
-    let linker = Linker::new();
-    linker.run(&args, &thread_pool)?;
-    drop(linker);
-    timing::finalise_perfetto_trace()?;
+    let thread_pool = args.common_mut().build_thread_pool()?;
+    thread_pool.pool.install(move || -> error::Result {
+        let linker = Linker::new();
+        linker.run(&args)?;
+        timing::finalise_perfetto_trace()?;
+        Ok(())
+    })?;
+
     Ok(())
 }
 
@@ -206,10 +210,6 @@ impl Linker {
     pub fn run<'layout_inputs>(
         &'layout_inputs self,
         args: &'layout_inputs Args,
-        // We don't actually use this, but take it as an argument to ensure that the caller has
-        // created it. We may decide to actually use it in future, if we stop using rayon's global
-        // thread pool.
-        _thread_pool: &crate::args::ThreadPool,
     ) -> error::Result<LinkerOutput<'layout_inputs>> {
         let identity = args.common().linker_identity();
         match args.common().version_mode {
@@ -451,8 +451,4 @@ pub fn init_timing() -> Result {
 
 pub fn should_fork(args: &Args) -> bool {
     args.common().should_fork()
-}
-
-pub fn activate_thread_pool(args: &mut Args) -> Result<crate::args::ThreadPool> {
-    args.common_mut().activate_thread_pool()
 }
--- a/libwild/src/lib.rs
+++ b/libwild/src/lib.rs
@@ -133,6 +133,7 @@ pub fn run(mut args: Args) -> error::Result {
     thread_pool.pool.install(move || -> error::Result {
         let linker = Linker::new();
         linker.run(&args)?;
+        drop(linker);
         timing::finalise_perfetto_trace()?;
         Ok(())
     })?;
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -344,12 +344,12 @@ impl CommonArgs {
         should_trace.then(|| tracing::trace_span!(crate::debug_trace::TRACE_SPAN_NAME).entered())
     }
 
-    /// Sets up the thread pool, using the explicit number of threads if specified,
+    /// Builds up the thread pool, using the explicit number of threads if specified,
     /// or falling back to the jobserver protocol if available.
     ///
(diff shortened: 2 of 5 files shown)


libwild/src/lib.rs
```diff
@@ -124,14 +124,18 @@ use tracing_subscriber::fmt;
...
 
-/// Runs the linker and cleans up associated resources. Only use this function if you've OK with
+/// Runs the linker in a Rayon thread pool configured from the supplied arguments or the available
+/// jobserver tokens, then cleans up associated resources. Only use this function if you've OK with
 /// waiting for cleanup.
 pub fn run(mut args: Args) -> error::Result {
-    let thread_pool = args.common_mut().activate_thread_pool()?;
-    let linker = Linker::new();
-    linker.run(&args, &thread_pool)?;
-    drop(linker);
```
marxin, line 134: Was the `drop` here used to get the information for the Perfetto output?

davidlattimore: Quite likely. I probably should have had a comment to that effect. When we drop the linker in relation to when we signal the parent process that linking is complete is also important from a performance perspective.

I just ran a quick benchmark. Linking with `--no-fork` appears unchanged, but linking with `--fork` (default) is showing a 19% slowdown on linking wild.

davidlattimore:
I feel like we need a grouping for this that isn't "features" - something for breaking API changes to libwild.

marxin:
> I feel like we need a grouping for this that isn't "features" - something for breaking API changes to libwild.

Looking at the spec: https://www.conventionalcommits.org/en/v1.0.0/, the exclamation mark (or `BREAKING CHANGE:`) would be an appropriate marker for such change. Which way do you like more?

libwild/src/lib.rs
marxin: Good, so restored the hunk.

> I just ran a quick benchmark. Linking with `--no-fork` appears unchanged, but linking with `--fork` (default) is showing a 19% slowdown on linking wild.

Good catch - I fixed the corresponding place where the parent process should have been informed earlier, but there's still ~5% diff right now - can you measure it as well?

davidlattimore:
The breaking-change classification is probably reasonable for something that's just a library. Given that what we distribute is primarily a tool rather than a library, I wonder if we should just have an API section in the release notes. We have some wild-only linker flags - removing them could be considered as breaking changes of the command-line interface, which is quite a different thing to breaking changes in the libwild API.

libwild/src/lib.rs
marxin: Note I see the different when linking the Mold linker binary - from my preliminary analysis based on the Perfetto UI traces, it seems the new thread pool usage makes it so that there's no longer a "main" thread does not participate in e.g. "Write group" and "Write prelude". As a consequence, the new partitioning of "Write group" is about +20ms longer. But still not the complete picture, now the main diff is the entire "Write output file" takes 680ms (compared to previous 625ms) making it the total **+5%** diff.

[fast.json.gz](https://github.com/user-attachments/files/30044318/fast.json.gz)
[slow.json.gz](https://github.com/user-attachments/files/30044319/slow.json.gz)

davidlattimore:
I've been doing some benchmarking and experiments to try to understand what's happening. The most pronounced effect seems to be with `--threads=2`. With 16 or 32 threads, I don't see any noticeable slowdown. At first, I thought that perhaps some work was happening on the main thread previously and now that work was in the worker pool, it was taking time away from other worker pool work. I'm less convinced now that this is the case. Running with taskset restricted to just 2 core, or even with all but two cores disabled (via chcpu), I still see the slowdown. So I think probably rayon is behaving differently depending on whether the job control is done from a thread in the worker pool or from an external thread.

marxin:
Well, I feel also pretty confused, where I collected numbers of 3 configurations (main, pool.install (this PR), pool.install + --no-fork) and there seems to be quite a consistent slowdown for the bigger number of the CPUs, but with some exceptions and frankly speaking, don't understand why it all starts with at least ~10 cores being used:
benchmark

libwild/src/lib.rs
```diff
@@ -124,14 +124,19 @@ use tracing_subscriber::fmt;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 
-/// Runs the linker and cleans up associated resources. Only use this function if you've OK with
+/// Runs the linker in a Rayon thread pool configured from the supplied arguments or the available
+/// jobserver tokens, then cleans up associated resources. Only use this function if you've OK with
 /// waiting for cleanup.
 pub fn run(mut args: Args) -> error::Result {
```
marxin, line 131: @davidlattimore  Right now, I’m leaning toward preserving the existing behavior by using the global thread pool, which would be used by the `wild` binary itself.
We could document that the function relies on the global pool and advise users who need explicit control to call `Linker::new().run(...)` from within their own `pool.install(...)`.
What do you think?

davidlattimore: I'd still like to eventually get to the bottom of what's going on, but for now, your suggestion of relying on the global pool sounds good to me.

mati865: Just make sure it solves the perf regression before investing too much time into that. We don't know yet the underlying reason.

marxin: There’s no rush on my side, so if @davidlattimore wants to experiment with it, I’m happy to wait before submitting the patch described above.

davidlattimore: From my experiments, it seems as though the performance difference is due to differences in how rayon handles task splitting when the tasks come from externally (e.g. the main thread) vs internally. This is likely due to [this line in rayon](https://github.com/rayon-rs/rayon/blob/1f9bb2538e50f1e6d1bc2e3d06a361ba2af0b632/src/iter/plumbing/mod.rs#L270) which changes splitting based on whether work was stolen or not.

The good news is that, at least based on my benchmarking, it looks like we can recover the performance loss with a single line:

```diff
diff --git a/libwild/src/elf_writer.rs b/libwild/src/elf_writer.rs
index 14b3bfeb26..3190ff0099 100644
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -274,6 +274,7 @@
     let groups_and_buffers = split_output_by_group(layout, &mut writable_buckets);
     groups_and_buffers
         .into_par_iter()
+        .with_max_len(1)
         .try_for_each(|(group, mut buffers)| -> Result {
             verbose_timing_phase!("Write group");
 
```

marxin: Nice observation! Lemme include the change in the PR and re-test it locally.

marxin: The number confirm your hypothesis - we're on par with the current `main` branch, cool!
benchmark

libwild/src/lib.rs
davidlattimore, line 208: So I assume the idea is that you either call `libwild::run` and it'll create a thread pool based on the arguments, run the linker then tear down the thread pool and other resources or you call this method and it'll use whatever thread pool you have already set up. It might be worthwhile extending the comment for this method to describe threadpool behaviour. In particular that it'll run in whatever rayon thread pool is currently installed or if none is installed that a default global rayon thread pool will be created and that in either case the value of `--threads` in `args` will have no effect.

davidlattimore (approved):

libwild/src/lib.rs
marxin: Exactly as you described! I feel it's a pretty nice design that given clients flexibility to either rely on the `--thread` family or arguments, or provide their own Rayon pool, and run the linker in it.

davidlattimore (approved):

marxin:
I wanted to thank you once again, @davidlattimore, for digging so deeply into this - my mental model of Rayon has definitely improved!

PR #2210, opened 2026-07-15
feat(API): Make zstd support optional at build time

marxin:
The change is motivated by the being able to cross compile the library for `+succinct` target: `riscv64im-succinct-zkvm-elf` ([^1]):
```
❯ cargo +succinct b --target=riscv64im-succinct-zkvm-elf --bin wild --no-default-features
info: `cargo` is unavailable for the active toolchain
info: falling back to "/home/marxin/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo"
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.11s
```

[^1]: https://docs.succinct.xyz/docs/sp1/introduction

### DIFF
--- a/libwild/src/fs.rs
+++ b/libwild/src/fs.rs
@@ -31,13 +31,6 @@ pub(crate) fn path_from_bytes(bytes: &[u8]) -> PathBuf {
         std::path::Path::new(OsStr::from_bytes(bytes)).to_path_buf()
     }
 
-    #[cfg(windows)]
-    {
-        use std::path::PathBuf;
-        let path = std::str::from_utf8(bytes).expect("Invalid UTF-8 in archive path name");
-        PathBuf::from(path)
-    }
-
     #[cfg(target_os = "wasi")]
     {
         use std::ffi::OsStr;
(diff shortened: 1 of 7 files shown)


libwild/src/fs.rs
```diff
@@ -44,4 +44,10 @@ pub(crate) fn path_from_bytes(bytes: &[u8]) -> PathBuf {
         use std::os::wasi::ffi::OsStrExt as _;
         std::path::Path::new(OsStr::from_bytes(bytes)).to_path_buf()
     }
+
+    #[cfg(not(any(unix, windows, target_os = "wasi")))]
+    {
+        let path = std::str::from_utf8(bytes).expect("Invalid UTF-8 in archive path name");
+        PathBuf::from(path)
+    }
 }
```
davidlattimore: Looks more or less identical to the windows branch. Do you think we should delete the windows branch and make this one apply to windows?

davidlattimore (approved):
I feel like "add zstd feature" in the "Features" section of the release notes might be misinterpreted. How about putting this in an API section and making the description something like "Make zstd support optional at build time"?

libwild/src/fs.rs
marxin: Yes, good idea!

marxin:
> How about putting this in an API section and making the description something like "Make zstd support optional at build time"?

Yes, good idea. Does it look good now?

davidlattimore (approved):

PR #2211, opened 2026-07-15
port(wasm): Support `--export` in Wasm

lapla-cogito:
This flag is passed by rustc when building a regular Rust program for wasm32-wasi.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -13,6 +13,7 @@ use crate::layout_rules::SectionRule;
 use crate::layout_rules::SectionRuleOutcome;
 use crate::output_section_id::SectionName;
 use crate::platform;
+use crate::platform::Args as _;
 use crate::symbol::UnversionedSymbolName;
 use crate::symbol_db::SymbolDb;
 use crate::timing_phase;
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -3214,6 +3226,68 @@ fn ensure_entry_export<'data>(
...
+    for name in symbol_db.args.force_export_symbol_names() {
+        let Some(symbol_id) =
+            symbol_db.get_unversioned(&UnversionedSymbolName::prehashed(name.as_bytes()))
+        else {
+            bail!("symbol exported via --export not found: {name}");
+        };
+        let def_id = symbol_db.definition(symbol_id);
+        let def_file_id = symbol_db.file_id_for_symbol(def_id);
+
+        let Some(def_obj_idx) = layout_inputs
```
davidlattimore, line 3247: I assume this linear search for a file will go away once the wasm code switches to using groups?

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: If we later keep remaps on per-object group state instead of parallel Vecs, this would go away.

PR #2212, opened 2026-07-16
chore: Don't try to run zstd test when zstd feature is disabled

davidlattimore:


marxin:
Thanks for the fix of the fallout caused by my change.

davidlattimore:
All good. My script that runs locally when I push changes, runs amongst other commands, `cargo clippy --no-default-features --all-targets -- -D warnings`. I've just sent #2217 to run something similar in our CI.

PR #2213, opened 2026-07-16
fix(zkvm): workaround weird behavior of `env::var`

marxin:
The SP1 zkVM currently returns `Ok("")` for all `env::var` calls. While this is somewhat understandable - since execution is intended to depend only on stdin - it would be more appropriate to return `Err(VarError::NotPresent)`. In any case, I’d like to propose a workaround.


libwild/src/env.rs
davidlattimore, line 1: We could consider calling the module `env`. Then we could call the function `var` - allowing us to call `crate::env::var` or just `env::var` (if we import `crate::env`). But I don't feel strongly, so feel free to disregard.

libwild/src/misc.rs
```diff
@@ -0,0 +1,14 @@
+use std::env::VarError;
+
+/// Reads an environment variable.
+///
+/// zkVM currently returns an empty string instead of an error for missing environment variables.
+/// See <https://github.com/succinctlabs/sp1/issues/2883>.
+pub(crate) fn get_env(key: &str) -> Result<String, VarError> {
+    let value = std::env::var(key)?;
+    if cfg!(target_os = "zkvm") && value.is_empty() {
```
davidlattimore: We could consider treating an empty value as not set for all platforms. I don't think we have any place where we need to distinguish between empty and unset and were we to introduce such a distinction, then we probably should think twice about doing so. But it's also fine to keep this platform-specific if you prefer.

davidlattimore (approved):

libwild/src/env.rs
marxin: I really like the idea!

Issue #2184, opened 2026-07-06
`rust-integration/linker-plugin-lto` test failure on Arch Linux


mati865:
Using https://github.com/wild-linker/wild/pull/2183 as base.

Repro steps:
1. `podman run -it --rm wild-dev-arch`
2. `rustup default nightly` (so, rustc bitcode matches system Clang)
3. `rustup toolchain uninstall 1.94.0`
4. `cargo t rust-integration/linker-plugin-lto`

It might take a few tries but the tests will fail in a different ways:
```
---- elf/x86_64/rust-integration/linker-plugin-lto ----
...
Validation failed.
wild: /wild/wild/tests/build/elf/x86_64/rust-integration/linker-plugin-lto/rust-integration.rs.wild
ld: /wild/wild/tests/build/elf/x86_64/rust-integration/linker-plugin-lto/rust-integration.rs.ld
eh_frame
  wild .eh_frame_hdr contains invalid info pointer 0x1428c
  ld OK
```

```
test elf/x86_64/rust-integration/linker-plugin-lto      ... ignored (Can't use rust+clang with linker plugin. LLVM version mismatch? Check that clang --version matches LLVM version reported by `rustc -vV`.
```

```
---- elf/riscv64/rust-integration/linker-plugin-lto ----
Test failed: `rust-integration.rs` with linker `wild` config `linker-plugin-lto`
  Caused by:
    Unexpected output on stderr:
Hard-float 'd' ABI can't be used for a target that doesn't support the D instruction set extension (ignoring target-abi)
```

This is despite rustc version matching Clang as much reasonable possible:
```
[root@4a28c0f5a3bc wild]# rustc -vV
rustc 1.99.0-nightly (3659db0d3 2026-07-05)
binary: rustc
commit-hash: 3659db0d3e2cd634c766fcda79ed118eca31a9fd
commit-date: 2026-07-05
host: x86_64-unknown-linux-gnu
release: 1.99.0-nightly
LLVM version: 22.1.8
[root@4a28c0f5a3bc wild]# clang --version
clang version 22.1.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

davidlattimore:
I think this is probably fixed now - at least elf/x86_64/rust-integration/linker-plugin-lto now passes in the docker image when I follow the repro steps.

mati865:
EDIT: Most of this comment is no longer relevant.

[...]

PR #2215, opened 2026-07-16
test: Fix race condition in Rust LTO support detection

mati865:
Different architectures might be tested in parallel, and we cannot use /dev/null in this specific case. That means we need a unique name for the each arch, otherwise one test might replace the outputs before the previous one has done reading them.

Observed in https://github.com/wild-linker/wild/issues/2184


marxin:
Very good job @mati865!

davidlattimore (approved):
Good find!

PR #2219, opened 2026-07-16
cli: drop unused flag --debug-address

marxin:
If I'm not mistaken, the option has been there basically since the beginning, but it hasn't been used.

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -53,7 +53,6 @@ pub struct ElfArgs {
     pub(crate) strip: Strip,
     pub(crate) merge_sections: bool,
     pub(crate) version_script_path: Option<PathBuf>,
-    pub(crate) debug_address: Option<u64>,
     pub(crate) should_write_eh_frame_hdr: bool,
     pub(crate) wrap: Vec<String>,
     pub(crate) rpath: Option<String>,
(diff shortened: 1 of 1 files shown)


mati865:
Do you have an opinion whether we should make args removal more visible? If we merge it as `chore`, it won't be even mentioned in the changelog. Chances for somebody to rely on it should be minimal, but still it's kinda breaking.

FWIW, it was introduced in https://github.com/wild-linker/wild/commit/16a1da957b89af9f3a899875291f0e1b606938e9

davidlattimore (approved):
I think I used it once, but only in throw-away code. But I think that was a year or two ago. I'm in favour of removing it.

marxin:
> Do you have an opinion whether we should make args removal more visible?

Looking here: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#type-enum, maybe `revert` would be more appropriate type?

mati865:
Perhaps we should introduce `cli` category? I don't think any of the predefined categories work well in this case.

mati865:
We probably should create an issue to add `cli` keyword to the changelog generator and make it a blocker for a release.

marxin:
> We probably should create an issue to add `cli` keyword to the changelog generator and make it a blocker for a release.

Yep, done as #2237.

PR #2220, opened 2026-07-16
CI: add cargo-deny job

marxin:


marxin:
@davidlattimore @mati865, the `cargo-deny` job will help identify duplicate crate versions.

mati865:
I've been using cargo-deny for license and vulnerability checking in a proprietary project. That works great.
I'm not entirely sold on failing CI when duplicated deps are introduced. Sometimes you have to go an extra mile to avoid that, and even if you decide to fix it upstream, you still have to wait for a release.

marxin:
> I'm not entirely sold on failing CI when duplicated deps are introduced. Sometimes you have to go an extra mile to avoid that, and even if you decide to fix it upstream, you still have to wait for a release.

Sure, I see it the same - we should not fail, but the cargo deny provides a warnings we can address if we want.

mati865:
Yeah, that's how I had configured it.

Regarding paste/pastey, I have checked pastey because of that advisory a few months back but didn't see any relevant fixes and there were some LLM commits. So, I decided to just ignore it still remembering https://www.reddit.com/r/rust/comments/1ibdxf9/beware_of_this_guy_making_slop_crates_with_ai/
Not saying that is a bad crate, but I haven't reviewed it.

marxin:
> Not saying that is a bad crate, but I haven't reviewed it.

Yep, no need for rushing and leaving `paste` crate right now given the surface is really small. So adjusting based on your advice.

davidlattimore (approved):
Thanks :)

PR #2222, opened 2026-07-16
chore: refactor RELR bitmap encoder into one

marxin:
The change unifies the encoding logic for RELR bitmap relocations into once place, where both the allocation and encoding can pretty neatly communicate via the provided `encode_fn` argument.

CC: @deepakshirkem


davidlattimore (approved):
Thanks :)

PR #2223, opened 2026-07-16
port(MachO): Write LC_UUID hash

### LINKED ISSUE #2120: Mach-O: fill up `LC_UUID` command
Similarly to ELF's `.build-id`, we should fill up hash based on the content of the emitted file.

joanna-jo:
Generates a BLAKE3 hash, then takes the first 16 bytes to fill out LC_UUID’s `uuid` field, along with setting the UUID version and variant bits. Two things I wanted to note were:

* The UUID hash is calculated by the contents (including the code signature), and the code signature page hashes depend on the UUID since one of the hashed pages will include `LC_UUID`. We can circumvent this by zeroing out the UUID and code signature when calculating the UUID hash, but then otherwise identical outputs linked under different filenames would produce the same UUID.
    * Looking at lld’s implementation, code signature metadata is written first ([phases](https://github.com/llvm/llvm-project/blob/7812fef524a0aef0b34599c2eb81b66843fbcdd8/lld/MachO/Writer.cpp#L1305-L1316), [code signature metadata](https://github.com/llvm/llvm-project/blob/7812fef524a0aef0b34599c2eb81b66843fbcdd8/lld/MachO/SyntheticSections.cpp#L1624-L1664), [code signature hashes](https://github.com/llvm/llvm-project/blob/7812fef524a0aef0b34599c2eb81b66843fbcdd8/lld/MachO/Writer.cpp#L1298-L1303)) so that this metadata (and specifically the identifier) differentiates identical files. I matched this implementation, but it required refactoring and splitting `write_code_signature` into separate metadata and hashing phases so that the ordering is now: `write_code_signature_metadata` -> `write_uuid` -> `write_code_signature_hashes`.

* [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) describes the specification for UUIDs. Likewise, I matched [what lld does](https://github.com/llvm/llvm-project/blob/95ff1fa00ed4c35b3589fa92b591e02444fcc5d5/lld/MachO/Writer.cpp#L554-L565) and used UUID Version 3 to specify the version and variant fields in our hash. Although the RFC states that these “values are created by computing an MD5 hash,” it seems like lld acknowledges this as the “least bad lie” and uses this version format for their XXH3 UUID hashes.

Testing, before vs. after:

**Before:**
```
running 1 test
test macho/aarch64/trivial/default ... FAILED

failures:

---- macho/aarch64/trivial/default ----
Output binary assertions failed. Binary `/Users/joannajo/wild/wild/tests/build/macho/aarch64/trivial/default/trivial.c.wild`. Relink with:
        cargo run --bin wild -- -flavor darwin -o /Users/joannajo/wild/wild/tests/build/macho/aarch64/trivial/default/trivial.c.wild /Users/joannajo/wild/wild/tests/build/macho/aarch64/trivial/default/trivial.c.o /Users/joannajo/wild/wild/tests/build/macho/aarch64/trivial/default/runtime.c.o --validate-output --write-layout --write-trace
  Caused by:
    LC_UUID does not match expected hash



failures:
    macho/aarch64/trivial/default

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.72s
```

**After:**
```
running 1 test
test macho/aarch64/trivial/default ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.55s
```

Resolves #2120

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2153,6 +2153,7 @@ version = "0.9.0"
 dependencies = [
  "ar",
  "bitflags",
+ "blake3",
  "dhat",
  "diff",
  "foldhash 0.2.0",
--- a/libwild/src/macho_writer.rs
+++ b/libwild/src/macho_writer.rs
@@ -24,6 +24,7 @@ use crate::macho::CS_BLOB_HEADERS_SIZE;
 use crate::macho::CS_BLOCK_SIZE;
 use crate::macho::CS_BLOCK_SIZE_EXP;
 use crate::macho::CS_HASH_SIZE;
+use crate::macho::CS_HEADERS_SIZE;
 use crate::macho::ChainedFixupsHeader;
 use crate::macho::CodeSignatureCommand;
 use crate::macho::DYLINKER_PATH;
@@ -100,6 +101,7 @@ use object::macho::LC_MAIN;
 use object::macho::LC_SEGMENT_64;
 use object::macho::LC_SYMTAB;
 use object::macho::LC_UUID;
+use object::macho::LoadCommand;
 use object::macho::MH_CIGAM_64;
 use object::macho::MH_EXECUTE;
 use object::macho::N_ABS;
@@ -163,7 +165,9 @@ pub(crate) fn write<'data, A: Arch<Platform = MachO>>(
     write_got_entries(layout, section_buffers.get_mut(output_section_id::GOT))?;
     write_plt_entries::<A>(layout, section_buffers.get_mut(output_section_id::PLT_GOT))?;
 
-    write_code_signature(layout, sized_output)?;
+    write_code_signature_metadata(layout, sized_output)?;
+    write_uuid(layout, sized_output)?;
+    write_code_signature_hashes(layout, sized_output)?;
 
     Ok(())
 }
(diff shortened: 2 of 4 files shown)


joanna-jo:
I'm not sure if these CI failures are because of my changes - could I get a maintainer's help in rerunning these checks?

davidlattimore:
From the logs, it looks unlikely that they are due to your change. But after rerunning, they're still failing, so we'll probably need to wait a bit for whatever infrastructure problem is happening at github to resolve itself.

lapla-cogito:
FYI this is probably the cause. https://www.githubstatus.com/incidents/gxycch3076xk
This also occurs in current main, so once resolved, I'll also re-run the Actions in this PR.

joanna-jo:
Sounds good, thank you! Looks like the CI is green now :)

davidlattimore (approved):

The testing for the hash just repeats the hash computation that's happening in the linker. This is a reasonable decision given the limited alternatives. I guess one alternative would be to not check the actual UUID, but just have the test framework verify that the UUID was different to the UUID of all the other test binaries produced. We could consider that in future if we change away from the current hashing algorithm for some reason, but there's probably no need to do it now. Another possibility would be to actually share code between the linker and the test framework. I don't have any real thoughts on whether this is a good thing to do in this instance or not.

Anyway, I'm happy for this to be merged as-is, but will give @marxin a chance to have a look first.

marxin (approved):
LGTM, thank you for the contribution.

PR #2224, opened 2026-07-17
port(wasm): Support `-z stack-size` for Wasm

lapla-cogito:
Support `-z stack-size` for Wasm. Unspecified stack size defaults to 64 KiB (wasm-ld compatible).

Part of #1431

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -28,13 +28,15 @@ pub(crate) const WASM_PAGE_ALIGNMENT: Alignment = Alignment { exponent: 16 };
 /// Default page size (in bytes) for a wasm linear memory page.
 pub(crate) const WASM_PAGE_SIZE: u64 = WASM_PAGE_ALIGNMENT.value();
 
+/// Default main stack size.
+pub(crate) const DEFAULT_STACK_SIZE: u32 = 64 * 1024;
+
 #[derive(Debug)]
 pub struct WasmArgs {
     pub(crate) common: super::CommonArgs,
     pub(crate) lib_search_path: Vec<Box<Path>>,
     pub(crate) export_symbols: Vec<String>,
-    /// If not set, the stack size will be set to 64KiB.
-    pub(crate) z_stack_size: Option<u64>,
+    pub(crate) z_stack_size: u32,
 }
 
 impl WasmArgs {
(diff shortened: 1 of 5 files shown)


libwild/src/wasm.rs
```diff
@@ -119,6 +119,15 @@ const LINKER_MEMORY_BASE: u32 = 1024;
 /// Default stack size reserved above `__data_end` for the main stack.
 const LINKER_STACK_SIZE: u32 = 64 * 1024;
```
davidlattimore: Do you think that this should be renamed to DEFAULT_STACK_SIZE given its new role?

libwild/src/args/wasm.rs
```diff
@@ -48,6 +52,7 @@ impl Default for WasmArgs {
             common: CommonArgs::default(),
             lib_search_path: Vec::new(),
             export_symbols: Vec::new(),
+            z_stack_size: None,
```
davidlattimore: Does this need to be an Option or could we just set the default value here?

libwild/src/args/wasm.rs
```diff
@@ -196,6 +202,24 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
...
+            "stack-size=",
+            "Set the main stack size in linear memory",
+            |args, _, value| {
+                let size = parse_number(value)?;
+                args.z_stack_size = Some(size);
+                Ok(())
+            },
+        )
+        .execute(|args, _modifier_stack, value| {
+            args.warn_unsupported(&(format!("-z {value}")))?;
```
lapla-cogito, line 221: Similar to ELF, when encountering an unknown `-z` option, we only display a warning.

libwild/src/args/wasm.rs
```diff
@@ -196,6 +201,21 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
...
+    parser
+        .declare_with_param()
+        .prefix("z")
+        .help("Linker options")
+        .sub_option_with_value(
+            "stack-size=",
+            "Set the main stack size in linear memory",
+            |args, _, value| {
+                let size = parse_number(value)?;
+                args.z_stack_size = Some(size);
```
davidlattimore: Any reason to store this as a u64 then later validate that it fits in a u32? Looks like we could validate here that it fits in a u32 and then store it as such

libwild/src/wasm.rs
lapla-cogito: Right, renamed.

davidlattimore (approved):

PR #2226, opened 2026-07-17
feat: support specifying an output section's fill pattern in a linker script

vishruth-thimmaiah:
This PR adds support for specifying the fill pattern for a given output section in a linker script, using `=fillexpr`.
Example:
```ld
.text {
    *(.text)
} =0x90909090
```


wild/tests/sources/elf/linker-script-fill/linker-script-fill.c
```diff
@@ -0,0 +1,37 @@
...
+
+//#Config:fillexpr:default
+//#LinkerScript:linker-script-fill.ld
+//#ExpectSectionBytes:.fill1=0x110000009000000022 0..9
+//#ExpectSectionBytes:.fill2=0x110000009000000022 0..9
+//#ExpectSectionBytes:.fill3=0x119090909090909022 0..9
+//#ExpectSectionBytes:.fill4=0x110000009000000022 0..9
+//#ExpectSectionBytes:.fill5=0x110000000900000022 0..9
+
+//#Config:sdf:default
```
davidlattimore: Something like "overflow-32" might be a more descriptive name

wild/tests/sources/elf/linker-script-fill/linker-script-fill.ld
```diff
@@ -0,0 +1,26 @@
+SECTIONS {
+    .fill1 : {
+        *(.fill1.first)
+        *(.fill1.second)
+    } =144
+
+    .fill2 : {
+        *(.fill2.first)
+        *(.fill2.second)
```
davidlattimore, line 9: If we have gaps between the merged sections - e.g. `. = . + 8`, should they be filled with the fill value?

wild/tests/sources/elf/linker-script-fill/linker-script-fill.ld
```diff
@@ -0,0 +1,26 @@
...
+
+    .fill4 : {
+        *(.fill4.first)
+        *(.fill4.second)
+    } = 0x0090
+
+    .fill5 : {
+        *(.fill5.first)
+        *(.fill5.second)
+    } = 0x9
```
davidlattimore, line 25: What about negative fill values? Is a value like -1 accepted by other linkers?

vishruth-thimmaiah: lld just bitcasts it into a u64, so it emits the same error as an overflow.

wild/tests/sources/elf/linker-script-fill/linker-script-fill.ld
vishruth-thimmaiah: This should be filled now. We also weren't filling in sections marked with `EXECINSTR` on x86 with nop bytes either, so this PR fixes that issue as well.

wild/tests/sources/elf/linker-script-fill/linker-script-fill.ld
```diff
@@ -0,0 +1,33 @@
...
+
+    .fill6 : {
+        . = . + 0x8;
+        *(.fill6.first)
+        . = . + 0x8;
+        *(.fill6.second)
```
davidlattimore, line 31: What about adding `. = . + 0x8;` after the last content rule and confirming that that gets filled correctly?

davidlattimore (approved):

PR #2230, opened 2026-07-17
feat: implement --sort-section flag

### LINKED ISSUE #1659: fails to link `musl` libc (`wild: error: unrecognized option(s): --sort-section`)
Steps to reproduce:

- `wget https://musl.libc.org/releases/musl-1.2.5.tar.gz`
- `tar -xf musl-1.2.5.tar.gz`
-  `cd musl-1.2.5`
- `export CFLAGS="-O2" LDFLAGS="-fuse-ld=wild"`
- `./configure && make`

```bash
gcc -std=c99 -nostdinc -ffreestanding -fexcess-precision=standard -frounding-math -fno-strict-aliasing -Wa,--noexecstack \
-D_XOPEN_SOURCE=700 -I./arch/x86_64 -I./arch/generic -Iobj/src/internal -I./src/include -I./src/internal -Iobj/include -I./include  \
-pipe -fomit-frame-pointer -fno-unwind-tables -fno-asynchronous-unwind-tables -ffunction-sections -fdata-sections \
-Wno-pointer-to-int-cast -Werror=implicit-function-declaration -Werror=implicit-int -Werror=pointer-sign -Werror=pointer-arith \
-Werror=int-conversion -Werror=incompatible-pointer-types -Werror=discarded-qualifiers -Werror=discarded-array-qualifiers -Waddress \
-Warray-bounds -Wchar-subscripts -Wduplicate-decl-specifier -Winit-self -Wreturn-type -Wsequence-point -Wstrict-aliasing -Wunused-function -Wunused-label -Wunused-variable \
-O2 -Wl,--sort-section,alignment -Wl,--sort-common -Wl,--gc-sections -Wl,--hash-style=both -Wl,--no-undefined -Wl,--exclude-libs=ALL -Wl,--dynamic-list=./dynamic.list -fuse-ld=wild -nostdlib -shared \
[ ... ]
wild: error: unrecognized option(s): --sort-section
make: *** [Makefile:162: lib/libc.so] Error 1
```

deepakshirkem:
Previously, Wild would warn that --sort-section was unsupported. Now: `--sort-section=name` sorts input sections alphabetically within each output section, matching GNU ld behavior. `--sort-section=alignment `is accepted silently, since Wild already places sections in alignment order by default.

Fixes #1659

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -4583,11 +4583,12 @@ fn get_symbol_attributes(
                 .and_then(|section_index| {
                     let slot = &obj.sections[section_index.0];
                     match slot {
-                        SectionSlot::Loaded(_) | SectionSlot::MergeStrings(_) => {
+                        SectionSlot::Loaded(_)
+                        | SectionSlot::MergeStrings(_)
+                        | SectionSlot::Sorted(_) => {
                             let output_section_id = obj
                                 .section_part_id(section_index, &layout.symbol_db.section_part_ids)
                                 .output_section_id();
-
                             layout
                                 .output_sections
                                 .output_index_of_section(output_section_id)
@@ -4819,8 +4820,11 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
             SectionSlot::Loaded(_) | SectionSlot::MergeStrings(_) => object
                 .section_part_id(section_index, &layout.symbol_db.section_part_ids)
                 .output_section_id(),
+            SectionSlot::Sorted(_) => object
+                .section_part_id(section_index, &layout.symbol_db.section_part_ids)
+                .output_section_id(),
             _ => bail!(
-                "Internal error: Defined symbols should always be for a loaded or merge-strings section"
+                "Internal error: Defined symbols should always be for a loaded, merge-strings or sorted section"
             ),
         };
         let output_section_id = layout
--- a/libwild/src/resolution.rs
+++ b/libwild/src/resolution.rs
@@ -1347,7 +1347,7 @@ fn resolve_section<'data, P: Platform>(
             must_load |= output_info.must_keep;
 
             unloaded_section = UnloadedSection::new();
-            unloaded_section.needs_sorting = output_info.sorted;
+            unloaded_section.needs_sorting = output_info.sorted || args.sort_sections_by_name();
         }
         SectionRuleOutcome::SortedSection(output_info) => {
             part_id = if output_info.section_id.is_regular() {
--- a/wild/tests/sources/elf/sort-section-alignments/sort-section-alignments.s
+++ b/wild/tests/sources/elf/sort-section-alignments/sort-section-alignments.s
@@ -1,0 +1,22 @@
+// Verify that --sort-section=alignment sorts sections by alignment (largest first).
+// Input order: .data.aaa (align 8) first, .data.zzz (align 64) second
+// With --sort-section=alignment: .data.zzz (align 64) should come first
+//#AbstractConfig:default
+//#LinkArgs:-shared --no-gc-sections --sort-section=alignment
+//#RunEnabled:false
+//#ReferenceLinkers:bfd
+//#DiffIgnore:.dynamic.DT_FLAGS_1.NOW
+//#DiffIgnore:segment.LOAD.RX.alignment
+//#Config:x86_64:default
+//#Arch:x86_64
+//#ExpectSectionBytes:.data=0x3300000000000000 0..8
+
+    .section .data.aaa,"aw",@progbits
+    .balign 8
+    .byte 0x11
+    .zero 7
+
+    .section .data.zzz,"aw",@progbits
+    .balign 64
+    .byte 0x33
+    .zero 63
--- a/wild/tests/sources/elf/sort-section-name/sort-section-name.s
+++ b/wild/tests/sources/elf/sort-section-name/sort-section-name.s
@@ -1,0 +1,24 @@
+// Verify that --sort-section=name sorts input sections alphabetically.
+// Also verifies that exported symbols in sorted sections work correctly.
+// Without sorting: .data.zzz (0x33) comes first (input order)
+// With --sort-section=name: .data.aaa (0x11) comes first (alphabetical)
+//#AbstractConfig:default
+//#LinkArgs:-shared --no-gc-sections --sort-section=name
+//#RunEnabled:false
+//#ReferenceLinkers:bfd
+//#DiffIgnore:.dynamic.DT_FLAGS_1.NOW
+//#DiffIgnore:segment.LOAD.RX.alignment
+//#Config:x86_64:default
+//#Arch:x86_64
+//#ExpectSectionBytes:.data=0x11000000000000003300000000000000 0..16
+    .section .data.zzz,"aw",@progbits
+    .balign 8
+    .byte 0x33
+    .zero 7
+
+    .globl exported
+    .section .data.aaa,"aw",@progbits
+    .balign 8
+exported:
+    .byte 0x11
+    .zero 7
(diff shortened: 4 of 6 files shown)


libwild/src/layout.rs
```diff
@@ -6046,6 +6047,7 @@ fn harvest_and_sort_script_sections<'data, P: Platform>(
     group_states: &mut [GroupState<'data, P>],
     output_sections: &OutputSections<P>,
     section_part_ids: &[PartId],
+    _sort_by_name: bool,
```
lapla-cogito: Is this necessary?

libwild/src/resolution.rs
```diff
@@ -1347,7 +1347,10 @@ fn resolve_section<'data, P: Platform>(
             must_load |= output_info.must_keep;
 
             unloaded_section = UnloadedSection::new();
-            unloaded_section.needs_sorting = output_info.sorted;
+            unloaded_section.needs_sorting = output_info.sorted
+                || (args.sort_sections_by_name()
+                    && !input_section.is_executable()
+                    && input_section.is_alloc());
```
lapla-cogito: I'm not entirely clear on why these two conditions are necessary 🤔. Both GNU ld and lld seem to include executable sections in their sorting (though lld only sorts when a linker script is present) and don't exclude non-alloc sections.

wild/tests/sources/elf/sort-section-name/sort-section-name.s
```diff
@@ -0,0 +1,22 @@
+// Verify that --sort-section=name sorts input sections alphabetically.
+// Two sections with same alignment but different names: .data.zzz and .data.aaa
+// Without sorting: .data.zzz (0x33) comes first (input order)
+// With --sort-section=name: .data.aaa (0x11) comes first (alphabetical)
+//#AbstractConfig:default
+//#LinkArgs:-shared --no-gc-sections --sort-section=name
```
lapla-cogito, line 6: Couldn't we also add tests for `--sort-section=alignment`?

wild/tests/sources/elf/sort-section/sort-section.s
```diff
@@ -0,0 +1,22 @@
...
+// With --sort-section=name: .data.aaa (0x11) comes first (alphabetical)
+//#AbstractConfig:default
+//#LinkArgs:-shared --no-gc-sections --sort-section=name
+//#RunEnabled:false
+//#ReferenceLinkers:bfd
+//#DiffIgnore:.dynamic.DT_FLAGS_1.NOW
+//#DiffIgnore:segment.LOAD.RX.alignment
+//#Config:x86_64:default
+//#Arch:x86_64
+//#ExpectSectionBytes:.data=0x11 0..1
```
davidlattimore: Should we assert more than just the first byte? It'd also be good to have some different alignments in the sorted section - to make sure that we're not just sorting within an alignment bucket

wild/tests/sources/elf/sort-section-name/sort-section-name.s
deepakshirkem: Done :)

wild/tests/sources/elf/sort-section/sort-section.s
deepakshirkem: For cross-alignment sorting sections with different alignments get different part_ids in Wild and end up in separate memory regions. Would you like me to implement this, or is sorting  within the same alignment bucket sufficient for the --sort-section use cases you have in mind?

davidlattimore: I'm not sure what use-cases people have for sorting sections by name. I guess one possible use-case is to try to give stuff in the file a more deterministic order to reduce how different files are between builds. I'm not sure really how well it works for this use-case though, since even without sorting, the sections will be in a deterministic order so long as the compiler outputs them in a consistent order - although perhaps it doesn't. Anyway, if that is a use-case, then it might not matter if we sort sections with different alignments separately. I'd say don't try to sort between alignments for now. It's likely to be non-trivial. It'd be good to better understand why people use the flag though.

davidlattimore (approved):

wild/tests/sources/elf/sort-section/sort-section.s
davidlattimore: Thinking some more about sorting. The main legitimate use-case for sorting would be when some things need to be registered in some custom section. In that case, I'd kind of expect that the items in the section would have not only the same alignment, but also the same size. I'm less sure about when the flag --sort-section would be used, since that seems to apply to all output sections, not just a particular special-purpose section.

wild/tests/sources/elf/sort-section-alignments/sort-section-alignments.s
lapla-cogito, line 1: How about testing alignments that are different?

lapla-cogito (commented):
i think we need to add a `SectionSlot::Sorted` match arm to matches like this:

https://github.com/wild-linker/wild/blob/c9209c2eee2fe346e9a82af484aa170572c75334/libwild/src/elf_writer.rs#L4826-L4833

I was able to verify that linking fails when exported symbols in shared libraries are subject to name sorting:

```
（▰╹◡╹）❯  cat repro.c

int exported = 42;
int foo(void) { return exported; }

（▰╹◡╹）❯  gcc -c -fPIC -o repro.o repro.c

（▰╹◡╹）❯  ~/repos/wild/target/debug/wild -shared --sort-section=name -o repro.so repro.o
wild: error: Failed copying from <epilogue> to output file
  Caused by:
    Internal error: Defined symbols should always be for a loaded or merge-strings section

（▰╹^╹）❯  ld -shared --sort-section=name -o repro.so repro.o
# success
```

wild/tests/sources/elf/sort-section/sort-section.s
deepakshirkem: Thank-You for the context. I'll leave cross-alignment sorting for now.

wild/tests/sources/elf/sort-section-alignments/sort-section-alignments.s
deepakshirkem: @lapla-cogito Thank-you. Addressed both review comments. Please take another look?

libwild/src/elf_writer.rs
```diff
@@ -4827,6 +4820,9 @@ fn write_regular_object_dynamic_symbol_definition<'data>(
             SectionSlot::Loaded(_) | SectionSlot::MergeStrings(_) => object
                 .section_part_id(section_index, &layout.symbol_db.section_part_ids)
                 .output_section_id(),
+            SectionSlot::Sorted(_) => object
+                .section_part_id(section_index, &layout.symbol_db.section_part_ids)
+                .output_section_id(),
             _ => bail!(
                 "Internal error: Defined symbols should always be for a loaded or merge-strings section"
```
```suggestion
                "Internal error: Defined symbols should always be for a loaded, merge-strings or sorted section"
```

wild/tests/sources/elf/sort-section-name/sort-section-name.s
```diff
@@ -0,0 +1,26 @@
...
+//#Arch:x86_64
+//#ExpectSectionBytes:.data=0x11000000000000003300000000000000 0..16
+
+    .globl exported
+    .section .data.zzz,"aw",@progbits
+    .balign 8
+    .byte 0x33
+    .zero 7
+
+    .globl exported2
```
lapla-cogito: It seems `exported2` is undefined 👀 Is there a specific reason for that?

lapla-cogito (commented):
Thanks! Overall looks good, just two nits. I also noticed that both the commit message and the PR description are outdated. The part "allocatable non-executable" seems incorrect at this point.

wild/tests/sources/elf/sort-section-name/sort-section-name.s
deepakshirkem: @lapla-cogito Thank-You.  Please take another look?

lapla-cogito (approved):
thanks again. I slightly updated the PR description (removed the "allocatable non-executable" part).

deepakshirkem:
> thanks again. I slightly updated the PR description (removed the "allocatable non-executable" part).

Thank you! I forgot to update that.

PR #2233, opened 2026-07-18
port(wasm): Emit the custom `name` section for function and global names

lapla-cogito:
Synthesize Wasm name-section function/global maps from imports, linker-defined symbols, linking-section symbols, and exports so tools like wasm-objdump and wasmtime can show names instead of bare indices.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -30,6 +30,8 @@ use linker_utils::utils::u32_from_slice;
 use rayon::prelude::*;
 use std::borrow::Cow;
 use std::ops::Range;
+use wasm_encoder::NameMap;
+use wasm_encoder::NameSection;
 use wasmparser::BinaryReader;
 use wasmparser::CodeSectionReader;
 use wasmparser::ConstExpr;
@@ -1372,7 +1374,8 @@ impl platform::ProgramSegmentDef for ProgramSegmentDef {
             | osid::WASM_ELEMENT
             | osid::WASM_DATA_COUNT
             | osid::WASM_CODE
-            | osid::WASM_DATA => SegmentType::Module,
+            | osid::WASM_DATA
+            | osid::WASM_NAME => SegmentType::Module,
             _ => SegmentType::Unused,
         };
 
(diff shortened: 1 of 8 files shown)


wild/tests/sources/wasm/export/export.c
```diff
@@ -1,7 +1,7 @@
 //#Config:export-functions
 //#LinkArgs: --export=foo
 //#Contains: foo
-//#DoesNotContain: not_exported
+//TODO(wasm): verify that `not_exported` is actually not exported
```
lapla-cogito: This PR causes `not_exported` to appear in the name section, making it impossible to verify using the `DoesNotContain` directive. A new directive for checking export status is required.

libwild/src/wasm.rs
```diff
@@ -1615,8 +1625,129 @@ fn encode_wasm_section(section: &impl wasm_encoder::Section) -> Vec<u8> {
...
+            crate::wasm_writer::OutputImportEntity::Global(_) => {
+                global_names.push((next_global_import, import.name));
+                next_global_import += 1;
+            }
+        }
+    }
+
+    // Linker-synthesised functions / globals.
+    if let Some(idx) = indices.memory_base_global {
+        global_names.push((idx, "__memory_base"));
```
davidlattimore, line 1654: The string "__memory_base" already appears twice in the linker. This brings it to 3. I gather you're intending to refactor soon and just wanted to confirm that all 3 will get dealt with in that refactoring?

davidlattimore (commented):
Would it make sense to have tests that check this?

libwild/src/wasm.rs
lapla-cogito: Yes. At least we can avoid placing literals directly in these places. While I can't predict exactly how much cleaner the code will become during the refactoring, we of course can expect some improvement.

lapla-cogito:
> Would it make sense to have tests that check this?

Added support for checking non-standard sections with `ExpectSection`/`NoSection`.

davidlattimore (approved):

PR #2238, opened 2026-07-19
port(wasm): Define Wasm linker symbols through SymbolDb

lapla-cogito:
Register Wasm linker-synthesized symbols in SymbolDb and resolve reserved imports from those definitions.

Part of #1431

### DIFF
--- a/libwild/src/parsing.rs
+++ b/libwild/src/parsing.rs
@@ -69,13 +69,13 @@ pub(crate) struct SyntheticSymbols {
 #[derive(Clone, derive_more::Debug)]
 pub(crate) struct InternalSymDefInfo<'data, P: Platform> {
     pub(crate) symbol: P::SymtabEntry,
-    pub(crate) placement: SymbolPlacement<'data>,
+    pub(crate) placement: SymbolPlacement<'data, P>,
     #[debug("{:?}", String::from_utf8_lossy(name))]
     pub(crate) name: &'data [u8],
 }
 
 #[derive(Clone, PartialEq, Eq, Debug)]
-pub(crate) enum SymbolPlacement<'data> {
+pub(crate) enum SymbolPlacement<'data, P: Platform> {
     /// Symbol 0 - the undefined symbol.
     Undefined,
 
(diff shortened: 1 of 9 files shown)


libwild/src/wasm.rs
```diff
@@ -3872,24 +3893,98 @@ struct LinkerDefinedDataAddress {
...
+    MemoryBase,
+    StackPointer,
+    TlsBase,
+    // Functions
+    CallCtors,
+    CallDtors,
+}
+
+impl WasmLinkerSymbol {
+    fn name(self) -> &'static [u8] {
```
marxin, line 3921: I welcome using the `strum` crate - what about using the `EnumString` for implementing `from_str` and similarly for the serialization: https://docs.rs/strum/latest/strum/derive.EnumString.html?
What do you think?

lapla-cogito: Thanks, that would be cleaner indeed.

marxin: Glad to hear that!

libwild/src/parsing.rs
```diff
@@ -98,6 +98,9 @@ pub(crate) enum SymbolPlacement<'data> {
 
     /// Symbol will point to the start of the first loadable segment.
     LoadBaseAddress,
+
+    /// A symbol defined by the linker.
+    LinkerDefined,
```
davidlattimore: Most of the other kinds of placements here are also linker-defined. Any reason to not include here the information needed to determine where the symbol should go? It looks like this is used sort of like a hook to say "parse the symbol name, then use the result of that parsing to look up where the symbol goes". What about renaming this to `PlatformSpecific`, then making `SymbolPlacement` be generic over `Platform` and having an associated type `P::PlatformSpecificSymbol` that is held by the variant?

lapla-cogito: That does sound promising, thanks. Implemented in https://github.com/wild-linker/wild/pull/2238/commits/cecf4bfc1e2b17747ff97c34a7d21fdcc6bf32c6

davidlattimore (approved):

davidlattimore:
Thanks :)

PR #2239, opened 2026-07-19
port(wasm): Speed up data symbol address lookup with O(1) segment indexing

lapla-cogito:
Data symbol memory addresses were resolved by linearly scanning an object's data segments. This can dominate link time on large Wasm objects.

Part of #1431


lapla-cogito:
Even moderately complex applications experience performance impacts from this change. In my environment, I saw a 20% speed improvement for ripgrep build.

davidlattimore (approved):

Issue #1221, opened 2025-10-16
Add support for --only-keep-debug


davidlattimore:
`--only-keep-debug` is a flag accepted by `strip`. Linkers already support the flag `--strip-debug`. `--only-keep-debug` is the opposite, keeping only the debug info. For docs, see the `strip` manpage.

The idea would be to run the linker twice. Once to produce the executable and once to produce the `.dbg` file. This is related to #1079, where a single linker invocation produces both files. I think it would be simpler to implement than #1079 and at the same time might be a stepping stone to being able to write both. Also, producing files in the background isn't suitable for some build systems. e.g. I'd guess that writing the dbg file in the background is not something you'd want if you were using either Bazel or Buck. For those sorts of build systems, I think having separate processes to produce each file would be better.

Looking at the behaviour of `strip --only-keep-debug` is useful for understanding how this flag should work. A few other observations:

* Non-debug sections that get discarded should have their type changed to NOBITS and their file-size changed to 0. They keep their address, alignment and other attributes.
* Debug symbols are emitted as per normal. i.e. we have identical debug symbols whether `--strip-debug` or `--only-keep-debug` is supplied.

I'm not 100% sure how this should be implemented. Despite being the "opposite of --strip-debug", the implementation is not at all similar. I don't think the eventual implementation will be too complex, but figuring out some of the details might require some work.

We need to know the addresses of all the symbols and sections. So we need to do most of the phases up to and including layout more or less as per usual. At some point during layout, if the flag is set, we need to set the file-size of many of the sections to 0.

Filing this issue to see what others think. If anyone is interested in working on this, let us know. I won't mark it as a good-first-issue because there are details that I'm not sure about that I think might be tricky to figure out.

YamasouA:
@davidlattimore 
I'm interested in taking a stab at this. It might take me some time because I’ll need to dig into the details and existing code paths, but I’d like to try implementing it. If there’s any guidance on tricky parts or preferred flag/behavior compatibility (e.g., matching strip/mold), please let me know.

lapla-cogito:
I've assigned you @YamasouA . I don't think other linkers support this flag equivalently; therefore, the relevant implementation should be in `strip`, and more specifically in [`objcopy.c`](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=binutils/objcopy.c).

davidlattimore:
[...]

PR #2241, opened 2026-07-20
WIP: --only-keep-debug flag plumbing

deepakshirkem:
Adds --only-keep-debug flag plumbing: CLI flag, debug-section detection by name, and file-size/sh_type overrides so non-debug sections report as NOBITS. Blocked because Wild writes section content directly into the mmap'd output file rather than a separate scratch buffer — zeroing a section's buffer at allocation time breaks relaxation and string merging, which still need real bytes regardless of the flag. 

@davidlattimore Opening as draft to get input on whether this needs a two-pass write (full normal link, then a second pass that copies only surviving bytes into a shrunk file) before continuing.

Issue #1221

### DIFF
--- a/libwild/src/args/elf.rs
+++ b/libwild/src/args/elf.rs
@@ -52,6 +52,7 @@ pub struct ElfArgs {
     pub(crate) lib_search_path: Vec<Box<Path>>,
     pub(crate) dynamic_linker: Option<Box<Path>>,
     pub(crate) strip: Strip,
+    pub(crate) only_keep_debug: bool,
     pub(crate) merge_sections: bool,
     pub(crate) version_script_path: Option<PathBuf>,
     pub(crate) should_write_eh_frame_hdr: bool,
@@ -274,6 +275,7 @@ impl Default for ElfArgs {
             should_output_partial_object: false,
             dynamic_linker: None,
             strip: Strip::Nothing,
+            only_keep_debug: false,
             // For now, we default to --gc-sections. This is different to other linkers, but other
             // than being different, there doesn't seem to be any downside to doing
             // this. We don't currently do any less work if we're not GCing sections,
@@ -788,6 +790,15 @@ fn setup_argument_parser() -> ArgumentParser<ElfArgs> {
         .help("Strip debug symbols")
         .execute(|args, _modifier_stack| {
             args.strip = Strip::Debug;
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("only-keep-debug")
+        .help("Keep only debug sections, discarding content of others")
+        .execute(|args, _modifier_stack| {
+            args.only_keep_debug = true;
             Ok(())
         });
 
@@ -1867,6 +1878,10 @@ impl platform::Args for ElfArgs {
 
     fn common_mut(&mut self) -> &mut crate::args::CommonArgs {
         &mut self.common
+    }
+
+    fn should_only_keep_debug(&self) -> bool {
+        self.only_keep_debug
     }
 
     // TODO: Some linkers like ld and mold cleanup debug symbols when linking with -r. For now, we
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -2161,7 +2161,7 @@ fn write_section_raw<'out, 'data, A: Arch<Platform = Elf>>(
     let part_id = object.section_part_id(section_index, &layout.symbol_db.section_part_ids);
     if layout
         .output_sections
-        .has_data_in_file(part_id.output_section_id())
+        .has_data_in_file(part_id.output_section_id(), layout.args().should_only_keep_debug())
     {
         let section_buffer = buffers.get_mut(part_id);
         let allocation_size = sec.capacity(part_id, &layout.output_sections) as usize;
@@ -5539,8 +5539,13 @@ fn write_section_headers(out: &mut [u8], layout: &ElfLayout) -> Result {
         let entry = entries.next().unwrap();
         let e = LittleEndian;
         entry.sh_name.set(e, name_offset);
-
-        let sh_type = if layout.args().use_android_relr_tags && section_type == sht::RELR {
+        let sh_type = if layout.args().should_only_keep_debug()
+            && section_type != sht::NULL
+            && section_id.is_regular()
+            && !layout.output_sections.is_debug_section(section_id)
+        {
+            sht::NOBITS
+        } else if layout.args().use_android_relr_tags && section_type == sht::RELR {
             object::elf::SHT_ANDROID_RELR
         } else {
             section_type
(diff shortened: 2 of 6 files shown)


davidlattimore:
Ideally we'd be able to make it work without any sort of two-pass arrangement. It also shouldn't be necessary to write the output into a scratch buffer. The only time when it's necessary to write to an internal buffer is when we're compressing debug info - since then we need to write the uncompressed debug info, compress it, then write the result.

Treating sections as no-bits is an interesting approach. I haven't thought too deeply about the implications of it. There might be other approaches too if that one doesn't work out. Something else to keep in mind is that we'd also like at some point to support writing both files - i.e. the file that would have been written with `--strip-debug` as well as the file that would have been written with `--only-keep-debug`. So that's something to keep in mind when deciding how to implement this.

deepakshirkem:
@davidlattimore  Link now succeeds, but found a correctness bug: addresses shift for every section after a hollowed one, not just file offsets (see screenshot — address column, not just offset, differs for `.init`/`.text`/`.dynamic`/`.got`).

Traced it to `align_load_segment_start` in `elf.rs`:
```rust
*mem_offset = segment_alignment.align_modulo(*file_offset as u64, *mem_offset);
```
Segment addresses get derived from the current file offset. Since NOBITS sections shrink file offset, every later segment's address shifts too.

Guessing the fix needs two file-offset counters through layout a real one and a hypothetical one. Wanted to check this matches your thinking before implementing it.


```
<   [ 3] .note.ABI-tag     NOTE             00000000004009b4  000009b4
---
>   [ 3] .note.ABI-tag     NOBITS           00000000004009b4  000009b4
14c14
<   [ 4] .gnu.hash         GNU_HASH         00000000004009d8  000009d8
---
>   [ 4] .gnu.hash         GNU_HASH         00000000004009d8  000009b8
16c16
<   [ 5] .dynsym           DYNSYM           00000000004009f8  000009f8
---
>   [ 5] .dynsym           DYNSYM           00000000004009f8  000009d8
18c18
<   [ 6] .dynstr           STRTAB           0000000000400a40  00000a40
---
>   [ 6] .dynstr           STRTAB           0000000000400a40  00000a20
20c20
<   [ 7] .gnu.version      VERSYM           0000000000400a82  00000a82
---
>   [ 7] .gnu.version      VERSYM           0000000000400a82  00000a62
22c22
<   [ 8] .gnu.version_r    VERNEED          0000000000400a88  00000a88
---
>   [ 8] .gnu.version_r    VERNEED          0000000000400a88  00000a68
24c24
<   [ 9] .rela.dyn         RELA             0000000000400aa8  00000aa8
---
>   [ 9] .rela.dyn         RELA             0000000000400aa8  00000a88
26c26
<   [10] .eh_frame         PROGBITS         0000000000400ad8  00000ad8
---
>   [10] .eh_frame         PROGBITS         0000000000400ad8  00000ab8
28c28
<   [11] .init             PROGBITS         0000000000401b70  00000b70
---
>   [11] .init             NOBITS           0000000000401b50  00000b50
30c30
<   [12] .fini             PROGBITS         0000000000401b8c  00000b8c
---
>   [12] .fini             NOBITS           0000000000401b6c  00000b50
32c32
<   [13] .text             PROGBITS         0000000000401ba0  00000ba0
---
>   [13] .text             NOBITS           0000000000401b80  00000b50
34c34
<   [14] .dynamic          DYNAMIC          0000000000402bf0  00000bf0
---
>   [14] .dynamic          DYNAMIC          0000000000402b50  00000b50
36c36
<   [15] .got              PROGBITS         0000000000402ea0  00000ea0
---
>   [15] .got              PROGBITS         0000000000402e00  00000e00
38,40c38,40
<   [16] .relro_padding    NOBITS           0000000000402eb0  00000eb0
<        0000000000000150  0000000000000000  WA       0     0     1
<   [17] .debug_info       PROGBITS         0000000000000000  00000eb0
---
>   [16] .relro_padding    NOBITS           0000000000402e10  00000e10
>        00000000000001f0  0000000000000000  WA       0     0     1
>   [17] .debug_info       PROGBITS         0000000000000000  00000e10
42c42
<   [18] .debug_abbrev     PROGBITS         0000000000000000  00000f04
---
>   [18] .debug_abbrev     PROGBITS         0000000000000000  00000e64
44c44
<   [19] .debug_aranges    PROGBITS         0000000000000000  00000f3c
---
>   [19] .debug_aranges    PROGBITS         0000000000000000  00000e9c
46c46
<   [20] .debug_line       PROGBITS         0000000000000000  00000f6c
---
>   [20] .debug_line       PROGBITS         0000000000000000  00000ecc
48c48
<   [21] .debug_str        PROGBITS         0000000000000000  00000fbf
---
>   [21] .debug_str        PROGBITS         0000000000000000  00000f1f
50c50
<   [22] .debug_line_str   PROGBITS         0000000000000000  00001052
---
>   [22] .debug_line_str   PROGBITS         0000000000000000  00000fb2
52c52
<   [23] .comment          PROGBITS         0000000000000000  00001085
---
>   [23] .comment          NOBITS           0000000000000000  00000fe5
54c54
<   [24] .shstrtab         STRTAB           0000000000000000  00001116
---
>   [24] .shstrtab         STRTAB           0000000000000000  00001076
56c56
<   [25] .symtab           SYMTAB           0000000000000000  00001230
---
>   [25] .symtab           SYMTAB           0000000000000000  00001190
58c58
<   [26] .strtab           STRTAB           0000000000000000  00001398
---
>   [26] .strtab           STRTAB           0000000000000000  000012f8

```

davidlattimore:
Hmm, yes, that bit where we compute memory offset based on both the previous memory offset and the file offset modulo the page size is likely unavoidable. I suspect it might work better to perform most of the layout phase as if we were going to write the full file, then at some point late in layout, if `--only-keep-debug` is active, adjust the layout. But you could also try keeping an extra counter and see how that looks. Just keep in mind what things will look like when we support writing both files in the one linker invocation.

Another complication that I just realised is build IDs. For a pair of files - executable and debug info - the build IDs are supposed to match. We currently compute the build ID by hashing the output after we've written it, which means if we invoke the linker twice, once with `--strip-debug` and once with `--only-keep-debug`, we'll end up with different build IDs. One option would be to say that for now, if `--only-keep-debug` is specified, then the build-id has to be explicitly passed via `--build-id=0x...` or disabled via `--build-id=none`. Another option would be to have a way to compute the build ID by hashing all the inputs rather than hashing the output. Whether we'd want this mode on by default would depend on how it compares performance-wise on a few real-world benchmarks.

Another option to consider, is to defer `--only-keep-debug` and go straight to writing split debug - i.e. two output files from the one line. I originally thought that `--only-keep-debug` would be an easier stepping stone to split debug, but given some of the complications, it might actually be that `--only-keep-debug` ends up more complex.

PR #2243, opened 2026-07-21
 fix: do not retain a linker script output section if a symbol is defined after it

vishruth-thimmaiah:
When defining a symbol between two output sections in a linker script, the symbol should reference any preceding section before it, that is retained. An empty section that would otherwise get discarded shouldn't be retained.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -4622,12 +4622,10 @@ fn get_symbol_attributes(
                     .output_sections
                     .output_index_of_nearest_section(section_id)
             } else {
-                def_info
-                    .section_id()
-                    .and_then(|section_id| {
-                        let section_id = layout.output_sections.primary_output_section(section_id);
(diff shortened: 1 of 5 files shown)


libwild/src/elf_writer.rs
```diff
@@ -4612,13 +4612,24 @@ fn get_symbol_attributes(
...
+            let shndx = if let crate::parsing::SymbolPlacement::Redirect(redirect) =
+                &def_info.placement
+                && let SymbolLoc::SectionEnd(section_id) = redirect.loc
+            {
+                let section_id = layout.output_sections.primary_output_section(section_id);
+                layout
+                    .output_sections
+                    .output_index_of_nearest_section(section_id)
+            } else {
+                script.internal_symbols.symbol_definitions[local_index.0]
```
```suggestion
                def_info
```

lapla-cogito (approved):
Looks good, thanks!

PR #2244, opened 2026-07-22
port(wasm): Accept constructors with non-void return types

lapla-cogito:
InitFuncs were required to be `() -> ()`, which rejects common patterns like `int ctor(void)`. Require empty parameter lists only and drop results in the synthesized `__wasm_call_ctors` body. Noticed while building CRuby.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -3117,12 +3117,17 @@ fn owned_linker_function_body(bytes: Vec<u8>) -> WasmFunctionBody<'static> {
     }
 }
 
-fn encode_call_sequence_body(func_indices: &[u32]) -> Vec<u8> {
+/// Encode a body that calls each function in order.
+///
+/// `calls` is `(function_index, result_count)`. Result values are dropped so that
+/// `__wasm_call_ctors` can stay `() -> ()` even when a constructor returns a value.
+fn encode_call_sequence_body(calls: &[(u32, usize)]) -> Vec<u8> {
     let mut bytes = vec![0x00]; // 0 locals
-    for &func_index in func_indices {
+    for &(func_index, result_count) in calls {
         bytes.push(0x10); // call
         leb128::write::unsigned(&mut bytes, u64::from(func_index))
             .expect("leb128 write to Vec cannot fail");
+        bytes.extend(std::iter::repeat_n(0x1a, result_count)); // drop each result
     }
     bytes.push(0x0b); // end
     bytes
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -3150,11 +3155,11 @@ fn function_type_for_symbol<'a>(
...
 }
 
-/// From InitFuncs to output function indices, sorted by ascending priority.
-fn collect_sorted_init_function_indices(
+/// From InitFuncs to `(output function index, result count)`, sorted by ascending priority.
+fn collect_sorted_init_function_calls(
     inputs: &[WasmObjectLayoutInput<'_>],
     object_index_maps: &[WasmObjectIndexMap],
-) -> Result<Vec<u32>> {
+) -> Result<Vec<(u32, usize)>> {
```
davidlattimore, line 3162: Could consider giving a name to the type here that is represented as a tuple - i.e. defining a struct. Up to you though.

libwild/src/wasm.rs
```diff
@@ -3166,16 +3171,20 @@ fn collect_sorted_init_function_indices(
             );
             let ty = function_type_for_symbol(input, sym)?;
             ensure!(
-                ty.params().is_empty() && ty.results().is_empty(),
-                "Wasm constructor must have type () -> ()"
+                ty.params().is_empty(),
+                "Wasm constructor must take no parameters (got {} param(s))",
+                ty.params().len()
             );
```
davidlattimore, line 3177: Removing this doesn't seem to cause any tests to fail. Should we test for this?

davidlattimore (commented):
Do we need a test that confirms that

libwild/src/wasm.rs
lapla-cogito: This is adequately explained in the doc comment for this function I think.

libwild/src/wasm.rs
lapla-cogito: Added, thanks.

lapla-cogito (commented):
> Do we need a test that confirms that

Hasn't a test for constructors that return values already been added in the `constructors-return-int` test?

davidlattimore:
> > Do we need a test that confirms that
> 
> Hasn't a test for constructors that return values already been added in the `constructors-return-int` test?

Yep, sorry. That was an incomplete sentence that was intended to refer to the no-parameters check, but then I ended up adding a comment there and forgot to delete the partial sentence. Sorry for the confusion.

davidlattimore (approved):

Issue #2247, opened 2026-07-22
AArch64 correctness against lld test suite


deepakshirkem:
Running lld's AArch64 ELF test suite against Wild to find correctness gaps

PR #2250, opened 2026-07-22
fix: Error on R_AARCH64_ABS32/ABS16 against symbols in shared objects

deepakshirkem:
Wild silently accepted R_AARCH64_ABS32 and R_AARCH64_ABS16 relocations when linking a shared object, producing a broken output -- these fields are too narrow to hold a runtime-resolved address.

This adds a check erroring with "recompile with -fPIC" when these relocation types are used against a shared object, matching lld and GNU ld's behavior.  implemented as a new is_disallowed_in_shared_object method on the Arch trait, shared between x86_64 and aarch64, rather than per-architecture matches in elf_writer.rs.

Issue #2247

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -3013,13 +3013,7 @@ fn apply_relocation<
         }
         RelocationKind::Absolute
             if layout.symbol_db.output_kind.is_shared_object()
-                && matches!(
-                    r_type,
-                    object::elf::R_X86_64_32
-                        | object::elf::R_X86_64_32S
-                        | object::elf::R_X86_64_8
-                        | object::elf::R_X86_64_16
-                ) =>
+                && A::is_disallowed_in_shared_object(r_type) =>
         {
             bail!(
                 "relocation type {} cannot be used when making a shared object; \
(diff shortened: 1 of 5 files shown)


libwild/src/elf_writer.rs
```diff
@@ -3027,6 +3027,19 @@ fn apply_relocation<
                 A::rel_type_to_string(r_type)
             );
         }
+        RelocationKind::Absolute
+            if layout.symbol_db.output_kind.is_shared_object()
+                && matches!(
+                    r_type,
+                    object::elf::R_AARCH64_ABS32 | object::elf::R_AARCH64_ABS16
```
davidlattimore: Relocation type codes are arch-specific, so we can't refer to them from code that is shared between architectures - e.g. elf_writer - otherwise we might get false positives when linking for other architectures that happen to use the same numeric values for different relocations.

I see that there's very similar code above from #2150 and I failed to notice the problem when reviewing it. Perhaps you could change the x86_64 code to call a new method on the `Arch` trait - e.g. `A::is_disallowed_in_shared_object(r_type)`, then implement that method for both x86_64 and aarch64? That would also allow us to avoid duplicating the message.

deepakshirkem: Done. Thank You .!!

davidlattimore (approved):

libwild/src/elf_writer.rs
lapla-cogito: Thanks for the refactoring!

PR #2256, opened 2026-07-23
feat: define multiple program headers per output section in linker scripts

vishruth-thimmaiah:
A single output section can be mapped to multiple program headers defined in a linker script.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -38,6 +38,7 @@ use crate::output_kind::OutputKind;
 use crate::output_section_id;
 use crate::output_section_id::CustomSectionIds;
 use crate::output_section_id::NUM_BUILT_IN_SECTIONS;
+use crate::output_section_id::OrderEvent;
 use crate::output_section_id::OutputOrder;
 use crate::output_section_id::OutputOrderBuilder;
 use crate::output_section_id::OutputSectionId;
@@ -68,7 +69,9 @@ use crate::platform::SectionType as _;
 use crate::platform::Symbol as _;
 use crate::platform::ThunkConfig;
 use crate::platform::VerneedTable as _;
+use crate::program_segments::ProgramSegmentId;
 use crate::program_segments::ProgramSegments;
+use crate::program_segments::SegmentEntry;
 use crate::resolution::LoadedMetrics;
 use crate::string_merging::MergedStringStartAddresses;
 use crate::string_merging::MergedStringsSection;
@@ -840,6 +843,20 @@ impl platform::Platform for Elf {
         &[STACK_SEGMENT_DEF]
     }
 
+    fn get_segment_flags_for_section(section_flags: &Self::SectionFlags) -> u32 {
+        let mut flags = 0;
+        if section_flags.contains(shf::ALLOC) {
+            flags |= pf::READABLE.0;
+        }
+        if section_flags.contains(shf::WRITE) {
+            flags |= pf::WRITABLE.0;
+        }
+        if section_flags.contains(shf::EXECINSTR) {
+            flags |= pf::EXECUTABLE.0;
+        }
+        flags
+    }
+
     fn create_linker_defined_symbols(
         symbols: &mut crate::parsing::InternalSymbolsBuilder<Elf>,
         output_kind: OutputKind,
@@ -976,9 +993,9 @@ impl platform::Platform for Elf {
                 min_alignment: d.min_alignment,
                 location_info: None,
                 secondary_order: None,
-                phdr_name: None,
                 region_name: None,
                 fill: None,
+                phdrs: Vec::new(),
             })
             .collect()
     }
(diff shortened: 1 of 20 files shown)


libwild/src/elf.rs
```diff
@@ -2017,90 +2033,148 @@ impl platform::Platform for Elf {
...
+                        has_explicit_flags: phdr.flags.is_some(),
+                    },
+                );
+            }
+
+            for id in &script.parsed.ordered_sections {
+                let info = output_sections.section_infos.get(*id);
+                for phdr in &info.phdrs {
+                    let segment = segments_map.get_mut(phdr).with_context(|| {
+                        format!(
```
davidlattimore, line 2083: If I apply the following change without this PR, the modified test passes, but fails with this PR. Do we want handle `:NONE`? We were sort of handling it before - albeit possibly accidentally.

```diff
diff --git a/wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.ld b/wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.ld
index 49e04e67b8..4dcead98e2 100644
--- a/wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.ld
+++ b/wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.ld
@@ -22,6 +22,10 @@
     .rodata : {
         *(.rodata*)
     } :rodata
+
+    /* :NONE suppresses the inherited :rodata program-header assignment. */
+    .empty : {
+    } :NONE
 }
 
 ASSERT(start_of_sections == 0x40 + ((3 + is_riscv) * 0x38), "Wrong start of sections");
```

libwild/src/elf.rs
```diff
@@ -2017,90 +2033,148 @@ impl platform::Platform for Elf {
...
+                    if Some(seg_id) == first_exec {
+                        builder.add_sections(&custom.exec);
+                    } else if Some(seg_id) == first_rw {
+                        builder.add_sections(&custom.tdata);
+                        builder.add_sections(&custom.tbss);
+                        builder.add_sections(&custom.data);
+                        builder.add_sections(&custom.bss);
+                    } else if Some(seg_id) == first_ro {
+                        builder.add_sections(&custom.ro);
+                    }
```
davidlattimore, line 2195: If there's no segment with the right permissions, then it seems that we drop the section. Can you add a test for that?

davidlattimore (commented):
Can you add a test where a segment gets defined, but not used?

libwild/src/elf.rs
```diff
@@ -2019,90 +2038,157 @@ impl platform::Platform for Elf {
...
+                }
+                ends[seg_idx] = Some(pos);
+                if entry.ptype == pt::LOAD.0 {
+                    if (entry.flags & pf::EXECUTABLE.0) != 0 {
+                        first_exec = Some(entry.id);
+                    } else if (entry.flags & pf::WRITABLE.0) != 0 {
+                        first_rw = Some(entry.id);
+                    } else {
+                        first_ro = Some(entry.id);
                     }
```
lapla-cogito: Since we're overwriting them every time here, naming them `first_*` might cause confusion.

vishruth-thimmaiah: I forgot to add a check to see if `first_*` was already assigned, which I have now done. Now we only reassign if we find a more favourable segment, i.e. if the original `first_rw` was a segment with the `RWX` flag, and we find a segment with `RW`, we reassign `first_rw` to the new segment.

libwild/src/elf.rs
vishruth-thimmaiah: I initially wanted to handle `:NONE` in a separate PR, but the implementation ended up being straightforward, so I've included it in this PR.

libwild/src/elf.rs
vishruth-thimmaiah: If there is a single `LOAD` segment all sections are included in that segment, regardless of the segment's flags([linker-script-phdrs-single-load-with-flag.ld‎](https://github.com/wild-linker/wild/pull/2256/changes#diff-8057ddbeeae971ca753021bd7b627662dd9236d7ff1e2e64defa3047e6089883)). If there is no `LOAD` segment defined ld just errors.

libwild/src/elf.rs
```diff
@@ -2019,90 +2038,189 @@ impl platform::Platform for Elf {
...
-        let mut insert_re = false;
-        let mut insert_rw = false;
-        let mut insert_r = false;
-        let mut first_load = false;
+        let mut segments_map = HashMap::new();
+        let mut ordered_sections = Vec::new();
+
+        let mut num_phdrs = 0;
+
+        let mut segment_has_explicit_flags = Vec::with_capacity(num_phdrs);
```
davidlattimore: Might as well do `Vec::new()`, since I think it will be equivalent to `Vec::with_capacity(0)`

libwild/src/elf.rs
```diff
@@ -2019,90 +2038,189 @@ impl platform::Platform for Elf {
...
+        }
 
-                    builder.add_segment_with_sections(
-                        ptype as u32,
-                        flags,
-                        sections,
-                        output_sections,
-                    );
+        if first_load.is_none() {
+            bail!("Final Link failed: bad value");
```
davidlattimore: I think I'd be somewhat confused seeing this error message. I assume this is the message that GNU ld emits, but I'd be inclined to see if we can give something more helpful - e.g. "Missing LOAD PHDR in linker script"?

vishruth-thimmaiah: I've updated the error message and test.

davidlattimore (approved):

lapla-cogito (approved):

PR #2262, opened 2026-07-24
port(wasm): Defer reloc entry decoding to layout

lapla-cogito:
While we needed to materialize symbols from the `linking` section when parsing the input file, we didn't need to decode the `reloc` section at this stage. Therefore record only the section header at parse time and decode entries when building layout inputs in this PR. This makes linking ripgrep about 10% faster.

Part of #1431


davidlattimore (approved):
Nice :)

PR #2263, opened 2026-07-24
refactor: use Platform::RelocationInfo for relocation types

philipc:
Make the `Relocation` trait generic over the platform and replace relocation type values with `Platform::RelocationInfo` instead of `u32`.

This is preparation for updating the `object` crate to use newtypes for relocation types. For ELF, the u32 will become `object::elf::RelocationType`.

Mach-O is already using `object::macho::RelocationInfo`. There will be a new `object::macho::RelocationType` as well, but it's simpler to use `Platform::RelocationInfo` everywhere rather than adding a new associated type.


davidlattimore (approved):
Thanks!

PR #2265, opened 2026-07-24
tests: drop parse_bool and provide more context

marxin:
Newly reported:
```
  Error: Failed to process test directive wild/tests/sources/elf/compressed-debug/compressed-debug.c:12
    Caused by:
      problematic directive line: 'RequiresZstdCompression:dog'
      provided string was not `true` or `false`
```

I fell it's much better as we provide enough information about the wrong directive.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -1880,6 +1880,7 @@ fn parse_configs(src_filename: &Path, default_config: &Config) -> Result<Vec<Con
                 &mut configs,
                 is_rust,
             )
+            .with_context(|| format!("problematic directive line: '{}'", rest))
             .with_context(|| {
                 format!(
                     "Failed to process test directive {}:{}",
@@ -2176,7 +2177,7 @@ fn process_directive(
         "EnableLinker" => {
             config.enabled_linkers.insert(arg.to_owned());
         }
-        "Cross" => config.cross_enabled = parse_bool(arg, "Cross")?,
+        "Cross" => config.cross_enabled = arg.to_lowercase().parse()?,
         "ExpectError" => {
             config.expect_stderr.push(ErrorMatcher::new(arg)?);
             config.should_error = true;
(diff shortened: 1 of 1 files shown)


wild/tests/integration_tests.rs
```diff
@@ -2176,7 +2177,7 @@ fn process_directive(
         "EnableLinker" => {
             config.enabled_linkers.insert(arg.to_owned());
         }
-        "Cross" => config.cross_enabled = parse_bool(arg, "Cross")?,
+        "Cross" => config.cross_enabled = arg.to_lowercase().parse()?,
```
davidlattimore, line 2180: Do we want to accept different cases for bools? Seems like it'd potentially result in less consistency in our test directives. That said, I guess we currently have other inconsistencies, like whether there's a space after the ':' or not. Anyway, I don't feel strongly, so happy to go with this if you want.

wild/tests/integration_tests.rs
```diff
@@ -1880,6 +1880,7 @@ fn parse_configs(src_filename: &Path, default_config: &Config) -> Result<Vec<Con
                 &mut configs,
                 is_rust,
             )
+            .with_context(|| format!("problematic directive line: '{}'", rest))
```
davidlattimore, line 1883: Nice addition :)

davidlattimore (approved):

wild/tests/integration_tests.rs
marxin: Agree, we can be definitely stricter when it comes to directive parsing - lemme address it in a separate PR then. This particular change is well aligned with other places in the function where we parse booleans.

marxin: The `//#Directive` form followed by a space appears in roughly 250 places, so I’m not convinced that changing all of them is a good idea.

davidlattimore: Yeah, that's fair enough

PR #2267, opened 2026-07-24
feat: introduce FileSystem abstraction

### LINKED ISSUE #2218: Support running on a system without file-system
One of the inherent limitations of the SP1 zkVM is the lack of a file system. This makes sense, since a proof must depend deterministically on its inputs, which in practice are provided only through stdin. I was therefore thinking about introducing a VFS abstraction - a virtual file-system trait or struct - that would take precedence whenever Wild reads from or writes to a file. The following commit is an early draft; for now, I am only interested in supporting input files and the output file: https://github.com/marxin/wild/commit/d6dffd1d4fe507c73cb16cb17c9b4f40d25b672d

The main question is whether this approach would be suitable for upstreaming and, if so, how it could be integrated cleanly into Wild.

marxin:
Based on the discussion in the linked issue, the PR introduces a new `FileSystem` trait that provides abstraction over a file system. All the current behavior (related to real FS: `OsFileSystem`) should be preserved and I tried really hard to preserve also all the valuable comments. One place where I intentionally left the current behavior is `save_dir.rs` as it's mostly focused on debugging functionality. I feel pretty happy about the trait itself and the number of places where an actual code change was necessary - hopefully something viable.

Resolves: #2218


libwild/src/fs.rs
```diff
@@ -1,6 +1,431 @@
...
+pub enum FileWriteMode {
+    Mmap,
+    BufferThenWrite,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct OutputOptions {
+    pub size: u64,
+    pub file_replacement_mode: FileReplacementMode,
+    pub write_mode: Option<FileWriteMode>,
```
davidlattimore, line 45: FileReplacementMode and FileWriteMode are public, but not exported, so users of libwild can't use these fields. Do you think we should add an example program that uses the public API, both to serve as a reference for the API and also to help pick up visibility issues?

davidlattimore (approved):
Fine to merge once the visibility issue is resolved. An example program would be good, as it might influence the API, but can also wait until another PR if you prefer.

libwild/src/fs.rs
```diff
@@ -1,6 +1,431 @@
...
+
+    /// Resolves symbolic links and returns the canonical absolute path.
+    fn canonicalize(&self, path: &Path) -> std::io::Result<PathBuf>;
+
+    /// Removes a file.
+    fn remove_file(&self, path: &Path) -> std::io::Result<()>;
+
+    /// Rename an existing file out of the way and remove it in a background task. Failures are
+    /// ignored because the subsequent output creation will report any relevant error.
+    fn remove_in_separate_thread(&self, path: &Path);
```
davidlattimore: This seems like an unfortunate method to have in the trait, since it leaks linker implementation details that an external implementor ideally shouldn't have to know about. e.g. the implementor needs to know that they're expected to spawn a task using rayon. I wonder if we should just have `rename` instead? Then the code that needs this can handle thread spawning.

libwild/src/fs.rs
marxin: Good catch!

libwild/src/fs.rs
marxin: Fully agree on this one, it's pretty unexpected API function.

marxin:
Included one more API change (`fn finish`) and included an in-memory example of the new API.

davidlattimore (approved):
Nice! I think this definitely expands the potential for using libwild as a library

PR #2270, opened 2026-07-25
tools: only move arguments after -- to the end in save-dir

### LINKED ISSUE #2259: Reproduction steps for failed tests with cargo no longer work after #2232
A failed test might print:
```
Relink with:
WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 OUT=/home/mateusz/Projects/wild/wild/tests/build/elf/aarch64/crtend-only-eh-frame/clang/shared.c.wild.so /home/mateusz/Projects/wild/wild/tests/build/elf/aarch64/crtend-only-eh-frame/clang/shared.c.wild.save/run-with cargo run --bin wild --
```
but that no longer works due to #2232 and instead fails with:
```
error: unexpected argument '-E' found

Usage: cargo [+toolchain] [OPTIONS] [COMMAND]
       cargo [+toolchain] [OPTIONS] -Zscript <MANIFEST_RS> [ARGS]...

For more information, try '--help'.
```
That's because this line:
```
L="$1"
shift
```
takes only `cargo`, so the execution looks like this: `cargo <linker-args> run --bin wild --`.

I think the tests should print:
```
Relink with:
WILD_WRITE_LAYOUT=1 WILD_WRITE_TRACE=1 LINKER_WRAPPER="cargo run --bin wild" OUT=/wild/wild/tests/build/elf/aarch64/crtend-only-eh-frame/clang/shared.c.wild.so /wild/wild/tests/build/elf/aarch64/crtend-only-eh-frame/clang/shared.c.wild.save/run-with --
```

mati865:
Fixes #2259

Terribly over-engineered, but it works even with spaces!

```
❯ mkdir '/tmp/dir with space'

❯ cp target/debug/wild /tmp/dir\ with\ space/

❯ cp /bin/cargo /tmp/dir\ with\ space/

❯ wild/tests/build/elf/x86_64/archive-activation/lto-clang/archive-activation.c.save/run-with cargo run --bin wild -q -- --

❯ wild/tests/build/elf/x86_64/archive-activation/lto-clang/archive-activation.c.save/run-with "/tmp/dir with space/cargo" run --bin wild -q -- --

❯ wild/tests/build/elf/x86_64/archive-activation/lto-clang/archive-activation.c.save/run-with wild

❯ wild/tests/build/elf/x86_64/archive-activation/lto-clang/archive-activation.c.save/run-with '/tmp/dir with space/wild'
```

Brief description:
The script now adds arguments before `run-with` arguments by default, similarly to how it used to work before the recent changes. Arguments after last `--` are added to the end.
This also means that when command needs its own `--`, it has to be sort of escaped by adding another `--`, as shown above with Cargo.

### DIFF
--- a/BENCHMARKING.md
+++ b/BENCHMARKING.md
@@ -6,7 +6,8 @@ If you decide to benchmark Wild against other linkers, in order to make it a fai
 should ensure that the other linkers aren't doing work on something that Wild doesn't support. In
 particular:
 
-* Wild defaults to `--gc-sections`, so for a fair comparison, that should be passed to all the linkers.
+* Wild defaults to `--gc-sections`, so for a fair comparison, that should be passed to all the
+  linkers.
 * Wild defaults to `-z now`, so best to pass that to all linkers.
 
 ## How to benchmark
@@ -23,8 +24,8 @@ comparisons you wish to make, or what change in wild you want to quantify.
 
 Follow-these steps:
 
-* Chose the crate that you wish to use in your benchmark, clone it, `cd` into it's root directory
-  and make sure it builds with `cargo build` (for a rust project)
+* Chose the crate that you wish to use in your benchmark, clone it, `cd` into its root directory and
+  make sure it builds with `cargo build` (for a rust project)
     * Examples: [`ripgrep`](https://github.com/BurntSushi/ripgrep.git)
 * Clean the build using `cargo clean`
 * To force the build of your chosen crate to link using wild, we have a couple of options:
@@ -51,7 +52,7 @@ rustflags = ["-Clink-arg=--ld-path=wild"]
 
 When you run `run-with wild`, the linker may print warnings for unsupported flags. It's a good idea
 to edit the `run-with` script to change / delete these flags. This will make comparison with other
-linkers more fair, since some of these unsupported flags may involve other linkers doing significant
+linkers fairer, since some of these unsupported flags may involve other linkers doing significant
 amounts of extra work.
 
 ### Run benchmark with hyperfine
@@ -86,9 +87,11 @@ Summary
 
 ### Run benchmark with poop
 
-An alternative tool to hyperfine, that reports some additional metrics is [`poop`](https://github.com/andrewrk/poop).
+An alternative tool to hyperfine, that reports some additional metrics is [
+`poop`](https://github.com/andrewrk/poop).
 
-Like hyperfine it takes a number of commands and runs each a number of times and gathers statistics about each tune.
+Like hyperfine it takes a number of commands and runs each a number of times and gathers statistics
+about each tune.
 
 ```shell
 poop '/tmp/wild/ripgrep/6/run-with ld' '/tmp/wild/ripgrep/6/run-with mold' '/tmp/wild/ripgrep/6/run-with wild'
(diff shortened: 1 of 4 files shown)


libwild/src/save-dir-prelude.sh
```diff
@@ -3,5 +3,22 @@ D=$(dirname $BASH_SOURCE)
...
+for arg in "$@"; do
+  if [[ "$arg" == "--" ]]; then
+    has_double_dash=true
+    break
+  fi
+done
+if $has_double_dash; then
+  for arg in "$@"; do
+    L+=("$arg")
+    [[ "$arg" == "--" ]] && break
```
davidlattimore: It looks like at the moment, you're passing through the `--` arg, which probably works for the `cargo run` use-case, since `cargo run` accepts and in fact needs the `--`, but for some other uses, the `--` wouldn't be accepted. e.g. use-cases like `./run-with pprof target/opt-debug/wild -- --strip-debug`. I'd be inclined to consume the `--`. It does mean that for the `cargo run` use-case, we'd need `--` twice - one for the run-with script and one for `cargo run`, but it gives us more flexibility.

davidlattimore (commented):
Nice! There's a documented use of the run-with script that probably needs updating in `BENCHMARKING.md`.

libwild/src/save-dir-prelude.sh
mati865: Hmm, that works but it needs to consume the last double-dash. I have tested it with following cases:
- `./run-with wild`
- `./run-with gdb --args wild --`
- `./run-with gdb --args wild -- --no-fork` (here `--no-fork` gets appended to the end)
- `./run-with lldb -- wild --`
- `./run-with lldb -- wild -- --no-fork` (again `--no-fork` gets appended to the end)
- `./run-with cargo run --bin wild -- --`
- `./run-with cargo run --bin wild -- -- --no-fork` (again `--no-fork` gets appended to the end)

Obviously moving `--no-fork` before the last double-dash gets it prepended before `run-with` args.

I couldn't figure out pprof to check it's arg handling. At least with Ubuntu 24.04 `google-perftools` package the args seem different. On Arch Linux it doesn't seem to be provided at all.

Pushed that change, and probably won't pick it up until tomorrow.

mati865:
Can you elaborate more on `BENCHMARKING.md` changes? I don't see what is outdated there.

I've been wondering why this branch didn't fail with neither yours, nor mine changes to `run-with`: https://github.com/wild-linker/wild/blob/6229e9df9318f9a9077502ba0dbac1568de11886/wild/tests/integration_tests.rs#L5855-L5874

Turns out this seems to be all dead code. Replacing it with just the panic doesn't cause any new failures and nothing stands out as ignore reason for the ignored ones.

BENCHMARKING.md
davidlattimore: At the moment, I think this line would need updating - actually it should have been updated with my original PR. But that got me thinking - right now, if there's no '--', we behave as if there was a '--' after the first argument. If we instead behaved as if there was '--' after the last argument - i.e. put all the arguments at the start, then this line wouldn't need updating. That would also mean that this PR was basically restoring the behaviour prior to my PR, but with an added way to append arguments at the end.

davidlattimore:
> Turns out this seems to be all dead code. Replacing it with just the panic doesn't cause any new failures and nothing stands out as ignore reason for the ignored ones.

That code is part of the `Display` implementation for a `LinkCommand`. It's only run if a `LinkCommand` gets printed, which only happens if a test fails. I was able to get it to run by making the following test edit:

```patch
diff --git a/wild/tests/sources/elf/rust-integration/rust-integration.rs b/wild/tests/sources/elf/rust-integration/rust-integration.rs
index 992ed99731..8727af0758 100644
--- a/wild/tests/sources/elf/rust-integration/rust-integration.rs
+++ b/wild/tests/sources/elf/rust-integration/rust-integration.rs
@@ -33,8 +33,6 @@
 //#RequiresRustMusl: true
 //#RequiresNightlyRustc: true
 //#Arch: aarch64
-//#DiffIgnore:section.debug_str.flags
-//#DiffIgnore:section.debug_str.entsize
 
 //#Config:llvm-dynamic:default
 //#SkipArch: ppc64le
```

Then running the test elf/aarch64/rust-integration/cranelift-static-aarch64.

BENCHMARKING.md
mati865: Huh? I don't know how many times I have looked at that line and saw nothing wrong, so good catch there.

> If we instead behaved as if there was '--' after the last argument - i.e. put all the arguments at the start, then this line wouldn't need updating.

I don't have a preference, as long as it works.

> That would also mean that this PR was basically restoring the behaviour prior to my PR, but with an added way to append arguments at the end.

Not exactly restores previous behaviour. If command requires `--`, like `cargo run`, it still would need "escaping" it with second `--`. It didn't need that previously.

mati865:
Oh, my bad. I thought they all would be used for printing and running tests, this branch is only used for printing.

BENCHMARKING.md
davidlattimore: That's a good point, cargo run would still need the extra `--`. I don't feel strongly, but am somewhat inclined to say that the behaviour for no `--` should probably be as if the `--` were at the end.

mati865:
I fear we will end up needing unit tests for this script with how complicated it gets 😆

davidlattimore (approved):

PR #2271, opened 2026-07-26
port(wasm): Synthesize GOT.mem globals for `R_WASM_GLOBAL_INDEX`

lapla-cogito:
With `-fPIC`, clang refers to data via linker-made i32 globals (import module GOT.mem / name-section `GOT.data.internal.*`). `R_WASM_GLOBAL_INDEX_*` then targets those Data symbols.

Now Wild synthesize one i32 global per unique defined data target, initialize it with the final linear-memory address after data layout, absorb matching GOT.mem.* imports, and resolve GLOBAL_INDEX relocs to the synthetic global index. True Global symbols keep the previous path.

This enables building CRuby for WebAssembly (see https://github.com/ruby/ruby/tree/master/wasm#readme). I verified with v4.0.6.

Part of #1431


lapla-cogito:
Running CRuby requires at least version 4.0.0 when using wasi-sdk23+. Without this, builds fail locally even with wasm-ld. This is because [a patch](https://github.com/ruby/ruby/commit/eac35edfd1101e8f7c34dbdd7b595fdac8f0ad4c) for mprotect isn't included before this version.

libwild/src/wasm.rs
```diff
@@ -2948,8 +3000,298 @@ struct LinkerDefinedIndices {
...
+            if sym.kind != WasmSymbolKind::Data {
+                continue;
+            }
+            let symbol_id = input.symbol_id_range.offset_to_id(sym_idx);
+            let def_id = symbol_db.definition(symbol_id);
+            let slot = if let Some(&slot) = def_to_slot.get(&def_id) {
+                slot
+            } else {
+                let Some(&def_obj_idx) =
+                    file_id_to_index.get(&symbol_db.file_id_for_symbol(def_id))
```
davidlattimore: What about symbols that aren't from files like `__heap_base`?

libwild/src/wasm.rs
```diff
@@ -3652,9 +4022,17 @@ where
     let wrap_entry = has_init_funcs
         && !call_ctors_used_in_objects(&layout_inputs)
         && entry_is_defined_function(&layout_inputs, symbol_db);
-    let indices = LinkerDefinedIndices::compute(
+
+    let file_id_to_index: HashMap<crate::input_data::FileId, usize> = layout_inputs
```
davidlattimore: Didn't we already build a file_id_to_index map elsewhere?

libwild/src/wasm.rs
```diff
@@ -2948,8 +3000,298 @@ struct LinkerDefinedIndices {
...
+    symbol_db: &SymbolDb<'_, Wasm>,
+    file_id_to_index: &HashMap<crate::input_data::FileId, usize>,
+) -> Result<(GotMem, GotMemSlotMap)> {
+    let mut def_to_slot: HashMap<SymbolId, usize> = HashMap::new();
+    let mut entries = Vec::new();
+    let mut per_object_slots = vec![Vec::new(); layout_inputs.len()];
+
+    for (obj_idx, input) in layout_inputs.iter().enumerate() {
+        let mut hits: Vec<(usize, usize)> = Vec::new();
+        for reloc in input
```
davidlattimore, line 3173: Ideally it'd be good if we could loop through the relocations just once during the layout phase and once during the write phase. Doing that would avoid repeating work and potentially repeating logic each time we iterate.

libwild/src/wasm.rs
lapla-cogito: Ah, you're right, I forgot to consider that.

libwild/src/wasm.rs
lapla-cogito: Changed to build a map in a single location.

libwild/src/wasm.rs
lapla-cogito: Thanks, done.

libwild/src/wasm.rs
```diff
@@ -3035,132 +3047,217 @@ impl GotMem {
...
+        let def_off = def_id.to_offset(def_input.symbol_id_range);
+        let def_ok = def_input
+            .symbols
+            .get(def_off)
+            .is_some_and(|s| s.kind == WasmSymbolKind::Data && !s.is_undefined());
+        ensure!(
+            def_ok,
+            "GOT.mem for `{}` requires a defined data symbol in the link",
+            symbol_db.symbol_name_for_display(def_id)
+        );
```
davidlattimore, line 3126: Might need to deal with undefined weak symbols here, including adding a test, but let's do that in a separate PR.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: Not handling undefined weak symbols currently causes runtime traps when running sqlite, so I planned to address this in a separate patch.

PR #2272, opened 2026-07-26
port(wasm): Support `--entry` / `--no-entry`

lapla-cogito:
`--no-entry` is commonly seen with targets like wasm32-unknown-unknown.

Additionally, this patch makes `ExpectSym` and `NoSym` test directives for ELF also usable for Wasm. This was necessary to improve the robustness of the tests added in this patch.

Part of #1431

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -3887,15 +3887,15 @@ impl LinkCommand {
                             }
                         }
                         PlatformKind::Wasm => {
-                            // TODO(wasm): Drop once Wild accepts the same flags as wasm-ld.
                             let primary_wat = config
                                 .test_src_dir
                                 .join(format!("{}.wat", config.test_name));
-                            if !linker.is_wild() && primary_wat.exists() {
-                                command
-                                    .arg("--no-entry")
-                                    .arg("--export-all")
-                                    .arg("--no-check-features");
+                            if primary_wat.exists() {
+                                command.arg("--no-entry");
+                                if !linker.is_wild() {
+                                    // TODO(wasm): Support these options
+                                    command.arg("--export-all").arg("--no-check-features");
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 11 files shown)


wild/tests/integration_tests.rs
```diff
@@ -558,6 +560,67 @@ fn wasm_section_names(wasm_file: &Path) -> Result<HashSet<String>> {
     Ok(sections)
 }
 
+fn wasm_export_names(wasm_file: &Path) -> Result<HashSet<String>> {
+    use wasmparser::Parser;
+    use wasmparser::Payload;
+
+    let bytes = std::fs::read(wasm_file)
```
davidlattimore: It seems wasteful that we read the entire file contents multiple times during the test - once for each kind of assertion. We also parse it multiple times, although fixing that would require more refactoring. Presumably, we'd need to read it once, parse it once and build some struct that holds everything we need for all the assertions. It's fine by me if you want to leave this as-is for now, but I think we should do something about it before we add any more kinds of wasm assertions.

wild/tests/sources/wasm/no-entry/no-entry.c
```diff
@@ -0,0 +1,33 @@
...
+//#NoSym: foo
+
+//#Config:entry-then-no-entry
+//#LinkArgs: --entry=custom_entry --no-entry --export=foo
+//#RunEnabled:false
+//#ExpectSym: foo
+//#ExpectSym: memory
+//#NoSym: _start
+//#NoSym: custom_entry
+
```
davidlattimore, line 24: Can you add a test for an `--entry` flag with a symbol that isn't defined?

wild/tests/integration_tests.rs
lapla-cogito: thanks, done

wild/tests/sources/wasm/no-entry/no-entry.c
lapla-cogito: done

davidlattimore (approved):

PR #2273, opened 2026-07-26
fix: Error on non-PIC AArch64 relocations against preemptible symbols in shared objects

deepakshirkem:
Wild silently accepted these AArch64 relocations when linking a shared  object, producing broken output:

- `R_AARCH64_ADD_ABS_LO12_NC`
- `R_AARCH64_LDST8_ABS_LO12_NC`
- `R_AARCH64_LDST32_ABS_LO12_NC`
- `R_AARCH64_LDST64_ABS_LO12_NC`

These relocations store absolute addresses in fixed-size 12-bit fields  that cannot hold runtime-resolved addresses in a shared object.

Issue #2247

### DIFF
--- a/wild/tests/sources/elf/aarch64-fpic-illegal-shared/aarch64-fpic-illegal-shared.s
+++ b/wild/tests/sources/elf/aarch64-fpic-illegal-shared/aarch64-fpic-illegal-shared.s
@@ -3,7 +3,7 @@
 // GNU ld cross linker silently succeeds so we skip it.
 //#Arch:aarch64
 //#Mode:dynamic
-//#SkipLinker:ld
+//#ReferenceLinkers:lld
 //#LinkArgs:-shared --no-gc-sections
 //#ExpectError:recompile with -fPIC
 //#RunEnabled:false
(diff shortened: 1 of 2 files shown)


wild/tests/sources/elf/aarch64-fpic-illegal-shared/aarch64-fpic-illegal-shared.s
```diff
@@ -0,0 +1,18 @@
+// Test that non-PIC AArch64 relocations error in shared objects.
+// Wild errors with "recompile with -fPIC" matching lld behavior.
+// GNU ld cross linker silently succeeds so we skip it.
+//#Arch:aarch64
+//#Mode:dynamic
+//#SkipLinker:ld
```
lapla-cogito: The `SkipLinker` directive is scheduled to be deprecated soon due to the introduction of `ReferenceLinkers`, so it's good to avoid using it. Additionally, based on the original issue being related to lld's test suite, it seems appropriate to set lld as the reference linker (I verified that this edit passes tests locally).

```suggestion
//#ReferenceLinkers:lld
```

deepakshirkem: @lapla-cogito Done. Thank You.

lapla-cogito (approved):
looks good, thanks!

PR #2274, opened 2026-07-26
fix: Error on R_AARCH64_PREL16/32/64 against preemptible symbols in shared objects

deepakshirkem:
Wild was producing an internal `validate_empty` error instead of a proper error message when these relocations were used against preemptible  symbols in shared objects:

- `R_AARCH64_PREL16`
- `R_AARCH64_PREL32`  
- `R_AARCH64_PREL64`

These are PC-relative relocations that store `symbol_address - place`.  In a shared object, the symbol may be preemptible (overridden at runtime by another library), so its address isn't known at link time and can't  be computed correctly.

Issue https://github.com/wild-linker/wild/issues/2247

### DIFF
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -72,7 +72,7 @@ impl crate::platform::Arch for ElfAArch64 {
         })
     }
 
-    fn is_illegal_in_shared_object(r_type: u32) -> bool {
+    fn is_disallowed_for_interposable_symbols(r_type: u32) -> bool {
         matches!(
             r_type,
             object::elf::R_AARCH64_ABS32
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -81,6 +81,9 @@ impl crate::platform::Arch for ElfAArch64 {
                 | object::elf::R_AARCH64_LDST8_ABS_LO12_NC
                 | object::elf::R_AARCH64_LDST32_ABS_LO12_NC
                 | object::elf::R_AARCH64_LDST64_ABS_LO12_NC
+                | object::elf::R_AARCH64_PREL16
+                | object::elf::R_AARCH64_PREL32
+                | object::elf::R_AARCH64_PREL64
         )
     }
 
--- a/wild/tests/sources/elf/aarch64-prel-shared/aarch64-prel-shared.s
+++ b/wild/tests/sources/elf/aarch64-prel-shared/aarch64-prel-shared.s
@@ -6,7 +6,6 @@
 //#ReferenceLinkers:lld
 //#LinkArgs:-shared --no-gc-sections
 //#ExpectError:recompile with -fPIC
-//#RunEnabled:false
 
 .globl foo
 foo:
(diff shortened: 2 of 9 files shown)


davidlattimore (commented):
What about if the symbol is declared as hidden or protected?

wild/tests/sources/elf/aarch64-prel-shared/aarch64-prel-shared.s
```diff
@@ -0,0 +1,16 @@
+// Test that R_AARCH64_PREL32 to a preemptible symbol errors in shared objects.
+// lld errors with "cannot be used against symbol; recompile with -fPIC"
+// GNU ld silently succeeds on this relocation.
+//#Arch:aarch64
+//#Mode:dynamic
+//#ReferenceLinkers:lld
+//#LinkArgs:-shared --no-gc-sections
+//#ExpectError:recompile with -fPIC
+//#RunEnabled:false
```
davidlattimore: Note that ExpectError automatically sets RunEnabled:false.

libwild/src/elf_aarch64.rs
davidlattimore: I see that this method is actually only called when the output is shared and the symbol is interposable.

The method name seems confusing to me, since it implies that listed relocations aren't allowed in shared objects regardless of the symbol visibility. The very similarly named `is_disallowed_in_shared_object` probably adds to the confusion.

Renaming it to something like `is_disallowed_in_shared_object_for_interposable_symbols` would reduce confusion, although it would likely set a new record for the longest function name in our codebase. One related question - are these relocations permitted for externally defined symbols in executables?  Since that feels like a similar situation to symbol interposition - in both cases the symbol is or at least might be provided by a different DSO. If they're not permitted there either, then `is_disallowed_for_interposable_symbols` might be a shorter name - assuming we can consider a dynamic reference to a symbol provided by a DSO as sort of interposition - it's interposing an undefined symbol.

wild/tests/sources/elf/aarch64-prel-shared/aarch64-prel-shared.s
deepakshirkem: Yes. Thank You

libwild/src/elf_aarch64.rs
deepakshirkem: Renamed the function to is_disallowed_for_interposable_symbols as you suggested.

>One related question - are these relocations permitted for externally defined symbols in executables? Since that feels like a similar situation to symbol interposition - in both cases the symbol is or at least might be provided by a different DSO.

When I tried removing is_position_independent() from the check in elf_writer.rs, it broke existing tests because _ITM_deregisterTMCloneTable has ABSOLUTE|DYNAMIC flags but is_interposable() returns true for it (NON_INTERPOSABLE is not set for absolute DSO symbols). Could you guide me on how to handle absolute DSO symbols correctly so we can extend the check to non-PIE executables too?

davidlattimore: Sounds like something that we should probably handle in a follow-up PR even if it means renaming the method again in that PR. I'd say that we'd need to do some experiments to see how both GNU ld and lld handle relocations against absolute dynamic symbols.

davidlattimore (approved):

PR #2275, opened 2026-07-27
fix: Avoid applying ELF-specific entry behaviour to other platforms

davidlattimore:
For ELF, we duplicate some of the dubious behaviours of GNU ld - e.g. if the entry symbol isn't defined, we warn, then use the start of the .text section or otherwise 0. We don't want to duplicate such behaviours to other platforms, so this PR moves this behaviour to be ELF-specific. In order to properly test the entry behaviour of the other platforms I needed to implement the `-e` flag for mac. A bunch of new tests were added to check the corner cases around entry handlings. Many of these fail without the rest of the change.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -440,7 +440,7 @@ fn populate_file_header<A: Arch<Platform = Elf>>(
     header.e_type.set(e, ty);
     header.e_machine.set(e, A::arch_identifier());
     header.e_version.set(e, object::elf::EV_CURRENT.0.into());
-    header.e_entry.set(e, layout.entry_symbol_address()?);
+    header.e_entry.set(e, elf_entry_address(layout)?);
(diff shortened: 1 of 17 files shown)


libwild/src/elf_writer.rs
```diff
@@ -486,6 +486,40 @@ fn populate_file_header<A: Arch<Platform = Elf>>(
...
+    let text_layout = layout.section_layouts.get(output_section_id::TEXT);
+    if text_layout.mem_size == 0 {
+        layout.symbol_db.warning(format!(
+            "cannot find entry symbol `{entry_name}` and .text is empty, not setting entry point"
+        ));
+        return Ok(0);
+    }
+
+    layout.symbol_db.warning(format!(
+        "cannot find entry symbol `{entry_name}`, defaulting to 0x{}",
```
```suggestion
        "cannot find entry symbol `{entry_name}`, defaulting to 0x{:x}",
```

PR #2276, opened 2026-07-27
ci: add Arch Linux cross CI image

marxin:
There are a new CI failures related to updated GCC 16.1 aarch64 cross compiler and thus I would like to have the platform in CI.
Note the `loongarch64` cross compiler in unavailable on the Arch distro.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -28,6 +28,10 @@ jobs:
             container: ci-alpine-amd64
             runs-on: ubuntu-24.04
             platform: linux/amd64
+          - file: docker/ci/arch.Dockerfile
+            container: ci-arch-cross-amd64
+            runs-on: ubuntu-24.04
+            platform: linux/amd64
           - file: docker/ci/opensuse.Dockerfile
             container: ci-opensuse-amd64
             runs-on: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


marxin:
Am I right I'll need 2 PRs, first one for the image addition (then Build images can be run) and the second one for adding the CI job (with the generated hash of the new images)?

mati865:
Yes, and building the image requires a manual trigger.

mati865:
If you don't mind waiting a bit, I can modify the workflow to run on PRs like this one, but without pushing the images.

marxin:
> If you don't mind waiting a bit, I can modify the workflow to run on PRs like this one, but without pushing the images.

Sure, will appreciate that!

.github/workflows/build-images.yml
```diff
@@ -28,6 +28,10 @@ jobs:
             container: ci-alpine-amd64
             runs-on: ubuntu-24.04
             platform: linux/amd64
+          - file: docker/ci/arch.Dockerfile
```
davidlattimore, line 31: I don't see this file, did you forget to add it?

davidlattimore (commented):
Sorry it's a bit of a painful process to add new containers and hook them up. My approach when testing was to trigger image builds on my fork, then update to the newly built images and run CI.

.github/workflows/build-images.yml
marxin: Yeah, sorry.

davidlattimore (approved):

PR #2277, opened 2026-07-27
ci: test-build images for PRs

mati865:
Every time I interact with GitHub Actions they prove to be insaner than I previously though. I used LLMs to make conditional `cache-to` and still could not come up with anything sane.
My idea is to make sure the images build (and in the future pass tests) on PR, before merging changes to them.

This got complicated a lot with base images being built in a distinct job. For PRs we really don't want to push untrusted stuff, so I followed https://docs.docker.com/build/ci/github-actions/share-image-jobs/ on how to move them between jobs.

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -89,8 +89,8 @@ jobs:
         if: github.event_name == 'pull_request'
         uses: actions/upload-artifact@v4
         with:
-          name: myimage
-          path: ${{ runner.temp }}/myimage.tar
+          name: base_image
+          path: ${{ runner.temp }}/base_image.tar
 
   # Build images that depend on other images.
   build-dependent:
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -76,7 +76,7 @@ jobs:
           cache-to: ${{ github.event_name != 'pull_request' && format('type=registry,ref=ghcr.io/{0}/{1}:buildcache,mode=max', github.repository, matrix.container) || '' }}
 
       - name: Export base image
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && contains(matrix.file, 'base')
         uses: docker/build-push-action@v7
         with:
           context: .
@@ -86,7 +86,7 @@ jobs:
           outputs: type=docker,dest=${{ runner.temp }}/base_image.tar
 
       - name: Upload base image archive
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && contains(matrix.file, 'base')
         uses: actions/upload-artifact@v4
         with:
           name: base_image-${{ matrix.container }}
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -79,6 +79,9 @@ jobs:
         if: github.event_name == 'pull_request'
         uses: docker/build-push-action@v7
         with:
+          context: .
+          file: ${{ matrix.file }}
+          platforms: ${{ matrix.platform }}
           tags: ghcr.io/${{ github.repository }}/${{ matrix.container }}:latest
           outputs: type=docker,dest=${{ runner.temp }}/base_image.tar
 
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -92,6 +92,7 @@ jobs:
           name: base_image-${{ matrix.container }}
           path: ${{ runner.temp }}/base_image.tar
           retention-days: 1
+          if-no-files-found: error
 
   # Build images that depend on other images.
   build-dependent:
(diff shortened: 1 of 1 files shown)


mati865:
Kudos to Docker for making loading local images so hard with buildx...

davidlattimore:
My approach to testing the images was to trigger the build workflow on my fork of the repo, then update the fork to use the newly built images. So that always an option if this turns out to be too complex.

mati865:
It was definitely annoying to get there, but I think it's not that bad once it works:
- diff of using `docker` driver for PR runs: https://github.com/wild-linker/wild/compare/main...mati865:wild:4e0bc11d50ab1d0f96c2fb9b31617354159c1026
- diff of using build-context to keep using `docker-container` driver: https://github.com/wild-linker/wild/compare/main...mati865:wild:1d8226c0e86f8ae82950246f50257044e1ae2d4c

Diff between the two approaches: https://github.com/wild-linker/wild/compare/1d8226c0e86f8ae82950246f50257044e1ae2d4c...mati865:wild:4e0bc11d50ab1d0f96c2fb9b31617354159c1026

Neither is particularly beautiful, but I cannot tell which is better. I'll leave it up to you, @marxin and @lapla-cogito to decide whether we care enough about testing the images on CI, and if so, which way to go.

There are also other possibilities like running a local registry, or creating dedicated workflow only for Docker images testing, but having spent hours on "15 minutes job" I'm not too keen on exploring them myself.

marxin:
Don't have a strong opinion about it - based on the current experience, using the 2 step approach works quite fine for me.
So leaving that up to you if the complexity overweight the benefits.

davidlattimore (commented):
Thanks for working on this! I agree that more needs to be done to make updating images easier. My original intention was to see how things went with our first image update and refine the process from there. We probably at least need some docs about how to do an update.

One possible concern is that although this verifies that the images build with a changed configuration, it doesn't verify that CI would pass with those new images. Testing that automatically seems impractical - or at least I have no idea how to do it. So if someone is making a substantial change to the images and they want to make sure that change is thoroughly tested, they likely need to do that testing in their own fork of the repo first anyway or at least test it as best they can on their local machine.

Another slight issue is that any change to any of the docker images will trigger all of them to be rebuilt, even if we just fixed a typo in a comment. This could be helped by finding ways to ensure that the resulting images have a short retention period, so that we're at least not keeping them around.

Anyway, I'm happy for this to go in if it's wanted and others find it useful - at least once retention periods are looked at.

mati865:
> One possible concern is that although this verifies that the images build with a changed configuration, it doesn't verify that CI would pass with those new images. Testing that automatically seems impractical - or at least I have no idea how to do it. So if someone is making a substantial change to the images and they want to make sure that change is thoroughly tested, they likely need to do that testing in their own fork of the repo first anyway or at least test it as best they can on their local machine.

I'm not sure if we can preform through testing without duplicating other pipelines. Or at least comparing them to conditionally depend on this one.
We could easily do basic tests, though. I didn't add them here to avoid failing Arch Linux build.

> Another slight issue is that any change to any of the docker images will trigger all of them to be rebuilt, even if we just fixed a typo in a comment.

I thought the changes in this area would be rare enough not to care about it.

> This could be helped by finding ways to ensure that the resulting images have a short retention period, so that we're at least not keeping them around.

Do you mean retention for the archives for the base images that are shared between steps? Note that, I disabled pushing of the images for pull requests, although for security concerns. So, they are not kept.

.github/workflows/build-images.yml
```diff
@@ -61,13 +67,30 @@ jobs:
...
+          platforms: ${{ matrix.platform }}
+          tags: ghcr.io/${{ github.repository }}/${{ matrix.container }}:sha-${{ github.sha }}
+          outputs: type=docker,dest=${{ runner.temp }}/base_image.tar
+
+      - name: Upload base image archive
+        if: github.event_name == 'pull_request' && contains(matrix.file, 'base')
+        uses: actions/upload-artifact@v4
+        with:
+          name: base_image-${{ matrix.container }}
+          path: ${{ runner.temp }}/base_image.tar
```
davidlattimore, line 93: > > This could be helped by finding ways to ensure that the resulting images have a short retention period, so that we're at least not keeping them around.
> 
> Do you mean retention for the archives for the base images that are shared between steps? Note that, I disabled pushing of the images for pull requests, although for security concerns. So, they are not kept.

Won't `Upload base image archive` get run and cause the base image to be kept around for whatever the default retention period is? https://github.com/actions/upload-artifact/blob/main/README.md#retention-period

mati865: That's true. By default it's 90 days, and looks like we cannot limit it more than to a single day.

davidlattimore (approved):

lapla-cogito (approved):

mati865:
> My original intention was to see how things went with our first image update and refine the process from there. We probably at least need some docs about how to do an update.

Nothing really changes about updating the images here. They still require manual trigger to be built and pushed, so I'd argue this is just a step towards making it easier, but the process still remains the same.

PR #2278, opened 2026-07-27
port(wasm): change default to --stack-first

marxin:
I noticed some failure tests on my Arch Linux system and apparently, starting with LLVM 22
(https://github.com/llvm/llvm-project/pull/166998), the default got adjusted to `--stack-first`. Given that, I am changing the default.


lapla-cogito (approved):
Ah, I hadn't noticed that, thanks! Regarding the PR title, while there's a discussion thread on Zulip about whether to use prefixes like "feat" for ports stuff, we currently use "port(wasm)" for all cases.

marxin:
> Zulip about whether to use prefixes like "feat" for ports stuff, we currently use "port(wasm)" for all cases.

Sure, let's use it then ;)

PR #2279, opened 2026-07-27
ci: build ubuntu-cross image natively

mati865:
It's a typo from https://github.com/wild-linker/wild/pull/2046


davidlattimore (approved):
Thanks for noticing it :)

PR #2282, opened 2026-07-28
test: add testing on Arch Linux

marxin:


marxin:
Blocked by #2283.

.github/workflows/ci.yml
davidlattimore: Do you think we should update all the other jobs to use the new images too? That way we're not running with a mix of images from two different builds. I did try to find a way to define the image version just once in the file, but never really found a good way.

marxin: Sure, adjusted.

marxin: Tried researching a bit, and it seems there's no easy way how to define a reusable constant.

davidlattimore (approved):

mati865 (approved):

PR #2292, opened 2026-07-29
feat: evaluate the modulo operator in linker scripts

vishruth-thimmaiah:
Implement the modulo operator(`%`) in arithmetic operations in linker scripts.

### DIFF
--- a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
+++ b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
@@ -14,7 +14,14 @@ ASSERT(1 == 1, "1 == 1 must always be true");
 
 /* Arithmetic assertion */
 ASSERT(1 + 1 == 2, "arithmetic must work");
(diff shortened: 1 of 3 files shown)


wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -14,6 +14,7 @@ ASSERT(1 == 1, "1 == 1 must always be true");
 
 /* Arithmetic assertion */
 ASSERT(1 + 1 == 2, "arithmetic must work");
+ASSERT(12 % 10 == 2, "modulo must work");
```
davidlattimore: Can you also test that we have the same behaviour as the other linkers with regard to negative numbers?

vishruth-thimmaiah: Fixed, I've modified the PR to use `wrapping_rem` instead, I also fixed division with negative numbers.

davidlattimore (approved):

PR #2296, opened 2026-07-30
feat: implement `SORT_BY_NAME` and `SORT_BY_ALIGNMENT`

vishruth-thimmaiah:
`SORT_BY_NAME` is an alias of `SORT`, and sorts by section names.
`SORT_BY_ALIGNMENT` sorts sections in decreasing order of alignment. Since wild already does this by default, it is ignored.

### DIFF
--- a/LINKER_SCRIPT_SUPPORT.md
+++ b/LINKER_SCRIPT_SUPPORT.md
@@ -53,7 +53,9 @@ end lists the features required to link the Linux kernel.
 | `=fillexp` | ✅ | |
 | `AT(addr)` load-address specifier on output sections | ✅ | |
 | Numeric address between section name and `:` (e.g. `name 0 : { ... }`) | 🧪 | Only numeric literals are currently supported |
-| `SORT_BY_NAME(...)`, `SORT_BY_ALIGNMENT(...)`, `SORT_BY_INIT_PRIORITY(...)` | 📅 | |
+| `SORT(...)`, `SORT_BY_NAME(...)` | ✅ | |
+| `SORT_BY_ALIGNMENT(...)` | ✅ | Parsed and ignored, as sections are sorted by alignment by default |
+| `SORT_BY_INIT_PRIORITY(...)` | 📅 | |
 | `EXCLUDE_FILE(...)` inside input section matchers | 📅 | |
 | `BYTE(expr)`, `SHORT(expr)`, `LONG(expr)`, `QUAD(expr)` output data | ❌ | |
 | `SUBALIGN(n)` forced input alignment | ❌ | |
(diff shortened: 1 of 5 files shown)


LINKER_SCRIPT_SUPPORT.md
```diff
@@ -119,7 +121,8 @@ see at a glance what remains before Wild can link the kernel.
 | `=fillexp` | ✅ | |
 | `AT(addr)` load-address specifier on output sections | ✅ | |
 | `>region` and `AT>region` memory region placement | 📅 | |
-| `SORT_BY_NAME(...)`, `SORT_BY_ALIGNMENT(...)`, `SORT_BY_INIT_PRIORITY(...)` | 📅 | |
```
davidlattimore, line 122: I assume you investigated and found that `SORT_BY_INIT_PRIORITY` isn't used by the kernel?

davidlattimore (commented):
How do these sort directives interact with the flag `--sort-section`? Should we add a test? If it doesn't work, I'm open to deferring fixing it until later if you like.

libwild/src/linker_script.rs
```diff
@@ -1315,32 +1276,29 @@ fn parse_matcher_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<Mat
 }
 
 fn parse_pattern<'input>(input: &mut &'input BStr) -> winnow::Result<SectionPattern<'input>> {
```
davidlattimore: Can sort directives be nested? It seems like it would be an obscure thing to do, so I'm fine with (a) assuming that won't happen (b) parse it and give an error saying that it's not supported or (c) if it's not too much work (seems unlikely), then actually supporting it.

vishruth-thimmaiah: Yes, two sorting commands can be defined, specifying a primary sort and secondary sort, i.e. `SORT_BY_ALIGNMENT(SORT_BY_NAME(...))` will sort the input sections by alignment first, then by name if two sections have the same alignment. For now I am just parsing and returning an error.

LINKER_SCRIPT_SUPPORT.md
vishruth-thimmaiah: Yes, only sorting by name and alignment are used by the kernel.

vishruth-thimmaiah:
If only a single sort command is defined in the linker script, the `--sort-sections` flag acts as the secondary sort, i.e. `SORT_BY_NAME(...)` with `--sort-sections=alignment` becomes `SORT_BY_NAME(SORT_BY_ALIGNMENT(...))`. I suppose we would first have to support two sorting commands for a section to implement support for `--sort-sections`.

davidlattimore (approved):

PR #2297, opened 2026-07-30
refactor:  Split ELF relocation GC into classify, materialize, and edges

lapla-cogito:
Split the existing ELF GC relocation path into smaller steps so other formats (especially Wasm) can reuse the edge-discovery half without taking on ELF materialize (GOT/PLT/sizes) or the section work queue.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -5614,175 +5614,270 @@ fn process_relocation<'data, 'scope, A: Arch<Platform = Elf>, R: Relocation<Plat
     scope: &Scope<'scope>,
     relr_writer: &mut RelrEncoder,
 ) -> Result<RelocationModifier> {
+    let Some(local_sym_index) = rel.symbol() else {
+        return Ok(RelocationModifier::Normal);
+    };
+
+    let mut classified =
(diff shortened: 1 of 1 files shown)


davidlattimore:
I haven't looked in detail at the changes yet, but I did run some benchmarks. I'm seeing about a 2.5% slowdown when linking wild for ELF. If I drill in on the "Find required sections" phase, I see about a 16% increase in wall time, 51% increase in cycles and more than 70% increase for instructions, cache misses and branch misses.

lapla-cogito:
> I haven't looked in detail at the changes yet, but I did run some benchmarks. I'm seeing about a 2.5% slowdown when linking wild for ELF. If I drill in on the "Find required sections" phase, I see about a 16% increase in wall time, 51% increase in cycles and more than 70% increase for instructions, cache misses and branch misses.

My initial measurements didn't show such a significant impact (though a bit regression). Anyway I decided to reconsider the abstraction approach and limit this PR to simply breaking down the existing ELF GC behavior.

davidlattimore (approved):
Thanks. I reran a benchmark and no longer see a performance loss.

PR #2307, opened 2026-08-01
port(wasm): Resolve undefined weak functions to unreachable stubs

lapla-cogito:
Synthesize a shared `unreachable` stub for direct calls and use table index 0 for null function-pointer checks. This prevents invalid WASI binaries from being created and triggering runtime traps. I've confirmed that this change enables SQLite to run correctly at runtime.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2329,6 +2329,9 @@ impl WasmObjectIndexMap {
                     "function {func_out} has no indirect table slot"
                 );
                 if reloc.ty == reloc_type::TABLE_INDEX_REL_SLEB {
+                    if slot == 0 {
+                        return Ok(0);
+                    }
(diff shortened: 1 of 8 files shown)


davidlattimore (commented):
How about undefined weak TBREL relocations?

```
	.functype	optional () -> ()
	.weak	optional

	.globl	_start
_start:
	.functype	_start () -> ()
	i32.const	optional@TBREL
	drop
	end_function
```

lapla-cogito:
> How about undefined weak TBREL relocations?
> 
> ```
> 	.functype	optional () -> ()
> 	.weak	optional
> 
> 	.globl	_start
> _start:
> 	.functype	_start () -> ()
> 	i32.const	optional@TBREL
> 	drop
> 	end_function
> ```

Ah right, we do need to address that. Thanks.

davidlattimore (approved):

PR #2308, opened 2026-08-01
port(wasm): Emit `target_features` custom section from linked objects

lapla-cogito:
Parse and union used features across inputs, reject used/disallowed conflicts, and write a sorted `target_features` section. Also see https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#target-features-section.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -1946,7 +1946,9 @@ fn build_target_features_section<'data>(
     }))
 }
 
-fn parse_target_features_payload<'data>(data: &'data [u8]) -> Result<Vec<WasmTargetFeature<'data>>> {
+fn parse_target_features_payload<'data>(
+    data: &'data [u8],
+) -> Result<Vec<WasmTargetFeature<'data>>> {
     let mut reader = BinaryReader::new(data, 0);
     let count = reader
         .read_var_u32()
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -5729,6 +5729,114 @@ mod tests {
     use super::*;
     use crate::args::wasm::DEFAULT_STACK_SIZE;
 
+    fn layout_input_with_features<'data>(
+        file: u32,
+        features: &'data [WasmTargetFeature<'data>],
+    ) -> WasmObjectLayoutInput<'data> {
+        WasmObjectLayoutInput {
+            data: &[],
+            types: Vec::new(),
+            function_imports: Vec::new(),
+            global_imports: Vec::new(),
+            memory_imports: Vec::new(),
+            table_imports: Vec::new(),
+            module_functions: Vec::new(),
+            globals: Vec::new(),
+            exports: Vec::new(),
+            function_bodies: Vec::new(),
+            memories: Vec::new(),
+            unsupported_output: Vec::new(),
+            code_relocations: Vec::new(),
+            data_segments: Vec::new(),
+            segment_alignments: Vec::new(),
+            data_relocations: Vec::new(),
+            symbols: &[],
+            init_funcs: &[],
+            target_features: features,
+            symbol_id_range: crate::symbol_db::SymbolIdRange::empty(),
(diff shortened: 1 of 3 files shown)


lapla-cogito:
With this patch, I've confirmed that linking CRuby with Wild now produces the same set of `target_features` as wasm-ld.

```
（▰╹◡╹）❯  wasm-objdump -x /tmp/wild-ruby-out/ruby.wild.wasm
...
Custom:
 - name: "target_features"
  - [+] bulk-memory
  - [+] multivalue
  - [+] mutable-globals
  - [+] reference-types
  - [+] sign-ext

（▰╹◡╹）❯  wasm-objdump -x /tmp/wild-ruby-out/ruby.lld.wasm
...
Custom:
 - name: "target_features"
  - [+] bulk-memory
  - [+] multivalue
  - [+] mutable-globals
  - [+] reference-types
  - [+] sign-ext
```

libwild/src/wasm.rs
```diff
@@ -1858,6 +1888,89 @@ fn wasm_symbol_name_str<'data>(data: &'data [u8], sym: &WasmSymbol) -> Option<&'
...
+                        feature.name
+                    );
+                }
+            }
+        }
+    }
+
+    for name in &used {
+        if let Some(&file_id) = disallowed.get(name) {
+            bail!(
```
davidlattimore, line 1920: Should we have a test for this? At the moment it seems that commenting this out doesn't cause any failures.

libwild/src/wasm.rs
```diff
@@ -1858,6 +1888,89 @@ fn wasm_symbol_name_str<'data>(data: &'data [u8], sym: &WasmSymbol) -> Option<&'
     core::str::from_utf8(bytes).ok()
 }
 
+/// Merge `target_features` from linked objects and encode the output custom section.
+fn build_target_features_section<'data>(
+    layout_inputs: &[WasmObjectLayoutInput<'data>],
+) -> Result<Option<wasm_encoder::CustomSection<'static>>> {
+    let mut used: HashSet<&'data str> = HashSet::new();
```
davidlattimore, line 1895: Do you think it's worth having a test that we deduplicate features between files? At the moment I can change this to a Vec and the tests still pass.

lapla-cogito:
Thanks, both issues are now covered by ~~integration~~ (typo: unit) tests.

davidlattimore (commented):
Generally I prefer integration tests over unit tests for a couple of reasons. Firstly, they generally can survive when the code under test gets substantially refactored, whereas unit tests often either need rewriting or even deleted when the code gets refactored. Secondly, because integration tests run against other linkers, they give us more assurance that the behaviour we're testing for actually matches the other linkers. Unit tests have their place - e.g. when it's not practical to test something with an integration test for some reason. Is there any reason these two tests can't be integration tests?

lapla-cogito:
> Is there any reason these two tests can't be integration tests?

We need hand-written `target_features` payloads if we verify them via integration tests. With WAT that means `@custom "target_features"`. However, `wat2wasm` places those customs before the synthetic linking section it appends, and wasm-ld rejects the object (`out of order section type: 0`). Who is right is still unclear to me but you can check with:

```wat
;; wat2wasm a.wat --relocatable --enable-annotations -o a.o
;; wasm-ld -m wasm32 --no-entry --export-all a.o -o a.wasm
;; # error: out of order section type: 0

(module
  (func (export "f") (result i32) i32.const 1)
  ;; +sign-ext
  (@custom "target_features" "\01\2b\08\73\69\67\6e\2d\65\78\74")
)
```

So verifying them via integration tests would need hacks (reorder customs in the harness, or commit prebuilt `.o`s) I think. Of course, we could restrict the validation to just Wild using `ReferenceLinkers`, but ultimately this wouldn't be much different from what unit tests are already doing.

davidlattimore (approved):
Makes sense. Thanks for the explanation :)

PR #2309, opened 2026-08-01
fix: OR GNU property notes within file before ANDing across files

deepakshirkem:
`merge_gnu_property_notes` was ANDing all property entries together, including multiple entries within the same file.

A file with separate BTI (0x1) and PAC (0x2) entries was merged as: `0x1 & 0x2 = 0x0` → filtered out → no `.note.gnu.property` in output!

- Within a file: OR the bits (accumulate features)
- Across files: AND the bits (intersection of features)

Also fixes `.note.gnu.property` not being output in `-r` mode, sinc the zero value was being filtered out.

Issue https://github.com/wild-linker/wild/issues/2247


wild/tests/sources/elf/aarch64-gnu-property-merge/aarch64-gnu-property-merge.s
davidlattimore, line 1: It looks like this test passes without the fix. Can you see what needs changing with the test so that it fully exercises the expected behaviour?

davidlattimore (commented):
Ca

wild/tests/sources/elf/aarch64-gnu-property-merge/aarch64-gnu-property-merge.s
deepakshirkem: You're right! Fixed. I updated the test to exercise the expected behavior and also updated the commit message.

wild/tests/sources/elf/aarch64-gnu-property-merge/aarch64-gnu-property-merge-input.s
```diff
@@ -0,0 +1,21 @@
+.text
+ret
+// Single .note.gnu.property section with TWO property entries in one note.
+// This triggers the bug where Wild ANDs BTI(1) & PAC(2) = 0 instead of OR-ing.
```
davidlattimore: We always want comments to be written so that make sense to someone reading them after the PR is merged. Once the PR is merged, wild shouldn't have that bug, so the comment could be misleading. I'd suggest wording like "This verifies that the linker does ___ not ___"

davidlattimore (commented):
Did you try the new version of your test without the fix? I just tried it and it still passes for me.

wild/tests/sources/elf/aarch64-gnu-property-merge/aarch64-gnu-property-merge-input.s
deepakshirkem: Updated the comment as suggested. I also updated the test and verified that it fails on upstream/main and passes with this change. Please take another look when you get a chance?

davidlattimore (approved):
Thanks :)

marxin:
Thank you for the fix ;)

PR #2311, opened 2026-08-01
chore: update object dependency

philipc:
There's a few breaking changes. Most of this commit is handling the change from `u32` to `object::elf::RelocationType`, but also:

- `elf::VersymIndex::new` replaced with `elf::VersionIndex::versym`
- `read::macho::SectionHeader::data` now takes the section offset (fix handles 32-bit file offsets only, which is fine for object files)


davidlattimore (approved):
Thanks!

Issue #2248, opened 2026-07-22
Mach-O: Identify sections by segment name and section name


joanna-jo:
Mach-O sections are identified by (segment name, section name), as also seen in [lld](https://github.com/llvm/llvm-project/blob/9b0e63af7c1026d7f4a1550753b7e1294af46d29/lld/MachO/OutputSegment.cpp#L94-L96) and [dyld](https://github.com/apple-oss-distributions/dyld/blob/fd8d0c4d52320ebf64db34f3cb280310d905c5ae/mach_o/UnsafeHeader.cpp#L2679-L2698). Currently, in our implementation, we resolve Mach-O input sections by only the section name, then map each resulting `OutputSectionId` to a fixed output segment. This means that sections with the same section name but different segment names are coalesced together, such as `__TEXT, __const` and `__DATA_CONST, __const`.

I was looking at #1795 (cc: @marxin) and saw that the supported segments and sections were made intentionally strict though - did we want to try refactoring this now?

davidlattimore:
Interesting. I didn't know that MachO did this. I don't see any reason why this would be better done later rather than sooner, so from my perspective, I'd say go ahead :) If there are preparatory refactorings that would help, feel free to do those first. It sounds like a challenging change to make. Good luck!

marxin:
Bery good observation @joanna-jo! And yes, we should do the proper (segment,section) assignment as soon as possible ;) If you need a help, just ping me.

joanna-jo:
Sounds great, thanks David and Martin! :) I'll get started working on this then.

joanna-jo:
Hmmm just to clarify, did we want to support section identification with the fixed segments we currently have, or did we want to move onto fully supporting dynamic segment (and section) identification? I was working on a design for the latter, but I realized the way I worded this could be a little vague. If we meant the former, I can shift gears and maybe make a separate issue.

davidlattimore:
Just so I understand, the two approaches you are talking about are:

1) Map segment name to segment ID then map (segment ID, section name) to OutputSectionId
2) Map (segment name, section name) to OutputSectionId

Is that right? Of those, the second sounds more correct, so it sounds like you're probably on the right track.

joanna-jo:
The second option is right for section identification, but I was wondering about the adjacent problem of whether we should support arbitrary segment names:

1) Should we only support the fixed set of output segment names we currently recognize and reject all other names?
[...]

PR #2312, opened 2026-08-02
port(MachO): Introduce `SectionIdentity`

joanna-jo:
Introduces `SectionIdentity`, which contains `SectionName` and a new `P::SectionIdentityExt` that provides format-specific section identification information. This replaces our previous section name-based identification and is in preparation for introducing (segment name, section name) section identification for Mach-O.

Part of #2248

### DIFF
--- a/libwild/src/platform.rs
+++ b/libwild/src/platform.rs
@@ -28,6 +28,7 @@ use crate::output_section_id::CustomSectionIds;
 use crate::output_section_id::OutputOrder;
 use crate::output_section_id::OutputSectionId;
 use crate::output_section_id::OutputSections;
+use crate::output_section_id::SectionIdentity;
 use crate::output_section_id::SectionName;
 use crate::output_section_map::OutputSectionMap;
 use crate::output_section_part_map::OutputSectionPartMap;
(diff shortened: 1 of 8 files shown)


libwild/src/platform.rs
```diff
@@ -950,6 +954,29 @@ pub(crate) trait Platform:
...
+    /// Constructs the complete identity of an input section, including all format-specific fields
+    /// that distinguish sections with the same name.
+    fn section_identity<'data>(
+        name: SectionName<'data>,
+        section: &Self::SectionHeader,
+    ) -> SectionIdentity<'data, Self>;
+
+    /// Constructs a section identity with only the section name.
+    /// Returns None when the name alone cannot determine the complete section identity.
+    fn section_identity_from_name<'data>(
```
davidlattimore, line 967: I feel like we might at some point want a different abstraction here, but at this stage I can't think of what it should be. The platforms where this function needs to be called are the ELF and likely eventually Windows, neither of which needs the platform-specific component of SectionIdentity. Anyway, I think this is fine for now, since as I mentioned, I can't think of an alternative.

davidlattimore (approved):

libwild/src/platform.rs
joanna-jo: I was thinking the same, that this is a bit awkward to have as a workaround. I also couldn't think of anything better right now, but I'll try to see if there's a cleaner way to do this while working on the rest of the issue.

PR #2313, opened 2026-08-02
port(wasm): Coalesce identical host imports across objects

lapla-cogito:
Unresolved function and global imports are now merged by module/name so each unique host import appears once in the output. Previously each object that referenced the same import got its own import row.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2929,8 +2929,6 @@ enum ImportResolution {
 struct ObjectImportResolutions {
     function_resolutions: Vec<ImportResolution>,
     global_resolutions: Vec<ImportResolution>,
-    unresolved_function_count: u32,
-    unresolved_global_count: u32,
 }
 
 #[derive(Debug, Clone)]
@@ -3130,7 +3128,7 @@ fn resolve_cross_object_imports<'data>(
         .par_iter()
         .map(|input| {
             verbose_timing_phase!("Resolve Wasm object imports");
-            let (function_resolutions, unresolved_function_count) = resolve_import_symbols(
+            let function_resolutions = resolve_import_symbols(
                 input.function_imports.len(),
                 WasmSymbolKind::Func,
                 input,
@@ -3138,7 +3136,7 @@ fn resolve_cross_object_imports<'data>(
                 symbol_db,
                 file_id_to_index,
             )?;
-            let (global_resolutions, unresolved_global_count) = resolve_import_symbols(
+            let global_resolutions = resolve_import_symbols(
                 input.global_imports.len(),
                 WasmSymbolKind::Global,
                 input,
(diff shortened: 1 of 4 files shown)


lapla-cogito:
Binaryen's Asyncify pass renames these `asyncify.*` imports to fixed internal names. This patch enables running  CRuby linked with Wild on wasmtime. Asyncify is [applied](https://github.com/ruby/ruby/blob/3ac5ce973ddc0c45ba3357b4d364844a7edf92a5/wasm/GNUmakefile.in#L30) after the linking process is complete.

```
（▰╹◡╹）❯  wasmtime ruby.wild.wasm -e 'puts 3+5'
8

（▰╹◡╹）❯  cat test.rb
print "Enter your name: "
name = gets.chomp

puts "Hello, #{name}!"

（▰╹◡╹）❯  wasmtime --dir=/tmp ./ruby.wild.wasm /tmp/wild-ruby-out/test.rb
Enter your name: aaa
Hello, aaa!

（▰╹◡╹）❯  wasmtime ruby.wild.wasm --version
ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [wasm32-wasi]
```

lapla-cogito:
Additionally, I'd like to share some performance metrics regarding link time performance. Below are the comparison results with wasm-ld. Note that, since Wild has not yet implemented section GC and debug information handling, I passed `--no-gc-sections` (`--gc-sections` is default for wasm-ld unlike lld) and `--strip-debug` options to wasm-ld. These results are from measurements conducted with CRuby, but similar wall time trends are observed with other software under test (including SQLite, ripgrep, etc.).

```
Benchmark 1: wild
  Time (mean ± σ):      36.9 ms ±   2.6 ms    [User: 33.5 ms, System: 29.1 ms]
  Range (min … max):    32.0 ms …  43.6 ms    86 runs

Benchmark 2: wasm-ld --strip-debug --no-gc-sections
  Time (mean ± σ):     132.1 ms ±   4.2 ms    [User: 100.5 ms, System: 36.2 ms]
  Range (min … max):   122.3 ms … 140.9 ms    22 runs

Summary
  wild ran
    3.58 ± 0.28 times faster than wasm-ld --strip-debug --no-gc-sections
```

libwild/src/wasm.rs
```diff
@@ -3792,14 +3957,6 @@ impl LinkerDefinedIndices {
             needs_table_base |= absorption.needs_table_base;
             needs_stack_pointer |= absorption.needs_stack_pointer;
             needs_tls_base |= absorption.needs_tls_base;
-
-            function_import_count = function_import_count
```
davidlattimore, line 3796: Are unresolved_function_count and unresolved_global_count still needed?

lapla-cogito: Didn't notice that, thanks.

davidlattimore (approved):

PR #2314, opened 2026-08-02
feat: implement `FILEHDR` and `PHDRS` commands in a linker script

vishruth-thimmaiah:
The `FILEHDR` and `PHDRS` commands are used within the PHDRS table in a linker script to specify which segments include the elf header and program header tables respectively. Example:
```ld
PHDRS {
  phdr PT_PHDR PHDRS;
  text PT_LOAD FILEHDR PHDRS;
  data PT_LOAD;
}
```


libwild/src/elf.rs
```diff
@@ -2329,15 +2329,25 @@ impl platform::Platform for Elf {
...
             num_phdrs += script.parsed.program_headers.len();
             for phdr in &script.parsed.program_headers {
                 let ptype = expression_eval::evaluate_const(&phdr.ptype)? as u32;
+                if ptype == pt::LOAD.0 {
+                    if (phdr.has_filehdr || phdr.has_phdrs) && load_without_hdrs {
+                        bail!(
+                            "PHDRS and FILEHDR are not supported when prior PT_LOAD headers lack them"
+                        );
+                    } else {
+                        load_without_hdrs = true;
```
lapla-cogito, line 2350: Would this accidentally bail if there were multiple `LOAD`s with headers?

libwild/src/elf.rs
```diff
@@ -2453,22 +2471,48 @@ impl platform::Platform for Elf {
...
-                    builder.push_event(OrderEvent::SegmentStart(ProgramSegmentId::new(seg_idx)));
+            if pos == 0 {
+                for (seg_idx, _) in starts.iter().enumerate().take(num_phdrs) {
+                    let entry = segment_entries[seg_idx];
+                    if entry.has_filehdr {
+                        builder
+                            .push_event(OrderEvent::SegmentStart(ProgramSegmentId::new(seg_idx)));
+                    }
+                }
+                builder.push_event(OrderEvent::Section(crate::output_section_id::FILE_HEADER));
```
lapla-cogito, line 2509: Segments with `has_filehdr` start before `FILE_HEADER` but never end before `PROGRAM_HEADERS`.

lapla-cogito (commented):
Also, shouldn't `LINKER_SCRIPT_SUPPORT.md` be updated?

wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.c
```diff
@@ -37,6 +37,19 @@
 //#LinkArgs:-shared -z now -T ./linker-script-phdrs-no-load.ld
 //#ExpectErrorWild:Missing LOAD PHDR in linker script
 
+//#Config:with-pt-phdr:default
+//#ExpectProgramHeader:PHDR sections=[],offset=0x40
```
davidlattimore: Do you think we could add a way to test that the PHDR segment covers only what it's supposed to? It looks like currently we might be including more than we should.

libwild/src/elf.rs
vishruth-thimmaiah: I've fixed the conditional check here.

libwild/src/elf.rs
vishruth-thimmaiah: Fixed, we now track both starting and ending offsets of both `FILEHDR` and `PHDRS`.

wild/tests/sources/elf/linker-script-phdrs/linker-script-phdrs.c
vishruth-thimmaiah: the section header was included when using `PHDRS`, which I've now fixed. I've added a size check as well, since the size of the program header table should be the same on all platforms except riscv. Would that suffice?

libwild/src/elf.rs
```diff
@@ -2368,21 +2381,29 @@ impl platform::Platform for Elf {
...
+                    let entry = &mut segment_entries[segment.as_usize()];
+                    entry.is_emitted = true;
+                    if !matches!(SegmentType(entry.ptype), pt::LOAD | pt::PHDR)
+                        && (entry.has_filehdr || entry.has_phdrs)
+                    {
+                        bail!(
+                            "Non-load segment {} includes file header and/or program header",
+                            segment.as_usize()
+                        );
+                    }
```
davidlattimore, line 2406: Is it right that this validation happens inside the sections loop? That presumably mean that we'll only reject such a segment if it is assigned a section?

libwild/src/elf.rs
```diff
@@ -2453,9 +2473,52 @@ impl platform::Platform for Elf {
...
+            if pos == 0 {
+                let mut filehdr_starts = Vec::new();
+                let mut filehdr_ends = Vec::new();
+                let mut phdrs_starts = Vec::new();
+                let mut phdrs_ends = Vec::new();
+
+                for (seg_idx, segment) in starts.iter().enumerate().take(num_phdrs) {
+                    let entry = segment_entries[seg_idx];
+                    let seg_id = ProgramSegmentId::new(seg_idx);
+                    if entry.has_filehdr {
```
davidlattimore, line 2488: Do we do the right thing with header-only segments? e.g.:

```
PHDRS {
    headers PT_LOAD FILEHDR PHDRS;
    text PT_LOAD;
}
```

libwild/src/elf.rs
vishruth-thimmaiah: Yes, non load/phdr type segments can have headers, as long as they don't have any sections assigned. I've added a test to verify this.

libwild/src/elf.rs
vishruth-thimmaiah: The segment wasn't being terminated correctly here, fixed.

LINKER_SCRIPT_SUPPORT.md
```diff
@@ -125,7 +125,7 @@ see at a glance what remains before Wild can link the kernel.
 | `SORT_BY_ALIGNMENT(...)` | ✅ | Parsed and ignored, as sections are sorted by alignment by default |
 | `EXCLUDE_FILE(...)` inside input section matchers | 📅 | |
 | `CONSTRUCTORS` command | ✅ | Parsed and ignored, it is a nop for ELF.  |
-| `PHDRS` command for explicit program header definition | 🧪 | The FILEHDR and PHDRS keywords aren't yet supported. |
+| `PHDRS` command for explicit program header definition | ✅ | |
```
davidlattimore, line 129: According to https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_node/ld_23.html - PHDRS supports an optional AT address. Should we record that this isn't supported somewhere?

vishruth-thimmaiah: I can't find any use for this in the kernel, so I've mentioned it under the general top level commands section.

davidlattimore (approved):

lapla-cogito (approved):
Thanks!

PR #2319, opened 2026-08-02
feat: Implement ADRP+ADD to NOP+ADR relaxation for AArch64

deepakshirkem:
When an ADRP+ADD pair references a symbol within ADR range (±1MB) and the instructions use the same register, relax to NOP+ADR.

This matches lld's `tryRelaxAdrpAdd` optimization and reduces code size for frequently accessed local symbols.

Issue #2247

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -6068,6 +6068,10 @@ fn classify_symbol_relocation<
         section_flags,
         true,
         None,
+        1,    // sym_addr: assume non-zero
+        0,    // section_address
+        0,    // rel_addend
+        None, // previous_relocation
     )
     .filter(|relaxation| args.should_relax() || relaxation.is_mandatory())
     {
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -10,14 +10,18 @@ use crate::layout::Layout;
 use crate::malfunction_point_ret;
 use crate::platform::ObjectFile as _;
 use crate::platform::Platform;
+use crate::platform::PreviousRelocationInfo;
 use linker_utils::aarch64::RelaxationKind;
 use linker_utils::aarch64::relocation_type_from_raw;
+use linker_utils::bit_misc::BitExtraction;
 use linker_utils::elf::AArch64Instruction;
+use linker_utils::elf::AllowedRange;
 use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::PAGE_MASK_4KB;
 use linker_utils::elf::RelocationKind;
 use linker_utils::elf::RelocationKindInfo;
 use linker_utils::elf::SIZE_4KB;
+use linker_utils::elf::Sign;
 use linker_utils::elf::aarch64_rel_type_to_string;
 use linker_utils::elf::shf;
 use linker_utils::relaxation::RelocationModifier;
@@ -163,6 +167,12 @@ impl crate::platform::Arch for ElfAArch64 {
         section_flags: linker_utils::elf::SectionFlags,
         _non_zero_address: bool,
         _relax_deltas: Option<&linker_utils::relaxation::SectionRelaxDeltas>,
+        sym_addr: u64,
+        section_address: u64,
+        rel_addend: i64,
+        previous_relocation: Option<
+            PreviousRelocationInfo<<Self::Platform as Platform>::RelocationInfo>,
+        >,
     ) -> Option<Self::Relaxation>
     where
         Self: std::marker::Sized,
@@ -307,7 +317,50 @@ impl crate::platform::Arch for ElfAArch64 {
                     mandatory: false,
                 });
             }
-
+            // Relax ADRP+ADD to NOP+ADR. Applied at ADD position when previous relocation
+            // was ADRP for the same symbol at the consecutive offset.
+            object::elf::R_AARCH64_ADD_ABS_LO12_NC
+                if !interposable && flags.is_address() && sym_addr != 0 =>
+            {
+                if let Some(prev) = previous_relocation
+                    && prev.kind == object::elf::R_AARCH64_ADR_PREL_PG_HI21
+                    && prev.offset + 4 == offset_in_section
+                    && prev.addend == 0
+                {
+                    let offset = offset_in_section as usize;
+                    if let Some(adrp_bytes) = offset
+                        .checked_sub(4)
+                        .and_then(|s| section_bytes.get(s..offset))
+                        && let Some(add_bytes) = section_bytes.get(offset..offset + 4)
+                    {
+                        let adrp_instr =
+                            u64::from(u32::from_le_bytes(adrp_bytes.try_into().unwrap()));
+                        let add_instr =
+                            u64::from(u32::from_le_bytes(add_bytes.try_into().unwrap()));
+                        if (adrp_instr as u32 & 0x9f000000) == 0x90000000
+                            && (add_instr as u32 & 0xffc00000) == 0x91000000
+                            && adrp_instr.extract_bit_range(0..5)
+                                == add_instr.extract_bit_range(0..5)
+                            && adrp_instr.extract_bit_range(0..5)
+                                == add_instr.extract_bit_range(5..10)
+                        {
+                            let add_place = section_address + offset_in_section;
+                            let diff = (sym_addr as i64)
+                                .wrapping_add(rel_addend)
+                                .wrapping_sub(add_place as i64);
+                            if AllowedRange::from_bit_size(21, Sign::Signed).contains(diff) {
+                                return Some(Relaxation {
+                                    kind: RelaxationKind::AdrpAddToNopAdr,
+                                    rel_info: rel_info_from_type!(
+                                        object::elf::R_AARCH64_ADR_PREL_LO21
+                                    ),
+                                    mandatory: false,
+                                });
+                            }
+                        }
+                    }
+                }
+            }
             _ => (),
         }
 
--- a/libwild/src/elf_riscv64.rs
+++ b/libwild/src/elf_riscv64.rs
@@ -9,6 +9,7 @@ use crate::error::Context as _;
 use crate::error::Result;
 use crate::platform::ObjectFile as _;
 use crate::platform::Platform;
+use crate::platform::PreviousRelocationInfo;
 use crate::platform::RelaxSymbolInfo;
 use crate::platform::Relocation;
 use itertools::Itertools;
@@ -194,6 +195,10 @@ impl crate::platform::Arch for ElfRiscV64 {
         section_flags: linker_utils::elf::SectionFlags,
         non_zero_address: bool,
         relax_deltas: Option<&SectionRelaxDeltas>,
+        _sym_addr: u64,
+        _section_address: u64,
+        _rel_addend: i64,
+        _previous_relocation: Option<PreviousRelocationInfo<object::elf::RelocationType>>,
     ) -> Option<Self::Relaxation>
     where
         Self: std::marker::Sized,
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -72,6 +72,7 @@ use crate::platform::Arch;
 use crate::platform::Args as _;
 use crate::platform::ObjectFile;
 use crate::platform::Platform;
+use crate::platform::PreviousRelocationInfo;
 use crate::platform::RawSymbolName as _;
 use crate::platform::Relaxation as _;
 use crate::platform::Relocation;
@@ -3125,6 +3126,19 @@ fn apply_relocation<
         section_info.section_flags,
         resolution.raw_value != 0,
         relax_deltas,
+        resolution.raw_value,
+        section_address,
+        rel.addend(),
+        relocation_cache
+            .previous
+            .as_ref()
+            .filter(|r| r.symbol() == rel.symbol())
+            .map(|r| PreviousRelocationInfo {
+                kind: r.raw_type(),
+                offset: r.offset(),
+                symbol: r.symbol(),
+                addend: r.addend(),
+            }),
     )
     .filter(|relaxation| layout.args().relax || relaxation.is_mandatory());
 
--- a/libwild/src/elf_x86_64.rs
+++ b/libwild/src/elf_x86_64.rs
@@ -11,6 +11,7 @@ use crate::error;
 use crate::error::Result;
 use crate::malfunction_point_ret;
 use crate::platform::Platform;
+use crate::platform::PreviousRelocationInfo;
 use crate::value_flags::ValueFlags;
 use linker_utils::elf::DynamicRelocationKind;
 use linker_utils::elf::RelocationKindInfo;
@@ -154,6 +155,10 @@ impl crate::platform::Arch for ElfX86_64 {
         section_flags: SectionFlags,
         _non_zero_address: bool,
         _relax_deltas: Option<&linker_utils::relaxation::SectionRelaxDeltas>,
+        _sym_addr: u64,
+        _section_address: u64,
+        _rel_addend: i64,
+        _previous_relocation: Option<PreviousRelocationInfo<object::elf::RelocationType>>,
     ) -> Option<Self::Relaxation> {
         let is_known_address = flags.is_address();
         let is_absolute = flags.is_absolute() && !flags.is_dynamic();
@@ -602,6 +607,10 @@ fn test_relaxation() {
             shf::EXECINSTR,
             true,
             None,
+            0,
+            0,
+            0,
+            None,
         ) {
             r.apply(&mut out, &mut offset, &mut 0);
 
@@ -618,6 +627,10 @@ fn test_relaxation() {
             OutputKind::StaticExecutable(RelocationModel::Relocatable),
             shf::EXECINSTR,
             true,
+            None,
+            0,
+            0,
+            0,
             None,
         ) {
             out.copy_from_slice(bytes_in);
--- a/linker-diff/src/lib.rs
+++ b/linker-diff/src/lib.rs
@@ -229,6 +229,9 @@ impl Config {
                 // LLD does some different relaxations to us
                 "rel.missing-opt.R_AARCH64_ADR_GOT_PAGE.ReplaceWithNop.*",
                 "rel.missing-opt.R_AARCH64_ADR_PREL_PG_HI21.ReplaceWithNop.*",
+                // Wild and lld relax ADRP+ADD to NOP+ADR, but GNU ld does not.
+                "rel.extra-opt.R_AARCH64_ADD_ABS_LO12_NC.AdrpAddToNopAdr.*",
+                "rel.extra-opt.R_AARCH64_ADR_PREL_PG_HI21.ReplaceWithNop.*",
                 // The other linkers set properties on sections if all input sections have that
                 // property. For sections like .rodata, this seems like an unimportant behaviour to
                 // replicate.
--- a/linker-utils/src/aarch64.rs
+++ b/linker-utils/src/aarch64.rs
@@ -44,6 +44,9 @@ pub enum RelaxationKind {
     /// Replace add with adr. We don't apply this, but lld does, so this is used by linker-diff.
     AddToAdr,
     LdrToAdr,
+    /// Relax ADRP+ADD pair to NOP+ADR. Applied at the ADD instruction position.
+    /// Writes NOP at ADRP position (offset-4) and ADR base at current position.
+    AdrpAddToNopAdr,
 }
 
 impl RelaxationKind {
@@ -94,6 +97,22 @@ impl RelaxationKind {
             }
             RelaxationKind::LdrToAdr => {
                 section_bytes[offset + 3] &= !0x89;
+            }
+            RelaxationKind::AdrpAddToNopAdr => {
+                let adrp_dest_reg = u64::from(u32_from_slice(&section_bytes[offset - 4..offset]))
+                    .extract_bit_range(0..5) as u8;
+                RelaxationKind::ReplaceWithNop.apply(
+                    section_bytes,
+                    &mut ((*offset_in_section).wrapping_sub(4)),
+                    _addend,
+                );
+                // Write ADR base at ADD position with same destination register.
+                section_bytes[offset..offset + 4].copy_from_slice(&[
+                    adrp_dest_reg,
+                    0x00,
+                    0x00,
+                    0x10,
+                ]);
             }
             RelaxationKind::AdrpX0 => {
                 section_bytes[offset..offset + 4].copy_from_slice(&[
--- a/wild/tests/sources/elf/aarch64-adrp-add-relax/aarch64-adrp-add-relax.s
+++ b/wild/tests/sources/elf/aarch64-adrp-add-relax/aarch64-adrp-add-relax.s
@@ -1,0 +1,36 @@
+// This verifies that Wild relaxes ADRP+ADD to NOP+ADR when the symbol is
+// within ADR range and registers match, but keeps ADRP+ADD when:
+// - the destination registers differ (x2 vs x3 in third pair)
+// - the second relocation references a different symbol (fourth pair)
+// - the second relocation is not ADD (fifth pair: ADRP+ADRP)
+// We use lld as reference linker since GNU ld does not implement this relaxation.
+//#Arch:aarch64
+//#Mode:static
+//#ReferenceLinkers:lld
+//#LinkArgs:--no-gc-sections -Ttext=0x200ffc
+//#RunEnabled:false
+//#DiffIgnore:file-header.entry
+//#ExpectSectionBytes:.text=0x1f2003d5 0..4
+//#ExpectSectionBytes:.text=0x1f2003d5 8..12
+
+.globl _start
+_start:
+    adrp x30, x
+    add  x30, x30, :lo12:x     // Relaxed (same register, same symbol, in range)
+    adrp x1, x
+    add  x1, x1, :lo12:x       // Relaxed (same register, same symbol, in range)
+    adrp x2, x
+    add  x3, x2, :lo12:x       // NOT relaxed (different dest register x3 vs x2)
+    adrp x4, x
+    add  x4, x4, :lo12:y       // NOT relaxed (different symbol: x vs y)
+    adrp x5, x
+    adrp x5, x                  // NOT relaxed (second reloc is not ADD)
+
+.globl x
+x:
+    .word 0
+
+.globl y
+y:
+    .word 0
+.data
--- a/libwild/src/elf_loongarch64.rs
+++ b/libwild/src/elf_loongarch64.rs
@@ -3,6 +3,7 @@ use crate::elf::PLT_ENTRY_SIZE;
 use crate::error;
 use crate::error::Result;
 use crate::platform::Platform;
+use crate::platform::PreviousRelocationInfo;
 use itertools::AllEqualValueError;
 use itertools::Itertools;
 use linker_utils::elf::DynamicRelocationKind;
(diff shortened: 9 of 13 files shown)


davidlattimore (commented):
I haven't looked in much detail as yet, but is there any reason why this can't be done similar to how the other aarch64 relaxations are implemented?

deepakshirkem:
The main challenge is that ADRP+ADD relaxation requires inspecting the NEXT relocation (ADD) to verify same symbol and register, which isn't available in new_relaxation. The existing AArch64 relaxations are single-instruction transformations.

If you have a preferred approach for handling two-instruction relaxations in Wild's existing infrastructure.

davidlattimore:
I think we'll need to adjust the API so that the code that decides relaxations can access the next relocation. Either that or add a new API for handling relaxations - e.g. one that always receives pairs of adjacent relocations. I'm not sure which would be best. But either way, aarch64 relaxation code definitely needs to be in the elf_aarch64 module. Where it is now, it gets run for all architectures and the relocation codes it's looking for are likely to be used for something completely different on those other architectures.

linker-diff/src/lib.rs
```diff
@@ -229,6 +229,7 @@ impl Config {
                 // LLD does some different relaxations to us
                 "rel.missing-opt.R_AARCH64_ADR_GOT_PAGE.ReplaceWithNop.*",
                 "rel.missing-opt.R_AARCH64_ADR_PREL_PG_HI21.ReplaceWithNop.*",
+                "rel.extra-opt.R_AARCH64_ADR_PREL_PG_HI21.ReplaceWithNop.*",
```
deepakshirkem, line 234: @davidlattimore Is this line correct? The CI is now green with this line. My understanding is that we're comparing the output against GNU ld, but Wild has an optimization that GNU ld doesn't. This line seems to skip that comparison.

wild/tests/sources/elf/aarch64-adrp-add-relax/aarch64-adrp-add-relax.s
```diff
@@ -0,0 +1,24 @@
+// This verifies that the linker relaxes ADRP+ADD to NOP+ADR when the symbol
+// is within ADR range (±1MB) and registers match, but keeps ADRP+ADD when
+// the destination registers differ (x2 vs x3 in the third pair).
+// We use lld as reference linker since GNU ld does not implement this relaxation.
+//#Arch:aarch64
+//#Mode:static
+//#ReferenceLinkers:lld
+//#LinkArgs:--no-gc-sections -Ttext=0x200ffc
+//#RunEnabled:false
+//#DiffIgnore:file-header.entry
```
deepakshirkem, line 12: Do you think we can improve this test as well?

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,59 @@ impl crate::platform::Arch for ElfAArch64 {
         None
     }
 
+    fn apply_pair_relaxation(
+        first_kind: object::elf::RelocationType,
+        _second_kind: object::elf::RelocationType,
```
davidlattimore: Why aren't you checking the type of the second relocation? Doesn't that mean that if it's a different kind of relocation, we might apply the relaxation when it shouldn't apply? Could be good to add a test for this.

linker-diff/src/lib.rs
davidlattimore: I think that's reasonable to add that global ignore. I guess ideally we'd have ignore rules that changed depending on which linker we were comparing against, but currently we don't. The placement of the ignore rule is possibly a bit misleading. It's under a comment that refers to "LLD", but from what I understand, LLD does this optimisation too and it's GNU ld that doesn't.

wild/tests/sources/elf/aarch64-adrp-add-relax/aarch64-adrp-add-relax.s
davidlattimore: Do you mean with regard to the ignore of `file-header.entry`? It looks like that's effectively a bug in linker-diff where it's not doing a good job of handling the fact that the entry point has multiple symbols that resolve to that address.

Usually I try to not disable running of the executable, but it's possible that it's not worth trying to run the executable in this case.

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,61 @@ impl crate::platform::Arch for ElfAArch64 {
         None
     }
 
+    fn apply_pair_relaxation(
```
marxin: To be honest, don't like the new entry point - can't we rather base the relaxation on what was suggested by David - passing the next relocation to the `A::next_relaxation` based on the following patch:

```patch
diff --git a/libwild/src/elf_writer.rs b/libwild/src/elf_writer.rs
index 97e18ca6..b80893ac 100644
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -2417,7 +2417,7 @@ fn apply_relocations<
     object: &ObjectLayout<'data, elf::Elf<C>>,
     out: &mut [u8],
     section_index: object::SectionIndex,
-    mut relocations: I,
+    relocations: I,
     layout: &ElfLayout<'data, C>,
     table_writer: &mut TableWriter<'_, '_, C>,
     trace: &TraceOutput,
@@ -2434,6 +2434,7 @@ fn apply_relocations<
     let relax_deltas = object.section_relax_deltas.get(section_index.0);
     let mut relax_cursor = relax_deltas.map(|deltas| deltas.cursor());

+    let mut relocations = relocations.peekable();
     while let Some(rel) = relocations.next() {
         let rel = rel?;
         relocation_count += 1;
@@ -2456,10 +2457,12 @@ fn apply_relocations<
             None => rel.offset(),
         };

+        let rel_next = relocations.peek().copied().transpose()?;
         modifier = apply_relocation::<C, A, R, _>(
             object,
             offset_in_section,
             &rel,
+            rel_next,
             SectionInfo {
                 section_address,
                 is_writable: object_section.is_writable(),
@@ -2718,6 +2721,9 @@ fn write_eh_frame_relocations<
                     object,
                     rel_offset - input_pos as u64,
                     rel,
+                    // Relaxations based on the ability to look-up the next relocation are not
+                    // expected to happen for the section.
+                    None,
                     SectionInfo {
                         section_address: output_pos as u64 + table_writer.eh_frame_start_address,
                         is_writable: false,
@@ -3040,6 +3046,7 @@ fn apply_relocation<
     object_layout: &ObjectLayout<'data, elf::Elf<C>>,
     mut offset_in_section: u64,
     rel: &R,
+    _next_rel: Option<R>,
     section_info: SectionInfo<linker_utils::elf::SectionFlags>,
     layout: &ElfLayout<'data, C>,
     out: &mut [u8],
```

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+            return false;
+        }
+        let offset = first_offset as usize;
+        if offset + 8 > out.len() {
+            return false;
+        }
+        // Get destination register from ADRP instruction
+        let adrp_instr = u32::from_le_bytes(out[offset..offset + 4].try_into().unwrap());
+        let add_instr = u32::from_le_bytes(out[offset + 4..offset + 8].try_into().unwrap());
+        let adrp_dest_reg = adrp_instr & 0x1f;
```
marxin: can we rather use already used pattern: `extract_bit_range`?

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+        let add_dest_reg = add_instr & 0x1f;
+        let add_src_reg = (add_instr >> 5) & 0x1f;
+        // Verify ADRP and ADD instructions use same register
+        if (adrp_instr & 0x9f000000) != 0x90000000
```
marxin: I see the code just extracted some source and destination registers - why are we then comparing the masks out of `adrp_instr` and `add_instr`? Don't get it.

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+            || (add_instr & 0xffc00000) != 0x91000000
+            || adrp_dest_reg != add_dest_reg
+            || adrp_dest_reg != add_src_reg
+        {
+            return false;
+        }
+        // Write NOP at ADRP position
+        out[offset..offset + 4].copy_from_slice(&[0x1f, 0x20, 0x03, 0xd5]);
```
marxin: You should rather define a new `RelaxationKind` enum variant and make the transformation of all 8B in `RelaxationKind::apply` function.

libwild/src/elf_aarch64.rs
```diff
@@ -314,6 +314,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+        // Write ADR base at ADD position
```
marxin: You might be able to reuse: `RelaxationKind::AddToAdr` (and so your `RelaxationKind` will be effectively a composition of 2 already supported relaxations)?

marxin (changes requested):

wild/tests/sources/elf/aarch64-adrp-add-relax/aarch64-adrp-add-relax.s
```diff
@@ -0,0 +1,27 @@
+// This verifies that Wild relaxes ADRP+ADD to NOP+ADR when the symbol is
+// within ADR range (±1MB) and registers match, but keeps ADRP+ADD when:
+// - the destination registers differ (x2 vs x3 in third pair)
```
marxin: These 2 lines are duplicate - they are later documented again at the assembly level.

deepakshirkem:
@marxin Thank you. I actually started with the peekable/next_rel approach but found it required changes to many architectures (updating all new_relaxation implementations). I then went with the apply_pair_relaxation approach to keep changes isolated.

However looking at your patch, it's much cleaner  passing next_rel through apply_relocation avoids the need for a new trait method entirely. I'll implement this approach later today and update the PR.

deepakshirkem:
@marxin @davidlattimore Please do one more cycle of review when you get a chance?

libwild/src/elf_aarch64.rs
```diff
@@ -307,13 +309,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+                    }
+                }
+            }
             _ => (),
         }
 
         None
     }
 
+    fn should_apply_relaxation(
```
marxin: I feel we should not introduce a new 2-step application mechanism, so I would rather drop this method and ...

libwild/src/elf_x86_64.rs
```diff
@@ -154,6 +154,7 @@ impl crate::platform::Arch for ElfX86_64 {
         section_flags: SectionFlags,
         _non_zero_address: bool,
```
marxin, line 156: Replace `_non_zero_address` with `resolution.raw_value`.

libwild/src/elf_aarch64.rs
```diff
@@ -307,13 +309,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+        relaxation: &Self::Relaxation,
+        sym_addr: u64,
+        section_address: u64,
+        offset_in_section: u64,
+    ) -> bool {
+        use linker_utils::aarch64::RelaxationKind;
+        match relaxation.kind {
+            RelaxationKind::AdrpAddToNopAdr => {
+                // Check symbol is within ADR range (±1MB) from ADD position
+                let add_place = section_address + offset_in_section + 4;
```
marxin: If you pass `resolution.raw_value` to `new_relaxation`, you should be able to do the `diff` calculate there (you already know the pair of instructions need to sit next to each other, the + 4. What might be missing is `section_address`, that can be added as an extra argument.

linker-utils/src/aarch64.rs
```diff
@@ -95,6 +99,22 @@ impl RelaxationKind {
             RelaxationKind::LdrToAdr => {
                 section_bytes[offset + 3] &= !0x89;
             }
+            RelaxationKind::AdrpAddToNopAdr => {
+                // Read destination register from ADRP before overwriting
+                let adrp_dest_reg = u64::from(u32_from_slice(&section_bytes[offset..offset + 4]))
+                    .extract_bit_range(0..5) as u8;
+                // Write NOP at ADRP position
+                section_bytes[offset..offset + 4].copy_from_slice(&[0x1f, 0x20, 0x03, 0xd5]);
```
marxin: `RelaxationKind::ReplaceWithNop.apply` can be used if I'm not mistaken.

linker-utils/src/aarch64.rs
```diff
@@ -95,6 +99,22 @@ impl RelaxationKind {
...
+                // Write ADR base at ADD position with same destination register
+                section_bytes[offset + 4..offset + 8].copy_from_slice(&[
```
marxin: Can't we use again already existing `RelaxationKind::AddToAdr`?

libwild/src/elf_aarch64.rs
```diff
@@ -307,13 +309,61 @@ impl crate::platform::Arch for ElfAArch64 {
...
+                let diff = (sym_addr as i64).wrapping_sub(add_place as i64);
+                (-(1024 * 1024)..(1024 * 1024)).contains(&diff)
```
marxin: Maybe I would consider using `AllowedRange::from_bit_size(X, Sign::Signed).contains(diff)`?

marxin (changes requested):
We're getting closer I think, please try to address my comments.

linker-utils/src/aarch64.rs
deepakshirkem: I tried using AddToAdr but it doesn't work here. AddToAdr only clears op bits (byte3 &= !0x89), but doesn't clear the ADD immediate field. For example ADD x30, x30, #0 (0x910003de) after AddToAdr becomes 0x100003de, but ADR x30, 0 should be 0x1000001e the immediate bits from the ADD instruction corrupt the ADR encoding.

deepakshirkem:
@marxin  Ping :)).

libwild/src/elf_riscv64.rs
```diff
@@ -211,7 +213,7 @@ impl crate::platform::Arch for ElfRiscV64 {
 
         match relocation_kind {
             object::elf::R_RISCV_CALL | object::elf::R_RISCV_CALL_PLT if !interposable => {
-                return if non_zero_address {
+                return if _sym_addr != 0 {
```
marxin: Should not be prefixed with `_`.

linker-utils/src/aarch64.rs
```diff
@@ -95,6 +99,24 @@ impl RelaxationKind {
             RelaxationKind::LdrToAdr => {
                 section_bytes[offset + 3] &= !0x89;
             }
+            RelaxationKind::AdrpAddToNopAdr => {
+                // Read destination register from ADRP before overwriting
+                let adrp_dest_reg = u64::from(u32_from_slice(&section_bytes[offset..offset + 4]))
+                    .extract_bit_range(0..5) as u8;
+                // Write NOP at ADRP position (reuse ReplaceWithNop)
```
marxin:
```suggestion
                // Write NOP at ADRP position.
```

linker-utils/src/aarch64.rs
marxin: Gotcha!

linker-utils/src/aarch64.rs
```diff
@@ -95,6 +99,24 @@ impl RelaxationKind {
...
+                RelaxationKind::ReplaceWithNop.apply(section_bytes, offset_in_section, _addend);
+                // Write ADR base instruction at ADD position with same destination register.
+                // We cannot use AddToAdr here since it only clears op bits but leaves the
+                // ADD immediate field intact, which would corrupt the ADR offset encoding.
```
marxin: That said, please remove the 2 lines.

libwild/src/elf.rs
```diff
@@ -6063,7 +6063,9 @@ fn classify_symbol_relocation<
         flags,
         symbol_db.output_kind,
         section_flags,
-        true,
+        1, // sym_addr: assume non-zero (actual address not available in layout phase)
```
marxin: @davidlattimore - what about using an `Option<(u64, u64)`, or a new type instead of the part of `u64` types? In particular, don't like the `1` special value which mimics the non-zero value, but is pretty non-intuitive.

marxin (commented):
Thanks for the iteration on the patch - I am pretty happy about it now. But let's make the final decision to @davidlattimore.

One general recommendation - please try to instruct your LLM (I guess you're using Claude) to emit more information intensive comments. Or ideally, try to compress the comments.

davidlattimore (commented):
I see a 2.5% performance loss when linking wild for aarch64:

```
Benchmark 1 (3495 runs): /home/david/save/wild-aarch64/run-with /home/d/wild-builds/2026-08-09.cg1 --strip-debug --no-fork
  wall_time          51.4ms ± 1.06ms    47.4ms … 59.1ms         31 ( 1%)        0%
  peak_rss            138MB ±  832KB     135MB …  142MB         28 ( 1%)        0%
  cpu_cycles         1.03G  ± 29.6M      826M  … 1.16G          84 ( 2%)        0%
  instructions       1.23G  ± 13.6M     1.16G  … 1.31G          78 ( 2%)        0%
  cache_references   25.5M  ±  550K     23.5M  … 27.6M          38 ( 1%)        0%
  cache_misses       5.93M  ±  144K     5.26M  … 6.49M          15 ( 0%)        0%
  branch_misses      2.26M  ± 80.4K     2.03M  … 2.87M         100 ( 3%)        0%
Benchmark 2 (3411 runs): /home/david/save/wild-aarch64/run-with target/cg1/wild --strip-debug --no-fork
  wall_time          52.7ms ± 1.05ms    48.8ms … 57.5ms         35 ( 1%)        💩+  2.5% ±  0.1%
  peak_rss            138MB ±  814KB     135MB …  141MB         22 ( 1%)          -  0.0% ±  0.0%
  cpu_cycles         1.04G  ± 30.3M      826M  … 1.20G          87 ( 3%)        💩+  1.2% ±  0.1%
  instructions       1.26G  ± 14.1M     1.18G  … 1.35G          77 ( 2%)        💩+  2.4% ±  0.1%
  cache_references   25.7M  ±  564K     23.3M  … 28.9M          37 ( 1%)          +  1.0% ±  0.1%
  cache_misses       5.92M  ±  149K     5.26M  … 6.63M          17 ( 0%)          -  0.2% ±  0.1%
  branch_misses      2.25M  ± 49.0K     1.98M  … 2.48M          62 ( 2%)          -  0.3% ±  0.1%
```

libwild/src/elf_aarch64.rs
```diff
@@ -307,7 +312,44 @@ impl crate::platform::Arch for ElfAArch64 {
...
+                        if (adrp_instr as u32 & 0x9f000000) == 0x90000000
+                            && (add_instr as u32 & 0xffc00000) == 0x91000000
+                            && adrp_instr.extract_bit_range(0..5)
+                                == add_instr.extract_bit_range(0..5)
+                            && adrp_instr.extract_bit_range(0..5)
+                                == add_instr.extract_bit_range(5..10)
+                        {
+                            // Check symbol is within ADR range (±1MB) from ADD position
+                            let add_place = section_address + next_offset;
+                            let diff = (sym_addr as i64).wrapping_sub(add_place as i64);
```
davidlattimore: Do we need to also take the addend into account here? Otherwise we might apply the relaxation when the addend puts us just outside the allowed range..

libwild/src/elf_aarch64.rs
```diff
@@ -161,8 +164,10 @@ impl crate::platform::Arch for ElfAArch64 {
         flags: crate::value_flags::ValueFlags,
         output_kind: crate::output_kind::OutputKind,
         section_flags: linker_utils::elf::SectionFlags,
-        _non_zero_address: bool,
+        sym_addr: u64,
+        section_address: u64,
         _relax_deltas: Option<&linker_utils::relaxation::SectionRelaxDeltas>,
+        next_relocation: Option<(object::elf::RelocationType, u64)>,
```
davidlattimore: It looks like we're not checking which symbol the relocation refers to, so if the second symbol was for a different symbol, we might still apply the relaxation when we shouldn't. It'd be a bit weird for the compiler to emit that, especially given that the instructions use the same register, but we still shouldn't relax it.

marxin:
Speaking of the numbers, I’d also be interested in how many such relaxations can be triggered for a project like Wild.
@davidlattimore am I right we should see a similar slowdown even for non-AArch64 architecture as we're passing more data to the relaxation parsing?

davidlattimore:
The slowdown for x86_64 was small enough that the benchmark didn't flag it. It showed it as 0.8%, which I guess is small enough that it didn't have high confidence that it wasn't noise. My guess is that the optimiser was able to eliminate some of the arguments since they were never used on the x86_64 version.

davidlattimore:
Note that I did benchmark an earlier version - I think the version that was adding the new API to `Platform` for relaxing a pair of relocations. That version didn't show a slowdown. I'm not saying that we necessarily need to go back to that version - there might be other options that can be experimented with.

deepakshirkem:
@marxin @davidlattimore For performance, I'll try a few other approaches as well. I'm very open to experimenting with different approaches to achieve the same result with this relaxation.  I'd be very happy to try another approach.

davidlattimore (commented):
That still shows about 2.4% slowdown on aarch64. What about the possibility of processing the pair when we get to the ADD rather than the ADRP? That would mean we'd need to pass in the previous relocation rather than the next relocation, which might be a bit easier to implement in a performant way. The `RelocationCache` already stores the previous relocation. The relocations are applied in order, so a relaxation on the later relocation can overwrite whatever the previous relocation did if it decides to do something different.

libwild/src/elf_aarch64.rs
davidlattimore: Any comment on this?

libwild/src/elf_aarch64.rs
```diff
@@ -163,6 +166,14 @@ impl crate::platform::Arch for ElfAArch64 {
         section_flags: linker_utils::elf::SectionFlags,
         _non_zero_address: bool,
         _relax_deltas: Option<&linker_utils::relaxation::SectionRelaxDeltas>,
+        sym_addr: u64,
+        section_address: u64,
+        previous_relocation: Option<(
+            <Self::Platform as Platform>::RelocationInfo,
```
marxin: Should be a new struct instead of the tuple.

libwild/src/elf_aarch64.rs
```diff
@@ -307,7 +318,46 @@ impl crate::platform::Arch for ElfAArch64 {
...
+            object::elf::R_AARCH64_ADD_ABS_LO12_NC
+                if !interposable && flags.is_address() && sym_addr != 0 =>
+            {
+                if let Some((prev_kind, prev_offset, _prev_sym, prev_addend)) = previous_relocation
+                    && prev_kind == object::elf::R_AARCH64_ADR_PREL_PG_HI21
+                    && prev_offset + 4 == offset_in_section
+                    && prev_addend == 0
+                {
+                    let offset = offset_in_section as usize;
+                    if offset >= 4 && offset + 4 <= section_bytes.len() {
```
marxin: Instead of the boundary checking, please use directly `section_bytes.get` and handle the Option return value.

libwild/src/elf_writer.rs
```diff
@@ -3125,6 +3125,13 @@ fn apply_relocation<
         section_info.section_flags,
         resolution.raw_value != 0,
         relax_deltas,
+        resolution.raw_value,
+        section_address,
+        relocation_cache
+            .previous
+            .as_ref()
+            .filter(|r| r.symbol() == rel.symbol() && r.addend() == 0 && rel.addend() == 0)
```
marxin: Seems the filtering is a duplicate logic of `new_relaxation` and it should not happen here.

marxin (commented):
Thanks for iterating on that change - in general looks fine to me, but I would like to see the performance numbers after the latest attempt @deepakshirkem?

libwild/src/elf_aarch64.rs
deepakshirkem: Done.

deepakshirkem:
>I would like to see the performance numbers after the latest attempt @deepakshirkem?
 
I don't have a setup to run the AArch64 benchmark locally. Could you or David run the benchmark on the latest version?

davidlattimore:
For x86_64, I see no change:

```
Benchmark 1 (1537 runs): /home/david/save/wild/run-with /home/d/wild-builds/2026-08-14.cg1 --strip-debug --no-fork
  wall_time          39.0ms ± 1.21ms    36.1ms … 43.8ms          9 ( 1%)        0%
  peak_rss            161MB ±  887KB     158MB …  164MB         10 ( 1%)        0%
  cpu_cycles          678M  ± 26.3M      563M  …  753M           6 ( 0%)        0%
  instructions        726M  ± 9.87M      701M  …  756M           6 ( 0%)        0%
  cache_references   19.6M  ±  542K     18.2M  … 21.2M           0 ( 0%)        0%
  cache_misses       4.55M  ±  135K     4.16M  … 5.13M           8 ( 1%)        0%
  branch_misses      1.92M  ± 32.0K     1.82M  … 2.03M          15 ( 1%)        0%
Benchmark 2 (1537 runs): /home/david/save/wild/run-with /home/d/wild-builds/2026-08-14.cg1.adrp-add-relax --strip-debug --no-fork
  wall_time          39.0ms ± 1.21ms    36.4ms … 44.2ms          9 ( 1%)          +  0.1% ±  0.2%
  peak_rss            161MB ±  865KB     159MB …  164MB         11 ( 1%)          -  0.0% ±  0.0%
  cpu_cycles          678M  ± 26.8M      598M  …  789M           1 ( 0%)          -  0.1% ±  0.3%
  instructions        727M  ± 10.3M      699M  …  788M          12 ( 1%)          +  0.1% ±  0.1%
  cache_references   19.9M  ±  559K     18.6M  … 21.7M           1 ( 0%)        💩+  1.4% ±  0.2%
  cache_misses       4.56M  ±  133K     4.19M  … 5.13M           7 ( 0%)          +  0.4% ±  0.2%
  branch_misses      1.92M  ± 35.0K     1.82M  … 2.10M          15 ( 1%)          +  0.2% ±  0.1%
```

For aarch64, it looks like there's probably a small change, but it's much smaller than it was before, so I'm inclined to say that it's acceptable performance-wise.

```
Benchmark 1 (3470 runs): /home/david/save/wild-aarch64/run-with /home/d/wild-builds/2026-08-14.cg1 --strip-debug --no-fork
  wall_time          51.8ms ± 1.02ms    47.7ms … 57.3ms         41 ( 1%)        0%
  peak_rss            138MB ±  813KB     135MB …  141MB         30 ( 1%)        0%
  cpu_cycles         1.03G  ± 27.9M      880M  … 1.14G          54 ( 2%)        0%
  instructions       1.23G  ± 13.4M     1.16G  … 1.30G          71 ( 2%)        0%
  cache_references   25.5M  ±  531K     23.0M  … 27.9M          39 ( 1%)        0%
  cache_misses       5.94M  ±  141K     5.41M  … 6.54M          17 ( 0%)        0%
  branch_misses      2.25M  ± 45.5K     2.05M  … 2.52M          40 ( 1%)        0%
Benchmark 2 (3451 runs): /home/david/save/wild-aarch64/run-with /home/d/wild-builds/2026-08-14.cg1.adrp-add-relax --strip-debug --no-fork
  wall_time          52.1ms ± 1.05ms    47.8ms … 56.6ms         51 ( 1%)          +  0.6% ±  0.1%
  peak_rss            138MB ±  841KB     135MB …  141MB         23 ( 1%)          +  0.0% ±  0.0%
  cpu_cycles         1.03G  ± 30.0M      831M  … 1.15G          88 ( 3%)          +  0.1% ±  0.1%
  instructions       1.25G  ± 14.1M     1.17G  … 1.32G          98 ( 3%)        💩+  1.1% ±  0.1%
  cache_references   27.1M  ±  617K     24.5M  … 30.5M          56 ( 2%)        💩+  6.4% ±  0.1%
  cache_misses       5.93M  ±  146K     5.12M  … 6.51M          24 ( 1%)          -  0.1% ±  0.1%
  branch_misses      2.27M  ± 49.6K     2.06M  … 2.53M          51 ( 1%)          +  0.7% ±  0.1%
```

davidlattimore (commented):
Can you add a test where the two relocations reference different symbols? It looks like that's not currently being checked, so the relaxation would be applied when it shouldn't.

davidlattimore (approved):
LGTM, but I'll wait for Martin to confirm that his suggestions have been addressed

marxin (approved):
LGTM as well

PR #2323, opened 2026-08-03
port(MachO): Add ld as a reference linker

davidlattimore:


davidlattimore:
I was thinking that it'd be good to not just compare with lld on Mac. Then I figured we might as well enable it by default. Open to opinions as to what's best to do if the SDK isn't installed. For now, in that case, we should only error if a test explicitly lists "ld" in its reference linkers directive.

marxin (approved):

PR #2325, opened 2026-08-03
port(wasm): Support `--initial-memory`

lapla-cogito:
Raise `memory.initial` to the requested page-aligned size. Reject values that are unaligned, too small for the layout, or larger than the wasm32 limit. This option is used when building CPython for WASI.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -4170,10 +4170,14 @@ fn wasm_page_size() -> u64 {
     crate::args::wasm::WASM_PAGE_SIZE
 }
 
+/// Maximum linear-memory size in bytes for wasm32.
+const WASM32_MAX_MEMORY_BYTES: u64 = 1u64 << 32;
+
 fn ensure_memory_covers(
     layout: &mut WasmLayout<'_>,
     stack_size: u32,
     stack_first: bool,
+    initial_memory: Option<u64>,
 ) -> Result<u64> {
     let page = wasm_page_size();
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -4170,10 +4170,14 @@ fn wasm_page_size() -> u64 {
     crate::args::wasm::WASM_PAGE_SIZE
 }
 
+/// Maximum linear-memory size in bytes for wasm32.
+const WASM32_MAX_MEMORY_BYTES: u64 = 1u64 << 32;
```
davidlattimore: I tried passing this value minus 1 page and it worked, but passing this value causes a 32 bit overflow

lapla-cogito: Thanks. After further investigation into wasm-ld's handling of this, I noticed that while wasm-ld allows specifying an initial memory size up to 2^32, it holds the value as a 64-bit number and rounds it down to 32 bits for the wasm32 target, resulting in `__heap_end` being wrapped to zero. I can't discern any specific benefit to this implementation. Therefore, I've chenged to accept values up to `2^32-page_size`. In practice, it's rare for users to specify an initial memory size of 2^32 I think.

lapla-cogito: Of course, we could also change how `__heap_end` is handled only when a maximum value is specified, but this seems somewhat ad-hoc.

davidlattimore (approved):

PR #2326, opened 2026-08-03
api(FS): use consistently Error type

marxin:
I feel it would be better having a unified `Error` type in the FS abstraction. It shows nicely in the provided example, where the nature of an error is not tightly connected to a `std::io` Error type, but can be e.g. `Mutex::lock` error.

### DIFF
--- a/libwild/src/fs.rs
+++ b/libwild/src/fs.rs
@@ -273,16 +273,16 @@ pub trait FileSystem: Send + Sync + 'static {
     ) -> Result<(Self::Input, Option<Arc<File>>)>;
 
     /// Returns the type of the file at `path`.
-    fn file_type(&self, path: &Path) -> std::io::Result<FileType>;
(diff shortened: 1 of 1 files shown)


libwild/src/fs.rs
```diff
@@ -140,50 +143,58 @@ pub trait OutputFileData: Send {
 ///         let bytes = self
 ///             .files
 ///             .lock()
-///             .unwrap()
+///             .expect("cannot lock in-memory FS")
```
davidlattimore: Note that an error when locking a Mutex means that the thread that previously locked the mutex panicked while it had it locked. Not sure if that changes how you'd want to handle it or what kind of message you might want. Usually I just unwrap in this case.

davidlattimore (approved):

libwild/src/fs.rs
marxin: Yeah, good observation, let's move it back to the `unwrap`.

Issue #2234, opened 2026-07-18
Embrace REUSE


mati865:
Continuing the licensing topic from our meeting. We should adapt https://reuse.software/ to make the life easier for downstreams (and hopefully for us) in terms of licensing.

Some links:
- Rust projec REUSEt: https://github.com/rust-lang/rust/blob/main/REUSE.toml
- Arch Linux REUSE for Wild package: https://gitlab.archlinux.org/archlinux/packaging/packages/wild/-/blob/main/REUSE.toml?ref_type=heads

cc @marxin

PR #2328, opened 2026-08-03
docs: Add REUSE.toml

mati865:
cc #2234


REUSE.toml
```diff
@@ -2,64 +2,62 @@ version = 1
...
+  ".git-blame-ignore-revs",
+  ".github/**",
+  ".gitignore",
+  ".gitmodules",
+  ".helix/**",
+  ".typos.toml",
+  ".vscode/**",
+  ".yamlfmt",
+  ".zed/**",
+  "BENCHMARKING.md",
```
davidlattimore, line 16: So each time we add a new file, we have to add it here? That seems a bit onerous. Is there some tool that checks that? Why do we need per-file license information?

davidlattimore (commented):
Thanks for looking into this. Does this help with collating the licenses for all our dependencies and putting them in the release packages?

REUSE.toml
mati865: Yeah, you are supposed to run `reuse lint` on CI or manually.

> Why do we need per-file license information?

TBH, I have no clue. I don't know even if we need that, but REUSE seems to require all files to have license and copyright.

mati865:
> Does this help with collating the licenses for all our dependencies and putting them in the release packages?

I thought it would, but I must have mistaken it with something else. REUSE apparently only helps with the code within the repo.

Rust seems to have a homegrown tool for dependencies license handling.

mati865:
This is how failed REUSE check looks like: https://github.com/wild-linker/wild/actions/runs/31602029051/job/94131575815?pr=2328

Are we interested in going on with this?

davidlattimore:
I'm unsure what I think about having to update it each time you add a file or directory to the root. I guess at least we don't have to for subdirectories and it would perhaps encourage us to put less files in the root - e.g. maybe we need a docs directory.

It's also a bit annoying that we have to add a second copy of the license files. i.e. we have them in the `LICENSES` directory when we already have them in the root. Relatedly, the license files in the root are themselves listed as being dual Apache2+MIT, which seems unlikely.

I guess this is mostly of use for people wanting to use libwild as part of some other project.

mati865:
> It's also a bit annoying that we have to add a second copy of the license files. i.e. we have them in the `LICENSES` directory when we already have them in the root. Relatedly, the license files in the root are themselves listed as being dual Apache2+MIT, which seems unlikely.

Yeah, that's annoying but on the upside `reuse` tool manages them by itself. So, you can just run `reuse download --all` when adding a new one.

> I guess this is mostly of use for people wanting to use libwild as part of some other project.

TBH I'm a bit lost here. Previously I thought REUSE would take care of the deps, so this would cover cases for both packaging Wild and using as a library. But now I don't know if this change is even worth pushing for.

davidlattimore:
> TBH I'm a bit lost here. Previously I thought REUSE would take care of the deps, so this would cover cases for both packaging Wild and using as a library.

Their [FAQ](https://reuse.software/faq/#dependencies-licenses) say that helping with licenses for dependencies is out of scope for REUSE.

> But now I don't know if this change is even worth pushing for.

Yeah, it does seem to be of pretty limited use - especially given that all of our code is under the same licenses. But the cost of maintaining it doesn't seem too onerous, so I don't object to merging this if you want to.

davidlattimore (approved):

mati865:
> Their [FAQ](https://reuse.software/faq/#dependencies-licenses) say that helping with licenses for dependencies is out of scope for REUSE.

Yeah, that was when I realised my mistake.

I'll wait a bit to give others time to comment if they want.

mati865:
cc @marxin @lapla-cogito ^

marxin (approved):

lapla-cogito (approved):

mati865:
I have checked source tarballs and we don't currently include any git submodules (those are empty dirs), so I don't know if they should be even included in `REUSE.toml`.

marxin:
> so I don't know if they should be even included in `REUSE.toml`.

I would probably exclude them from the `REUSE.toml` file then.

marxin:
@mati865 can we please finish the PR - I feel there's a consensus and the only remainder is the question of sub-modules. I would leave it up to you how to handle it.

davidlattimore (approved):

PR #2332, opened 2026-08-04
port(wasm): Synthesize GOT.func globals for `R_WASM_GLOBAL_INDEX`

lapla-cogito:
Similar to the earlier GOT.mem work for Data symbols, handle `R_WASM_GLOBAL_INDEX_*` relocations against Func symbols by synthesizing immutable i32 globals (`GOT.func.internal.*`) whose initializers are the function's funcref table index. Absorb `GOT.func.*` imports, place referenced functions in the indirect table, and use table index 0 for undefined weak functions.

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -3286,13 +3286,20 @@ impl GotMem {
     }
 }
 
+/// Where a GOT.func slot's table index comes from.
+#[derive(Debug, Clone, Copy)]
+enum GotFuncTarget {
+    Object {
+        object_index: usize,
+        symbol_offset: usize,
+    },
+    LinkerDefined(WasmLinkerSymbol),
+}
+
 #[derive(Debug, Clone, Copy)]
 struct GotFuncEntry {
     def_symbol_id: SymbolId,
-    /// Layout-input object that holds the definition (or the undefined weak import).
-    object_index: usize,
-    /// Offset of that function symbol within the object's linking symbol table.
-    symbol_offset: usize,
+    target: GotFuncTarget,
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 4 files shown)


lapla-cogito:
This patch enables SpiderMonkey to build (though it still fails to run).

libwild/src/wasm.rs
```diff
@@ -3679,6 +3858,134 @@ fn finalize_got_mem_import_resolutions(
     Ok(())
 }
 
+/// Resolve GOT.func placeholders after weak-undef function imports are absorbed.
+fn resolve_got_func_entry_defs(
```
davidlattimore: Do we need resolve_got_func_entry_defs? Could each GOT.func entry retain its originating object and symbol index then use the object’s function_indices mapping to look up the final slot in function_table_slots?

libwild/src/wasm.rs
```diff
@@ -3679,6 +3858,134 @@ fn finalize_got_mem_import_resolutions(
...
+}
+
+fn resolve_got_func_def(
+    def_id: SymbolId,
+    layout_inputs: &[WasmObjectLayoutInput<'_>],
+    resolutions: &[ObjectImportResolutions],
+    symbol_db: &SymbolDb<'_, Wasm>,
+    file_id_to_index: &HashMap<crate::input_data::FileId, usize>,
+) -> Result<GotFuncDef> {
+    let def_file_id = symbol_db.file_id_for_symbol(def_id);
```
davidlattimore: Do we care about linker-defined symbols here? e.g.

```c
//#CompArgs:-fPIC

__attribute__((constructor)) static void init(void) {}

void __wasm_call_ctors(void);

typedef void (*Fn)(void);

static Fn get_ctors(void) { return __wasm_call_ctors; }

void _start(void) { get_ctors()(); }
```

libwild/src/wasm.rs
lapla-cogito: I might have tried to implement it too closely following the same approach as GOT.mem...

libwild/src/wasm.rs
lapla-cogito: While such code might be rare I think, I've implemented support for it.

libwild/src/wasm.rs
```diff
@@ -3249,6 +3286,56 @@ impl GotMem {
     }
 }
 
+/// Where a GOT.func slot's table index comes from.
+#[derive(Debug, Clone, Copy)]
+enum GotFuncTarget {
```
davidlattimore: I feel like the code could be simplified further if we got rid of this enum, but maybe it's easier if I send a follow-up PR.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: Indeed GOT.func only needs to retrieve the table slot from the output function index, so this was unnecessary.

davidlattimore (approved):
Much simpler. Thanks!

PR #2333, opened 2026-08-04
port(wasm32): add version and help options

marxin:


lapla-cogito (approved):
Thanks! Regarding the PR title, if the main sentence is "add common options," I would suggest using "chore" as the prefix rather than "port." This prevents ambiguous entries in the changelog. If maintaining "port" is preferred, the main sentence should be more specific I think.

PR #2340, opened 2026-08-04
port(wasm): warn on -mllvm option values

marxin:
Similarly to Mach-O port, let's warn on these options for now. If someone links without the LTO, then the options have no impact on the final binary.

Issue: #1431

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -246,6 +246,12 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
 
     parser
         .declare_with_param()
+        .short("mllvm")
+        .help("Pass an LLVM option")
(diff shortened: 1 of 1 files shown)


libwild/src/args/wasm.rs
```diff
@@ -244,6 +244,12 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
             Ok(())
         });
 
+    parser
+        .declare_with_param()
+        .short("mllvm")
```
lapla-cogito: Is this not a long option?

```
（▰╹◡╹）❯  wasm-ld --help|rg mllvm
  --mllvm=<value>         Additional arguments to forward to LLVM's option processing
```

```suggestion
        .long("mllvm")
```

marxin: Right, good point!

lapla-cogito (approved):

PR #2341, opened 2026-08-04
chore: Bump MSRV to 1.97.1

lapla-cogito:


lapla-cogito:
Ah, since we now use prebuilt Docker images for CI, do we need to update the Docker image before raising the MSRV?

davidlattimore:
Yeah, I'm afraid so

lapla-cogito:
Nix's newer GCC (via flake update) now emits TLSDESC for some local-dynamic tests so I pinned the TLS dialect to gnu to match expectations.

davidlattimore (approved):
Thanks for doing this!

PR #2342, opened 2026-08-04
feat: implement setting section attributes in linker scripts

vishruth-thimmaiah:
Adds support for section attributes in a section output definition in a linker script.

Support for the `TYPE = type` command is not included in this PR, as we do not have any way currently to define an input section without an implicit type of it's own (for example, using the `LONG` command).

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2910,7 +2910,7 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
         Some(SectionIdentity::new(name, ()))
     }
 
-    fn set_section_attributes(
+    fn apply_linker_script_attributes(
         linker_script_attributes: &linker_script::SectionAttributes,
         mut output_attributes: Self::SectionAttributes,
     ) -> Self::SectionAttributes {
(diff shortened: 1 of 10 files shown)


libwild/src/elf.rs
```diff
@@ -2906,6 +2909,28 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
     ) -> Option<SectionIdentity<'data, Self>> {
         Some(SectionIdentity::new(name, ()))
     }
+
+    fn set_section_attributes(
```
davidlattimore: I'm not sure about the name "set" here, since to me it implies mutation. "merge" might fit better. Or perhaps even include "linker_script" in the name given that it's linker-script specific. Maybe something like "apply_linker_script_attributes"?

libwild/src/elf.rs
```diff
@@ -2906,6 +2909,28 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
...
+        linker_script_attributes: &linker_script::SectionAttributes,
+        mut output_attributes: Self::SectionAttributes,
+    ) -> Self::SectionAttributes {
+        match linker_script_attributes {
+            linker_script::SectionAttributes::Noload => {
+                output_attributes.ty = sht::NOBITS;
+                output_attributes.overrides.has_fixed_type = true;
+            }
+            linker_script::SectionAttributes::Readonly => {
+                output_attributes.overrides.avoid_progpogation = shf::WRITE | shf::EXECINSTR;
```
davidlattimore: Can you add a test to make sure that we handle "AX" sections correctly?

vishruth-thimmaiah: Done, added tests for `AX` as well as `A` sections.

davidlattimore (approved):

PR #2343, opened 2026-08-05
port(MachO): Propagate input section flags

joanna-jo:
I'm planning on removing the built-in regular sections and instead having them go through custom section handling to support arbitrary sections. Basic section flags functionality (namely `is_executable`, `is_no_bits`, and `apply`) is needed to propagate output section attributes from the input section flags.

The new linker-diff output also caught a pre-existing bug where `__got` wasn't being defined with `S_NON_LAZY_SYMBOL_POINTERS`, so this includes a fix for that as well.

Part of #2248

### DIFF
--- a/libwild/src/macho.rs
+++ b/libwild/src/macho.rs
@@ -53,6 +53,14 @@ use object::macho::N_EXT;
 use object::macho::N_PEXT;
 use object::macho::N_SECT;
 use object::macho::N_WEAK_DEF;
+use object::macho::S_ATTR_EXT_RELOC;
+use object::macho::S_ATTR_LOC_RELOC;
+use object::macho::S_ATTR_PURE_INSTRUCTIONS;
+use object::macho::S_ATTR_SOME_INSTRUCTIONS;
+use object::macho::S_GB_ZEROFILL;
+use object::macho::S_THREAD_LOCAL_ZEROFILL;
+use object::macho::S_ZEROFILL;
+use object::macho::SECTION_ATTRIBUTES;
 use object::macho::SEG_LINKEDIT;
 use object::macho::Section64;
 pub use object::macho::SectionFlags;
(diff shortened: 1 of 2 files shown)


libwild/src/macho.rs
```diff
@@ -769,9 +783,11 @@ impl platform::SectionAttributes for SectionAttributes {
 
     fn apply(
         &self,
-        _output_sections: &mut crate::output_section_id::OutputSections<Self::Platform>,
-        _section_id: crate::output_section_id::OutputSectionId,
+        output_sections: &mut crate::output_section_id::OutputSections<Self::Platform>,
+        section_id: crate::output_section_id::OutputSectionId,
     ) {
+        let info = output_sections.section_infos.get_mut(section_id);
+        info.section_attributes.flags |= self.flags.without(SECTION_FLAGS_PROPAGATION_MASK);
```
davidlattimore: Looking at the docs for `SectionFlags`, it sounds like its composed of a type and attributes. Bitwise or is fine for the attributes, but not for the type. If there are two different types, their bits will or together and we'll get some unrelated third type. The same problem seems to exist in the `merge` method above.

davidlattimore (commented):
Are there any tests that could be added for this?

libwild/src/macho.rs
joanna-jo: Thanks for the catch! I mistakenly thought it could be or'd together because of how `merge` was implemented. I think representing the type and attributes separately in `SectionAttributes` seem cleaner, so I included that in the latest change.

davidlattimore (commented):
A few tests fail when I run them locally. The failure is:

```
    section.__got.type
      wild 0x0
      ld 0x6
      lld 0x6
```

This wasn't picked up by our CI because we were missing some configuration in our CI. That is fixed by #2348. Can you sync your change past that? You possibly need to create a test-config.toml on your local machine to see the failures too.

joanna-jo:
> A few tests fail when I run them locally. The failure is:
> 
> ```
>     section.__got.type
>       wild 0x0
>       ld 0x6
>       lld 0x6
> ```
> 
> This wasn't picked up by our CI because we were missing some configuration in our CI. That is fixed by #2348. Can you sync your change past that? You possibly need to create a test-config.toml on your local machine to see the failures too.

Ah sorry I somehow missed that - thanks again for adding this into CI! It looks like this wasn't a regression, but the new linker-diff output in this change caught a pre-existing bug where `__got` wasn't being defined with `S_NON_LAZY_SYMBOL_POINTERS`. The tests are passing on my end now that this is added.

davidlattimore (approved):

PR #2344, opened 2026-08-05
chore: Don't special-case __wasm_call_dtors

davidlattimore:


lapla-cogito (approved):
Ah, I didn't know [this](https://github.com/llvm/llvm-project/blob/3d6ac5a9ba252cbaf8d12cecef89d015f1ffe0f7/lld/wasm/Driver.cpp#L1407-L1410). Thanks.

PR #2345, opened 2026-08-05
port(wasm): Error on undefined strong symbols by default and add `--allow-undefined`

lapla-cogito:
commits:

1. Define the linker-defined `__global_base` at the start of static data (after the stack when `--stack-first`), as expected by wasi-libc and similar inputs once undefined symbols are enforced.

2. Non-weak undefined functions/globals without `EXPLICIT_NAME`, and unresolved data symbols, are link errors by default.

3. Support `--allow-undefined` which leave unresolved strong functions (as host imports) / undefined data.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -3483,6 +3483,10 @@ fn report_disallowed_unresolved_imports<'data>(
     resolutions: &[ObjectImportResolutions],
     symbol_db: &SymbolDb<'data, Wasm>,
 ) -> Result {
+    if symbol_db.args.allow_undefined {
+        return Ok(());
+    }
+
     let mut errors: Vec<String> = Vec::new();
     let mut seen: HashSet<(String, String)> = HashSet::new();
 
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -3521,6 +3521,7 @@ fn report_disallowed_unresolved_imports<'data>(
                 WasmSymbolKind::Data => {
                     let symbol_id = input.symbol_id_range.offset_to_id(sym_offset);
                     symbol_db.definition(symbol_id) == symbol_id
+                        && live_relocs_reference_symbol(input, sym_offset)
                 }
                 _ => false,
             };
@@ -3543,6 +3544,17 @@ fn report_disallowed_unresolved_imports<'data>(
         return Ok(());
     }
     bail!("{}", errors.join("\n"));
+}
+
+fn live_relocs_reference_symbol(input: &WasmObjectLayoutInput<'_>, sym_offset: usize) -> bool {
+    let Ok(idx) = u32::try_from(sym_offset) else {
+        return false;
+    };
+    input
+        .code_relocations
+        .iter()
+        .chain(input.data_relocations.iter())
+        .any(|reloc| reloc.index == idx)
 }
 
 fn collect_shared_unresolved_imports<'data>(
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -3518,11 +3518,6 @@ fn report_disallowed_unresolved_imports<'data>(
                             .get(idx)
                             .is_some_and(|r| matches!(r, ImportResolution::Unresolved))
                 }
-                WasmSymbolKind::Data => {
-                    let symbol_id = input.symbol_id_range.offset_to_id(sym_offset);
-                    symbol_db.definition(symbol_id) == symbol_id
-                        && live_relocs_reference_symbol(input, sym_offset)
-                }
                 _ => false,
             };
             if !is_unresolved {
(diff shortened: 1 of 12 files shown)


wild/tests/sources/wasm/allow-undefined/allow-undefined.c
```diff
@@ -0,0 +1,21 @@
+//#Config:function
+//#LinkArgs: --allow-undefined
+//#RunEnabled: false
+//#ExpectFuncImport: env/missing_host_fn=1
+//#ExpectFuncImportCount: 1
+
+//#Config:data
+//#CompArgs: -DTEST_DATA
+//#LinkArgs: --allow-undefined
```
davidlattimore, line 9: If I add -fPIC to the flags, this test fails. Same for the data test in `undefined-strong.c`.

libwild/src/wasm.rs
```diff
@@ -5126,6 +5191,8 @@ pub(crate) enum WasmLinkerSymbol {
     // Data
     #[strum(serialize = "__data_end")]
     DataEnd,
+    #[strum(serialize = "__global_base")]
```
davidlattimore, line 6031: I'd be somewhat tempted to have at least the "__global_base" commit as a separate PR, unless there's some reason why it needs to be together with the other changes? But I don't feel strongly, so if you want to keep all three commits together in a single PR, that's also fine.

davidlattimore, line 6031: It'd be good if we had integration tests for things like `__global_base` for the usual reasons - mostly it let's us verify against wasm-ld. But if it's not practical for some reason, then the unit tests will do.

wild/tests/sources/wasm/allow-undefined/allow-undefined.c
lapla-cogito: Changed to allocate GOT.mem slots even for undefined data

libwild/src/wasm.rs
lapla-cogito: Added.

lapla-cogito: When actually splitting into multiple PRs, we'd include the `__global_base` change first before adding the remaining changes. Otherwise, we'd get linker errors on CI. So this approach would work for splitting, but either method seems equally valid. If you have a strong preference, I'll split accordingly.

davidlattimore (commented):
It looks like wasm-ld doesn't emit errors for undefined symbols if the references to those symbols get GCed. Should we do the same? e.g.:

```c
extern int missing_data;

void dead_function(void) { missing_data = 1; }

void _start(void) {}
```

lapla-cogito:
> It looks like wasm-ld doesn't emit errors for undefined symbols if the references to those symbols get GCed. Should we do the same? e.g.:
> 
> ```c
> extern int missing_data;
> 
> void dead_function(void) { missing_data = 1; }
> 
> void _start(void) {}
> ```

Thanks, we should match that behavior.

wild/tests/sources/wasm/undefined-strong/undefined-strong.c
```diff
@@ -0,0 +1,62 @@
...
+
+//#Config:data-pic-gc-dead-no-gc
+//#CompArgs: -DTEST_DATA -DTEST_GC_DEAD -fPIC
+//#LinkArgs: --no-gc-sections
+//#ExpectError: undefined symbol: missing_data
+
+#ifdef TEST_DATA
+extern int missing_data;
+
+#ifdef TEST_GC_DEAD
```
davidlattimore: All these preprocessor directives make this kind of hard to read. Do you think we'd be better off with a separate test?

libwild/src/wasm.rs
```diff
@@ -3474,6 +3478,85 @@ impl<'data> SharedUnresolvedImports<'data> {
...
+    bail!("{}", errors.join("\n"));
+}
+
+fn live_relocs_reference_symbol(input: &WasmObjectLayoutInput<'_>, sym_offset: usize) -> bool {
+    let Ok(idx) = u32::try_from(sym_offset) else {
+        return false;
+    };
+    input
+        .code_relocations
+        .iter()
```
davidlattimore: Iterating over all the relocations each time we encounter an undefined symbol seems like it has the potential to be problematic from a performance perspective. For ELF, we just error when we process the relocation. Why can't we do the same for Wasm?

Also, I think this approach, due to only scanning the relocations for the current object, can have us miss errors that we should emit. e.g.:

```patch
diff --git a/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object.c b/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object.c
new file mode 100644
index 0000000000..94eec87acc
--- /dev/null
+++ b/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object.c
@@ -0,0 +1,6 @@
+//#Object:undefined-data-cross-object2.c
+//#ExpectError: undefined symbol: missing_data
+
+extern int missing_data;
+
+void _start(void) { missing_data = 1; }
diff --git a/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object2.c b/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object2.c
new file mode 100644
index 0000000000..f59e340cf9
--- /dev/null
+++ b/wild/tests/sources/wasm/undefined-data-cross-object/undefined-data-cross-object2.c
@@ -0,0 +1,3 @@
+extern int missing_data;
+
+void dead_function(void) { missing_data = 2; }
```

lapla-cogito: Aligned with that.

wild/tests/sources/wasm/undefined-strong/undefined-strong.c
lapla-cogito: Divided.

davidlattimore (approved):

PR #2347, opened 2026-08-05
refactor: Make GC phase more platform-neutral

davidlattimore:


davidlattimore:
@lapla-cogito  - The idea of this PR is that instead of always working with section indexes during the GC phase, we can work with a platform-specific GcUnit. For ELF (and currently Mach-O) this is still a section index. For Wasm it's currently an empty enum, but can hopefully at some point be an enum that selects between functions, data, globals etc. Loading of GC roots is done in `activate_object_gc`. For ELF and Mach-O, this calls into a common section-based loading function that remains in `layout`.

lapla-cogito (approved):
Thanks! This would help me a lot to implement GC for Wasm.

PR #2350, opened 2026-08-06
feat: support redefining linker script and defsym symbols

vishruth-thimmaiah:
This PR adds support for redefining symbols in a linker script or via the `--defsym` flag.

### DIFF
--- a/wild/tests/sources/elf/defsym/defsym.c
+++ b/wild/tests/sources/elf/defsym/defsym.c
@@ -33,6 +33,11 @@
 //#ReferenceLinkers:
 //#ExpectError:Symbol 'notexist' referenced by --defsym
 
+//#Config:repeat_defs:default
+//#LinkArgs:--defsym=bar=0x1000 --defsym=bar=bar*2 --defsym=baz=0x5000
+//#ExpectSym:bar address=0x2000
+//#ExpectSym:baz address=0x5000
+
 int foo(void) { return 0; }
 
 void _start(void) { foo(); }
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -514,13 +514,10 @@ fn update_redirect_resolutions<'data, P: Platform>(
     verbose_timing_phase!("Update symdef resolutions");
 
     for group in &symbol_db.groups {
-        let mut symbol_id = group.symbol_id_range().start();
-
         match group {
             Group::Prelude(prelude) => {
                 for def_info in &prelude.symbol_definitions {
                     update_defsym_symbol_resolution(
-                        symbol_id,
                         def_info,
                         symbol_db,
                         resolutions,
@@ -530,14 +527,12 @@ fn update_redirect_resolutions<'data, P: Platform>(
                         sizeof_headers,
                         &[],
                     )?;
-                    symbol_id = symbol_id.next();
                 }
             }
             Group::LinkerScripts(scripts) => {
                 for script in scripts {
                     for def_info in &script.parsed.symbol_defs {
                         update_defsym_symbol_resolution(
-                            symbol_id,
                             def_info,
                             symbol_db,
                             resolutions,
@@ -547,7 +542,6 @@ fn update_redirect_resolutions<'data, P: Platform>(
                             sizeof_headers,
                             resolved_location_counters,
                         )?;
-                        symbol_id = symbol_id.next();
                     }
                 }
             }
@@ -561,7 +555,6 @@ fn update_redirect_resolutions<'data, P: Platform>(
 }
 
 fn update_defsym_symbol_resolution<'data, P: Platform>(
-    symbol_id: SymbolId,
     def_info: &InternalSymDefInfo<'data, P>,
     symbol_db: &SymbolDb<'data, P>,
     resolutions: &mut [Option<Resolution<P>>],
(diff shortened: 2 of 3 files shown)


libwild/src/symbol_db.rs
```diff
@@ -1636,8 +1637,13 @@ fn load_linker_script_symbols<'data, P: Platform>(
...
 ) {
-    for (offset, definition) in script.parsed.symbol_defs.iter().enumerate() {
-        let symbol_id = script.symbol_id_range.offset_to_id(offset);
+    let mut next_expected_offset = 0;
+    for (definition, offset) in &script.parsed.symbol_defs {
+        if offset.as_usize() < next_expected_offset {
+            continue;
+        }
+        next_expected_offset = offset.as_usize() + 1;
+        let symbol_id = script.symbol_id_range.offset_to_id(offset.as_usize());
```
davidlattimore: The fact that we're converting the symbol ID to a usize, then using that as an offset to produce a new SymbolId suggests that `offset` wasn't really representing a SymbolId to begin with - it's an offset. But it'd actually be good to see if there are alternatives to having it there at all. Is this here mostly so that we can handle multiple assignments to the same symbol? Do we actually need a SymbolId for earlier overridden definitions of the symbol? For example, a rough outline that comes to mind - I'm likely missing important considerations - is:

* Keep a hashmap from name to definition.
* As a name is redefined, replace its definition in the map after making use of the old definition if it was referenced. So `a=1; a=a+2` becomes just `a=1+2` and we end up with just a single symbol definition.

At some point we need to put those definitions into a deterministic order. The two ways that come to mind are (a) based on the order in which we first saw a definition or (b) some kind of sorted order, e.g. alphabetical. I suspect it doesn't matter which we use, so long as it's deterministic.

wild/tests/sources/elf/defsym/defsym.c
```diff
@@ -33,6 +33,10 @@
 //#ReferenceLinkers:
 //#ExpectError:Symbol 'notexist' referenced by --defsym
 
+//#Config:repeat_defs:default
+//#LinkArgs:--defsym=bar=0x1000 --defsym=bar=bar*2
```
davidlattimore: If I make the following change to this test it shows that an additional --defsym after `bar` doesn't work properly:

```patch
diff --git a/wild/tests/sources/elf/defsym/defsym.c b/wild/tests/sources/elf/defsym/defsym.c
index 68d0fe1101..0709247275 100644
--- a/wild/tests/sources/elf/defsym/defsym.c
+++ b/wild/tests/sources/elf/defsym/defsym.c
@@ -21,8 +21,9 @@
 //#ExpectSym:bar address=0x3333
 
 //#Config:symbol_expr:default
-//#LinkArgs:--defsym=bar=0x1111 --defsym=baz=bar+0x2222
+//#LinkArgs:--defsym=bar=0x1111 --defsym=baz=bar+0x2222 --defsym=foo=0x4444
 //#ExpectSym:baz address=0x3333
+//#ExpectSym:foo address=0x4444
 
 //#Config:symbol_expr2:default
 //#LinkArgs:--defsym=bar=0x2222 --defsym=baz=-0x1111+bar
```

I suspect the same bug exists for linker scripts.

davidlattimore (commented):
It's unfortunate that there's duplication of logic between how we handle these flags and the equivalent concepts in linker scripts. I think we should do some refactoring at some point to reduce that duplication. It doesn't necessarily have to be done now though and plausibly should be a separate PR either before or after this one. I'm also happy to have a go at doing the deduplication if you'd like.

libwild/src/symbol_db.rs
vishruth-thimmaiah: Replacing the expression might work for `--defsym`, however this might not work in a linker script when we need to use both variants of a symbol - for example:
```ld
a = 0x1000;
.text a : {
  *(.text)
}

a = a * 2;
.data a : {
  *(.data)
}
```

While wild doesn't yet support symbols in section properties and we could technically use this approach, we would have to eventually track each expression individually for cases like the above.

wild/tests/sources/elf/defsym/defsym.c
vishruth-thimmaiah: This fails without this PR's changes as well - foo is defined in the c file as well as through `--defsym`. The linker currently uses the value provided by the object instead of using `--defsym`. Perhaps this should be fixed in another PR?

vishruth-thimmaiah:
I can work on deduplication in a follow up PR.

wild/tests/sources/elf/defsym/defsym.c
davidlattimore: I agree, let's leave that one until a later PR.

wild/tests/sources/elf/defsym/defsym.c
```diff
@@ -33,6 +33,10 @@
...
+//#ExpectSym:bar address=0x2000
+
```
davidlattimore, line 40: Here's another test that fails. I'm less sure that this one should be left until a later PR since it seems to be a bug in the feature that this PR is implementing:

```patch
diff --git a/wild/tests/sources/elf/defsym/defsym.c b/wild/tests/sources/elf/defsym/defsym.c
index 68d0fe1101..ef5bbfcc82 100644
--- a/wild/tests/sources/elf/defsym/defsym.c
+++ b/wild/tests/sources/elf/defsym/defsym.c
@@ -34,8 +34,9 @@
 //#ExpectError:Symbol 'notexist' referenced by --defsym
 
 //#Config:repeat_defs:default
-//#LinkArgs:--defsym=bar=0x1000 --defsym=bar=bar*2
+//#LinkArgs:--defsym=bar=0x1000 --defsym=bar=bar*2 --defsym=baz=0x5000
 //#ExpectSym:bar address=0x2000
+//#ExpectSym:baz address=0x5000
 
 int foo(void) { return 0; }
 
```

libwild/src/symbol_db.rs
davidlattimore: Fair enough that replacing the expression might not work, but I still think we should find an alternative to the design change that this PR makes to `symbol_definitions`. Having it not be indexed by local symbol index seems most unfortunate. It seems to be because we're trying to do two things at once with `symbol_definitions`. It's being used both for symbol definitions (it's original purpose) and also now as a way of storing an ordered assignment history of a defsym name. Might there be a way to separate these two uses so that indexing into `symbol_definitions` by local symbol offset can remain valid and we don't need the extra usize to be stored?

davidlattimore (approved):

PR #2351, opened 2026-08-06
CI: factor out image-tag to a single place

marxin:
Not fully convinced if it's a straightforward improvement - the prerequisite job takes **4s**.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,18 @@ env:
   RUSTFLAGS: '-D warnings'
 
 jobs:
+  ci-env:
+    name: CI environment
+    runs-on: ubuntu-24.04
+    outputs:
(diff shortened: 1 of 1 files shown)


davidlattimore (commented):
Thanks for looking into this. One of many annoying things about github actions that it doesn't provide a nice way to do this. I'm also unsure that this would be worth it. I think perhaps we can just have instructions that say "do a replace-all from the old version to the new version". There's probably some things we could do to streamline it a little. e.g. I suspect we should consider folding external_tests.yml into ci.yml. That will make it less likely that we forget to move it to the new image (which I think has already happened). Some of the build jobs look very similar and could probably be turned into a single matrix job - e.g. PPC64LE, LoongArch64 and RISC-V.

marxin:
All right - I have a similar feeling the suggested approach makes the CI more complicated. In any case, I can iterate on the CI job simplification over the weekend.

PR #2227, opened 2026-07-17
port(wasm): Synthesize `__wasm_first_page_end` and `__heap_end` for WASI heap

lapla-cogito:
Basic Rust programs linked with Wild were previously trapped at runtime:

```
（▰╹◡╹）❯  echo 'fn main() { println!("foo"); }' > wasm_rust.rs

（▰╹◡╹）❯  rustc --target wasm32-wasip1 wasm_rust.rs -o wasm_rust.wild.wasm -C linker=../../repos/wild/target/debug/wild -C linker-flavor=wasm-ld

（▰╹◡╹）❯  wasmtime wasm_rust.wild.wasm
Error: failed to run main module `wasm_rust.wild.wasm`

Caused by:
    0: failed to invoke command default
    1: error while executing at wasm backtrace:
    0:  0x36160 - <unknown>!<wasm function 1171>
    1:  0x356e6 - <unknown>!<wasm function 1170>
    2:   0x80b1 - <unknown>!<wasm function 212>
    3:   0x1bbc - <unknown>!<wasm function 103>
    4:   0x1c4e - <unknown>!<wasm function 110>
    5:  0x15396 - <unknown>!<wasm function 408>
    6:   0x1b61 - <unknown>!<wasm function 100>
    7:   0x1a7d - <unknown>!<wasm function 95>
    8:   0x1a51 - <unknown>!<wasm function 94>
    9:   0x1a9c - <unknown>!<wasm function 96>
   10:  0x13e17 - <unknown>!<wasm function 379>
   11:   0x1a21 - <unknown>!<wasm function 93>
   12:   0x1bae - <unknown>!<wasm function 102>
   13:   0x19d2 - <unknown>!<wasm function 92>
    2: wasm trap: wasm `unreachable` instruction executed
```

The function executing the `unreachable` instruction is shown below:

```
（▰╹◡╹）❯  wasm-objdump -d wasm_rust.wild.wasm | sed -n '/func\[1170\]/,/func\[1172\]/p' | head
0356e3 func[1170]:
 0356e4: 20 00                      | local.get 0
 0356e6: 10 93 89 80 80 00          | call 1171
 0356ec: 0b                         | end
0356ef func[1171]:
 0356f0: 0b 7f                      | local[1..11] type=i32
 0356f2: 23 81 80 80 80 00          | global.get 1
 0356f8: 41 10                      | i32.const 16
 0356fa: 6b                         | i32.sub
 0356fb: 22 01                      | local.tee 1
```

[...]

PR #2352, opened 2026-08-06
CI: build more Docker images (for Wasm target builds)

marxin:


davidlattimore (commented):
Thanks for doing this! The wasm builds were failing for a while in CI the other day due to an infra issue, so it'll be good to have them using docker images to avoid that. It'll be necessary to add the new image in one PR, merge it, do a build, then switch to using the new image in a later PR.

mati865:
Do we need up to date Wasm tools? Ubuntu is probably a bui outdated.

marxin:
> Do we need up to date Wasm tools? Ubuntu is probably a but outdatedm

@lapla-cogito :point_up: What do you think?

lapla-cogito:
> > Do we need up to date Wasm tools? Ubuntu is probably a but outdatedm
> 
> @lapla-cogito ☝️ What do you think?

Considering the current Wasm port's target (wasm32/wasip1), using Ubuntu 26.04 should be fine (though Ubuntu 24.04 did have known issues: https://github.com/wild-linker/wild/pull/2227#discussion_r3602727302).

davidlattimore (approved):

PR #2354, opened 2026-08-06
port(MachO): Add support for (segment name, section name) identification

joanna-jo:
Adds support for (segment name, section name) section identification with our fixed supported segments. (Full arbitrary segments will be implemented in a later diff.) Regular Mach-O sections are now dynamically discovered and routed through custom section handling, with only linker-generated or otherwise "special" sections in the default built-in section definitions/rules for Mach-O instead.

Additionally, in Mach-O, section flags don't contain a writability bit and section writability can only be determined by the segment. Since we support only a fixed number of segments right now, `__DATA` and `__DATA_CONST` are classified as writable, matching the protections passed to their  `write_segment` calls in `write_segment_commands`. 

There are also a few TODOs that remained or were introduced by this change, which I documented.

Part of #2248

### DIFF
--- a/libwild/src/macho_writer.rs
+++ b/libwild/src/macho_writer.rs
@@ -926,20 +926,21 @@ fn write_chained_fixup_table(layout: &MachOLayout, chained_fixup_table: &mut [u8
     header.imports_format.set(DYLD_CHAINED_IMPORT);
     header.symbols_format.set(0);
 
-    let data_const_segment_index = active_segments
-        .iter()
-        .position(|segment_type| segment_type.segment_type == SegmentType::DataConstSections);
-
     // 2) fill up dyld_chained_starts_in_image, which is `seg_count` (u32) followed by
     //    `seg_info_offset` ([u32; seg_count]); only __DATA_CONST,__got segment is covered
     starts_in_image[0].set(LE, segment_count as u32);
     starts_in_image[1..].fill(U32::new(LE, 0));
 
     // Early exit if we don't have any GOT entry to be encoded.
-    let Some(data_const_segment_index) = data_const_segment_index else {
+    if layout.section_layouts.get(output_section_id::GOT).mem_size == 0 {
         rest.zero();
         return Ok(());
-    };
+    }
+
+    let data_const_segment_index = active_segments
+        .iter()
+        .position(|segment| segment.segment_type == SegmentType::DataConstSections)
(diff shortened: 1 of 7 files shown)


libwild/src/macho_writer.rs
```diff
@@ -926,20 +926,21 @@ fn write_chained_fixup_table(layout: &MachOLayout, chained_fixup_table: &mut [u8
...
-        .position(|segment_type| segment_type.segment_type == SegmentType::DataConstSections);
-
     // 2) fill up dyld_chained_starts_in_image, which is `seg_count` (u32) followed by
     //    `seg_info_offset` ([u32; seg_count]); only __DATA_CONST,__got segment is covered
     starts_in_image[0].set(LE, segment_count as u32);
     starts_in_image[1..].fill(U32::new(LE, 0));
 
     // Early exit if we don't have any GOT entry to be encoded.
-    let Some(data_const_segment_index) = data_const_segment_index else {
+    if layout.section_layouts.get(output_section_id::GOT).mem_size == 0 {
```
joanna-jo, line 935: Now that the presence of a `__DATA_CONST` segment doesn't mean that `__got` exists, this needed to be changed so that we check explicitly for  `__got` instead.

joanna-jo:
I don't have permissions to request reviews until a reviewer reviews first, but would it be possible to get your feedback on this change @marxin @davidlattimore? 

(There's also currently an issue with GitHub Actions and I see that CI didn't run. Everything passes locally for me, but I can resubmit later once this is fixed.)

marxin:
> (There's also currently an issue with GitHub Actions and I see that CI didn't run. Everything passes locally for me, but I can resubmit later once this is fixed.)

Usually what help's is force pushing the commit (just `git commit -a --amend`), typically it helps.

marxin:
> Usually what help's is force pushing the commit (just `git commit -a --amend`), typically it helps.

Apparently connected to an incident: https://www.githubstatus.com/incidents/qcvjkzcs7j74

davidlattimore (approved):

davidlattimore:
> I don't have permissions to request reviews until a reviewer reviews first, but would it be possible to get your feedback on this change

Interesting. It looks like github only lets people request reviews if they have write access. Github permissions seem weirdly restrictive to me. I usually look through the open PRs for stuff that needs reviewing, although sometimes one slips my attention for a day or two. It's always fine to ping me.

PR #2355, opened 2026-08-06
chore(--help): align columns of all the help texts

marxin:
Addresses:
```
OPTIONS:
    @<VALUE>                        Read options from a file
    -L <VALUE>                      Add directory to library search path
    -R <VALUE>                      Add runtime library search path
    -l <VALUE>                      Link with library
      -l :filename                      Link with specific file
      -l libname                        Link with library libname.so or libname.a
    -m <VALUE>                      Set target architecture
      -m aarch64elf                     AArch64 ELF target
      -m aarch64linux                   AArch64 ELF target (Linux)
      -m elf64loongarch                 LoongArch 64-bit ELF target
      -m elf64lppc                      PowerPC64 LE ELF target
```

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -885,7 +885,7 @@ impl<T: platform::Args> ArgumentParser<T> {
 
         // TODO: This is ad-hoc
         help.push_str(&format!(
-            "    {:<31} Read options from a file\n",
+            "    {:<30} Read options from a file\n",
             "@<VALUE>",
         ));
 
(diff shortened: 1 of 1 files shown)


libwild/src/args.rs
```diff
@@ -935,8 +932,9 @@ impl<T: platform::Args> ArgumentParser<T> {
                     } else {
                         sub_name.to_string()
                     };
+                    let option_name = format!("-{prefix} {display_name}");
                     help.push_str(&format!(
-                        "      -{prefix} {display_name:<30} {sub_help}\n",
+                        "     {option_name:<29} {sub_help}\n",
```
davidlattimore: I assume the two 30s and the 29 are all related? Should we extract these to a constant, e.g. `" {option_name:<width$}`", width=HELP_COL1_WIDTH - 1`?

marxin: Sure, like the idea.

davidlattimore (approved):

PR #2356, opened 2026-08-07
perf: Parallelize archive member open/parse

lapla-cogito:
Walk regular and thin archive headers serially, then process members with rayon. This has minimal impact on ELF builds, but provides perf benefits when building for Wasm, where large archive files are common.

### DIFF
--- a/libwild/src/input_data.rs
+++ b/libwild/src/input_data.rs
@@ -500,32 +500,45 @@ fn process_archive<'data, P: Platform, F: FileSystem>(
     file: Option<&Arc<std::fs::File>>,
     state: &TemporaryState<'data, P, F>,
 ) -> Result<LoadedFileState<'data, P, F::Input>> {
-    let mut outputs = Vec::new();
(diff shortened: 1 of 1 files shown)


libwild/src/input_data.rs
```diff
@@ -500,33 +500,46 @@ fn process_archive<'data, P: Platform, F: FileSystem>(
...
+                data: &archive_data[start_offset..end_offset],
+                entry: Some(EntryMeta {
+                    identifier: ident,
+                    start_offset,
+                    end_offset,
+                }),
+            };
+            state.process_input(member_ref, file, kind)
+        })
+        .collect::<Result<Vec<_>>>()?;
```
davidlattimore, line 541: I suspect this will mean that if there are multiple errors, it'll be non-deterministic which one gets returned. That said, we have plenty of other places where our error reporting is non-deterministic, so one more isn't the end of the world. We should probably do something about them at some point.

davidlattimore (approved):

PR #2353, opened 2026-08-06
port(wasm): Support GC for functions, globals, data, and imports

lapla-cogito:
In ELF, the GC unit is the input section, but in Wasm it encompasses multiple types: defined functions, defined globals, data segments, and function/global import slots.

Part of #1431

PR #2357, opened 2026-08-07
port(wasm): Ignore `--no-merge-data-segments`

mati865:
Noticed it being used by wasm-component-ld tests, and upon inspecting the binaries with hex viewer Wild appears not to merge data segments.

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -352,6 +352,22 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
             Ok(())
         });
 
+    parser
+        .declare()
+        .long("merge-data-segments")
+        .help("Merge data segments")
+        .execute(|args, _modifier_stack| {
+            // TODO: implement and make it default.
+            args.warn_unsupported("--merge-data-segments")?;
+            Ok(())
+        });
+
+    parser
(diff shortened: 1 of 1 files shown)


lapla-cogito (approved):
Thanks!

> Wild appears not to merge data segments.

Yeah we should do that. https://github.com/wild-linker/wild/pull/2353#issuecomment-5207324880

libwild/src/args/wasm.rs
```diff
@@ -352,6 +352,22 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
...
+        .execute(|args, _modifier_stack| {
+            // TODO: implement and make it default.
+            args.warn_unsupported("--merge-data-segments")?;
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("no-merge-data-segments")
+        .help("Disable symbol demangling")
```
lapla-cogito, line 368: Ah, I noticed the help text isn't quite right... Either way, this feature should be implemented soon (though I'm not sure if I'll be responsible for it), so it would be good to be able to fix it then.

mati865: Oh, shoot. I abandoned wrong revision, will post the fix. Sorry and good catch.

PR #2359, opened 2026-08-07
nrel-fix: Update description of `--no-merge-data-segments`

mati865:


lapla-cogito (approved):
How about making the PR title "Update description of `--no-merge-data-segments`" or something like that?

PR #2360, opened 2026-08-07
CI: port more CI jobs to Docker images

marxin:


lapla-cogito (commented):
It'd be good to also update the revision in `external_tests.yml`.

lapla-cogito (approved):

PR #2367, opened 2026-08-08
CI: unify cross builds with a matrix

marxin:


davidlattimore (approved):
Thanks for the cleanup :)

PR #2369, opened 2026-08-08
chore: use CodeSignatureEncoder from object crate

marxin:
We can actually rely on one of the latest additions to the lovely crate called `object`.

CC: @philipc


davidlattimore (approved):
Nice simplification :)

libwild/src/macho_writer.rs
```diff
@@ -1072,72 +1070,51 @@ fn write_code_signature_metadata(
...
+    };
+
+    let mut rest: &mut [u8] = code_signature;
+    encoder.signature_super_blob(&mut rest, code_signature_section.file_size as u32, 1);
+    encoder.blob_index(&mut rest, CSSLOT_CODEDIRECTORY, CS_BLOB_HEADERS_SIZE as u32);
+    let padding_size = CS_BLOB_HEADERS_SIZE as usize
+        - encoder.super_blob_size() as usize
+        - encoder.blob_index_size() as usize;
+    let (padding, mut rest) = rest.split_at_mut(padding_size);
+    padding.zero();
```
philipc, line 1111: Thanks for the CC, it's nice to see how the APIs get used in practice. This looks like we need to add something to make it simpler, but I'm also curious if this padding is actually needed. I guess you are matching lld here, but I didn't see any padding like this in the Apple code.

marxin: Good observation, let's simplify it then: #2378.

Issue #2320, opened 2026-08-02
Windows support (PE/COFF): offering to build a first version


thewh1teagle:
Windows is on the README's list of larger unsupported features, and I couldn't find an existing issue or discussion for it, so here's one. To be clear up front: I'm not asking you to build this. I want to build it myself, and CONTRIBUTING.md says to reach out before starting larger work, so I'm reaching out.

What I have in mind for a first version: x86-64 only, linking real MSVC-produced COFF objects, archives and import libraries into exes and DLLs that actually run. Imports, exports, base relocations, the common link.exe-style flags (/ENTRY, /SUBSYSTEM, defaultlib handling), and nothing else. In particular no PDB output at first. Debug info is by far the biggest and least documented part of a Windows linker, and I'd rather have that design conversation separately once plain binaries link.

Questions before I write code:

1. Do you want this in-tree at all right now? Windows sits behind other things on your priority list, and if you'd rather land more architectures or the incremental core first, I'd like to know before I entrench anything.

2. How would you want the format boundary drawn? The codebase is ELF throughout today. A parallel coff/pe module sharing the parallel infrastructure is one option, a deeper trait-based abstraction is another. If you have a preference I'll follow it, since this decision is the expensive one to redo.

3. Incremental linking is wild's stated end goal, and Windows might be where it matters most: link.exe's incremental mode is still the one thing that beats a fast full relink in the edit-compile-run loop, and it's also the main reason Windows developers tolerate link.exe. Should the PE writer be shaped for incrementality from the start, or would you rather keep v1 boring and revisit?

On development setup: I'm on macOS, so the plan is cross-development. clang-cl with xwin for genuine MSVC-ABI test inputs, lld-link as the local reference to diff against, llvm-readobj for inspection, and a Windows CI job as the "does the exe run" gate.

If you'd prefer this start as a Zulip thread, or as a smaller warm-up PR so you can see how I work first, that's fine by me.

davidlattimore:
Thanks! I definitely appreciate communication and coordination for such large bits of work!

> Do you want this in-tree at all right now? Windows sits behind other things on your priority list, and if you'd rather land more architectures or the incremental core first, I'd like to know before I entrench anything.

In tree is definitely preferable. Doing significant work outside of the tree is likely to lead to merge conflicts if changes need to be made to common infrastructure - and they almost certainly will. Also, we prefer to review code in manageable chunks. That will also allow us to provide feedback on the design before it's too set in stone.
 
[...]

PR #2374, opened 2026-08-08
port(coff): add MSVC link.exe-style argument parsing

thewh1teagle:
First step toward Windows PE/COFF support as discussed in #2320. This is argument parsing only, no linking yet (`Args::Coff` bails with "not yet implemented").

* Adds a `Coff` platform flavor, selected via `-flavor link` or a `link`/`lld-link` executable name.
* Parses the core link.exe style options: `/OUT`, `/ENTRY`, `/SUBSYSTEM`, `/DLL`, `/MACHINE` (X64 only), `/LIBPATH`, `/DEFAULTLIB` and `/NODEFAULTLIB`. Option names and values are case insensitive and accept either `/` or `-` prefixes, with values attached as `NAME:value`.
* The remaining options that rustc emits for `x86_64-pc-windows-msvc` (`/DEBUG`, `/OPT`, `/NXCOMPAT`, `/PDBALTPATH`, `/IMPLIB`, `/WHOLEARCHIVE`, `/NATVIS`, `/DEF`, `/guard` and so on) are recognised and currently ignored, so a real rustc link line parses cleanly. Anything else is reported through the shared `report_unrecognized()` path.
* MSVC options don't fit the declarative GNU style parser, so this uses a small hand rolled token loop. Happy to restructure if you'd prefer a different shape.
* I went with MSVC style first per the discussion in #2320. GNU style (MinGW) args can be added later.
* Not included yet: `@file` response files (rustc emits them on long link lines) and everything past parsing.

Tested beyond the unit tests by replaying real link invocations through the parser: a captured rustc hello world link line for `x86_64-pc-windows-msvc`, and the response files from four real release links (Rust std, ripgrep, rust-analyzer and uv, up to ~250 arguments) previously captured with lld-link's `/reproduce`. All of them parse cleanly and reach the intended "not yet implemented" bail, while injected bogus flags are still reported as unrecognized.

### DIFF
--- a/libwild/src/args.rs
+++ b/libwild/src/args.rs
@@ -27,6 +27,7 @@ use itertools::Itertools;
 use jobserver::Acquired;
 use jobserver::Client;
 use rayon::ThreadPoolBuilder;
+use std::borrow::Cow;
 use std::fmt::Display;
 use std::io::Write;
 use std::num::NonZeroUsize;
@@ -635,10 +636,62 @@ pub(crate) enum UnresolvedSymbols {
     IgnoreAll,
 }
 
+/// Describes how a platform spells its options. GNU-style platforms use the default, whereas
+/// link.exe-style platforms accept a `/` prefix, ignore case and attach values with `:`.
+#[derive(Clone, Copy)]
+struct OptionSyntax {
+    /// Prefixes that introduce an option. Matched in order, so longer prefixes must come first.
+    prefixes: &'static [&'static str],
+
+    /// The character that attaches a value to an option name, e.g. `=` in `--foo=bar`.
+    value_separator: char,
+
+    case_insensitive: bool,
+
+    /// Whether an option that takes a value may instead take it from the following token. When
+    /// false, as for link.exe, the value must be attached to the option name, so a missing value is
+    /// an error, as is a value supplied to an option that doesn't take one.
+    allows_separate_value: bool,
+}
+
+impl Default for OptionSyntax {
+    fn default() -> Self {
+        Self {
+            prefixes: &["--", "-"],
+            value_separator: '=',
+            case_insensitive: false,
+            allows_separate_value: true,
+        }
+    }
+}
+
+impl OptionSyntax {
+    /// Strips the first matching option prefix from `arg`, returning the rest of the argument.
+    fn strip_prefix<'a>(&self, arg: &'a str) -> Option<&'a str> {
+        self.prefixes
+            .iter()
+            .find_map(|prefix| arg.strip_prefix(prefix))
+    }
+
+    fn starts_with_prefix(&self, arg: &str) -> bool {
+        self.strip_prefix(arg).is_some()
+    }
+
+    /// Returns the key under which an option with this name should be stored and looked up.
+    fn lookup_key<'a>(&self, name: &'a str) -> Cow<'a, str> {
+        if self.case_insensitive {
+            Cow::Owned(name.to_ascii_lowercase())
+        } else {
+            Cow::Borrowed(name)
+        }
+    }
+}
+
 struct ArgumentParser<T> {
-    options: HashMap<&'static str, OptionHandler<T>>, // Long option lookup
-    short_options: HashMap<&'static str, OptionHandler<T>>, // Short option lookup
+    options: HashMap<String, OptionHandler<T>>, // Long option lookup
+    short_options: HashMap<String, OptionHandler<T>>, // Short option lookup
     prefix_options: HashMap<&'static str, PrefixOptionHandler<T>>, // For options like -L, -l, etc.
+    syntax: OptionSyntax,
 }
 
 impl<T: platform::Args> Default for ArgumentParser<T> {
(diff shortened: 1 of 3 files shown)


libwild/src/args/coff.rs
```diff
@@ -0,0 +1,616 @@
...
+                }
+                _ => args.no_default_libraries = true,
+            },
+            name if IGNORED_NO_VALUE_OPTIONS.contains(&name) => {
+                no_value(arg, value)?;
+            }
+            name if IGNORED_OPTIONS.contains(&name) => {}
+            // A prefixed token whose name isn't in the table above is an input only if that name
+            // looks like a path, since on Unix hosts absolute paths begin with `/` too.
+            _ if name.contains(['/', '\\']) => add_input(args, arg),
```
davidlattimore: Does link.exe do anything like this? It might be better to leave such behaviour our for now. We can always add it in later if we conclude that we definitely need to do this.

libwild/src/args/coff.rs
```diff
@@ -0,0 +1,616 @@
...
+    "manifest",
+    "manifestinput",
+    "manifestuac",
+    "wx",
+];
+
+/// Ignored options that take no value, so `/NOLOGO:x` is an error just like `/DLL:x` is.
+const IGNORED_NO_VALUE_OPTIONS: &[&str] = &["nologo"];
+
+/// MSVC options don't fit the GNU-style declarative parser used by the other platforms: names are
```
davidlattimore: It doesn't seem to me that there's anything stopping us from having a declarative parser like we have for the other platforms. The main difference is that flags are prefixed with '/' rather than '-' or "--". I'd hope that we could also share most of the code that handled that declarative parsing.

libwild/src/args/coff.rs
```diff
@@ -0,0 +1,616 @@
...
+            "/NXCOMPAT",
+            "/NXCOMPAT:NO",
+            "/DYNAMICBASE:NO",
+            "/OPT:REF,NOICF",
+            "/PDBALTPATH:%_PDB%",
+            "/PDB:out.pdb",
+            "/IMPLIB:out.lib",
+            "/NATVIS:types.natvis",
+            "/DEF:exports.def",
+            "/WHOLEARCHIVE",
```
davidlattimore: Do we issue warnings when unsupported options are passed? I'm thinking in particular of options like "/WHOLEARCHIVE". If we ignore it, then the library that we were supposed to be linking will be dropped from the inputs. Likely we'll then get an unresolved symbol error. It'll be pretty hard to trace that back to the fact that "/WHOLEARCHIVE" isn't yet supported. So, I'd say we should at least issue a warning, if not error for things that we don't yet support.

davidlattimore (commented):
I'd say at this point, we should mostly error if we encounter something that we don't yet support. So for example, if we are passed a response file, we should give an error saying that it's not supported. I think at the moment it will treat it like an input, which will likely give a confusing error. Do you think we should identify response file arguments and give a better error?

libwild/src/args/coff.rs
marxin: Wanted to notice that we would benefit from the option prefix customization in the Mach-O platforms. Right now, we support both `-` and `--`, but only the former one is suppose to be supported by the official linker.

mati865: MSVC's link.exe supports both `/` and `-`, although docs mention only `/`.

libwild/src/args/coff.rs
thewh1teagle: After digging I found that it's not. Windows paths never start with `/`. Will remove

libwild/src/args/coff.rs
thewh1teagle: Thanks, moving to the shared declarative parser. Kept both `/` and `-`

libwild/src/args/coff.rs
thewh1teagle: Not yet, adding warnings for those.

libwild/src/args.rs
```diff
@@ -621,10 +636,62 @@ pub(crate) enum UnresolvedSymbols {
...
+        } else {
+            Cow::Borrowed(name)
+        }
+    }
+}
+
 struct ArgumentParser<T> {
-    options: HashMap<&'static str, OptionHandler<T>>, // Long option lookup
-    short_options: HashMap<&'static str, OptionHandler<T>>, // Short option lookup
+    options: HashMap<String, OptionHandler<T>>, // Long option lookup
```
davidlattimore: I'd probably have a slight preference for keeping this as a `'static str`. The methods that insert into these maps can then assert that if case-insensitive mode is active, that the supplied str is already lowercase.

davidlattimore (commented):
Looks like you need to run `cargo +nightly fmt`. BTW, once this first PR is merged, CI should run automatically with your subsequent PRs.

libwild/src/args.rs
thewh1teagle: fixed!

davidlattimore (approved):

PR #2377, opened 2026-08-09
port(wasm): Redirect defined weak functions to the winning definition

lapla-cogito:
Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2240,6 +2240,7 @@ pub(crate) struct WasmObjectIndexMap {
     pub(crate) data_addresses: Vec<u32>,
     pub(crate) got_mem_globals: Vec<Option<u32>>,
     pub(crate) got_func_globals: Vec<Option<u32>>,
+    pub(crate) function_symbol_redirects: Vec<Option<u32>>,
 }
 
 impl WasmObjectIndexMap {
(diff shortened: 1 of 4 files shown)


libwild/src/wasm.rs
```diff
@@ -4290,6 +4305,47 @@ fn apply_got_func_to_index_maps(object_index_maps: &mut [WasmObjectIndexMap], go
...
 
+/// Map defined weak functions to the winning definition's output index.
+fn fill_function_symbol_redirects(
+    object_index_maps: &mut [WasmObjectIndexMap],
+    layout_inputs: &[WasmObjectLayoutInput<'_>],
+    symbol_db: &SymbolDb<'_, Wasm>,
+    file_id_to_index: &HashMap<crate::input_data::FileId, usize>,
+) {
+    for (obj_idx, input) in layout_inputs.iter().enumerate() {
+        let mut redirects = vec![None; input.symbols.len()];
```
davidlattimore, line 4316: The generic layout code has `symbol_resolutions`, which serves the same purpose. At some point we should consider making the wasm port use it, since I'm not aware of any reason it couldn't. But I think it'd be a larger refactoring than would be appropriate for this PR.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: I don't think we can use that as a drop-in here. Wasm relocations are applied from the per-object index maps during `build_output_module_layout`, which runs before `symbol_resolutions` is populated. Even afterwards, a function symbol's `raw_value` is the input function index, not the output one.

Putting the output function index on `Resolution` (like in `format_specific`) and routing Wasm relocs through `merged_symbol_resolution` would remove this table. If this aligns with what you intended, I can understand but yeah it would certainly be a major refactoring.

PR #2381, opened 2026-08-10
port(wasm): Support `--export-memory`

lapla-cogito:
This option defaults to exporting "memory" in wasm-ld:

```
（▰╹◡╹）❯  wasm-ld --help | rg export-memory
  --export-memory=<value> Export the module's memory with the passed name
  --export-memory         Export the module's memory with the default name of "memory"
```

Part of #1431


davidlattimore (commented):
How do exports with `--export-memory` interact with memory exports listed by the input objects, both those with the same name and with different names? Can you add some tests for this?

davidlattimore (approved):
To what extent are we worried about `--export-memory=` with a special function name - e.g. `_start`? Assuming that this is enough of a corner-case, then this looks good to me.

PR #2383, opened 2026-08-10
port(MachO): Add support for arbitrary segments

### LINKED ISSUE #2248: Mach-O: Identify sections by segment name and section name

joanna-jo:
Adds support for arbitrary segments. Segments and their sections are ordered from `__PAGEZERO` -> `__TEXT` -> `__DATA_CONST` -> `__DATA` -> arbitrary segments -> `__LINKEDIT`. For all arbitrary segments, we assign RW protections, matching ld and lld behavior.

I also (hopefully) simplified the design a bit. Since we have `SectionIdentityExt` to differentiate between actual section identities from synthetic output regions (such as `FILE_HEADER`, `LOAD_COMMANDS`), I replaced `SegmentType` and `count_as_segment` from `ProgramSegmentDef`, instead making `ProgramSegmentDefs` the source of truth for a single `LC_SEGMENT_64` with `SegmentName`, `VmProt`, and `SegmentFlags`.  Synthetic regions are still represented with `OutputSectionId` and explicitly associated with their corresponding segment.

I wanted to note:
* Since we have arbitrary segments, we can no longer use `P::program_segment_defs` in `OutputOrderBuilder` for Mach-O - this has been replaced with `self.segment_defs`, although ELF and Wasm still pass in `program_segment_defs`.
* Similarly, having arbitrary segments means that we need to use `layout.segment_layouts.segments` throughout various places in the Mach-O writer.
* This means we have direct access to the segment bounds/sizes, and we no longer need `get_segment_sections` to both return the sizes and the sections of the segment. `get_segment_sections` now only collects the section information for each segment, and I removed `SegmentSectionsInfo`.

Resolves #2248

### DIFF
--- a/libwild/src/macho.rs
+++ b/libwild/src/macho.rs
@@ -22,7 +22,13 @@ use crate::layout::SymbolCopyInfo;
 use crate::layout::SymbolResolutions;
 use crate::layout_rules::SectionKind;
 use crate::layout_rules::SectionRule;
+use crate::macho::output_section_id::CHAINED_FIXUP_TABLE;
+use crate::macho::output_section_id::CODE_SIGNATURE;
+use crate::macho::output_section_id::LOAD_COMMANDS;
+use crate::macho::output_section_id::STRTAB;
+use crate::macho::output_section_id::SYMTAB_GLOBAL;
 use crate::macho_writer;
+use crate::output_section_id::FILE_HEADER;
 use crate::output_section_id::OrderEvent;
 use crate::output_section_id::OutputOrderBuilder;
 use crate::output_section_id::OutputSectionId;
@@ -34,6 +40,8 @@ use crate::part_id::PartId;
 use crate::platform;
 use crate::platform::Args;
 use crate::platform::ObjectFile;
+use crate::program_segments::ProgramSegmentId;
+use crate::program_segments::ProgramSegments;
 use crate::resolution;
 use crate::symbol_db::SymbolId;
 use crate::symbol_db::Visibility;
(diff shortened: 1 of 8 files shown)


libwild/src/macho.rs
davidlattimore, line 158: Are we still fine with this being 6?

davidlattimore (approved):
Looks good! Just one question below...

libwild/src/macho.rs
joanna-jo: Thanks so much for looking this over! Since this was already getting a bit big in terms of changes, I left that as a TODO for chained fixups since `MAX_SEGMENT_COUNT` was only used there as a temporary workaround. I think this would likely require some extra plumbing to get the exact retained segments, although I'm still happy to add this to the diff if we'd like everything bundled together.

davidlattimore: Dealing with it later in separate PR sounds good to me.

davidlattimore (approved):

PR #2384, opened 2026-08-10
feat: support compound assignment operators in linker scripts

vishruth-thimmaiah:
Support the use of complex assignment operators such as `+=` and `-=` in linker scripts, when defining a symbol/location counter.

### DIFF
--- a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
+++ b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
@@ -5,7 +5,16 @@ SECTIONS {
         *(.text*)
         symbol3 = 0x3000;
     }
-    .data : { *(.data*) }
+    symbol11 = 0x100;
+    symbol11 -= 0x20;
+    symbol11 &= 0x80;
+    .data : {
+        *(.data*)
+        symbol10 = 0x100;
+        symbol10 *= 0x100;
+        symbol10 ^= 0x40;
+        ASSERT(symbol10 == 0x10040, "symbol10 must be 0x100");
+    }
     .bss  : { *(.bss*)  }
 }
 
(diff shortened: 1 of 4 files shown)


wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -45,3 +45,10 @@ ASSERT(!DEFINED(symbol5), "symbol5 should not be defined");
 ASSERT(DEFINED(symbol6), "symbol6 must be defined");
 ASSERT(DEFINED(symbol7), "symbol7 must be defined");
 ASSERT(!DEFINED(symbol8), "symbol8 should not be defined");
+
+symbol9 = 0;
```
davidlattimore: Can you make the initial value non-zero? Otherwise if `+=` behaved as `=`, the test would still pass.

davidlattimore (commented):
Can you test that the operators work correctly inside `SECTIONS` too?

wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -5,7 +5,16 @@ SECTIONS {
...
-    .data : { *(.data*) }
+    symbol11 = 0x100;
+    symbol11 -= 0x20;
+    symbol11 &= 0x80;
+    .data : {
+        *(.data*)
+        symbol10 = 0x100;
+        symbol10 *= 0x100;
+        symbol10 ^= 0x40;
+        ASSERT(symbol10 == 0x10040, "symbol10 must be 0x100");
```
davidlattimore: The message here doesn't match what's being asserted. Same with the assert in the next file.

davidlattimore (approved):

PR #2386, opened 2026-08-10
port(MachO): Support dynamic export of symbols

davidlattimore:


libwild/src/trie.rs
```diff
@@ -0,0 +1,401 @@
...
+                uncompressed.push(UncompressedNode {
+                    representative: symbol_index,
+                    depth,
+                    ..Default::default()
+                });
+                uncompressed[parent].children.push(child);
+                previous_path.push(child);
+                parent = child;
+            }
+            debug_assert!(uncompressed[parent].symbol.replace(symbol_index).is_none());
```
joanna-jo: I think the CI failures are from this line and how the symbol isn't written when debug assertions are disabled?

davidlattimore: Well spotted! You're quite right. Thanks!

PR #2387, opened 2026-08-11
port(wasm): Support `--max-memory`

lapla-cogito:
This option allows specifying the maximum size of linear memory.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -5059,11 +5059,14 @@ const fn wasm_page_size() -> u64 {
 /// Maximum linear-memory size in bytes for wasm32.
 const WASM32_MAX_MEMORY_BYTES: u64 = (1u64 << 32) - wasm_page_size();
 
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -5059,11 +5059,14 @@ const fn wasm_page_size() -> u64 {
 /// Maximum linear-memory size in bytes for wasm32.
 const WASM32_MAX_MEMORY_BYTES: u64 = (1u64 << 32) - wasm_page_size();
 
+const WASM32_MAX_MEMORY_LIMIT_BYTES: u64 = 1u64 << 32;
```
davidlattimore: The name of this constant is very similar to the one above. Are there names that would make the difference between them more apparent? At the moment the difference is the word "limit", but "limit" and "max" have pretty similar meaning.

lapla-cogito: renamed to `WASM32_MAX_INITIAL_MEMORY_BYTES` and `WASM32_ADDRESS_SPACE_BYTES`

davidlattimore (approved):

PR #2389, opened 2026-08-11
port(wasm): Speed up name section construction

lapla-cogito:
Replace collect/sort/dedup with dense first-wins tables, and collect per-object symbol names in parallel.

Part of #1431


davidlattimore (approved):
Nice speedup :)

PR #2390, opened 2026-08-11
fix: Follow GNU ld mandatory relaxations logic better

mati865:
Found by trial and error, so they might be not entirely correct.

You might wonder why not also add static PIE test case. Well, GNU ld (version 2.47) produces broken binary, so I wouldn't trust it:
```
(gdb) disas
Dump of assembler code for function _start:
   0x00007ffff7c26460 <+0>:	endbr64
   0x00007ffff7c26464 <+4>:	xor    %ebp,%ebp
   0x00007ffff7c26466 <+6>:	mov    %rdx,%r9
   0x00007ffff7c26469 <+9>:	pop    %rsi
   0x00007ffff7c2646a <+10>:	mov    %rsp,%rdx
   0x00007ffff7c2646d <+13>:	and    $0xfffffffffffffff0,%rsp
   0x00007ffff7c26471 <+17>:	push   %rax
   0x00007ffff7c26472 <+18>:	push   %rsp
   0x00007ffff7c26473 <+19>:	xor    %r8d,%r8d
   0x00007ffff7c26476 <+22>:	xor    %ecx,%ecx
   0x00007ffff7c26478 <+24>:	mov    0x22d1a9(%rip),%rdi        # 0x7ffff7e53628
=> 0x00007ffff7c2647f <+31>:	call   *0x22e403(%rip)        # 0x7ffff7e54888
   0x00007ffff7c26485 <+37>:	hlt
End of assembler dump.
(gdb) si
0x00000000001216f0 in ?? ()
(gdb) disas
❌️ No function contains program counter for selected frame.
(gdb) si

Program received signal SIGSEGV, Segmentation fault.
0x00000000001216f0 in ?? ()
```

### DIFF
--- a/libwild/src/elf_x86_64.rs
+++ b/libwild/src/elf_x86_64.rs
@@ -212,7 +212,7 @@ impl crate::platform::Arch for ElfX86_64 {
                             return Some(Relaxation {
                                 kind: RelaxationKind::RexMovIndirectToAbsolute(inst_offset),
                                 rel_info: rel_info_from_type!(object::elf::R_X86_64_32),
-                                mandatory: output_kind.is_static_executable(),
+                                mandatory: output_kind.is_static_executable() && is_absolute,
                             });
                         }
                         // sub *x(%rip), reg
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):
Thanks for figuring these out :)

PR #2391, opened 2026-08-12
port(MachO): Use `UnlinkAndReplace` for outputs

### LINKED ISSUE #2375: Mach-O: In-place relinking can trigger code signature `SIGKILL`
I've been noticing that Mach-O integration tests intermittently fail with `SIGKILL (Code Signature Invalid), namespace CODESIGNING, code 2` and `rejecting invalid page ... (cs_mtime:1786214319... != mtime:1786220483...)`. After investigating, it seems this is because the tests reuse the existing output paths with `UpdateInPlaceWithFallback`, preserving its inode. If the executable has already been run, the macOS kernel may retain code signature state associated with its vnode; if Wild then writes different contents and a new code signature in-place, this invalidates that state and kills the process the next time it runs. See https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/kern/ubc_subr.c#L5586-L5591.

I also verified that `codesign --verify --strict` is successful and that a copy linked using a new inode executes successfully. The simplest fix for this looks like changing Mach-O output to use `UnlinkAndReplace`, although I wonder if there's a better way to do this?

joanna-jo:
On macOS, Wild now uses `UnlinkAndReplace` as the default replacement mode for regular Mach-O output files because of the issue described in #2375. I considered putting this in `OsOutputFile` instead since that's where we also implement the logic for `MS_INVALIDATE`, but I couldn't find a cleaner way here to see if Mach-O was being emitted.

Resolves #2375

### DIFF
--- a/libwild/src/file_writer.rs
+++ b/libwild/src/file_writer.rs
@@ -8,6 +8,7 @@ use crate::error;
 use crate::error::Context as _;
 use crate::error::Result;
 use crate::fs::FileReplacementMode;
+use crate::fs::FileType;
 use crate::fs::FileWriteMode;
 use crate::layout::GroupLayout;
 use crate::layout::Layout;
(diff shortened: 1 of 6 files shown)


librust_out.rlib
davidlattimore, line 1: Accidentally added file?

joanna-jo: Ack sorry about that! I was testing something last minute and didn't see that it got added in - should be removed now.

davidlattimore (approved):

Issue #2380, opened 2026-08-09
Integrate LLD ELF test suite into Wild's CI (similar to Mold tests)


deepakshirkem:
As suggested by @marxin, it would be useful to integrate the LLD ELF test suite into Wild's CI similar to how Mold tests are currently integrated via `wild/tests/external_tests/mold_tests.rs`. This would allow automatic tracking of LLD compatibility across architectures (x86_64, AArch64, RISC-V, etc.) and prevent regressions.

Main challenges to discuss:
- LLD tests use LLVM's lit framework (llvm-mc, split-file, FileCheck)
- Need a skip list for unsupported features
- Which architectures to support initially

marxin:
Thanks for opening the topic - I feel @mati865 made an integration experiments with the LLD linker some time ago. Can you summarize your investigations and what would be the best design?

mati865:
I wrote how to run them using CMake (it sill builds LLD) in: https://github.com/wild-linker/wild/issues/932
There was also some discussion on Zulip: https://wild.zulipchat.com/#narrow/channel/508469-general/topic/Using.20LLD.20tests.20with.20Wild/with/525747530

I didn't dig much deeper into that.

> LLD tests use LLVM's lit framework (llvm-mc, split-file, FileCheck)

Maybe we can use system one? This probably will limit it to bleeding edge distros like Arch Linux, but at least should make it easy for the initial phase.

deepakshirkem:
Based on the discussion, I think the vendoring approach makes the most sense:

1. Vendor just `lld/test/ELF/` into `external_test_suites/lld/` (much smaller than the full LLVM repo).
2. Write an update script to sync it periodically.
3. Use system LLVM tools (`llvm-mc`, `FileCheck`, `split-file`) that are available on Arch Linux and similar distros.
4. Replace `ld.lld` with `wild` in the `RUN:` lines.
5. Use a skip list for unsupported features.

We have already manually evaluated 127 AArch64 and 70 x86_64 LLD tests, so we have a good starting point for the skip list.

We could start with the x86_64 tests since we have the most data there, and then expand to AArch64 and RISC-V.

Does this approach sound reasonable? I'm still open to discussing other approaches or anything else we could add to this plan. I'd really appreciate your input before I start, but if this looks good, I can put together a proof of concept.

@mati865 @marxin @davidlattimore

deepakshirkem:
[...]

PR #2392, opened 2026-08-12
feat: Add infrastructure for LLD ELF test integration

deepakshirkem:
Adds basic infrastructure to run LLD's ELF test suite with Wild as the linker, similar to the existing Mold test integration.
Tests are vendored (not submoduled) to avoid pulling in the full LLVM repository. A script is provided to sync tests from llvm-project.

Part of   #2380

### DIFF
--- a/external_test_suites/lld/test/ELF/README.md
+++ b/external_test_suites/lld/test/ELF/README.md
@@ -2,6 +2,6 @@
 
 This directory contains tests vendored from llvm-project/lld/test/ELF/.
 
-To update these tests, run: scripts/update-lld-tests.sh
+To update these tests, run: external_test_suites/lld/update-lld-tests.sh
 
 These tests are run with Wild substituted for ld.lld to verify compatibility.
--- a/external_test_suites/lld/test/ELF/README.md
+++ b/external_test_suites/lld/test/ELF/README.md
@@ -1,0 +1,7 @@
+# LLD ELF Tests
+
+This directory contains tests vendored from llvm-project/lld/test/ELF/.
+
+To update these tests, run: scripts/update-lld-tests.sh
+
+These tests are run with Wild substituted for ld.lld to verify compatibility.
--- a/wild/tests/external_tests/lld_skip_tests.toml
+++ b/wild/tests/external_tests/lld_skip_tests.toml
@@ -1,0 +1,3 @@
+[skipped_groups.unsupported_features]
+reason = "These tests use features not yet supported by Wild"
+tests = []
--- a/wild/tests/external_tests/lld_skip_tests.toml
+++ b/wild/tests/external_tests/lld_skip_tests.toml
@@ -1,3 +1,0 @@
-[skipped_groups.unsupported_features]
-reason = "These tests use features not yet supported by Wild"
-tests = []
--- a/wild/tests/external_tests/lld_tests.rs
+++ b/wild/tests/external_tests/lld_tests.rs
@@ -1,0 +1,106 @@
+//! Integration tests that run LLD's ELF test suite with Wild as the linker.
+//! Tests are vendored from llvm-project/lld/test/ELF/.
+//!
+//! Requires system LLVM tools: llvm-mc, FileCheck, split-file.
+//! These are available on distributions like Arch Linux.
+
+use crate::Result;
+use libtest_mimic::Failed;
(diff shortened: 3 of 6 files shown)


wild/tests/external_tests/mod.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "lld_tests")]
```
davidlattimore: I think it's better if we always compile the test infra code even when the feature is off. That makes it easier to work with the code in the IDE because it's not behind a cfg flag that's disabled. i.e. similar to how it's current done with the mold tests where we just skip test registration if the feature is off.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -0,0 +1,106 @@
...
+        .unwrap()
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap()
+        .join("wild");
+
+    cmd.replace("llvm-mc", llvm_mc)
+        .replace("FileCheck", filecheck)
+        .replace("ld.lld", &format!("{} -m elf_x86_64", wild.display()))
```
davidlattimore: `.display()` is only really intended for printing a path to show to a user. If the path contains non-UTF-8 content, then it'll get replaced with placeholders, so will end up referring to a file that doesn't exist. Better to just crash here rather than continue and end up with a more confusing error when the command is invoked.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+fn substitute_vars(cmd: &str, test_file: &Path, tmpdir: &Path) -> String {
+    cmd.replace("%s", &test_file.display().to_string())
+        .replace("%t", &tmpdir.join("test").display().to_string())
+        .replace("%p", &test_file.parent().unwrap().display().to_string())
+}
+
+fn substitute_tools(cmd: &str, llvm_mc: &str, filecheck: &str) -> String {
+    let wild = std::env::current_exe()
```
davidlattimore: I think you can probably use `env!("CARGO_BIN_EXE_wild")` instead.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -0,0 +1,106 @@
...
+    let filecheck = find_tool("FileCheck")?;
+
+    let content = std::fs::read_to_string(test_file)?;
+    let tmpdir = tempfile::tempdir()?;
+
+    // Extract and execute RUN: lines
+    for line in content.lines() {
+        let Some(cmd) = line
+            .trim()
+            .strip_prefix("// RUN:")
```
davidlattimore, line 66: I think we'll also need to support multi-line RUN commands. See https://www.llvm.org/docs/TestingGuide.html

wild/tests/external_tests/mod.rs
davidlattimore, line 19: Might as well call the `collect_tests` function here so that it's hooked up.

external_test_suites/lld/test/ELF/README.md
```diff
@@ -0,0 +1,7 @@
+# LLD ELF Tests
+
+This directory contains tests vendored from llvm-project/lld/test/ELF/.
+
+To update these tests, run: scripts/update-lld-tests.sh
```
davidlattimore: This script doesn't seem to exist yet. It'd be good to add it in this PR as well, but leave the actual vendoring for a subsequent PR.

davidlattimore (commented):
Thanks for working on this!

external_test_suites/lld/test/ELF/README.md
deepakshirkem: Added the script, but placed it at external_test_suites/lld/update-lld-tests.sh instead of scripts/update-lld-tests.sh since there's no existing scripts/ dir and this is specific to the LLD suite. Updated the README reference to match.

wild/tests/external_tests/lld_skip_tests.toml
davidlattimore, line 1: I'm not seeing anything hooked up to process this file. Unless I missed something?

deepakshirkem: Removed it here will add it back in the follow-up PR once the skip-list logic is wired up.

davidlattimore (approved):

PR #2394, opened 2026-08-13
fix: Don't consume the next argument for `--trace` / `-t`

lapla-cogito:
These options take no value, but declaring them with `declare_with_param()` causes them to consume the next argument.


wild/tests/external_tests/mold_skip_tests.toml
lapla-cogito, line 1: I discovered this issue while investigating the two tests. For the "Undefined symbol main" error, this patch resolves it, but the failure still occurs due to differences in the `--trace` output format. Regardless, since these tests also fail with GNU ld, the statement "passes with GNU ld" seems inappropriate.

davidlattimore (approved):
Good find!

PR #2399, opened 2026-08-15
feat: Wire up LLD tests via wild-linker/lld-tests submodule

deepakshirkem:
Replaces the directly-committed test files from earlier commits on this PR with a proper git submodule pointing at [wild-linker/lld-tests](https://github.com/wild-linker/lld-tests), which vendors the full `lld/test/` directory (ELF, COFF, MachO, MinGW, Unit, wasm) and has its own sync script.

Test activation is intentionally left as a no-op for now `collect_tests` returns early without registering any tests. The full LLD suite includes many untriaged tests, and deciding what should run vs. be skip-listed is left to a follow-up PR. This PR is scoped to just the submodule wiring.

Part of #2380.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -7,3 +7,7 @@ aranges = "aranges"
 
 [files]
 extend-exclude = ["external_test_suites", "wild/tests/bins"]
+
+[type.external-test-list]
+extend-glob = ["*_skip_tests.toml"]
+extend-ignore-re = ['(?m)^\s*tests\s*=\s*\[[^\]]*\]\s*$']
--- a/wild/tests/external_tests/lld_skip_tests.toml
+++ b/wild/tests/external_tests/lld_skip_tests.toml
@@ -1,0 +1,101 @@
+[skipped_groups.design_differences]
+reason = "Wild uses immediate binding; PLT/GOT layout differs from lld"
+tests = [
+  "x86-64-tlsdesc-ld.s",
+  "x86-64-tlsdesc-gd-mixed.s",
+  "x86-64-reloc-gotoff64.s",
+  "x86-64-reloc-gotpc64.s",
+  "x86-64-reloc-pltoff64.s",
+  "x86-64-static-tls-model.s",
+  "x86-64-combined-dynrel.s",
+  "x86-64-got-plt-header.s",
+  "x86-64-plt.s",
+  "x86-64-rela.s",
+  "x86-64-tls-gd-local.s",
+  "x86-64-tls-ld-local.s",
+  "x86-64-zrel-zrela.s",
+  "x86-64-gotpc-relax-nopic.s",
+  "x86-64-tlsdesc-gd.s",
+  "x86-64-tls-ie-local.s",
+  "x86-64-tls-ie.s",
+  "x86-64-tls-gdie.s",
+  "x86-64-tls-dynamic.s",
+]
+
+[skipped_groups.error_message_wording]
+reason = "Wild errors correctly but message text differs from lld"
+tests = [
+  "x86-64-reloc-tpoff32-error.s",
+  "x86-64-pc32-overflow-large.s",
+  "x86-64-reloc-error2.s",
+  "x86-64-reloc-debug-overflow.s",
+  "x86-64-reloc-range-debug-loc.s",
+]
+
+[skipped_groups.noinhibit_exec_unsupported]
+reason = "Needs --noinhibit-exec, unsupported by Wild"
+tests = ["x86-64-tls-le-undef.s"]
+
+[skipped_groups.unsupported_features]
+reason = "Feature not yet implemented in Wild"
+tests = [
+  "x86-64-feature-cet.s",
+  "x86-64-reloc-size.s",
+  "x86-64-reloc-size-shared.s",
+  "x86-64-split-stack-prologue-adjust-shared.s",
+  "x86-64-split-stack-prologue-adjust-silent.s",
+  "x86-64-tls-opt-noplt.s",
+  "x86-64-tls-pie.s",
+]
+
+[skipped_groups.missing_flag_support]
+reason = "Wild doesn't recognize a flag this test uses"
+tests = [
+  "x86-64-reloc-32.s",
+  "x86-64-branch-to-branch.s",
+  "x86-64-gotpc-relax.s",
+  "x86-64-plt-high-addr.s",
+  "x86-64-relax-jump-tables.s",
+]
+
+[skipped_groups.missing_validation]
+reason = "Wild doesn't error where lld does; validation gap"
+tests = [
+  "x86-64-dyn-rel-error3.s",
+  "x86-64-dyn-rel-error.s",
+  "x86-64-dyn-rel-error5.s",
+  "x86-64-reloc-error.s",
+  "x86-64-reloc-pc32.s",
+  "x86-64-reloc-range.s",
+  "x86-64-reloc-8-16.s",
+  "x86-64-tls-ie-err.s",
+  "x86-64-split-stack-prologue-adjust-fail.s",
+  "x86-64-gotpc-no-relax-err.s",
+]
+
+[skipped_groups.relocation_bugs]
+reason = "Wild fails to correctly apply a valid relocation"
+tests = ["x86-64-gotpc-relax-too-far-relr.s", "x86-64-gotpc-relax-too-far.s"]
+
+[skipped_groups.objdump_missing_x86_target]
+reason = "Requires an llvm-objdump build with x86_64 target support"
+tests = [
+  "x86-64-tls-ie-opt-local.s",
+  "x86-64-relax-got-abs.s",
+  "x86-64-retpoline.s",
+  "x86-64-retpoline-linkerscript.s",
+  "x86-64-retpoline-znow.s",
+  "x86-64-retpoline-znow-linkerscript.s",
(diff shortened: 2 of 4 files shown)


marxin (changes requested):
Frankly speaking, I don't like us taking the copy of the LLVM's `lld/test/ELF` sub-folder and tracking them as part of the main git repository. If we include all files in the aforementioned folder, we're speaking about ~2400 files.

I would rather see us using an existing submodule https://github.com/wild-linker/test-files, where we can also put the synchronization script.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -17,11 +33,18 @@ pub(crate) fn collect_tests(tests: &mut Vec<Trial>, filter: &crate::Filter) -> R
         return Ok(());
     }
 
+    // Test activation is not yet implemented. The submodule wiring
```
marxin: Well, it does not make much sense to merge the PR that is actually not running any of the tests.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -128,3 +172,41 @@ fn execute_command(cmd: &str) -> Result {
     }
     Ok(())
 }
+
+fn load_skip_tests_config() -> &'static Option<Vec<String>> {
```
marxin, line 190: Similarly here, it's a code that is not actually run.

marxin (changes requested):
I think we should include running of a selected subset of tests as part of this PR - either by explicitly listing them, or similarly to Mold, by skipping the unsupported ones.

.typos.toml
```diff
@@ -4,6 +4,7 @@ rela = "rela"
 ot = "ot"
 FRE = "FRE"
 aranges = "aranges"
+preemptable = "preemptable"
```
davidlattimore: This is a typo that I've been known to make myself, so it'd be good if we could avoid adding it here. I just did some experimentation, and it looks like we can skip typo checking specifically for the excluded test lists with something like this:

```toml
[type.external-test-list]
extend-glob = ["*_skip_tests.toml"]
extend-ignore-re = ['(?m)^\s*tests\s*=\s*\[[^\]]*\]\s*$']
```

davidlattimore (commented):
Do you think you could split adding the git submodule out to a separate PR and get that in first? Especially if we might iterate on this a bit. I often like to pull down PRs to try them out, but having a submodule addition makes it tricky because if I initialise the submodule, then once I move off the commit that adds it, things get messy. Basically it'd be good to get the submodule in quickly, then I can more easily review the remainder of the change.

I'm not sure that the "feat" prefix is ideal for this change. It would mean that this would appear in the "features" section of the release notes - but from the perspective of a user of the linker, this isn't a feature. "test" is probably more appropriate.

wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -0,0 +1,104 @@
+[skipped_groups.lld_specific]
+reason = "Test exercises an lld-specific flag or behavior not applicable to Wild"
+tests = ["x86-64-branch-to-branch.s"]
+
+[skipped_groups.requires_newer_lld]
+reason = "Test requires a newer lld/relocation type than our local llvm-mc supports"
```
marxin: If that's so - we can introduce a `llvm-mc` detection and require a specific version.

wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -0,0 +1,104 @@
...
+  "x86-64-gotpc-relax.s",
+  "x86-64-gotpc-relax-nopic.s",
+  "x86-64-plt-high-addr.s",
+  "x86-64-tlsdesc-gd.s",
+  "x86-64-tls-ie-err.s",
+  "x86-64-tls-ie-local.s",
+]
+
+[skipped_groups.design_differences]
+reason = "Wild intentionally differs from lld's PLT/GOT layout and relocation binding strategy (immediate binding vs lazy binding)"
```
marxin: Can we by default pass `-z now` to the reference linker (`lld`) and achieve a parity?

wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -0,0 +1,104 @@
...
+  "x86-64-reloc-tpoff32-error.s",
+  "x86-64-pc32-overflow-large.s",
+  "x86-64-reloc-error2.s",
+  "x86-64-reloc-debug-overflow.s",
+  "x86-64-reloc-range-debug-loc.s",
+  "x86-64-tls-ie-opt-local.s",
+]
+
+[skipped_groups.gc_sections_behavior]
+reason = "Test relies on lld's default --gc-sections behavior differing from Wild's"
```
marxin: Similarly here.

wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -0,0 +1,104 @@
...
+  "x86-64-tls-opt-noplt.s",
+  "x86-64-tls-pie.s",
+]
+
+[skipped_groups.missing_flag_support]
+reason = "Wild does not yet recognize the --image-base flag used by this test"
+tests = ["x86-64-reloc-32.s"]
+
+[skipped_groups.previously_believed_working]
+reason = "Manual testing previously suggested Wild handles this correctly, but it fails under this harness's exact FileCheck comparison. Needs investigation to reconcile: possible regression, an imprecise manual check, or a local tooling gap (e.g. objdump missing x86_64 target support)."
```
marxin: In general, the generated LLM wording for all the `reasons` is too long - please reduce it.

wild/tests/external_tests/lld_tests.rs
```diff
@@ -7,11 +7,37 @@
...
+    tests: Vec<String>,
+}
+
+static SKIP_TESTS_NAME: OnceLock<Option<Vec<String>>> = OnceLock::new();
 
 const PREFIX: &str = "external_test_suites/lld";
 
+/// Architecture prefixes used by LLD's test file naming convention
+/// (e.g. `x86-64-pcrel.s`, `aarch64-abs16.s`).
+const SUPPORTED_ARCH_PREFIXES: &[&str] = &["x86-64-"];
```
marxin: I would call it `SUPPORTED_ARCHS` and put there `x86-64`, then I would use it directly without the extra `is_supported_arch_test`.

marxin (changes requested):
Thanks for iterating on that - generally speaking, I would like to see more passing tests being included under the test harness, but I guess it's something you're planning to happen once this PR gets merged, right?

davidlattimore (commented):
When I run these tests on my machine, two of the tests fail due to missing FileCheck. If there's system configuration required, can you document it somewhere in `CONTRIBUTING.md`?

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: Done.

wild/tests/external_tests/lld_tests.rs
deepakshirkem: Done.

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: Done.

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: @marxin Sorry, I miscategorized the tests this wasn't actually an issue with a newer lld. I've now fixed and re-verified all the tests, but I'd like your input on the categorization

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: lld, default:

```
Relocation section '.rela.plt' contains 2 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
00000000002033f8  0000000100000007 R_X86_64_JUMP_SLOT     0000000000000000 weak + 0
0000000000203400  0000000200000007 R_X86_64_JUMP_SLOT     0000000000000000 bar + 0
```

lld, with -z now:

```
Relocation section '.rela.plt' contains 2 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
00000000002023e8  0000000100000007 R_X86_64_JUMP_SLOT     0000000000000000 weak + 0
00000000002023f0  0000000200000007 R_X86_64_JUMP_SLOT     0000000000000000 bar + 0
```

Wild:

```
Relocation section '.rela.dyn' contains 2 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
00000000004028d8  0000000100000006 R_X86_64_GLOB_DAT      0000000000000000 weak + 0
00000000004028e0  0000000200000006 R_X86_64_GLOB_DAT      0000000000000000 bar + 0
```

-z now only changes when lld's dynamic linker resolves the JUMP_SLOT relocation (lazy vs eager)  it stays R_X86_64_JUMP_SLOT in .rela.plt either way.

marxin (approved):
I am fine with the initial integration, thanks!

@deepakshirkem given the existing skipped tests listing, do we have any rough estimation of how big porting will be actually possible to fully run with Wild as the linker?

davidlattimore (approved):
Looks good. Thanks for doing this!

I've got some ideas for making the tooling situation a bit easier that I'll follow up with later today.

PR #2402, opened 2026-08-15
feat: parse ASSERT in linker script expressions

vishruth-thimmaiah:
Treat `ASSERT` as a symbol definition without a symbol name. This lets us do the following:
- An `ASSERT` statement can be used within a symbol definition/location counter
- Multiple asserts on the same symbol, with different values now work, example:
```ld
foo = 1;
ASSERT(foo == 1, "foo should be 1");
foo = 10;
ASSERT(foo == 10, "foo should be 10");
```

### DIFF
--- a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
+++ b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
@@ -9,7 +9,7 @@ SECTIONS {
     symbol11_assert = ASSERT(symbol11 == 0x100, "symbol11 must be 0x100");
     symbol11 -= 0x20;
     symbol11 &= 0x80;
-    symbol11_assert = ASSERT(symbol11 == 0x80, "symbol11 must be 0x80");
+    ASSERT(symbol11 == 0x80, "symbol11 must be 0x80")
     .data : {
         *(.data*)
         symbol10 = 0x100;
--- a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
+++ b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
@@ -6,11 +6,14 @@ SECTIONS {
         symbol3 = 0x3000;
     }
     symbol11 = 0x100;
+    symbol11_assert = ASSERT(symbol11 == 0x100, "symbol11 must be 0x100");
     symbol11 -= 0x20;
     symbol11 &= 0x80;
+    symbol11_assert = ASSERT(symbol11 == 0x80, "symbol11 must be 0x80");
     .data : {
         *(.data*)
         symbol10 = 0x100;
+        symbol10_assert = ASSERT(symbol10 == 0x100, "symbol10 must be 0x100");
         symbol10 *= 0x100;
         symbol10 ^= 0x40;
         ASSERT(symbol10 == 0x10040, "symbol10 must be 0x10040");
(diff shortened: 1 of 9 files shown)


wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -6,11 +6,14 @@ SECTIONS {
...
     symbol11 -= 0x20;
     symbol11 &= 0x80;
+    symbol11_assert = ASSERT(symbol11 == 0x80, "symbol11 must be 0x80");
     .data : {
         *(.data*)
         symbol10 = 0x100;
+        symbol10_assert = ASSERT(symbol10 == 0x100, "symbol10 must be 0x100");
         symbol10 *= 0x100;
         symbol10 ^= 0x40;
         ASSERT(symbol10 == 0x10040, "symbol10 must be 0x10040");
```
davidlattimore, line 19: I just tried changing the constant value here, the test fails when run against GNU ld, but if I then disable the reference linker, the test passes, suggesting that wild isn't actually checking this assert. But this is a pre-existing bug, since I just tried the same thing on main and got the same behaviour. Could be good to fix this in a separate PR - unless there's a reason to also fix it in this one.

wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -6,11 +6,14 @@ SECTIONS {
         symbol3 = 0x3000;
     }
     symbol11 = 0x100;
+    symbol11_assert = ASSERT(symbol11 == 0x100, "symbol11 must be 0x100");
     symbol11 -= 0x20;
     symbol11 &= 0x80;
+    symbol11_assert = ASSERT(symbol11 == 0x80, "symbol11 must be 0x80");
     .data : {
```
davidlattimore, line 13: If I make the following change to this test, it fails. Is this something that we should support?

```patch
diff --git a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
index 9cbe49dca6..7beb1202ae 100644
--- a/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
+++ b/wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
@@ -11,9 +11,11 @@
     symbol11 &= 0x80;
     symbol11_assert = ASSERT(symbol11 == 0x80, "symbol11 must be 0x80");
     .data : {
+        data_start = .;
         *(.data*)
         symbol10 = 0x100;
         symbol10_assert = ASSERT(symbol10 == 0x100, "symbol10 must be 0x100");
+        symbol13_assert = ASSERT(data_start == ADDR(.data), "data_start must equal ADDR");
         symbol10 *= 0x100;
         symbol10 ^= 0x40;
         ASSERT(symbol10 == 0x10040, "symbol10 must be 0x10040");
```

wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
vishruth-thimmaiah: Looks like `ASSERT` commands within an output section were previously unsupported, we were parsing it as an input section instead. I've included parsing support within this PR, as it is a simple fix.

wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
vishruth-thimmaiah: I believe `ADDR` here is being treated as a relative value, so its final value is computedd to `. + ADDR(.data)`. This is an unrelated issue, I can raise another PR to fix this issue.

davidlattimore (approved):

PR #2411, opened 2026-08-17
fix: Apply ADD/SUB relocations correctly when reading debug sections for source info

wszqkzqk:
## Bug

These 2 tests failed on Arch Linux for Loong64:

```
     test elf/loongarch64/source-info-from-debug/non-lto ... FAILED
     test elf/loongarch64/source-info-from-debug/lto     ... FAILED
```

The test assertion in stderr requires an error with `source-info-from-debug...:27`, but wild actually reports line number `:26`.

LoongArch's `.debug_line` uses `DW_LNS_fixed_advance_pc` to encode the address advance, and its 2 byte operand is filled with a pair of R_LARCH_ADD16/R_LARCH_SUB16 relocations. The correct value should be 0x18 - 0x10 = 8.

However, wild's diagnostic code treats this as an absolute overwrite. The advance changes from 8 to 16, causing line 27 to be incorrectly mapped to address 0x20 instead of 0x18.

Therefore, when querying the `foo` call point, the range of the previous line is hit, resulting in line 26 being reported.

## Fix

Just apply the correct read-modify-write semantics. Read the current N bytes from the buffer first, then perform wrapping_add/wrapping_sub, and finally write them back.

Tested on Arch Linux for Loong64.

### DIFF
--- a/libwild/src/dwarf_address_info.rs
+++ b/libwild/src/dwarf_address_info.rs
@@ -167,11 +167,7 @@ fn apply_section_relocations<
             continue;
         };
 
-        let rel_out = section_data
-            .get_mut(data_offset..data_offset + num_bytes)
-            .context("Invalid relocation offset")?;
-
-        match r_type.kind {
+        let new_value = match r_type.kind {
             RelocationKind::Absolute => {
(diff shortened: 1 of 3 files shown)


davidlattimore (approved):
Thanks :)

PR #2413, opened 2026-08-17
fix: correctly resolve linker script functions that return addresses

vishruth-thimmaiah:
Functions such as `ADDR`, `LOADADDR`, `SEGMENT_START` and `ORIGIN` return an address instead of a number. Addresses must be treated as an absolute value, instead of being relative to an output section.

Additionally, comparison operators like `==`, etc must resolve any relative values to their absolute value before performing a comparison.

### DIFF
--- a/wild/tests/sources/elf/linker-script-region/linker-script-region.ld
+++ b/wild/tests/sources/elf/linker-script-region/linker-script-region.ld
@@ -8,7 +8,7 @@ SECTIONS {
     .data.flash : {
         KEEP(*(.data.flash))
     } > FLASH
-	ASSERT(ORIGIN(FLASH) == 0x40000000, "ROM region must start at 0x20000000")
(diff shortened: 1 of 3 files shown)


wild/tests/sources/elf/linker-script-region/linker-script-region.ld
```diff
@@ -8,6 +8,7 @@ SECTIONS {
     .data.flash : {
         KEEP(*(.data.flash))
     } > FLASH
+	ASSERT(ORIGIN(FLASH) == 0x40000000, "ROM region must start at 0x20000000")
```
davidlattimore: Message doesn't match assertion. Feel free to not include the address in the message if that makes it easier :)

vishruth-thimmaiah: fixed!

davidlattimore (approved):

PR #2414, opened 2026-08-18
port(wasm): Support `--export` of linker-defined symbols

### LINKED ISSUE #2334: Wasm: --export not found: `__wasm_call_ctors`
Noticed while testing the Wasm32 port:

```
❯ cat main.C
int main() {
  return 42;
}
❯ clang --target=wasm32-wasip1 --sysroot=/usr/share/wasi-sysroot main.C -Wl,--export=__wasm_call_ctors
❯ clang --target=wasm32-wasip1 --sysroot=/usr/share/wasi-sysroot main.C -Wl,--export=__wasm_call_ctors -fuse-ld=/home/marxin/Programming/wild/fakes/wasm-ld
wild: error: symbol exported via --export not found: __wasm_call_ctors
```

lapla-cogito:
Look up linker-synthesized symbols for `--export` / `--export-if-defined` and materialize them.

Fixes #2334
Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -1672,6 +1672,9 @@ fn build_name_section<'data>(
     if let Some(idx) = indices.tls_base_global {
         set_name_first_wins(&mut global_names, idx, "__tls_base");
     }
+    for &(known, idx) in &indices.data_address_globals {
+        set_name_first_wins(&mut global_names, idx, <&str>::from(known));
+    }
     if let Some(got_base) = indices.got_mem_global_base {
         got_mem_names.reserve(got_mem.entries.len());
         for (i, entry) in got_mem.entries.iter().enumerate() {
@@ -3945,6 +3948,7 @@ struct LinkerDefinedIndices {
     /// First module global index for GOT.func entries.
     got_func_global_base: Option<u32>,
     got_func_count: u32,
+    data_address_globals: Vec<(WasmLinkerSymbol, u32)>,
 }
 
 /// Where a GOT.mem slot's final linear-memory address comes from.
@@ -4061,8 +4065,10 @@ fn setup_got_mem_and_indices<'data>(
             shared_imports.function_count(),
             shared_imports.global_count(),
             weak_undef_stubs,
-            LinkerDefinedIndexRequest {
+            &LinkerDefinedIndexRequest {
                 has_init_funcs,
+                export_symbols: requested_linker_export_symbols(symbol_db.args),
+                has_memory: any_object_needs_linker_memory(layout_inputs),
                 wrap_entry,
                 got_mem_count: scan.got_mem.len(),
                 got_func_count: scan.got_func.len(),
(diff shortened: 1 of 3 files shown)


libwild/src/wasm.rs
```diff
@@ -4845,13 +4886,37 @@ impl LinkerDefinedIndices {
...
         let mut needs_stack_pointer = false;
         let mut needs_tls_base = false;
         let mut needs_ctors = request.has_init_funcs;
+        let mut export_data = Vec::new();
+        for &sym in &request.export_symbols {
+            if !sym.materialize_on_export() {
+                continue;
+            }
+            match sym {
+                WasmLinkerSymbol::CallCtors => needs_ctors = true,
```
davidlattimore: Isn't this repeating logic that already exists elsewhere? Any idea how to deduplicate?

libwild/src/wasm.rs
```diff
@@ -4740,6 +4746,39 @@ fn fill_got_mem_inits(
...
+                crate::error!(
+                    "linker-defined symbol `{}` has no address to export",
+                    std::str::from_utf8(known.name()).unwrap_or("?")
+                )
+            })?;
+        let defined_slot = (global_index - indices.global_import_count) as usize;
+        let global = layout.globals.get_mut(defined_slot).ok_or_else(|| {
+            crate::error!("exported data global slot {defined_slot} out of range")
+        })?;
+        let addr_i32 = i32::try_from(addr).map_err(|_| {
```
davidlattimore: Why are addresses encoded as i32 rather than u32? Seems surprising. Could it ever make sense for a negative value?

libwild/src/wasm.rs
```diff
@@ -5499,17 +5586,52 @@ fn ensure_entry_export<'data>(
     });
 }
 
+fn requested_linker_export_symbols(args: &WasmArgs) -> Vec<WasmLinkerSymbol> {
+    let mut symbols = Vec::new();
+    for name in args.force_export_symbol_names() {
+        let Some(sym) = WasmLinkerSymbol::parse(name) else {
```
davidlattimore, line 5599: It seems odd to me that we're parsing the symbol name both here and in `ensure_force_exports`. Do we need to do both?

libwild/src/wasm.rs
lapla-cogito: Both processes have been consolidated into the `LinkerImportAbsorption` method.

libwild/src/wasm.rs
lapla-cogito: I was using the raw encoding type for Wasm instructions, so I've fixed.

libwild/src/wasm.rs
lapla-cogito: Changed to not re-parse.

libwild/src/wasm.rs
```diff
@@ -3945,6 +3962,12 @@ struct LinkerDefinedIndices {
     /// First module global index for GOT.func entries.
     got_func_global_base: Option<u32>,
     got_func_count: u32,
+    data_address_globals: Vec<(WasmLinkerSymbol, u32)>,
+    // Linker symbols named by `--export` / `--export-if-defined`.
+    requested_exports: Vec<WasmLinkerSymbol>,
+    // True when `__memory_base` is the PIC value (`LINKER_MEMORY_BASE`). False when it exists
+    // only because of `--export`.
+    memory_base_is_pic: bool,
```
davidlattimore: This field feels like it could do with some refactoring, but I'm not sure what exactly.

libwild/src/wasm.rs
```diff
@@ -7624,6 +7808,12 @@ mod tests {
         assert_eq!(features[1].name, "atomics");
     }
 
+    #[test]
+    fn memory_base_export_only_is_zero() {
```
davidlattimore: This test seems to pretty much just repeat the implementation of the function that it's testing. That said, if I change the function that it's testing to always return `LINKER_MEMORY_BASE_INIT_EXPR`, this is the only test that fails. I assume the behaviour is important and so incorrect behaviour should have some observable impact on the produced wasm binary. If that's the case, is an integration test instead of this test practical?

If the code involved here were refactored, there would be a good chance that this test would just be deleted, but an integration test can survive refactorings, making sure that we keep the important behaviours.

libwild/src/wasm.rs
```diff
@@ -5160,14 +5251,16 @@ fn emit_reserved_linker_definitions(
 ) {
     let mut linker_globals = Vec::with_capacity(indices.num_defined_globals as usize);
     if indices.memory_base_global.is_some() {
-        layout.memory_base = LINKER_MEMORY_BASE;
+        if indices.memory_base_is_pic {
+            layout.memory_base = LINKER_MEMORY_BASE;
+        }
```
davidlattimore: Is this bit of code needed? The tests still all pass if I remove it.

libwild/src/wasm.rs
lapla-cogito: Merged the separate bool and init value definitions into a single definition.

libwild/src/wasm.rs
lapla-cogito: I was struggling with how to add integration tests until I realized that in Wasm, the `ExpectSym` address assertion could be interpreted as a `i32.const` global variable.

libwild/src/wasm.rs
lapla-cogito: Removed.

davidlattimore (approved):

PR #2417, opened 2026-08-19
port(wasm): Drop unused function types from the type section

lapla-cogito:
wasm-ld keeps only types used by live functions, imports, or `R_WASM_TYPE_INDEX_LEB`. While previous implementation supported GC for non-type objects, this patch adds support for removing unnecessary types.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2826,7 +2826,6 @@ struct WasmObjectLayoutInput<'data> {
 
 #[derive(Debug, Clone, Copy)]
 struct WasmObjectIndexBases {
-    type_index_base: u32,
     defined_function_base: u32,
     defined_global_base: u32,
 }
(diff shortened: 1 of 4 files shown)


libwild/src/wasm.rs
```diff
@@ -5057,77 +5050,113 @@ fn encode_i32_const_u32(value: u32) -> Vec<u8> {
...
         }
     }
-
-    for index_map in &mut layout.object_index_maps {
-        for type_index in &mut index_map.type_indices {
-            *type_index = remap(*type_index);
+    for reloc in input
+        .code_relocations
+        .iter()
+        .chain(input.data_relocations.iter())
```
davidlattimore, line 5106: If I comment out this chain call, the tests still pass. Is that something we should test?

libwild/src/wasm.rs
```diff
@@ -5057,77 +5050,113 @@ fn encode_i32_const_u32(value: u32) -> Vec<u8> {
...
+    }
+    for (i, import) in input.function_imports.iter().enumerate() {
+        if !input.live_function_imports.get(i).copied().unwrap_or(false) {
+            continue;
+        }
+        match resolutions.function_resolutions.get(i) {
+            Some(
+                ImportResolution::Unresolved
+                | ImportResolution::WeakUndefStub { .. }
+                | ImportResolution::LinkerDefined(_),
```
davidlattimore, line 5098: If I comment out these two lines, the tests still pass. Do you think it'd be practical to test them?

PR #2418, opened 2026-08-19
feat: lookup object symbol values in linker script location counters

vishruth-thimmaiah:
This PR adds support for symbol lookup within location counters in linker scripts. For now we only fetch the first definition of the symbol, if it was defined in a linker script.

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -295,6 +295,7 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>, F: FileSystem>(
 
     let (mut section_part_layouts, mut section_layouts, mut resolved_location_counters) =
         compute_layout_sections::<A::Platform>(
+            &group_states,
             &section_part_sizes,
             &output_sections,
             &program_segments,
@@ -4970,13 +4971,44 @@ impl SymbolOutputInfos {
     }
 }
 
+fn compute_object_section_addresses<'data, P: Platform>(
+    obj: &ObjectLayoutState<'data, P>,
+    offsets: &mut OutputSectionPartMap<u64>,
+    symbol_db: &SymbolDb<'data, P>,
+    output_sections: &OutputSections<'data, P>,
+) -> Vec<u64> {
+    let mut addresses = vec![0u64; obj.sections.len()];
+    for (sec_idx, slot) in obj.sections.iter().enumerate() {
+        match slot {
+            SectionSlot::Loaded(sec) => {
+                let part_id =
+                    obj.section_part_id(object::SectionIndex(sec_idx), &symbol_db.section_part_ids);
+                addresses[sec_idx] = offsets.get(part_id);
+                *offsets.get_mut(part_id) += sec.capacity(part_id, output_sections);
+            }
+            SectionSlot::LoadedDebugInfo(sec) => {
+                // Advance offsets so subsequent sections are placed correctly, but we don't need
+                // the address for relaxation.
+                let part_id =
+                    obj.section_part_id(object::SectionIndex(sec_idx), &symbol_db.section_part_ids);
+                *offsets.get_mut(part_id) += sec.capacity(part_id, output_sections);
+            }
+            _ => {}
+        }
+    }
+
+    P::compute_object_addresses(obj, offsets);
+
+    addresses
+}
+
 /// Compute the output address of every loaded input section and every symbol in a single parallel
 /// pass over groups.
 fn compute_section_and_symbol_addresses<'data, P: Platform>(
     group_states: &[GroupState<'data, P>],
     section_part_layouts: &OutputSectionPartMap<OutputRecordLayout>,
     symbol_db: &SymbolDb<'data, P>,
-    output_sections: &OutputSections<P>,
+    output_sections: &OutputSections<'data, P>,
 ) -> (Vec<Vec<Vec<u64>>>, SymbolOutputInfos) {
     timing_phase!("Compute section and symbol addresses");
     let mem_offsets: OutputSectionPartMap<u64> = starting_memory_offsets(section_part_layouts);
@@ -4997,33 +5029,12 @@ fn compute_section_and_symbol_addresses<'data, P: Platform>(
                 .iter()
                 .map(|file| match file {
                     FileLayoutState::Object(obj) => {
-                        let mut addresses = vec![0u64; obj.sections.len()];
-                        for (sec_idx, slot) in obj.sections.iter().enumerate() {
-                            match slot {
-                                SectionSlot::Loaded(sec) => {
-                                    let part_id = obj.section_part_id(
-                                        object::SectionIndex(sec_idx),
-                                        &symbol_db.section_part_ids,
-                                    );
-                                    addresses[sec_idx] = offsets.get(part_id);
-                                    *offsets.get_mut(part_id) +=
-                                        sec.capacity(part_id, output_sections);
-                                }
-                                SectionSlot::LoadedDebugInfo(sec) => {
-                                    // Advance offsets so subsequent sections are placed
-                                    // correctly, but we don't need the address for relaxation.
-                                    let part_id = obj.section_part_id(
-                                        object::SectionIndex(sec_idx),
-                                        &symbol_db.section_part_ids,
-                                    );
-                                    *offsets.get_mut(part_id) +=
-                                        sec.capacity(part_id, output_sections);
-                                }
-                                _ => {}
-                            }
-                        }
-
-                        P::compute_object_addresses(obj, &mut offsets);
+                        let addresses = compute_object_section_addresses(
+                            obj,
+                            &mut offsets,
+                            symbol_db,
+                            output_sections,
+                        );
 
                         // While we have the section addresses, also resolve symbol
                         // output addresses for this file's canonical definitions.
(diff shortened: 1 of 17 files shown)


libwild/src/layout.rs
```diff
@@ -5324,6 +5324,97 @@ fn perform_iterative_relaxation<'data, A: Arch>(
     Ok(())
 }
 
+fn lookup_symbol_address<'data, P: Platform>(
```
davidlattimore: From the name of this function, one might infer that this is the way to look up symbol addresses, but it's scope is presumably more limited than that, since we already have preexisting ways to look up symbol addresses. I assume the reason this exists is that you're needing to determine some symbol addresses at an earlier stage than the addresses are normally known. If so, can we reflect that in the name of the function and/or in a comment on the function?

Relatedly, what kinds of symbols need to be supported here? Can we make sure that we give a suitable error if any unsupported symbols are looked up here?

libwild/src/layout.rs
```diff
@@ -5324,6 +5324,97 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+        } else {
+            bail!(
+                "Symbol `{}` defined in unsupported source for location counter lookup",
+                String::from_utf8_lossy(name)
+            );
+        }
+    };
+
+    match &def_info.placement {
+        SymbolPlacement::Redirect(redirect) => crate::expression_eval::evaluate_expression(
```
davidlattimore: What happens if we have cyclic dependency between symbols here? e.g.

```
ENTRY(begin_here)

first = second;
second = first;

SECTIONS
{
    . = first;
    .text : {
        *(.text)
    }
    .data : {
        *(.data)
    }
    .bss : {
        KEEP(*(.bss))
    }
}
```

libwild/src/layout.rs
```diff
@@ -5324,6 +5324,97 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+                    .parsed
+                    .object
+                    .symbol_offset_in_section(local_symbol, section_index)?;
+                let input_section_id = obj.section_id_range.input_to_id(section_index);
+                let part_id = symbol_db.section_part_ids[input_section_id.as_usize()];
+                let output_section_id = part_id.output_section_id::<P>();
+                let primary_id = output_sections.primary_output_section(output_section_id);
+                let section_layout = section_layouts.get(primary_id);
+                let base_address = section_layout.mem_offset;
+                return Ok(base_address + input_offset);
```
davidlattimore: I'm not sure this will do what you want when the input section isn't the first in the PartId. Can you add a test for that to check?

libwild/src/layout.rs
vishruth-thimmaiah: Both ld and lld seem to return a 0 address if there is a cyclic dependency, so I've matched that behavior.

libwild/src/layout.rs
vishruth-thimmaiah: For now I've only added support for symbols originating from objects, linker scripts and via `--defsym`. If we encounter any other symbols we return an error.

libwild/src/layout.rs
```diff
@@ -5324,6 +5324,132 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+) -> Result<u64> {
+    let prehashed = UnversionedSymbolName::prehashed(name);
+    let Some(target_symbol_id) = symbol_db.get_unversioned(&prehashed) else {
+        bail!(
+            "Undefined symbol `{}` in linker script location counter expression",
+            String::from_utf8_lossy(name)
+        );
+    };
+    let canonical_id = symbol_db.definition(target_symbol_id);
+    if !visited.insert(canonical_id) {
```
davidlattimore: It looks like you're only inserting into this set. What happens if the same symbol is referenced more than once? Won't that get a false positive here?

libwild/src/layout.rs
```diff
@@ -5324,6 +5324,132 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+                    .symbol_offset_in_section(local_symbol, section_index)?;
+                let input_section_id = obj.section_id_range.input_to_id(section_index);
+                let part_id = symbol_db.section_part_ids[input_section_id.as_usize()];
+                let output_section_id = part_id.output_section_id::<P>();
+                let primary_id = output_sections.primary_output_section(output_section_id);
+                if source_section.is_some_and(|s| s == primary_id) {
+                    return Ok(input_offset);
+                }
+                let part_layout = part_layouts.get(part_id);
+                if part_layout.mem_offset == 0 {
```
davidlattimore: Will this give false positives if a symbol actually has the value 0?

libwild/src/layout.rs
davidlattimore: Here's a test that shows what I was meaning:

```patch
diff --git a/wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter.c b/wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter.c
index 6608597b70..2e5ecbe628 100644
--- a/wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter.c
+++ b/wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter.c
@@ -27,6 +27,14 @@
 // RISC-V: BFD complains about missing __global_pointer$ (defined in the default linker script)
 //#SkipArch:riscv64,ppc64le
 
+//#Config:object_symbol_offset
+//#LinkerScript:linker-script-object-symbol-offset.ld
+//#Object:runtime.c
+//#Object:object-symbol-prefix.c
+//#Object:object-symbol-target.c
+// RISC-V: BFD complains about missing __global_pointer$ (defined in the default linker script)
+//#SkipArch:riscv64,ppc64le
+
 //#Config:underflow
 //#Object:runtime.c
 //#LinkerScript:linker-script-location-counter-underflow.ld
diff --git a/wild/tests/sources/elf/linker-script-location-counter/linker-script-object-symbol-offset.ld b/wild/tests/sources/elf/linker-script-location-counter/linker-script-object-symbol-offset.ld
new file mode 100644
index 0000000000..b9c50a7d9e
--- /dev/null
+++ b/wild/tests/sources/elf/linker-script-location-counter/linker-script-object-symbol-offset.ld
@@ -0,0 +1,18 @@
+ENTRY(begin_here)
+
+SECTIONS
+{
+    . = 0x600000;
+    .text : {
+        *(.text)
+    }
+    .symbol_lookup : {
+        KEEP(*(.symbol_lookup))
+    }
+    .data : AT(ASSERT(displaced_symbol == ADDR(.symbol_lookup) + 32, "wrong address")) {
+        *(.data)
+    }
+    .bss : {
+        KEEP(*(.bss))
+    }
+}
diff --git a/wild/tests/sources/elf/linker-script-location-counter/object-symbol-prefix.c b/wild/tests/sources/elf/linker-script-location-counter/object-symbol-prefix.c
new file mode 100644
index 0000000000..c096ce3e00
--- /dev/null
+++ b/wild/tests/sources/elf/linker-script-location-counter/object-symbol-prefix.c
@@ -0,0 +1,1 @@
+__attribute__((section(".symbol_lookup"), aligned(1))) unsigned char symbol_prefix[32];
diff --git a/wild/tests/sources/elf/linker-script-location-counter/object-symbol-target.c b/wild/tests/sources/elf/linker-script-location-counter/object-symbol-target.c
new file mode 100644
index 0000000000..695dc2b4ed
--- /dev/null
+++ b/wild/tests/sources/elf/linker-script-location-counter/object-symbol-target.c
@@ -0,0 +1,1 @@
+__attribute__((section(".symbol_lookup"), aligned(1))) unsigned char displaced_symbol;
```

libwild/src/layout.rs
```diff
@@ -5324,6 +5324,132 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+                        section_layouts,
+                        part_layouts,
+                        memory_regions,
+                        sizeof_headers,
+                        resolved_lc,
+                        visited,
+                        source_section,
+                    );
+                    if let Some(section_base) = current_section_base {
+                        address.map(|addr| addr - section_base)
```
davidlattimore: I think this can underflow? e.g.:

```
ENTRY(begin_here)

absolute_symbol = 1;

SECTIONS
{
    . = 0x600000;
    .text : {
        section_relative_symbol = absolute_symbol;
        *(.text)
        *(.text.*)
    }

    . += ASSERT(section_relative_symbol != 0, "must be not be zero");
    .data : {
        *(.data)
        *(.data.*)
    }
    .bss : {
        KEEP(*(.bss))
    }
}
```

libwild/src/layout.rs
```diff
@@ -5319,11 +5324,194 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+                    group_states,
+                    symbol_db,
+                    output_sections,
+                    input_section_id,
+                    part_id,
+                ) else {
+                    bail!("Symbol is in a section that we didn't load");
+                };
+                if source_section.is_some_and(|s| s == primary_id) {
+                    return Ok((input_offset + intra_part_offset, false));
```
davidlattimore: What about if there are other parts in the section before this one? It doesn't seem that is take into account

libwild/src/layout.rs
```diff
@@ -5319,11 +5324,194 @@ fn perform_iterative_relaxation<'data, A: Arch>(
...
+
+        let object_part_ids = &symbol_db.section_part_ids[obj.section_id_range.as_usize()];
+        for (slot_idx, (slot, &part_id)) in obj.sections.iter().zip(object_part_ids).enumerate() {
+            let input_sec_id = obj.section_id_range.start().add_usize(slot_idx);
+            if input_sec_id == target_input_section_id {
+                if !matches!(
+                    slot,
+                    SectionSlot::Loaded(_)
+                        | SectionSlot::LoadedDebugInfo(_)
+                        | SectionSlot::Sorted(_)
```
davidlattimore: Does this do the right thing when sorted sections are involved? If so, can we test it? It'd also be perfectly fine to say it's out of scope provided we give a suitable error (which we should probably test)

libwild/src/layout.rs
```diff
@@ -605,8 +601,17 @@ fn update_defsym_symbol_resolution<'data, P: Platform>(
                     .as_ref()
                     .ok_or_else(|| redirect.missing_resolution(name))?;
 
+                let is_positional = symbol_db
```
davidlattimore: Does this mean that absolute symbols from regular objects will be treated as positional?

davidlattimore (commented):
Thanks. Still a few possible issues to look into. I think part of the issue is that we're trying to implement address resolution logic, but earlier, before we have all the info that we possibly need for address resolution. It'd be good if we could reduce duplication of logic where possible and for cases where we can't reliably determine the address that we need, give an error. Perhaps there's some scope reduction that could be done? e.g. reduce what kinds of symbols we'll support this for, with suitable errors for the others.

davidlattimore (approved):
Thanks :)

Issue #2071, opened 2026-06-14
Mach-O: TLS support


marxin:
Haven't figured out the requirements yet, but the platform also provides TLS by using a special 2 sections in the output:
```
6: Section { segment: "__DATA", name: "__thread_vars", address: 100008000, size: 18, align: 8, kind: TlsVariables, flags: MachO { flags: 13 } }
7: Section { segment: "__DATA", name: "__thread_bss", address: 100008018, size: 4, align: 4, kind: UninitializedTls, flags: MachO { flags: 12 } }
```

marxin:
Note the current ELF test-suite contains quite some tests related to TLS and we can copy (port) these to macOS as well.

joanna-jo:
Could I start looking into this next for the port? (Although it might take me a little bit to ramp up on TLS-related things if that's okay!)

davidlattimore:
That'd be great! It's fine if it takes a bit. It took quite a while with the ELF TLS implementation. There was quite a long period when every time a program misbehaved, it almost always turned out to be yet another missing bit in our TLS implementation.

PR #2420, opened 2026-08-19
port(MachO): Order TLV template sections

joanna-jo:
The generic layout implements ELF semantics so that NOBITS TLS sections 1) overlap non-TLS sections in the virtual address space and 2) extend `PT_TLS.p_memsz` without extending a non-TLS segment. For Mach-O, `S_THREAD_LOCAL_ZEROFILL` is treated as a zerofill section (typically `__DATA,__thread_bss`) that occupies virtual address space. This diff introduces a new `SectionAttributes::is_tls_only` so that Mach-O can then use `is_tls` to classify `S_THREAD_LOCAL_REGULAR` and `S_THREAD_LOCAL_ZEROFILL` sections, since it shouldn't follow the behavior above.

[`dyld` also constructs a TLV initialization template from the first `S_THREAD_LOCAL_REGULAR` or `S_THREAD_LOCAL_ZEROFILL` section it sees to the last](https://github.com/apple-oss-distributions/dyld/blob/fd8d0c4d52320ebf64db34f3cb280310d905c5ae/libdyld/ThreadLocalVariables.cpp#L62-L90). These sections therefore need to be laid out contiguously, with `S_THREAD_LOCAL_REGULAR` before `S_THREAD_LOCAL_ZEROFILL`. `__thread_vars` and `__thread_ptrs` (proper support coming in a future diff) don't have to be contiguous. 

I plan on adding [TLV template alignment](https://github.com/llvm/llvm-project/blob/main/lld/MachO/Writer.cpp#L987-L1008) in the following diff.

**Before**:
```
  Caused by:
    Mach-O TLV template sections must be contiguous: __thread_data (S_THREAD_LOCAL_REGULAR), __thread_vars (S_THREAD_LOCAL_VARIABLES), __thread_bss (S_THREAD_LOCAL_ZEROFILL)



failures:
    macho/aarch64/tls/default

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.93s
```

**After**:
```
running 1 test

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 24 filtered out; finished in 1.10s
```

Part of #2071

### DIFF
--- a/libwild/src/platform.rs
+++ b/libwild/src/platform.rs
@@ -1383,6 +1383,8 @@ pub(crate) trait SectionAttributes:
 
     fn is_tls(&self) -> bool;
 
+    fn occupies_only_tls_address_space(&self) -> bool;
+
     fn is_writable(&self) -> bool;
 
     fn is_no_bits(&self) -> bool;
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -354,6 +354,8 @@ use object::ObjectSymbol as _;
 use object::macho::LC_CODE_SIGNATURE;
 use object::macho::LC_DYLD_CHAINED_FIXUPS;
 use object::macho::LC_DYLD_EXPORTS_TRIE;
+use object::macho::S_THREAD_LOCAL_REGULAR;
+use object::macho::S_THREAD_LOCAL_ZEROFILL;
 use object::macho::SEG_LINKEDIT;
 use object::macho::SEG_TEXT;
 use object::read::elf::ProgramHeader;
@@ -4933,6 +4935,7 @@ impl Assertions {
         verify_no_overlapping_sections(obj)?;
         verify_no_overlapping_segments(obj)?;
         verify_chained_fixups_segment_offsets(obj, bytes)?;
+        verify_macho_tlv_template_layout(obj)?;
 
         if linker_used.is_wild() {
             verify_uuid(obj, bytes)?;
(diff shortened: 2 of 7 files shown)


wild/tests/sources/macho/trivial-tls/trivial-tls.c
```diff
@@ -0,0 +1,15 @@
+//#LinkerDriver:clang
```
joanna-jo, line 1: I noticed that we don't pass in `-syslibroot <SDK> -lSystem` to lld and Wild in the macOS linker catalog, which means that we have an unresolved reference for `__tlv_bootstrap`. I'll add this in a separate diff.

wild/tests/integration_tests.rs
```diff
@@ -5971,6 +5974,52 @@ fn verify_macho_exports(obj: &object::File, bytes: &[u8]) -> Result<HashMap<Vec<
...
+    let is_tlv_template = |ty| matches!(ty, S_THREAD_LOCAL_REGULAR | S_THREAD_LOCAL_ZEROFILL);
+
+    let (Some(start), Some(end)) = (
+        sections.iter().position(|&(_, ty)| is_tlv_template(ty)),
+        sections.iter().rposition(|&(_, ty)| is_tlv_template(ty)),
+    ) else {
+        return Ok(());
+    };
+
+    let template = &sections[start..=end];
```
marxin, line 6000: Maybe use can use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunk_by + `.collect_vec()` and make an assertion `len() == 2` && `chunks.first().0 == true`?

joanna-jo: Thanks for the suggestion! Should we maybe consider `dedup` since we also have to account for `len() == 1` (either just `S_THREAD_LOCAL_REGULAR` or `S_THREAD_LOCAL_ZEROFILL`) and edge cases like `REGULAR, unrelated section, ZEROFILL` if the `chunk_by` key was `ty == S_THREAD_LOCAL_REGULAR`? I was thinking of something like:

```
let type_runs = template.iter().map(|&(_, ty)| ty).dedup().collect_vec();

    ensure!(
        matches!(
            type_runs.as_slice(),
            [S_THREAD_LOCAL_REGULAR]
                | [S_THREAD_LOCAL_ZEROFILL]
                | [S_THREAD_LOCAL_REGULAR, S_THREAD_LOCAL_ZEROFILL]
        ),
```

libwild/src/platform.rs
```diff
@@ -1383,6 +1383,10 @@ pub(crate) trait SectionAttributes:
 
     fn is_tls(&self) -> bool;
 
+    /// Returns whether the section reserves space only in each thread's TLS block,
+    /// without reserving space in the load image.
+    fn is_tls_only(&self) -> bool;
```
davidlattimore: The current name to me doesn't quite capture the intended meaning. What about something like `occupies_only_tls_address_space`? But I don't feel strongly if you'd rather keep the current name.

joanna-jo: I was also feeling a bit on the fence about it, thanks for pointing this out! I agree that `occupies_only_tls_address_space` sounds clearer. I removed the doc comment as well since the name is self-explanatory now.

davidlattimore (approved):

wild/tests/integration_tests.rs
marxin: Even nicer - yes, let's use it.

PR #2421, opened 2026-08-20
chore: Add test config option llvm_tools_dir

davidlattimore:
On Ubuntu, the package llvm-21-tools installs e.g. `/usr/bin/FileCheck-21` and `/usr/lib/llvm-21/bin/FileCheck`. Just adding `/usr/lib/llvm-21/bin` globally to my path doesn't seem appealing, especially since it contains a binary called "not", which seems a bit special-purpose. Instead, we now require that for running lld tests, `lld_tools_dir` in `test-config.toml` needs to point to the directory containing these tools.

I also got rid of some code that was using `.display().to_string()` to produce a string that would be executed. If the path had contained non-UTF-8 characters, then this would have given a confusing error. Given that this is a test, it's better to panic with a reasonable error message.

I also removed code that was substituting `FileCheck` and `llvm-mc` for absolute paths to these tools. I couldn't figure out a reason why we'd need to do this given that they'll be found on the path which was where we were getting the absolute paths from. But maybe I'm missing some reason why this was necessary.

`x86-64-tls-gd-got.s` was passing, so I removed it from the exclusion list.

Tested on Ubuntu 26.04 and on Arch Linux.

### DIFF
--- a/wild/tests/external_tests/lld_skip_tests.toml
+++ b/wild/tests/external_tests/lld_skip_tests.toml
@@ -87,7 +87,6 @@ tests = [
   "x86-64-retpoline-znow.s",
   "x86-64-retpoline-znow-linkerscript.s",
   "x86-64-retpoline-znow-static-iplt.s",
-  "x86-64-tls-gd-got.s",
   "x86-64-gotpc-offset.s",
   "x86-64-tls-le-align.s",
   "x86-64-tls-ld-preemptable.s",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -211,9 +211,13 @@ WILD_IGNORE_SKIP=mold cargo test --features mold_tests
 WILD_IGNORE_SKIP=all cargo test --features external_tests
 ```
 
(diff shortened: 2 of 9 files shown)


wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -87,7 +87,6 @@ tests = [
   "x86-64-retpoline-znow.s",
   "x86-64-retpoline-znow-linkerscript.s",
   "x86-64-retpoline-znow-static-iplt.s",
-  "x86-64-tls-gd-got.s",
```
deepakshirkem, line 90: x86-64-tls-gd-got.s still fails for me, but that's my local llvm-objdump build missing X86 target support (built from a personal fork, not stable) not an issue with your change.

deepakshirkem (approved):
Looks good to me. Thank You.

wild/tests/external_tests/lld_skip_tests.toml
davidlattimore: Hopefully the option `ignore_external_tests` will be useful for you if you need to ignore that one test on your system - although it looks like it still needs to be hooked up to the lld test runner.

PR #2426, opened 2026-08-20
feat: support aligning a given expression in linker scripts

vishruth-thimmaiah:
The `ALIGN` function in linker scripts can have a second expression passed, which is used in place of the default, which is the current location counter.


wild/tests/sources/elf/linker-script-assert-pass/linker-script-assert-pass.ld
```diff
@@ -71,3 +71,4 @@ ASSERT(symbol9 == 0x160, "symbol9 must be 0x160");
 
 symbol12 = ASSERT(ASSERT(0x5000 < 0x10000, "0x5000 < 0x10000 must be true") == 1, "ASSERT must return 1");
 . = ASSERT(symbol12 == 1, "must be 1");
+ASSERT(ALIGN(0x38, 16) == 0x40, "ALIGN value is incorrect");
```
davidlattimore, line 74: What do the other linkers do when an alignment isn't a power of 2? Can we add a test for this?

vishruth-thimmaiah: ld supports aligning on any value, not just powers of two. I've added a test for this.

davidlattimore (approved):

PR #2427, opened 2026-08-20
test: Pass macOS SDK and libSystem args to direct linkers

joanna-jo:
Passes in `-syslibroot <SDK> -lSystem` args to all linkers when linking directly on macOS. Clang provides these automatically, but when linking directly in our integration harness, all linkers need them explicitly passed in to resolve references to symbols found in system libraries/frameworks.

**Before**:
```
running 1 test
  test macho/aarch64/trivial-tls/default ... FAILED

  failures:

  ---- macho/aarch64/trivial-tls/default ----
  Test failed: `trivial-tls.c` with linker `lld` config `default`
    Caused by:
      Linker failed:
  ld64.lld: error: undefined symbol: __tlv_bootstrap
```
```
running 1 test
  test macho/aarch64/trivial-tls/default ... FAILED

  failures:

  ---- macho/aarch64/trivial-tls/default ----
  Test failed: `trivial-tls.c` with linker `wild` config `default`
    Caused by:
    Undefined symbol __tlv_bootstrap,
```

**After**:
```
  running 1 test

  test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.23s
```

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -4304,6 +4304,9 @@ impl LinkCommand {
                             }
                         }
                         PlatformKind::MachO => {
+                            let (_, sdk) = macos_toolchain()
+                                .map_err(|err| error!("Unable to locate macOS toolchain: {err}"))?;
+                            command.arg("-syslibroot").arg(sdk).arg("-lSystem");
                             if linker.is_lld() {
                                 let arch = cross_arch.unwrap_or(arch);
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):
I was initially unsure, since we don't link against libc by default for ELF. But it seems that Mac is just a bit different in the extent to which it can create executables that don't link against libSystem, so I think the direction you're taking this is right. Thanks :)

PR #2428, opened 2026-08-21
chore(wasm): Cache encoded size of input data segments

lapla-cogito:
Compute `wasm_data_segment_encoded_size()` once when building `WasmDataSegment` and reuse it for section offsets, GC spans, and data-reloc classification.


davidlattimore (approved):
Nice :)

PR #2432, opened 2026-08-21
port(wasm): Parallelize layout relocation scan

lapla-cogito:
Scan each object's relocations in parallel, then assign GOT slots in object order.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -4256,7 +4256,7 @@ fn note_undefined_data_from_reloc(
     seen: &mut HashSet<(String, String)>,
     errors: &mut Vec<String>,
 ) -> Result {
-    if symbol_db.args.allow_undefined || reloc.ty == RelocationType::TypeIndexLeb {
+    if reloc.ty == RelocationType::TypeIndexLeb {
         return Ok(());
     }
     let Some(sym) = input.symbols.get(reloc.index as usize) else {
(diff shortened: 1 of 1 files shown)


libwild/src/wasm.rs
```diff
@@ -4287,146 +4287,209 @@ fn note_undefined_data_from_reloc(
...
     let mut needs_table_base = false;
-    let mut needs_table = layout_inputs
-        .iter()
-        .any(|input| !input.table_imports.is_empty());
-    let mut table_index_symbol_indices = vec![Vec::new(); layout_inputs.len()];
-    let mut undefined_data_errors: Vec<String> = Vec::new();
-    let mut seen_undefined_data: HashSet<(String, String)> = HashSet::new();
+    let mut needs_table = !input.table_imports.is_empty();
+    let mut undefined_data_errors = Vec::new();
+    let mut seen_undefined_data = HashSet::new();
```
davidlattimore, line 4322: It looks like we're possibly now deduplicating undefined data errors on a per-object basis rather than globally. Up to you if you think this matters.

davidlattimore (approved):

libwild/src/wasm.rs
lapla-cogito: This patch should handle it without any issues I think.

davidlattimore (approved):

PR #2434, opened 2026-08-22
port(wasm): Hold relocations as per-object ranges

lapla-cogito:
Give each function body and data segment a range into one object-level reloc list instead of its own Vec. The writer subtracts `code_offset` / `payload_start`.

Part of #1431


davidlattimore (approved):
Nice :)

PR #2440, opened 2026-08-24
ci: run all external tests

marxin:
FYI: @deepakshirkem

### DIFF
--- a/wild/tests/external_tests/lld_tests.rs
+++ b/wild/tests/external_tests/lld_tests.rs
@@ -127,12 +127,6 @@ fn run_lld_test(test_file: &Path, test_config: &TestConfig) -> Result {
     let tmpdir = tempfile::tempdir()?;
 
     for cmd in extract_run_lines(&content) {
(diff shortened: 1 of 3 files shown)


wild/tests/external_tests/lld_tests.rs
```diff
@@ -127,6 +127,12 @@ fn run_lld_test(test_file: &Path, test_config: &TestConfig) -> Result {
     let tmpdir = tempfile::tempdir()?;
 
     for cmd in extract_run_lines(&content) {
+        // Ubuntu's LLVM packages don't provide an unversioned `not` executable, so use
```
davidlattimore: I have `/usr/lib/llvm-21/bin/not` from package llvm-21-tools

davidlattimore: But you'd probably need to add `llvm_tools_dir = "/usr/lib/llvm-21/bin"` to test-config-ci.toml

marxin: Thank you - let me try with the config value being set.

davidlattimore (approved):

PR #2445, opened 2026-08-25
fix: Support LTO with `clang -ffat-lto-objects`

mati865:
Previously we were just using the regular non-LTO symbols in this case.

Also added GCC counterpart test which seems fine, but there is a strange symbol there:
```
Symbol table '.symtab' contains 9 entries:
...
3: 0000000000000000     0 FILE    LOCAL  DEFAULT  ABS <artificial>
```

### DIFF
--- a/libwild/src/file_kind.rs
+++ b/libwild/src/file_kind.rs
@@ -54,6 +54,8 @@ impl FileKind {
                 object::elf::ET_REL => {
                     if is_gcc_bitcode(bytes, header).unwrap_or(false) {
                         Ok(FileKind::GccIr)
+                    } else if is_llvm_bitcode(bytes, header).unwrap_or(false) {
+                        Ok(FileKind::LlvmIr)
                     } else {
                         Ok(FileKind::ElfObject)
                     }
@@ -136,6 +138,25 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
     const MAX_SCAN: usize = 200;
     let strings = data.get(start_offset + START..start_offset + (START + MAX_SCAN).min(len))?;
     Some(memchr::memmem::find(strings, b"\0.gnu.lto_.").is_some())
+}
+
+fn is_llvm_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool> {
+    // If we don't have plugin support, then we skip checking if the file contains LLVM IR. If it
+    // is, then we'll figure that out later on and report an error. We do this because this code
+    // has a measurable performance impact.
+    if !cfg!(all(feature = "plugins", unix)) {
+        return Some(false);
+    }
+    let e = LittleEndian;
+    let section_headers = header.section_headers(e, data).ok()?;
+    let sh_str_index = header.shstrndx(e, data).ok()?;
+    let strings_section_header = section_headers.get(sh_str_index as usize)?;
+    let start_offset = strings_section_header.sh_offset(e) as usize;
+    let len = strings_section_header.sh_size(e) as usize;
+    const START: usize = 0;
+    const MAX_SCAN: usize = 100;
+    let strings = data.get(start_offset + START..start_offset + (START + MAX_SCAN).min(len))?;
+    Some(memchr::memmem::find(strings, b"\0.llvm.lto").is_some())
 }
 
 impl std::fmt::Display for FileKind {
--- a/libwild/src/file_kind.rs
+++ b/libwild/src/file_kind.rs
@@ -140,6 +140,9 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
     Some(memchr::memmem::find(strings, b"\0.gnu.lto_.").is_some())
 }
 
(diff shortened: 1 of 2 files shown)


libwild/src/file_kind.rs
```diff
@@ -138,6 +140,25 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
...
+    let sh_str_index = header.shstrndx(e, data).ok()?;
+    let strings_section_header = section_headers.get(sh_str_index as usize)?;
+    let start_offset = strings_section_header.sh_offset(e) as usize;
+    let len = strings_section_header.sh_size(e) as usize;
+    const START: usize = 0;
+    const MAX_SCAN: usize = 100;
+    let strings = data.get(start_offset + START..start_offset + (START + MAX_SCAN).min(len))?;
+    Some(memchr::memmem::find(strings, b"\0.llvm.lto").is_some())
+}
+
```
mati865, line 157: I think this function could be merged with `is_gcc_bitcode` up to the consts part and return an enum.

libwild/src/file_kind.rs
```diff
@@ -138,6 +140,25 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
...
+        return Some(false);
+    }
+    let e = LittleEndian;
+    let section_headers = header.section_headers(e, data).ok()?;
+    let sh_str_index = header.shstrndx(e, data).ok()?;
+    let strings_section_header = section_headers.get(sh_str_index as usize)?;
+    let start_offset = strings_section_header.sh_offset(e) as usize;
+    let len = strings_section_header.sh_size(e) as usize;
+    const START: usize = 0;
+    const MAX_SCAN: usize = 100;
```
mati865: Dunno what to do about these. It works with the test added in this PR, while `START = 100` did not, so just lowered them both compared to `is_gcc_bitcode`.

libwild/src/file_kind.rs
marxin: Good idea!

mati865: Dunno if merging still makes sense given how simple is the updated check with Clang. Unfortunately we cannot apply the same for GCC:
```
❯ readelf -S wild/tests/build/elf/x86_64/linker-plugin-lto/gcc-fat-objects/linker-plugin-lto.c.o | rg \.lto
  [ 5] .gnu.lto_.pr[...] PROGBITS         0000000000000000  0000005a
  [ 6] .gnu.lto_.in[...] PROGBITS         0000000000000000  00000069
  [ 7] .gnu.lto_.jm[...] PROGBITS         0000000000000000  000000b0
  [ 8] .gnu.lto_.pu[...] PROGBITS         0000000000000000  000000ca
  [ 9] .gnu.lto_.ip[...] PROGBITS         0000000000000000  000000db
  [10] .gnu.lto_.lt[...] PROGBITS         0000000000000000  000000f0
  [11] .gnu.lto__st[...] PROGBITS         0000000000000000  000000f8
  [12] .gnu.lto_.sy[...] PROGBITS         0000000000000000  000001f2
  [13] .gnu.lto_.re[...] PROGBITS         0000000000000000  0000024c
  [14] .gnu.lto_.de[...] PROGBITS         0000000000000000  0000025a
  [15] .gnu.lto_.sy[...] PROGBITS         0000000000000000  0000056f
  [16] .gnu.lto_.ex[...] PROGBITS         0000000000000000  000005d0
  [17] .gnu.lto_.opts    PROGBITS         0000000000000000  000005d9

❯ readelf -S wild/tests/build/elf/x86_64/linker-plugin-lto/clang-fat-objects/linker-plugin-lto.c.o | rg \.lto
  [ 4] .llvm.lto         LLVM_LTO         0000000000000000  00000055
```

libwild/src/file_kind.rs
```diff
@@ -138,6 +140,21 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
     Some(memchr::memmem::find(strings, b"\0.gnu.lto_.").is_some())
 }
 
+// FIXME: Add to object crate.
+const SHT_LLVM_LTO: object::elf::SectionType = object::elf::SectionType(0x6fff4c0c);
```
marxin, line 144: Nice observation - seems it's there since LLVM 17 (https://github.com/llvm/llvm-project/commit/a484e020d75d398049e1fcbeb0157d57011714b3) which is a solid version coverage.

libwild/src/file_kind.rs
```diff
@@ -138,6 +140,21 @@ fn is_gcc_bitcode(data: &[u8], header: &crate::elf::FileHeader64) -> Option<bool
     Some(memchr::memmem::find(strings, b"\0.gnu.lto_.").is_some())
 }
 
+// FIXME: Add to object crate.
```
mati865: Philip has merged change to the object. I'll update the comment before merging this PR.

davidlattimore (approved):
Thanks :)

marxin (approved):

PR #2447, opened 2026-08-25
test: support split-file and persistent cd in LLD test harness

deepakshirkem:
Some LLD tests use split-file to split a single .s file into multiple named sub-files, then cd into that directory so subsequent RUN lines can reference them by bare name. The harness had no support for either, causing these tests to fail with "No such file or directory" a harness gap, not a Wild bug.

Part of https://github.com/wild-linker/wild/issues/2380

### DIFF
--- a/wild/tests/external_tests/lld_skip_tests.toml
+++ b/wild/tests/external_tests/lld_skip_tests.toml
@@ -20,6 +20,7 @@ tests = [
   "x86-64-tls-ie.s",
   "x86-64-tls-gdie.s",
   "x86-64-tls-dynamic.s",
+  "x86-64-section-layout.s",
 ]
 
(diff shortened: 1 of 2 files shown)


davidlattimore (commented):
I thought there were a bunch of tests that were failing because didn't support this. Shouldn't this change therefore remove those tests from the skip list?

davidlattimore (commented):
It feels like the direction we're going with parsing RUN commands, we'll end up having to implement a lot of LLVM's test runner in order to properly support their tests. I'm pretty unfamiliar with how LLVM / LLD does testing, but I gather they use a tool called lit. Could we see if we can use that to help drive the tests rather than reimplementing this stuff ourselves?

PR #2449, opened 2026-08-26
port(wasm): Support `--shared-memory`

### LINKED ISSUE #2336: Wasm: support `--shared-memory` option
```
wasm-ld --help
--shared-memory         Use shared linear memory
...
clang --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot m.c -Wl,--shared-memory
wasm-tools print ...
...
(module $a.out
...
  (memory (;0;) 2 2 shared)
```

lapla-cogito:
This option specifies the use of shared linear memory. Note that wasm-ld emits errors when the option is specified and both `atomics` and `bulk-memory` features absent.

Part of #1431

Close #2336

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -46,6 +46,9 @@ pub struct WasmArgs {
     // page-aligned). `None` means size is derived from data / stack layout only.
     pub(crate) initial_memory: Option<u64>,
     pub(crate) max_memory: Option<u64>,
+    // Emit a shared linear memory (`memory.shared`). Requires the `atomics` and `bulk-memory`
+    // target features.
+    pub(crate) shared_memory: bool,
     pub(crate) gc_sections: bool,
     pub(crate) allow_undefined: bool,
 }
@@ -77,6 +80,7 @@ impl Default for WasmArgs {
             entry: Some(DEFAULT_ENTRY.to_owned()),
             initial_memory: None,
             max_memory: None,
+            shared_memory: false,
             export_memory: None,
             gc_sections: true,
             allow_undefined: false,
@@ -363,6 +367,15 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
         .help("Maximum size of the linear memory in bytes")
         .execute(|args, _modifier_stack, value| {
             args.max_memory = Some(parse_number(value)?);
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("shared-memory")
+        .help("Use shared linear memory")
+        .execute(|args, _modifier_stack| {
+            args.shared_memory = true;
             Ok(())
         });
 
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -85,6 +85,9 @@
 //!
 //! ExpectFuncImportCount:{count} (Wasm) Asserts the total number of function imports.
 //!
+//! ExpectSharedMemory:{bool} (Wasm) When true, asserts that the output module defines a shared
+//! linear memory with a maximum.
+//!
 //! ExpectSection:{section_name} [properties] Checks that the specified section exists in the
 //! output binary. Optional properties:
 //!   max_entries=N: Asserts the section has at most N entries (uses the section's sh_entsize,
@@ -546,6 +549,7 @@ struct WasmModuleInfo {
     imported_global_count: u32,
     defined_global_i32: Vec<Option<i32>>,
     export_global_indices: HashMap<String, u32>,
+    memories: Vec<wasmparser::MemoryType>,
 }
 
 fn i32_const_from_expr(expr: wasmparser::ConstExpr<'_>) -> Option<i32> {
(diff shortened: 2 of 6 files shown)


wild/tests/integration_tests.rs
```diff
@@ -838,6 +873,7 @@ fn validate_wasm(wasm_file: &Path, linker_name: &str) -> Result {
 fn run_wasm_with_wasmtime(wasm_file: &Path, linker_name: &str) -> Result {
     let output = Command::new("wasmtime")
         .arg("run")
+        .args(["-W", "threads,shared-memory"])
```
lapla-cogito, line 876: I'm wondering if we should always apply this.

wild/tests/integration_tests.rs
```diff
@@ -2435,6 +2473,10 @@ fn process_directive(
                     .with_context(|| format!("Invalid ExpectFuncImportCount: `{arg}`"))?,
             );
         }
+        "ExpectSharedMemory" => {
+            config.assertions.expect_shared_memory =
+                arg.parse().context("Invalid bool for ExpectSharedMemory")?;
```
marxin: As seen in #2453, an extra context is unnecessary.

wild/tests/integration_tests.rs
marxin: Yes, seems fine to me.

wild/tests/sources/wasm/shared-memory-disallowed/shared-memory-disallowed.s
```diff
@@ -0,0 +1,13 @@
...
+//#ExpectError: --shared-memory is disallowed
+
+  .globl _start
+_start:
+  .functype _start () -> ()
+  end_function
+
+  .section .custom_section.target_features,"",@
+  .int8 1
+  .int8 45
```
marxin: Can you please comment these constants `# (+/-)` in the assembly files.

libwild/src/args/wasm.rs
```diff
@@ -46,6 +46,9 @@ pub struct WasmArgs {
     // page-aligned). `None` means size is derived from data / stack layout only.
     pub(crate) initial_memory: Option<u64>,
     pub(crate) max_memory: Option<u64>,
+    // Emit a shared linear memory (`memory.shared`). Requires the `atomics` and `bulk-memory`
```
marxin, line 49: The `atomics` makes sense, but are you sure about the `bulk-memory` extension?

lapla-cogito: IIUC, this option is primarily used with thread extensions. In this context, to ensure safe initialization of shared memory between threads, the data segment must be passive, and since this feature is required for that, wasm-ld enforces this requirement (though Wild currently doesn't support passive segments, aligning them now won't cause any issues).


https://github.com/llvm/llvm-project/blob/e89363f9fffd1972fbf2c2bd9816a3454480dee5/lld/wasm/Writer.cpp#L1060-L1062

https://github.com/llvm/llvm-project/blob/e89363f9fffd1972fbf2c2bd9816a3454480dee5/lld/wasm/Writer.cpp#L666-L668

wild/tests/integration_tests.rs
```diff
@@ -817,6 +828,30 @@ impl WasmModuleInfo {
...
         Ok(())
     }
+
+    fn ensure_shared_memory(&self, expected: bool, linker_name: &str) -> Result {
+        if !expected {
+            return Ok(());
+        }
+        ensure!(
+            !self.memories.is_empty(),
+            "Expected a shared memory in {linker_name} output ({}), but the module has no memory",
```
davidlattimore, line 838: I feel like we should add the linker name as context somewhere up the call stack so as we don't need to remember to include it in every error message. Although some of the other functions look to be similary structured, so it's fine to do this cleanup later.

wild/tests/sources/wasm/shared-memory/shared-memory.c
```diff
@@ -0,0 +1,22 @@
...
+//#Config:missing-atomics
+//#LinkArgs: --shared-memory
+//#ExpectError: 'atomics' feature must be used in order to use shared memory
+
+//#Config:unaligned
+//#LinkArgs: --shared-memory --max-memory=100000
+//#CompArgs: -matomics
+//#ExpectError: maximum memory must be aligned
+
+void _start(void) {}
```
davidlattimore: Could be good to add a variable that will go into the memory and check that is has the expected value.

wild/tests/integration_tests.rs
lapla-cogito: Let me clean up the other similar ones later.

wild/tests/integration_tests.rs

wild/tests/sources/wasm/shared-memory-disallowed/shared-memory-disallowed.s
lapla-cogito: Done.

wild/tests/sources/wasm/shared-memory/shared-memory.c
lapla-cogito: Done.

davidlattimore (approved):

libwild/src/args/wasm.rs
marxin: Thank you for the clarification - let's implement it the same as wasm-ld does.

marxin (approved):
Good job @lapla-cogito!

PR #2453, opened 2026-08-26
chore(tests): do not create extra context for directives

marxin:
An example before:
```
  --- stderr:
  Error: Failed to process test directive /home/marxin/Programming/wild/wild/tests/sources/elf/range-extension-thunks/range-extension-thunks.c:22
    Caused by:
      problematic directive line: 'MaxThunks:aa'
      Invalid MaxThunks value
      invalid digit found in string
```

and after:
```
  --- stderr:
  Error: Failed to process test directive /home/marxin/Programming/wild/wild/tests/sources/elf/range-extension-thunks/range-extension-thunks.c:22
    Caused by:
      problematic directive line: 'MaxThunks:aa'
      invalid digit found in string
```

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -2337,10 +2337,7 @@ fn process_directive(
             if config.variant_num.is_some() {
                 bail!("Variant can only be specified once per config");
             }
-            config.variant_num = Some(
-                arg.parse()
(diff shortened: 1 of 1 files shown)


davidlattimore (approved):
Thanks for the cleanup :)

PR #2455, opened 2026-08-27
port(MachO): Align `__thread_data` and `__thread_bss`

joanna-jo:
TLV template sections (`__thread_data` and `__thread_bss`) are copied contiguously at runtime to initialize thread-local storage, and each descriptor in `__thread_vars` contains an offset from the start of this storage to its corresponding thread-local variable. If a later TLV template section requires a stricter alignment than the previous one, then the offsets might not preserve that alignment in each thread-local copy. Therefore, this requires that we find the maximum alignment in all TLV template sections and normalize each section's alignment to this maximum.

`__thread_vars` is not included in this normalization and requires its own alignment requirement, which I plan on implementing next.

**Before**:
```
failures:

---- macho/aarch64/trivial-tls/default ----
  Caused by:
    Mach-O TLV template sections must use maximum alignment 64: __thread_data alignment: 4, __thread_bss alignment: 64
```

**After:**
```
running 1 test

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 24 filtered out; finished in 0.05s
```

Part of #2071

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -291,6 +291,8 @@ pub fn compute<'data, P: Platform, A: Arch<Platform = P>, F: FileSystem>(
         unreachable!();
     };
 
+    P::finalise_output_section_alignments(&section_part_sizes, &mut output_sections);
+
     let (mut section_part_layouts, mut section_layouts, mut resolved_location_counters) =
         compute_layout_sections::<A::Platform>(
             &section_part_sizes,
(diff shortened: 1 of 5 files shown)


libwild/src/layout.rs
```diff
@@ -5539,7 +5541,11 @@ fn compute_layout_sections<'data, P: Platform>(
 
                 for (part_id, part_size) in part_sizes.chain(empty_part) {
                     let part_layout = records_out.get_mut(part_id);
-                    let alignment = part_id.alignment(output_sections).min(max_alignment);
+                    let alignment = if is_first_part {
+                        max_alignment
+                    } else {
+                        part_id.alignment(output_sections).min(max_alignment)
+                    };
```
joanna-jo, line 5548: It looks like we previously relied on dense map behavior where the layout always visited the highest-alignment bucket first, which aligned the section start to its maximum alignment. But sparse maps skip absent buckets, which means the first stored bucket might be of a lower alignment than the maximum. Here, `__thread_data` needs a section-level alignment of 64 from `__thread_bss`, but it doesn't contain that bucket in its sparse map.

davidlattimore: Thanks for figuring that out and fixing it! I'll follow up with a change to an ELF test that would have picked this up.

davidlattimore (approved):

PR #2457, opened 2026-08-27
port(wasm): Always emit memory and `__stack_pointer`

lapla-cogito:
wasm-ld defines a linear memory and a `__stack_pointer` global for executables even when nothing in the input references them.

Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -4028,7 +4028,8 @@ fn setup_got_mem_and_indices<'data>(
                 &LinkerDefinedIndexRequest {
                     has_init_funcs,
                     export_symbols: requested_linker_export_symbols(symbol_db.args),
-                    has_memory: any_object_needs_linker_memory(layout_inputs),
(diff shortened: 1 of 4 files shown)


libwild/src/wasm.rs
```diff
@@ -4028,7 +4028,8 @@ fn setup_got_mem_and_indices<'data>(
                 &LinkerDefinedIndexRequest {
                     has_init_funcs,
                     export_symbols: requested_linker_export_symbols(symbol_db.args),
-                    has_memory: any_object_needs_linker_memory(layout_inputs),
+                    // Executables always get a defined linear memory.
+                    has_memory: true,
```
davidlattimore, line 4032: Is `has_memory` ever false? If not, are there plans for it to ever be false in future, or could we remove it?

lapla-cogito: I think it might be `false` when linking relocatable binaries.

davidlattimore (approved):

PR #2459, opened 2026-08-27
test: add lit integration for LLD tests

deepakshirkem:
Runs LLD's ELF test suite via lit with Wild substituted for ld.lld.

lit is already available in the CI container — no Dockerfile changes needed. A small `wild-lit.site.cfg.py` stub provides the required config variables. Currently 4 tests pass; failing tests use `--ignore-fail` pending fixes in Wild.

The existing Rust harness is kept for now, to be removed in a follow-up PR.

Part of #2380


davidlattimore (commented):
Thanks for looking into this. I think it'd still be good to invoke the test from a rust test. I'm unsure what granularity we should use. There's likely some overhead for each invocation of lit, especially if it has to scan a large directory full of tests. I guess one very course option would be to have one test per arch. That'd solve the issue that at the moment, it's only running the aarch64 tests.

wild/tests/external_tests/lld_lit_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+                .map_err(|e| Failed::from(e.to_string()))
+        }));
+    }
+
+    Ok(())
+}
+
+fn find_lit_binary() -> Option<PathBuf> {
+    // CI location
+    let ci_path = PathBuf::from("/usr/lib/llvm-21/bin/lit");
```
davidlattimore: Probably best to use `llvm_tools_dir` from `test_config` to find this

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,37 @@
...
+config.sizeof_void_p = 8
+config.has_plugins = False
+config.build_examples = False
+config.linked_bye_extension = False
+config.enable_threads = True
+config.ld_lld_default_mingw = False
+config.llvm_shlib_dir = "/tmp"
+config.llvm_shlib_ext = ".so"
+config.host_triple = "x86_64-unknown-linux-gnu"
+config.target_triple = "aarch64-unknown-linux-gnu"
```
davidlattimore: Don't we need to change this based on which arch we're running for?

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,37 @@
...
+config.enable_backtrace = False
+config.sizeof_void_p = 8
+config.has_plugins = False
+config.build_examples = False
+config.linked_bye_extension = False
+config.enable_threads = True
+config.ld_lld_default_mingw = False
+config.llvm_shlib_dir = "/tmp"
+config.llvm_shlib_ext = ".so"
+config.host_triple = "x86_64-unknown-linux-gnu"
```
davidlattimore: Could be good to pass this in via environment variable rather than hard coding it. We want to be able to run the tests on other hosts too.

wild/tests/external_tests/lld_lit_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+    lit_binary: &PathBuf,
+) -> Result {
+    let repo_root = crate::base_dir().join("..");
+    let cfg_src = repo_root.join("wild/tests/external_tests/wild-lit.site.cfg.py");
+    let cfg_link = test_dir.join("wild-lit.site.cfg.py");
+
+    // Create symlink so lit can find the config alongside lit.cfg.py.
+    // Use a temp name + rename for atomic replacement to avoid race conditions
+    // when multiple arch tests run concurrently.
+    let tmp_link = test_dir.join(format!("wild-lit.site.cfg.py.tmp.{}", std::process::id()));
```
davidlattimore: If the tests are run with `cargo test` rather than nextest, then we'll have multiple threads in the one process, so this path won't be unique and the processes will race.

wild/tests/external_tests/lld_lit_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+    std::fs::rename(&tmp_link, &cfg_link)?;
+
+    let wild_bin = std::path::Path::new(env!("CARGO_BIN_EXE_wild"));
+
+    let llvm_tools_dir = test_config.llvm_tools_dir.to_str().unwrap();
+
+    let output = Command::new(lit_binary)
+        .arg("--config-prefix")
+        .arg("wild-lit.site")
+        .arg("--ignore-fail")
```
davidlattimore: With this flag, the rust test will pass regardless of which lld tests fail. I'd be fine with either filtering the tests (--filter / filter-out) so that we only run those that we expected to pass, or passing `--xfail` with a list of tests that we expect to fail.

wild/tests/external_tests/lld_lit_tests.rs
davidlattimore, line 1: I'd be inclined to put this code in `lld_tests.rs` and delete the old code that's there that is no longer needed.

wild/tests/external_tests/lld_lit_tests.rs
```diff
@@ -0,0 +1,94 @@
...
+    filter: &crate::Filter,
+    test_config: &TestConfig,
+) -> Result {
+    let test_dir = crate::base_dir().join("../external_test_suites/lld/test");
+    if !test_dir.exists() {
+        return Ok(());
+    }
+
+    let Some(lit_binary) = find_lit_binary() else {
+        return Ok(());
```
davidlattimore: If the developer tries to run the external tests, but lit can't be found, we should probably fail rather than just not running any tests.

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,37 @@
...
+import lit.llvm
+
+LLVM_TOOLS_DIR = os.environ.get("LLVM_TOOLS_DIR", "/usr/lib/llvm-21/bin")
+WILD = os.environ.get("WILD_BIN", "wild")
+
+lit.llvm.initialize(lit_config, config)
+config.llvm_tools_dir = LLVM_TOOLS_DIR
+config.llvm_build_dir = LLVM_TOOLS_DIR
+
+config.lld_obj_root = "/tmp/lld-obj"
```
davidlattimore: Is it OK that multiple concurrent runs will be sharing an output directory? I'd be somewhat inclined to make this be a subdirectory of the temp dir and pass the temp dir in via environment variable.

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,37 @@
...
+config.host_triple = "x86_64-unknown-linux-gnu"
+config.target_triple = "aarch64-unknown-linux-gnu"
+
+lit_config.load_config(config, os.path.join(os.path.dirname(__file__), 'lit.cfg.py'))
+
+# Override ld.lld with Wild
+new_subs = []
+for pat, sub in config.substitutions:
+    if re.search(r'ld\.lld', sub):
+        new_subs.append((pat, f'{WILD} -m aarch64elf'))
```
davidlattimore: The architecture is hard-coded again here

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,37 @@
...
+config.llvm_shlib_dir = "/tmp"
+config.llvm_shlib_ext = ".so"
+config.host_triple = "x86_64-unknown-linux-gnu"
+config.target_triple = "aarch64-unknown-linux-gnu"
+
+lit_config.load_config(config, os.path.join(os.path.dirname(__file__), 'lit.cfg.py'))
+
+# Override ld.lld with Wild
+new_subs = []
+for pat, sub in config.substitutions:
```
davidlattimore: From what I can see, your approach to overriding the linker is to let lit find lld, then to text substitutions to override the config to substitute wild for lld. I feel like perhaps a more robust mechanism might be to create a fakes directory, then point `config.lld_tools_dir` to it. From what I can tell, the fakes directory will need to have all the different linker names - `ld.lld`, `lld-link`, `ld64.lld` and `wasm-ld`, with all being symlinks to wild.

wild/tests/external_tests/lld_lit_tests.rs
deepakshirkem: Done.

wild/tests/external_tests/wild-lit.site.cfg.py
deepakshirkem: Done.

wild/tests/external_tests/wild-lit.site.cfg.py
deepakshirkem: Done.

wild/tests/external_tests/lld_lit_tests.rs
deepakshirkem: Done.

wild/tests/external_tests/lld_lit_tests.rs
deepakshirkem: Two options: --filter (only run known passing tests, simpler) or --xfail (run all, mark known failures as expected, closer to existing expect_failure mechanism). Which do you prefer?

wild/tests/external_tests/lld_lit_tests.rs
deepakshirkem: Done.

wild/tests/external_tests/lld_lit_tests.rs
deepakshirkem: Done.

wild/tests/external_tests/wild-lit.site.cfg.py
deepakshirkem: Done.

wild/tests/external_tests/wild-lit.site.cfg.py
deepakshirkem: Done.

wild/tests/external_tests/lld_lit_tests.rs
davidlattimore: Either is fine. We will however want a way to discover tests that are passing that our toml file lists as expected to fail. So if we don't do xfail now, then we'll probably need to do it before too long.

wild/tests/external_tests/wild-lit.site.cfg.py
deepakshirkem: Done

wild/tests/external_tests/wild-lit.site.cfg.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Set lld_tools_dir BEFORE load_config so use_lld() inside lit.cfg.py
+# picks up our fakes directory containing symlinks to Wild.
+FAKES_DIR = os.environ.get("WILD_FAKES_DIR", "")
+if FAKES_DIR:
+    config.lld_tools_dir = FAKES_DIR
+
+lit_config.load_config(config, os.path.join(os.path.dirname(__file__), 'lit.cfg.py'))
+
+# If no fakes dir, fall back to text substitution
```
davidlattimore: When would we not have a fakes dir? Shouldn't we just fail?

wild/tests/external_tests/lld_tests.rs
```diff
@@ -45,238 +21,101 @@ pub(crate) fn collect_tests(
...
-    let file_name = path
-        .file_name()
-        .expect("Must be a valid filename")
-        .to_str()
-        .expect("Expected valid string name");
+    let tmpdir = tempfile::tempdir()?;
+    let output = Command::new(lit_binary)
+        .arg("--config-prefix")
+        .arg("wild-lit.site")
+        .arg("--ignore-fail")
```
davidlattimore: It seems that we're still ignoring failures

wild/tests/external_tests/lld_tests.rs
```diff
@@ -45,238 +21,101 @@ pub(crate) fn collect_tests(
...
+    lit_binary: &PathBuf,
+) -> Result {
+    let repo_root = crate::base_dir().join("..");
+    let cfg_src = repo_root.join("wild/tests/external_tests/wild-lit.site.cfg.py");
+    let cfg_link = test_dir.join("wild-lit.site.cfg.py");
+
+    // Create symlink so lit can find the config alongside lit.cfg.py.
+    // Use a temp name + rename for atomic replacement to avoid race conditions
+    // when multiple arch tests run concurrently.
+    let tmp_link = test_dir.join(format!("wild-lit.site.cfg.py.tmp.{arch}"));
```
davidlattimore: It seems like we're modifying files inside the git submodule. I don't think we should be doing that. If we absolutely need to do it, then they should be git-ignored. Ideally, any files we create should be in a temporary directory.

davidlattimore: It doesn't look like you've addressed this comment

davidlattimore (commented):
Can you verify that removing a failing test from the skip list causes a failure and that adding a passing test to the skip list also causes a failure?

wild/tests/external_tests/lld_tests.rs
deepakshirkem: https://github.com/wild-linker/lld-tests/pull/4

davidlattimore: Why can't we create the files in a temporary directory?

davidlattimore (commented):
Thanks. I think this is getting close to where we can merge it.

I tried adding the passing test "aarch64-relative.s" to the skip list as you did. When I run the test, it shows that the test unexpected passed, but then the rust test passes, when it should have failed.

```
Total Discovered Tests: 2052
  Excluded           : 1918 (93.47%)
  Passed             :    4 (0.19%)
  Expectedly Failed  :  129 (6.29%)
  Unexpectedly Passed:    1 (0.05%)
lit: /usr/lib/llvm-21/utils/lit/lit/llvm/config.py:520: note: using ld.lld: /tmp/.tmpufIF63/ld.lld
lit: /usr/lib/llvm-21/utils/lit/lit/llvm/config.py:520: note: using lld-link: /tmp/.tmpufIF63/lld-link
lit: /usr/lib/llvm-21/utils/lit/lit/llvm/config.py:520: note: using ld64.lld: /tmp/.tmpufIF63/ld64.lld
lit: /usr/lib/llvm-21/utils/lit/lit/llvm/config.py:520: note: using wasm-ld: /tmp/.tmpufIF63/wasm-ld
test external_test_suites/lld_lit/aarch64 ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2156 filtered out; finished in 1.42s
```

The tests are also quite noisy when run under `cargo test`. Is there anything we can do about that?

I expect that we'll want an easy way to run a single lld test. Not sure what that way should be. Any ideas?

davidlattimore (approved):
Thanks for working on this :)

PR #2461, opened 2026-08-28
feat: implement --emit-relocs

vishruth-thimmaiah:
Adds support for `--emit-relocs`. This flag is used to retain all relocations from input files, similar to a partial object.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2072,14 +2072,31 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
                     num_globals += 1;
                 }
                 strings_size += symtab_name_for_strtab(info.name).len() + 1;
-            } else if symbol_db.args.should_output_partial_object
+            } else if (symbol_db.args.emit_relocs())
                 && sym.is_undefined()
-                && symbol_db.is_canonical(symbol_id)
                 && let Ok(name) = state.object.symbol_name(sym)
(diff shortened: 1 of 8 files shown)


libwild/src/elf.rs
```diff
@@ -2071,14 +2072,31 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
...
+                    .get_atomic(canonical)
+                    .get()
+                    .contains(ValueFlags::DYNAMIC);
+                if (symbol_db.is_canonical(symbol_id) || is_dynamic)
+                    && def_file != crate::input_data::PRELUDE_FILE_ID
+                {
+                    let old_flags = per_symbol_flags
+                        .get_atomic(canonical)
+                        .fetch_or(ValueFlags::SYMTAB_INSTALLED);
+                    if !old_flags.contains(ValueFlags::SYMTAB_INSTALLED) {
```
davidlattimore, line 2092: Is this deterministic? It looks like it might not be. I wonder if it might be simpler if you just set the flag on the canonical definition and made it responsible for writing the symtab entry.

davidlattimore (commented):
Does this get offsets correct when riscv size-reducing relaxations occur? If not, it might be reasonable to error if the flag is enabled concurrently with size-reducing relaxations - unless there's an easy way to make them work.

PR #2464, opened 2026-08-30
fix: Use relaxed size for exported dynsym and allow dynsym debug refs

davidlattimore:
This started out as a refactoring, but then I found that several of the differences between the symtab caller of `copy_symbol` and the dynsym callers of `copy_symbol` were actually bugs. They're all probably pretty obscure. Things like getting the symbol size in dynsym wrong if the function was reduced in size by a risc-v relaxation.

### DIFF
--- a/libwild/src/elf_writer.rs
+++ b/libwild/src/elf_writer.rs
@@ -1363,6 +1363,30 @@ struct VersionedDynsymWriter<'layout, 'out, C: ElfClass> {
     versym: Option<&'out mut [Versym]>,
 }
 
+fn object_symbol_size<C: ElfClass>(
+    sym: &elf::SymtabEntry<C>,
+    sym_index: SymbolIndex,
+    object: &ObjectLayout<elf::Elf<C>>,
+) -> Result<u64> {
+    let e = LittleEndian;
+    let st_size: u64 = sym.st_size(e).into();
+    if st_size == 0 {
+        return Ok(0);
+    }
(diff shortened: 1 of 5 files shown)


libwild/src/elf_writer.rs
```diff
@@ -4924,120 +4961,79 @@ fn write_regular_object_dynamic_symbol_definition<'data, C: ElfClass>(
...
+            layout.symbol_debug(symbol_id)
+        )
+    })?;
+
+    // For non-PIE executables, export IFUNC symbols as STT_FUNC pointing to PLT stub.
+    // For PIE executables, keep IFUNC as-is.
+    if section_index.is_some()
+        && resolution.flags.is_ifunc()
+        && layout.symbol_db.output_kind.is_executable()
+        && !layout.symbol_db.output_kind.is_relocatable()
```
mati865: Doesn't `!is_relocatable()` already cover `is_exectuable()` case?

davidlattimore: The call is needed, but is confusing, since `is_relocatable` doesn't do what one might expect - that's `is_partial_object`. I added a test that fails if the call is removed. I also changed the call to the much less confusing `is_position_independent`. I think I might see if I do a follow-up to get rid of `is_reloctable`.

mati865: Oh, now I see what you mean. I looked at that code but it didn't occur to me that `is_relocatable` could mean literally `OutputKind::Relocatable` rather than what I used to remember.

mati865 (approved):

PR #2465, opened 2026-08-30
fix: don't panic on unsupported R_AARCH64_FUNCINIT64 relocation

deepakshirkem:
new_relaxation() called unwrap() on the result of relocation_from_raw(), causing a panic when encountering unknown relocation types such as R_AARCH64_FUNCINIT64 (0x13d). Changed to ok()? so unknown relocations return None (no relaxation available) instead of panicking.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -6080,6 +6080,18 @@ fn classify_symbol_relocation<
         next_modifier = relaxation.next_modifier();
         relaxation.rel_info()
     } else {
+        // R_AARCH64_FUNCINIT64 has specific restrictions.
+        if r_type == object::elf::RelocationType(0x13d) {
+            if flags.is_ifunc() {
+                let sym_name = symbol_db.symbol_name_for_display(local_symbol_id);
+                bail!(
+                    "relocation R_AARCH64_FUNCINIT64 cannot be used against ifunc symbol '{sym_name}'"
+                );
+            } else if !flags.is_dynamic() && !flags.is_interposable() {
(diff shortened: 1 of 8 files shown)


libwild/src/elf.rs
```diff
@@ -6080,6 +6080,18 @@ fn classify_symbol_relocation<
         next_modifier = relaxation.next_modifier();
         relaxation.rel_info()
     } else {
+        // R_AARCH64_FUNCINIT64 has specific restrictions.
+        if r_type == object::elf::RelocationType(0x13d) {
```
marxin: I think we should map this specific relocation to a new type `RelocationKind::NonPreemptiveFnCheck` (or something similar) and then handle it in the big switch in this function (`match rel_info.kind`). Note we have a similar relocation `RelocationKind::Alignment`, where we should at least make a validation on LoongArch64 platform (we're not doing it now).

Btw. have you experienced any toolchain that produces the `R_AARCH64_FUNCINIT64` relocation, or is it coming from a synthetic test-case?

marxin (changes requested):

libwild/src/elf.rs
deepakshirkem: >Btw. have you experienced any toolchain that produces the R_AARCH64_FUNCINIT64 relocation, or is it coming from a synthetic test-case?

It came from the LLD test suite (aarch64-funcinit64-invalid.s). 

But I’m not sure if adding this fix will improve Wild. Still, this test satisfies what LLD has, but we still can’t remove it from the skip list. So do you think adding this is helping Wild or not?

Issue #2016, opened 2026-06-04
Support --separate-debug


agrieve:
When looking at mold vs wild for chrome, this came up as a really nice feature that I was lld had. Figured I'd file a feature request for wild to consider it as well! It's particularly useful for Android dev, where you always want to push a stirpped .so to the device, and always want a symbols file locally for symbolization, but actually never really want a .so with both. This flag both saves a call to `strip`, and makes debug .so files smaller by having them not include the code.

PR #2472, opened 2026-09-01
feat: add support for --only-keep-debug

### LINKED ISSUE #1221: Add support for --only-keep-debug

Blazearth:
Implements `--only-keep-debug` for ELF outputs.

The debug-only output is produced as a post-layout transformation: allocatable non-NOTE sections are converted to SHT_NOBITS`, while their addresses and debug information are preserved. File offsets and segment sizes are recalculated accordingly.

Also handle the related writer paths and build-id restrictions.
It also includes standalone tests for ELF structure invariants, address preservation, build-id matching, and compression interaction.
I came across the existing draft PR #2241 while working on #1221. This takes a post-layout approach based on the discussion in that PR and avoids the address shifting issue in the current implementation.
Related to #2016
Closes #1221


wild/tests/only_keep_debug_test.rs
```diff
@@ -0,0 +1,380 @@
+#![cfg(all(target_os = "linux", target_arch = "x86_64"))]
+//! Standalone tests for --only-keep-debug ELF output properties.
```
davidlattimore: I think it'd be better to extend the existing integration test infrastructure to support what we need in order to test this feature rather than creating a whole parallel test setup just for this feature. e.g. a test directive, that causes the linker to be run twice, once with `--strip-debug` and once with `--only-keep-debug`, then performs consistency checks on the two outputs. That would make it easier, if we find bugs in the implementation, to add new integration tests that directly demonstrate those bugs. It'd also allow us to directly use existing test directives in those same tests.

davidlattimore (commented):
Thanks for working on this. I suspect I might later have comments on the implementation, but for now I've just got a comment on the tests.

wild/tests/only_keep_debug_test.rs
Blazearth: Thanks for the feedback. I have removed the standalone test file and integrated `--only-keep-debug` testing directly into the integration test framework.
The reference build is intentionally unstripped because using `--strip-debug` before linking can change section GC/reachability and therefore the final layout and symbol addresses.
I have also incorporated your feedbacks from the draft pr on this.

davidlattimore: Thanks, that direction for testing definitely looks better.

`--strip-debug` shouldn't change reachability. What I suspect you're seeing is that `--strip-debug` means we don't include debug sections in the sections table and since the sections table is near the start of the file and is addressable, changing its size shifts the addresses of everything that follows.

But having `--strip-debug` and `--only-keep-debug` match address layout is the point of the feature, so I think we need to find a way for that to work. One option would be to move the section headers table to the end of the file and don't include it in a LOAD segment. If we're doing that, then it should probably be done first as a separate PR.

Blazearth: You're completely right. I confirmed the behavior and checked how other linkers handle this.

1. **Address shift in Wild:** Reachability was unchanged. The shift is caused by `SECTION_HEADERS` being an allocatable internal output section inside the first `PT_LOAD`. Stripping the 6 debug sections removes 6 section-header entries,shrinking the table by 384 bytes (`6 * 64`), which shifts the following virtual addresses and offsets by exactly `0x180`.

2. **Other linkers:** I tested `ld.bfd`, `ld.lld`, `ld.gold`, and `mold`.
In each case, the ELF section-header table is placed at the end of the file and outside the `LOAD` segments. Consequently, removing debug sections does not shift the tested memory addresses between the normal, `--strip-debug`,and debug-only outputs.

I'll put together a separate PR first to move Wild's section-header table to the end of the file and keep it out of `PT_LOAD`, as you suggested. Once that lands, I'll update and rebase this PR to compare `--only-keep-debug` directly against `--strip-debug`.

davidlattimore (commented):
Looks like there are some merge conflicts. I'll take another look after they're resolved.

davidlattimore (commented):
Can you add or extend tests to make sure this works in conjunction with dynamic linking, range-extension thunks and non-debug merged strings?

Blazearth:
>Can you add or extend tests to make sure this works in conjunction with dynamic linking, range-extension thunks and non-debug merged strings?

Thanks for the feedback! I've added tests for dynamic linking, range-extension thunks, and non-debug merged strings, and fixed the related write paths so `--only-keep-debug` doesn't try to write into allocatable sections that have been converted to `SHT_NOBITS`.

davidlattimore (commented):
The following patch adds a couple of tests that currently fail. Can you take a look at fixing them?

```patch
diff --git a/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c b/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
index 22d120a4a0..18fd40acf5 100644
--- a/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
+++ b/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
@@ -44,6 +44,26 @@
 //#TestOnlyKeepDebug:true
 //#ExpectSym:_start
 
+//#Config:only-keep-debug-copy-relocations:only-keep-debug-dynamic
+//#Arch:x86_64,aarch64,riscv64
+//#CompArgs:-g -fno-pic -fno-pie -DCOPY_RELOCATION
+//#CompSoArgs:-fPIC
+//#LinkArgs:-z now
+//#WildExtraLinkArgs:--only-keep-debug
+//#SoSingleLinker:lld
+//#ExpectSym:shared_data section=".bss"
+
+//#Config:only-keep-debug-got-plt-syms:only-keep-debug-dynamic
+//#SkipArch:ppc64le
+//#LinkArgs:-z now
+//#WildExtraLinkArgs:--only-keep-debug --got-plt-syms
+//#SoSingleLinker:lld
+//#ExpectSym:shared_func$got
+
+//#Config:only-keep-debug-android-relr:only-keep-debug
+//#LinkArgs:--only-keep-debug -pie --pack-dyn-relocs=relr --use-android-relr-tags
+//#ExpectSection:.relr.dyn type=8
+
 #include "../common/runtime.h"
 
 int global_var = 42;
@@ -53,9 +73,16 @@
 
 __attribute__((weak)) int shared_func(int x);
 
+#ifdef COPY_RELOCATION
+extern int shared_data;
+#endif
+
 void _start(void) {
   runtime_init();
   int val = global_var;
+#ifdef COPY_RELOCATION
+  val += shared_data;
+#endif
   if (shared_func) {
     val = shared_func(val);
   }
diff --git a/wild/tests/sources/elf/only-keep-debug/shared.c b/wild/tests/sources/elf/only-keep-debug/shared.c
index 59587c8972..fc735e134f 100644
--- a/wild/tests/sources/elf/only-keep-debug/shared.c
+++ b/wild/tests/sources/elf/only-keep-debug/shared.c
@@ -1,1 +1,3 @@
 int shared_func(int x) { return x + 10; }
+
+int shared_data = 7;
```

wild/tests/integration_tests.rs
```diff
@@ -213,6 +213,14 @@
 //! having been pre-filled with random data. It then compares the output of the two runs to verify
 //! that they're the same.
 //!
+//! TestOnlyKeepDebug:{bool} Whether to perform additional testing of the --only-keep-debug flag.
+//! When true, re-links the same inputs without --only-keep-debug and compares the two outputs:
+//! symbol addresses and section addresses must match, allocatable non-NOTE sections must be
+//! SHT_NOBITS, debug sections must exist and be non-empty, .symtab must exist, and the debug output
+//! must be smaller than the normal output. The reference link is unstripped (identical arguments
```
davidlattimore, line 220: Can you check over this description. The bit about debug output being smaller than the normal output isn't something that's checked anymore right?

libwild/src/elf_writer.rs
```diff
@@ -6254,13 +6307,17 @@ fn write_dynamic_file<'data, C: ElfClass, A: Arch<Platform = elf::Elf<C>>>(
                 }
             }
 
-            table_writer
-                .process_resolution::<A>(Some(layout), layout.args(), res)
-                .with_context(|| format!("Failed to write {}", layout.symbol_debug(symbol_id)))?;
+            if !only_keep_debug {
```
davidlattimore, line 6310: This guard looks to be unnecessary since `process_resolution` already checks this

libwild/src/only_keep_debug.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::error::Result;
+use crate::layout::Layout;
+use crate::platform::Arch;
+use crate::platform::Args as _;
+use crate::platform::SectionFlags as _;
+use crate::timing_phase;
+use linker_utils::elf::sht;
+
+#[allow(clippy::unnecessary_wraps)]
+pub(crate) fn maybe_only_keep_debug_elf<C: ElfClass, A: Arch<Platform = elf::Elf<C>>>(
```
davidlattimore, line 16: The generic parameter A appears to be unused. Can you remove it?

libwild/src/only_keep_debug.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::elf::ElfClass;
+use crate::error::Result;
+use crate::layout::Layout;
+use crate::platform::Arch;
+use crate::platform::Args as _;
+use crate::platform::SectionFlags as _;
+use crate::timing_phase;
+use linker_utils::elf::sht;
+
+#[allow(clippy::unnecessary_wraps)]
```
davidlattimore, line 15: Is there any reason we couldn't remove the return value rather than supressing the warning?

libwild/src/platform.rs
```diff
@@ -1517,6 +1523,10 @@ pub(crate) trait Args: std::fmt::Debug + Send + Sync + 'static {
 
     fn should_strip_all(&self) -> bool;
 
+    fn only_keep_debug(&self) -> bool {
```
davidlattimore, line 1526: Is this actually called from any non-ELF code? If it's not, this could be removed and we can just have the elf-specific implementation

libwild/src/compression.rs
```diff
@@ -472,7 +472,115 @@ fn update_allocation_sizes<P: Platform>(layout: &mut Layout<P>) {
...
+                for (_, seg_file_end) in &mut seg_stack {
+                    *seg_file_end = (*seg_file_end).max(file_offset);
+                }
+            }
+            _ => {}
+        }
+    }
+}
+
+pub(crate) fn update_file_offset<P: Platform>(layout: &mut Layout<P>) -> Result {
```
davidlattimore, line 583: It doesn't look like this needs to be `pub(crate)`

davidlattimore (commented):
Here's another case that fails:

```patch
diff --git a/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c b/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
index 77347095fe..96f2ad42ec 100644
--- a/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
+++ b/wild/tests/sources/elf/only-keep-debug/only-keep-debug.c
@@ -60,6 +60,12 @@
 //#SoSingleLinker:lld
 //#ExpectSym:shared_func$got
 
+//#Config:only-keep-debug-prelude-got-syms:only-keep-debug
+//#Arch:x86_64
+//#CompArgs:-g -fPIC -DPRELUDE_GOT
+//#LinkArgs:--only-keep-debug --got-plt-syms --no-relax
+//#ExpectSym:__ehdr_start$got section=".got"
+
 //#Config:only-keep-debug-android-relr:only-keep-debug
 //#LinkArgs:--only-keep-debug -pie --pack-dyn-relocs=relr --use-android-relr-tags
 //#ExpectSection:.relr.dyn type=8
@@ -80,12 +86,19 @@
 extern int shared_data;
 #endif
 
+#ifdef PRELUDE_GOT
+extern const unsigned char __ehdr_start;
+#endif
+
 void _start(void) {
   runtime_init();
   int val = global_var;
 #ifdef COPY_RELOCATION
   val += shared_data;
 #endif
+#ifdef PRELUDE_GOT
+  val += __ehdr_start;
+#endif
   if (shared_func) {
     val = shared_func(val);
   }
```

PR #2473, opened 2026-09-01
test(aarch64): verify and categorize non-PIC shared object failures

deepakshirkem:
Adds a Wild integration test verifying that LDST ABS_LO12 relocations are correctly rejected in shared objects.

Improves skip list categorization with a new gc_sections_skips_validation group for tests that fail due to Wild's intentional gc-sections-before- validation design.

Part of #2380

### DIFF
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -94,7 +94,17 @@ impl crate::platform::Arch for ElfAArch64 {
     fn is_disallowed_in_shared_object(r_type: object::elf::RelocationType) -> bool {
         matches!(
             r_type,
-            object::elf::R_AARCH64_ABS32 | object::elf::R_AARCH64_ABS16
+            object::elf::R_AARCH64_ABS32
+                | object::elf::R_AARCH64_ABS16
+                | object::elf::R_AARCH64_PREL16
+                | object::elf::R_AARCH64_PREL32
+                | object::elf::R_AARCH64_PREL64
+                | object::elf::R_AARCH64_ADD_ABS_LO12_NC
+                | object::elf::R_AARCH64_ADR_PREL_LO21
+                | object::elf::R_AARCH64_ADR_PREL_PG_HI21
+                | object::elf::R_AARCH64_LDST8_ABS_LO12_NC
+                | object::elf::R_AARCH64_LDST32_ABS_LO12_NC
+                | object::elf::R_AARCH64_LDST64_ABS_LO12_NC
         )
     }
 
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -94,12 +94,7 @@ impl crate::platform::Arch for ElfAArch64 {
     fn is_disallowed_in_shared_object(r_type: object::elf::RelocationType) -> bool {
         matches!(
             r_type,
-            object::elf::R_AARCH64_ABS32
-                | object::elf::R_AARCH64_ABS16
-                | object::elf::R_AARCH64_ADD_ABS_LO12_NC
-                | object::elf::R_AARCH64_LDST8_ABS_LO12_NC
-                | object::elf::R_AARCH64_LDST32_ABS_LO12_NC
-                | object::elf::R_AARCH64_LDST64_ABS_LO12_NC
+            object::elf::R_AARCH64_ABS32 | object::elf::R_AARCH64_ABS16
         )
     }
 
--- a/libwild/src/elf_aarch64.rs
+++ b/libwild/src/elf_aarch64.rs
@@ -96,12 +96,7 @@ impl crate::platform::Arch for ElfAArch64 {
             r_type,
             object::elf::R_AARCH64_ABS32
                 | object::elf::R_AARCH64_ABS16
-                | object::elf::R_AARCH64_PREL16
-                | object::elf::R_AARCH64_PREL32
-                | object::elf::R_AARCH64_PREL64
                 | object::elf::R_AARCH64_ADD_ABS_LO12_NC
-                | object::elf::R_AARCH64_ADR_PREL_LO21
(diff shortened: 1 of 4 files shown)


libwild/src/elf_aarch64.rs
```diff
@@ -94,7 +94,17 @@ impl crate::platform::Arch for ElfAArch64 {
     fn is_disallowed_in_shared_object(r_type: object::elf::RelocationType) -> bool {
         matches!(
             r_type,
-            object::elf::R_AARCH64_ABS32 | object::elf::R_AARCH64_ABS16
+            object::elf::R_AARCH64_ABS32
```
marxin: Intentionally leaving that part to @davidlattimore.

wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -236,3 +225,19 @@ tests = [
   "lto/aarch64-pac-got-func.ll",
   "aarch64-reloc-implicit-addend.test",
 ]
+
+[skipped_groups.gc_sections_skips_validation]
+reason = "Wild skips validation for GC'd sections by design. These tests expect errors on relocations in sections removed by --gc-sections."
```
marxin, line 236: Great, if that's the case, then we can add an extra configuration for the group and pass `--no-gc-section` to the Wild linker. By doing that, we should hopefully make majority of the tests green. Can you give it a try?

marxin (changes requested):

libwild/src/elf_aarch64.rs
```diff
@@ -94,7 +94,17 @@ impl crate::platform::Arch for ElfAArch64 {
...
+                | object::elf::R_AARCH64_ABS16
+                | object::elf::R_AARCH64_PREL16
+                | object::elf::R_AARCH64_PREL32
+                | object::elf::R_AARCH64_PREL64
+                | object::elf::R_AARCH64_ADD_ABS_LO12_NC
+                | object::elf::R_AARCH64_ADR_PREL_LO21
+                | object::elf::R_AARCH64_ADR_PREL_PG_HI21
+                | object::elf::R_AARCH64_LDST8_ABS_LO12_NC
+                | object::elf::R_AARCH64_LDST32_ABS_LO12_NC
+                | object::elf::R_AARCH64_LDST64_ABS_LO12_NC
```
davidlattimore: Can you add tests for these? I'd like where possible not to rely on external tests, but rather have them as an extra layer of defence and to help us find gaps.

libwild/src/elf_aarch64.rs
```diff
@@ -94,7 +94,17 @@ impl crate::platform::Arch for ElfAArch64 {
     fn is_disallowed_in_shared_object(r_type: object::elf::RelocationType) -> bool {
         matches!(
             r_type,
-            object::elf::R_AARCH64_ABS32 | object::elf::R_AARCH64_ABS16
+            object::elf::R_AARCH64_ABS32
+                | object::elf::R_AARCH64_ABS16
+                | object::elf::R_AARCH64_PREL16
```
davidlattimore: This sounds like a relative relocation, but it looks like the function is only called when we have RelocationKind::Absolute. Does adding this here actually do anything?

libwild/src/elf_aarch64.rs
deepakshirkem: Done.

libwild/src/elf_aarch64.rs
deepakshirkem: Yes, Initially I wanted to handle all relocation kinds, but extending the check to RelocationKind::Relative caused regressions by rejecting legitimate relative relocations in shared objects. Keeping it to Absolute only for now, with PREL relocations to be handled in a follow-up

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: @marxin @davidlattimore  We still have one issue the error message wording differs from lld's. Wild wraps errors with Failed to copy section  Caused by: context which prevents FileCheck from matching. 


Screenshot from 2026-09-04 00-42-38

davidlattimore: I mean we could have a wrapper around wild that does things like adds --no-gc-sections and applies regular expressions to search/replace error message, but I'm doubtful if it's worth it to make get the tests to pass.

libwild/src/elf_aarch64.rs
davidlattimore: Makes sense. I'd be somewhat inclined to remove the relocations that aren't yet actually getting checked, otherwise the code is misleading as to what's happing.

wild/tests/external_tests/lld_skip_tests.toml
deepakshirkem: I think this is one-time work. Once we build this infrastructure, we don't have to revisit it any future test with the same situation benefits automatically. It's not just these 3 tests; there could be many more requiring this kind of handling. Adding them all to the skip list is not a good long-term solution. The confidence we gain from passing these tests is worth the investment. I'm happy with whatever decision you make.

libwild/src/elf_aarch64.rs
deepakshirkem: Sorry for the confusion. Removed the relocations that weren't actually being checked.

wild/tests/external_tests/lld_skip_tests.toml
davidlattimore: If you're keen to go the route of a compatibility wrapper, then it's fine by me. We can see how it works out and if it's proving to be a maintenance burden, then we can revisit the decision later.

davidlattimore (commented):
I'm likely happy to merge whatever categorisations you think are appropriate. I'll have a final look once the merge conflicts are resolved.

PR #2475, opened 2026-09-01
feat: fetch symbols defined in a linker script in a location counter

vishruth-thimmaiah:
Adds support for fetching the value of a symbol defined in a linker script, within a location counter. For now, we only support fetching the first value, i.e. if a linker script has multiple symbol redefinitions, we only fetch the first one.

### DIFF
--- a/libwild/src/expression_eval.rs
+++ b/libwild/src/expression_eval.rs
@@ -9,19 +9,20 @@ use crate::error::Result;
 use crate::grouping::Group;
 use crate::input_data::InputRef;
 use crate::layout;
-use crate::layout::EarlyObjectSymbolValue;
 use crate::layout::FileLayoutState;
 use crate::layout::GroupState;
 use crate::layout::InputSectionPositions;
 use crate::layout::MemoryRegion;
 use crate::layout::OutputRecordLayout;
 use crate::linker_script::Expression;
+use crate::output_section_id::OutputSectionId;
 use crate::output_section_id::OutputSections;
 use crate::output_section_id::SectionName;
 use crate::output_section_map::OutputSectionMap;
 use crate::output_section_part_map::OutputSectionPartMap;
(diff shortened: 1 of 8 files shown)


libwild/src/expression_eval.rs
```diff
@@ -385,6 +397,174 @@ pub(crate) fn evaluate_const<'data>(expr: &Expression<'data>) -> Result<u64> {
     }
 }
 
+pub(crate) fn evaluate_early_expression<'data, P: Platform>(
```
davidlattimore, line 485: Is there any scope for sharing more logic with late evaluation?

wild/tests/sources/elf/linker-script-location-counter/linker-script-location-counter-cyclic-symbol.ld
davidlattimore, line 1: Should something like the following be expected to work?

```
ENTRY(begin_here)

first = second;
second = 0x600000;

SECTIONS
{
    . = first;
    .text : {
        *(.text)
        *(.text.*)
    }
    .data : {
        *(.data)
        *(.data.*)
    }
    .bss : {
        KEEP(*(.bss))
    }
}

ASSERT(first == second, "forward-defined symbol has the wrong value");
```

vishruth-thimmaiah: This would require late evaluation to perform deferred symbol resolution, which we do not yet support.

libwild/src/expression_eval.rs
vishruth-thimmaiah: The earlier way of handling relative symbols in late evaluation was not entirely correct, I've now unified the section relative conversion between early and late evaluation.

davidlattimore (approved):

PR #2485, opened 2026-09-03
port(wasm): Support `--demangle` and `--no-demangle`

### LINKED ISSUE #2339: Wasm: support `--demangle` and `--no-demangle`
As opposed to the other platforms (like ELF), the options not only influence the error messages:
```
❯ clang++ --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot missing.C
wasm-ld: error: /tmp/missing-a5a2db.o: undefined symbol: Foo::meh()
❯ clang++ --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot missing.C -Wl,-no-demangle
wasm-ld: error: /tmp/missing-fafba7.o: undefined symbol: _ZN3Foo3mehEv
```

but it's also responsible for demangling of the symbol names in the symbol table:
```
clang++ --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot main.C
❯ llvm-nm a.out | rg Foo
000002ba t Foo::Foo()
000002fd t Foo::meh()
clang++ --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot main.C -Wl,-no-demangle
❯ llvm-nm a.out | rg Foo
000002fd t _ZN3Foo3mehEv
000002ba t _ZN3FooC2Ev
```

Note there's one quirk in Clang, where `-Wl,--no-demangle` option is not properly converted to `wasm-ld` option, will open a PR. Apparently, one can use `-Wl` syntax with a single dash and it's passed (and interpreted by `wasm-ld`) correctly: https://github.com/llvm/llvm-project/pull/213878.

lapla-cogito:
As mentioned in the original issue, when demangling occurs, the contents of the name section also follow suit.

Close #2339 
Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -1636,12 +1636,21 @@ fn encode_wasm_section(section: &impl wasm_encoder::Section) -> Vec<u8> {
     bytes
 }
 
+fn demangle_symbol_name(name: &str, demangle: bool) -> Cow<'_, str> {
+    if demangle {
+        symbolic_demangle::demangle(name)
+    } else {
+        Cow::Borrowed(name)
+    }
+}
+
 fn build_name_section<'data>(
     layout: &WasmLayout<'data>,
     layout_inputs: &[WasmObjectLayoutInput<'data>],
     indices: &LinkerDefinedIndices,
     got_mem: &GotMem,
     got_func: &GotFunc,
+    demangle: bool,
 ) -> Option<wasm_encoder::NameSection> {
     let (n_func_imports, n_global_imports) = count_output_imports(layout);
     let n_funcs = n_func_imports + layout.function_type_indices.len();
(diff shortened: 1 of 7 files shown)


libwild/src/wasm.rs
```diff
@@ -1700,7 +1709,7 @@ fn build_name_section<'data>(
                     })
                     .map_or_else(
                         || format!("GOT.data.internal.{i}"),
-                        |sym| format!("GOT.data.internal.{sym}"),
+                        |sym| format!("GOT.data.internal.{}", demangle_symbol_name(sym, demangle)),
```
davidlattimore, line 1712: How easy would it be to test this one too? Same question with the `undefined_data_errors` code below.

davidlattimore (commented):
I'd be slightly tempted to use an enum - e.g. SymbolDisplayMode instead of a bool, but I don't feel strongly

libwild/src/wasm.rs
lapla-cogito: Done.

davidlattimore (approved):

PR #2486, opened 2026-09-03
fix(tests): require GCC 14+ for linker plug-in

marxin:
Noticed while testing on my RISC-V board, where I didn't get recent enough version of the GCC compiler.

### DIFF
--- a/wild/tests/integration_tests.rs
+++ b/wild/tests/integration_tests.rs
@@ -7594,6 +7594,33 @@ fn verify_linker_plugin_requirements(
 
             verify_command_success(&mut command)
                 .context("Can't use compiler with linker-plugin")?;
+
+            // Our linker plugin implementation requires at least GCC 14 (where the LAPI_V1 version
+            // got added).
+            let output = Command::new(&linker_driver_string)
+                .arg("--version")
+                .output()
+                .context("Can't execute compiler with --version")?;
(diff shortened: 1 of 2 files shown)


wild/tests/integration_tests.rs
```diff
@@ -7634,6 +7594,33 @@ fn verify_linker_plugin_requirements(
...
+            let compiler_version = String::from_utf8_lossy(&output.stdout);
+            if compiler_version.contains("gcc") || compiler_version.contains("g++") {
+                let mut command = Command::new(&linker_driver_string);
+                let output = command
+                    .arg("-dumpfullversion")
+                    .output()
+                    .context("Can't execute compiler with -dumpversion")?;
+                let compiler_version = String::from_utf8_lossy(&output.stdout)
+                    .split_once(".")
+                    .context("expected comma in the compiler version")?
```
davidlattimore: Looks like a full stop, not a comma. Perhaps just skip this error return and chain into the next one. e.g.

```rust
                let compiler_version = String::from_utf8_lossy(&output.stdout)
                    .split_once(".")
                    .and_then(|(major, _)| major.parse::<u64>().ok())
                    .context("Can't parse compiler version")?;
```

davidlattimore (approved):

wild/tests/integration_tests.rs
marxin: Yep, done.

PR #2488, opened 2026-09-03
port(MachO): Fix TLV descriptor offset

joanna-jo:
Currently in the `ARM64_RELOC_UNSIGNED`/`RelocationKind::Absolute` case, the full address is being written into the TLV descriptor offset when it should actually be calculated by the `S + A - first TLV data section address` when handling this relocation for TLV descriptors. See [ld](https://github.com/apple-oss-distributions/ld64/blob/f60a74eaa2c99585de1dc0f2820e7a9f8aaf522c/src/ld/OutputFile.cpp#L540-L543) and [lld](https://github.com/llvm/llvm-project/blob/main/lld/MachO/InputSection.cpp#L262C1-L271C8).

Part of #2071

### DIFF
--- a/libwild/src/macho_writer.rs
+++ b/libwild/src/macho_writer.rs
@@ -109,6 +109,9 @@ use object::macho::N_ABS;
 use object::macho::N_SECT;
 use object::macho::PLATFORM_MACOS;
 use object::macho::RelocationInfo;
+use object::macho::S_THREAD_LOCAL_REGULAR;
+use object::macho::S_THREAD_LOCAL_VARIABLES;
+use object::macho::S_THREAD_LOCAL_ZEROFILL;
 use object::macho::SegmentFlags;
 use object::slice_from_bytes_mut;
 use object::write::macho::CodeDirectory;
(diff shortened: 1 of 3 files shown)


libwild/src/macho_writer.rs
```diff
@@ -681,6 +684,17 @@ fn apply_relocation<'data, A: Arch<Platform = MachO>>(
...
+                && flags.has_link_time_address()
+                && is_tlv_template_referent(layout, local_symbol_id) =>
+        {
+            // TODO: Once addends are supported, remember to change this to S + A -
+            // tlv_data_start_address().
+            resolution
+                .raw_value
+                .wrapping_sub(layout.tlv_data_start_address())
+        }
         RelocationKind::Absolute => resolution.raw_value.bitand(mask.symbol_plus_addend),
```
joanna-jo, line 698: Technically, there could be a local resolver in the TLV descriptor, although normal compiler outputs will emit `__tlv_bootstrap` and not enter this case since `!has_link_time_address()`. `ld` and `lld` diverge in behavior here where for a local resolver, `ld` will emit a fixup and `lld` treats any relocation from `S_THREAD_LOCAL_VARIABLES` as an offset. But the actual bytes emitted don't seem to matter since [any resolver slot will be overwritten by `dyld` with `_tlv_get_addr`](https://github.com/apple-oss-distributions/dyld/blob/fd8d0c4d52320ebf64db34f3cb280310d905c5ae/libdyld/ThreadLocalVariables.cpp#L154-L168).

davidlattimore (approved):

PR #2493, opened 2026-09-04
port(MachO): support `__init_offsets` and `__dso_handle` symbol

### LINKED ISSUE #2379: Mach-O: add synthetic `___dso_handle` symbol for C++ ABI
Seems the symbol can point to a beginning of the image and is essential for test-cases like:

```c++
#include <iostream>

class Tracer {
public:
    ~Tracer() {
        std::cout << "destroy\n";
    }
};

Tracer globalA;

Tracer& functionStatic() {
    static Tracer object{};
    return object;
}

int main() {

```

marxin:
The PR adds support for constructors in a form of function points in the `__mod_init_func` section. We could just concatenate the section to the output file, but as Mach-O is PIC by default, we would have to emit also chained fix-ups. Rather than that, we can use a more modern `__init_offsets` section, which contains relative (to __TEXT segment) 32-bit offsets and thus the fixup are unnecessary. On top of that, I'm also adding a synthetic symbol `___dso_handle` used as part of C++ ABI function `__cxa_atexit`.

Part of #757
Resolves: #2379

### DIFF
--- a/libwild/src/layout.rs
+++ b/libwild/src/layout.rs
@@ -4152,6 +4152,7 @@ impl<'data, P: Platform<GcUnit = SectionGcUnit>> ObjectLayoutState<'data, P> {
         let mut frame_section_indices = SmallVec::<[SectionIndex; 2]>::new();
         let mut note_gnu_property_section = None;
         let mut riscv_attributes_section = None;
+        let mut init_func_section = None;
 
         let no_gc = !resources.symbol_db.args.should_gc_sections();
 
@@ -4186,6 +4187,9 @@ impl<'data, P: Platform<GcUnit = SectionGcUnit>> ObjectLayoutState<'data, P> {
                 SectionSlot::RiscvVAttributes(index) => {
                     riscv_attributes_section = Some(*index);
                 }
+                SectionSlot::InitFunc(index) => {
+                    init_func_section = Some(*index);
+                }
                 _ => (),
             }
         }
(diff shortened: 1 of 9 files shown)


libwild/src/layout.rs
```diff
@@ -4186,6 +4187,9 @@ impl<'data, P: Platform<GcUnit = SectionGcUnit>> ObjectLayoutState<'data, P> {
                 SectionSlot::RiscvVAttributes(index) => {
                     riscv_attributes_section = Some(*index);
                 }
+                SectionSlot::InitFunc(index) => {
+                    init_func_section = Some(*index);
```
davidlattimore: Can there be more than one?

libwild/src/macho.rs
```diff
@@ -1427,6 +1462,45 @@ impl platform::Platform for MachO {
...
+                info.r_extern
+                    && !info.r_pcrel
+                    && info.r_length == 3
+                    && info.r_type == macho::ARM64_RELOC_UNSIGNED,
+                "unsupported Mach-O initializer relocation"
+            );
+            object.format_specific.init_functions.push(
+                object
+                    .symbol_id_range
+                    .input_to_id(SymbolIndex(info.r_symbolnum as usize)),
```
davidlattimore, line 1502: We're just using the local symbol index here. Is it possible for a symbol to be non-canonical - i.e. we need to use a definition from a different object file?

libwild/src/layout.rs
marxin: Very likely not, but let's make it a vector instead that we iterate.

libwild/src/macho.rs
marxin: That can be definitely possible, though I could not come up with a non-synthetic test-case. I added the `symbol_db.definition` into `create_finalise_sizes_ext`, hope it's a good spot for doing the canonical symbol resolution?

davidlattimore: For a test, how about something like this?

```c
__attribute__((used, section("__DATA,__mod_init_func,mod_init_funcs")))
static void (*const initializer)(void) = externally_defined_initializer;
```

davidlattimore (approved):

libwild/src/macho.rs
marxin: Good idea, but problematic for other reason - will file a new issue shortly.

PR #2498, opened 2026-09-05
fix: emit _GLOBAL_OFFSET_TABLE_ as local hidden symbol

deepakshirkem:
_GLOBAL_OFFSET_TABLE_ was being emitted as a global visible symbol. Per the ELF spec and consistent with lld, it should be local and hidden. Removes `global-offset-table-position-aarch64.s` and `aarch64-gnu-ifunc-nosym.s` from the LLD test skip list (5 → 7 passing aarch64 LLD tests).

Part of #2380


wild/tests/sources/elf/got-hidden-symbol/got-hidden-symbol.s
```diff
@@ -0,0 +1,28 @@
...
+//#Arch:aarch64
+//#Mode:dynamic
+//#ReferenceLinkers:lld
+//#LinkArgs:--hash-style=sysv -shared
+//#DiffIgnore:.dynamic.DT_FLAGS_1*
+//#DiffIgnore:section.got.plt.entsize
+//#RunEnabled:false
+//#ExpectSym:_GLOBAL_OFFSET_TABLE_ binding=local
+//#NoDynSym:_GLOBAL_OFFSET_TABLE_
+.globl f
```
davidlattimore: It looks like this test is a modified copy of one of the lld tests. We can't just copy code without proper attribution. How about writing the test in C? I don't think it needs to do any special, just make sure there's at least one GOT entry.

davidlattimore (approved):

PR #2502, opened 2026-09-07
test(lld): categorize aarch64 skip list entries

deepakshirkem:
Moves 113 tests out of aarch64_untriaged into specific groups with documented reasons. No tests added or removed purely organizational.

Part of #2380


wild/tests/external_tests/lld_skip_tests.toml
```diff
@@ -96,141 +194,49 @@ tests = [
...
-  "aarch64-prel32.s",
+  "aarch64-thunk-reuse.s",
+  "aarch64-thunk-script.s",
+  "aarch64-thunk-sort.s",
+  "aarch64-variant_pcs.s",
+  "eh-frame-nonzero-offset-aarch64.s",
+]
+
+[skipped_groups.range_thunks_not_implemented]
+reason = "Range extension thunks not implemented"
```
davidlattimore: Range extension thunks are implemented. It's possible however that the tests are synthetically causing functions to be far apart in a way that our implementation doesn't support.

deepakshirkem: moved those test. Thank you

davidlattimore (approved):

PR #2503, opened 2026-09-07
fix: Define hidden `__dso_handle` linker symbol

lapla-cogito:
Noticed while building llvm-min-tblgen (as part of the TensorFlow build).

### DIFF
--- a/wild/tests/sources/elf/dso-handle/dso-handle.c
+++ b/wild/tests/sources/elf/dso-handle/dso-handle.c
@@ -2,7 +2,6 @@
 //#Object:runtime.c
 //#Object:ptr_black_box.c
 //#ReferenceLinkers:lld
-//#DiffMatchAny:true
 //#ExpectSym:__dso_handle
 //#NoDynSym:__dso_handle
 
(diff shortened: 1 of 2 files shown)


wild/tests/sources/elf/dso-handle/dso-handle.c
```diff
@@ -0,0 +1,39 @@
+//#AbstractConfig:default
+//#Object:runtime.c
+//#Object:ptr_black_box.c
+//#ReferenceLinkers:lld
```
lapla-cogito, line 4: GNU ld doesn't synthesize this symbol.

wild/tests/sources/elf/dso-handle/dso-handle.c
```diff
@@ -0,0 +1,39 @@
...
+//#DiffMatchAny:true
```
davidlattimore: DiffMatchAny will only have an effect if there's more than one reference linker

davidlattimore (approved):

wild/tests/sources/elf/dso-handle/dso-handle.c
lapla-cogito: Forgot to remove it. Thanks.

PR #2504, opened 2026-09-07
port(MachO): relax GOT-relative relocations

marxin:
FYI: @joanna-jo

Part of #757

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,7 +168,9 @@ jobs:
     env:
       WILD_TEST_IGNORE_FORMAT: "1" # Checked by other jobs
     steps:
-      - run: brew install lld
+      - run: brew install lld llvm
+      - run: echo "PATH=/opt/homebrew/opt/llvm/bin:$PATH" >> $GITHUB_ENV
+      - run: clang --version
       - uses: actions/checkout@v7
         with:
           persist-credentials: false
--- a/libwild/src/macho_aarch64.rs
+++ b/libwild/src/macho_aarch64.rs
@@ -242,6 +242,26 @@ impl crate::platform::Arch for MachOAArch64 {
                     rel_info: relocation,
                 })
             }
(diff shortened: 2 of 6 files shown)


joanna-jo (commented):
Thanks for the FYI! Was that related to my question on #2493? If so, I was asking a slightly different one from @davidlattimore - in the test case, if `externally_defined_initializer` comes from a dylib and not an object file, then do we need to emit a stub and stores its offset in `__init_offsets`? I looked into this briefly just now and it seems like [lld behavior suggests that we do](https://github.com/llvm/llvm-project/blob/9fafedaa908f1844f4150ae2ba6e0eec938341c4/lld/MachO/SyntheticSections.cpp#L2052-L2077).

libwild/src/macho_aarch64.rs
```diff
@@ -242,6 +242,26 @@ impl crate::platform::Arch for MachOAArch64 {
                     rel_info: relocation,
                 })
             }
+            object::macho::ARM64_RELOC_GOT_LOAD_PAGE21
+                if flags.has_link_time_address() && !interposable =>
```
davidlattimore, line 246: Note that `is_symbol_non_interposable` currently returns just true on mac. Does that matter for the purposes of this PR?

davidlattimore (commented):
I'll wait until the failing test passes to take a proper look

.github/workflows/ci.yml
```diff
@@ -168,7 +168,9 @@ jobs:
     env:
       WILD_TEST_IGNORE_FORMAT: "1" # Checked by other jobs
     steps:
-      - run: brew install lld
+      - run: brew install lld llvm
```
marxin, line 171: The previous CI error was related to #2514 - I am sticking to the latest clang for now.

libwild/src/macho_aarch64.rs
marxin: Decided to create a dedicated issue #2515 where we can approach it more systematically.

davidlattimore (approved):

PR #2505, opened 2026-09-07
port(wasm): Support --import-memory

### LINKED ISSUE #2337: Wasm: support `--import-memory` option
```
clang --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot m.c -Wl,--import-memory
...
wasm-tools print ...
(module $a.out
...
  (import "env" "memory" (memory (;0;) 2 2 shared))
```

joissriram99:
Adds `--import-memory`: As mentioned in the original issue, the output module imports its
linear memory from `env.memory` instead of defining one.

Closes https://github.com/wild-linker/wild/issues/2337
Part of https://github.com/wild-linker/wild/issues/1431

### DIFF
--- a/libwild/src/args/wasm.rs
+++ b/libwild/src/args/wasm.rs
@@ -30,6 +30,12 @@ pub(crate) const DEFAULT_ENTRY: &str = "_start";
 /// Default export name for the module's linear memory.
 pub(crate) const DEFAULT_MEMORY_EXPORT_NAME: &str = "memory";
 
+/// Default module name used when importing the linear memory (`--import-memory`).
+pub(crate) const DEFAULT_MEMORY_IMPORT_MODULE: &str = "env";
+
+/// Default field name used when importing the linear memory (`--import-memory`).
+pub(crate) const DEFAULT_MEMORY_IMPORT_NAME: &str = "memory";
+
 #[derive(Debug)]
 pub struct WasmArgs {
     pub(crate) common: super::CommonArgs,
@@ -50,6 +56,7 @@ pub struct WasmArgs {
     // Emit a shared linear memory (`memory.shared`). Requires the `atomics` and `bulk-memory`
     // target features.
     pub(crate) shared_memory: bool,
+    pub(crate) import_memory: bool,
     pub(crate) gc_sections: bool,
     pub(crate) allow_undefined: bool,
     pub(crate) allow_multiple_definition: bool,
@@ -84,6 +91,7 @@ impl Default for WasmArgs {
             initial_memory: None,
             max_memory: None,
             shared_memory: false,
+            import_memory: false,
             export_memory: None,
             gc_sections: true,
             allow_undefined: false,
@@ -419,6 +427,15 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
         .help("Use shared linear memory")
         .execute(|args, _modifier_stack| {
             args.shared_memory = true;
+            Ok(())
+        });
+
+    parser
+        .declare()
+        .long("import-memory")
+        .help("Import the module's memory from the default module of env with the name memory")
+        .execute(|args, _modifier_stack| {
+            args.import_memory = true;
             Ok(())
         });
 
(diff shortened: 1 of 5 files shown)


libwild/src/args/wasm.rs
```diff
@@ -422,6 +430,15 @@ fn setup_argument_parser() -> ArgumentParser<WasmArgs> {
             Ok(())
         });
 
+    parser
+        .declare()
```
lapla-cogito, line 434: wasm-ld also accepts the `--import-memory=module,name` format.

libwild/src/wasm.rs
```diff
@@ -6060,6 +6077,11 @@ where
         } else {
             Some(heap_end_from_initial_pages(initial_pages)?)
         };
+        if import_memory {
+            // The output memory is imported rather than defined using --import-memory,
+            // so remove the memory section and add an import instead.
```
lapla-cogito, line 6082: I think this comment is redundant because it says almost the same thing in the `import_output_memory` doc comment.

libwild/src/wasm.rs
```diff
@@ -6042,7 +6059,7 @@ where
                 .memories
                 .push(linker_output_memory_type(&layout_inputs, shared_memory));
         }
-        if !layout.memories.is_empty() {
+        if !layout.memories.is_empty() && (!import_memory || export_memory.is_some()) {
```
lapla-cogito, line 6062: When `--import-memory` is used without `--export-memory`, this skips `ensure_memory_export` entirely. That function is also what strips input memory exports.

wild/tests/integration_tests.rs
```diff
@@ -836,6 +852,66 @@ impl WasmModuleInfo {
...
+        let Some(exp) = expected else {
+            return Ok(());
+        };
+        let matching = self
+            .memory_imports
+            .iter()
+            .filter(|(module, name, _)| module == &exp.module && name == &exp.name)
+            .collect::<Vec<_>>();
+        ensure!(
+            matching.len() == 1,
```
marxin, line 869: Use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one.

libwild/src/args/wasm.rs
joissriram99: i've made the changes, thank you 🙏

libwild/src/wasm.rs
joissriram99: Removed, thanks 🙏

libwild/src/wasm.rs
joissriram99: i've now split this into strip_memory_exports and push_memory_export, thanks 🙏

wild/tests/integration_tests.rs
joissriram99: add the changes, thank you 🙏

wild/tests/sources/wasm/import-memory/import-memory.c
```diff
@@ -0,0 +1,56 @@
...
+//#LinkArgs: --import-memory
+//#NoSection: Memory
+//#ExpectSection: Import
+//#ExpectMemoryImport: env/memory
+//#NoSym: memory
+//#RunEnabled: false
+
+//#Config:shared
+//#CompArgs: -matomics
+//#LinkArgs: --import-memory --shared-memory --initial-memory=131072 --max-memory=196608
```
davidlattimore, line 11: I don't think this test is sufficient to be sure that --import-memory and --shared-memory are actually working correctly together. For that, I think you'd need to load some modules and check that the memory is being shared as expected. This would require an extension to the test framework to be able to preload those modules, then this test would need to not set RunEnabled:false and would need to check the values. One option if you want to skip this for now would be to disallow --import-memory and --shared-memory together and issue an error saying that it's not yet supported. That would allow proper support and testing to be done in a later PR.

davidlattimore (commented):
If current lld releases don't support the feature as you've implemented it, then it'd be fine to disable the reference linker for that test with a comment as to why and a TODO to enable it once lld > some version is released.

wild/tests/sources/wasm/import-memory/import-memory.c
joissriram99: created - https://github.com/wild-linker/wild/issues/2540

davidlattimore (approved):

lapla-cogito (approved):
Thanks!

PR #2509, opened 2026-09-08
perf: enable mmap on Btrfs with sufficiently new kernel

### LINKED ISSUE #1499: BTRFS write performance
We've previously observed bad write performance when the linker's output file is on a BTRFS filesystem.

There are likely limits to how fast we can get with btrfs, but we can probably make things faster than they are at the moment. [Benchmarks](https://gist.github.com/mati865/7817cc637f15435f536b81f05575bb21) that @mati865 has run previously have shown a significant slowdown as we add more threads, especially when the parition is small. So, worst-case-scenario, we could detect that the output file is on btrfs and limit the number of threads used during the write phase.

There was some [discussion](https://www.reddit.com/r/rust/comments/1qe85w1/comment/nzvomyj/) about this on reddit where The_8472 said:

> I was asking about one vs. multiple files because I was wondering whether writing the extents to separate files and then using copy_file_range (extent cloning) to merge them into one file would speed things up, in case it was some per-file-lock stuff.
> 
> And check with filefrag -v <filename> if it does anything silly like creating lots of tiny extents due to random access. Maybe preallocate one large extent before mmaping.

mati865:
Together with huge pages enabled by default fixes https://github.com/wild-linker/wild/issues/1499 for real.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ memchr = { version = "2.7.0", default-features = false }
 memmap2 = { version = "0.9.0", default-features = false }
 mimalloc = { version = "0.1", default-features = false }
 nix = { version = "0.31.1", default-features = false, features = [
+  "feature",
   "resource",
   "fs",
 ] }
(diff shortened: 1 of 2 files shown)


davidlattimore (approved):
Thanks for doing this and all the benchmarks you ran!

PR #2511, opened 2026-09-09
refactor: move section-header table to end of file, outside PT_LOAD

Blazearth:
This PR is pre-requisite to #2472 
Wild models the ELF section-header table as an internal output section named `SECTION_HEADERS`. Previously, it was marked `SHF_ALLOC` and placed inside the first `PT_LOAD`, so changing the number of sections could change the loadable layout and shift virtual addresses.

This PR moves `SECTION_HEADERS` to the end of the file, removes `SHF_ALLOC`, and takes `e_shoff` from its actual laid-out file offset. This matches the layout used by other linkers.

This is a prerequisite for `--only-keep-debug`, since `--strip-debug` and `--only-keep-debug` need to preserve the same address layout.

### DIFF
--- a/libwild/src/elf.rs
+++ b/libwild/src/elf.rs
@@ -2507,7 +2507,6 @@ impl<C: ElfClass> platform::Platform for Elf<C> {
 
         builder.add_section(crate::output_section_id::FILE_HEADER);
         builder.add_section(output_section_id::PROGRAM_HEADERS);
-        builder.add_section(output_section_id::SECTION_HEADERS);
         builder.add_section(output_section_id::NOTE_GNU_PROPERTY);
         builder.add_section(output_section_id::NOTE_GNU_BUILD_ID);
         builder.add_section(output_section_id::INTERP);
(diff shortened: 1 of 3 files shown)


davidlattimore (commented):
It looks as though with this change, we're no longer correctly aligning the section headers. It seems that we weren't explicitly aligning them before, but they ended up correctly aligned by virtue of where we were putting them. The following patch adds a test that passed before, but fails with this change:

```patch
diff --git a/wild/tests/sources/elf/section-header-alignment/section-header-alignment.c b/wild/tests/sources/elf/section-header-alignment/section-header-alignment.c
new file mode 100644
index 0000000000..e6a3f981a5
--- /dev/null
+++ b/wild/tests/sources/elf/section-header-alignment/section-header-alignment.c
@@ -0,0 +1,18 @@
+//#ReferenceLinkers:bfd,lld
+//#Object:runtime.c
+
+#include <elf.h>
+
+
+extern const Elf64_Ehdr __ehdr_start;
+
+void _start(void) {
+  runtime_init();
+
+  if (__ehdr_start.e_shoff % _Alignof(Elf64_Shdr) != 0) {
+    exit_syscall(1);
+  }
+
+  exit_syscall(42);
+}
```

davidlattimore (approved):

PR #2513, opened 2026-09-09
feat(args): implement --image-base flag

deepakshirkem:
Adds --image-base=<address> to set the base load address of the output binary. Wild previously always used the platform default (0x400000 for x86-64).

Updates the lld skip list to move tests that were in missing_flag_support due to --image-base into their correct failure categories now that the flag is implemented.'

Part of #2380

### DIFF
--- a/libwild/src/lib.rs
+++ b/libwild/src/lib.rs
@@ -336,8 +336,17 @@ impl<F: FileSystem> Linker<F> {
 
         let mut output = file_writer::Output::new::<P>(args, output_kind, self.file_system.clone());
 
-        let mut output_sections =
-            OutputSections::with_base_address(A::start_memory_address(output_kind), output_kind);
+        let mut output_sections = OutputSections::with_base_address(
+            args.image_base()
+                .unwrap_or_else(|| A::start_memory_address(output_kind)),
+            output_kind,
+        );
+        if let Some(base) = args.image_base() {
+            let page_size = args.loadable_segment_alignment().value();
+            if base % page_size != 0 {
+                bail!("--image-base: address isn't multiple of page size: {base:#x}");
+            }
+        }
         output_sections.set_rosegment(args.rosegment());
 
         let mut layout_rules_builder = LayoutRulesBuilder::default();
--- a/wild/tests/sources/elf/image-base/image-base.c
+++ b/wild/tests/sources/elf/image-base/image-base.c
@@ -1,0 +1,19 @@
+// Verify that --image-base sets the load address of the output binary.
+// Also verifies that an error is emitted when the address is not page-aligned.
+//#AbstractConfig:default
+//#Mode:static
+//#RunEnabled:false
+//#Object:runtime.c
+
+//#Config:aligned:default
+//#ReferenceLinkers:bfd,lld
+//#LinkArgs:--image-base=0x10000000 --undefined=__executable_start
+//#ExpectSym:__executable_start address=0x10000000
+
+//#Config:unaligned:default
+//#ReferenceLinkers:
+//#LinkArgs:--image-base=0x10000001
+//#ExpectError:--image-base: address isn't multiple of page size: 0x10000001
+
+#include "../common/runtime.h"
+void _start(void) { exit_syscall(42); }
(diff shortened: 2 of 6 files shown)


wild/tests/sources/elf/image-base/image-base.s
```diff
@@ -0,0 +1,18 @@
+// Verify that --image-base sets the load address of the output binary.
+// We verify __executable_start is at exactly the image base address.
+//#AbstractConfig:default
+//#Mode:static
+//#ReferenceLinkers:
+//#LinkArgs:--image-base=0x10000000
+//#RunEnabled:false
+//#ExpectSym:__executable_start address=0x10000000
+
+//#Config:x86_64:default
```
davidlattimore: Having separate configurations for different architectures is only needed if you have different options. So it should be fine to just fold these two configurations into the main configuration above.

wild/tests/sources/elf/image-base/image-base.c
davidlattimore, line 1: What do bfd, lld and wild do if the image base isn't a multiple of the value supplied to `-z max-page-size=0x2000`? Can you test and make sure that we have similar behaviour to one if not both reference linkers?

wild/tests/sources/elf/image-base/image-base.c
```diff
@@ -0,0 +1,18 @@
...
+//#RunEnabled:false
+//#Object:runtime.c
+
+//#Config:aligned:default
+//#LinkArgs:--image-base=0x10000000
+//#ExpectSym:__executable_start address=0x10000000
+
+//#Config:unaligned:default
+//#LinkArgs:--image-base=0x10000001
+//#ExpectWarningWild:--image-base: address isn't multiple of page size: 0x10000001
```
davidlattimore: With a warning, we'll still emit the binary. Can you check that the address of `__executable_start` is the same for us and the reference linker(s)?

libwild/src/lib.rs
```diff
@@ -336,8 +336,18 @@ impl<F: FileSystem> Linker<F> {
...
+        let mut output_sections = OutputSections::with_base_address(
+            args.image_base()
+                .unwrap_or_else(|| A::start_memory_address(output_kind)),
+            output_kind,
+        );
+        // Warn if image_base is not a multiple of the page size, matching lld behavior.
+        if let Some(base) = args.image_base() {
+            let page_size = args.loadable_segment_alignment().value();
+            if base % page_size != 0 {
+                eprintln!("warning: --image-base: address isn't multiple of page size: {base:#x}");
```
davidlattimore: Warnings should be emitted with `args.warning`

deepakshirkem: Done.

wild/tests/sources/elf/image-base/image-base.c
deepakshirkem: Wild rounds up the unaligned address to the next page boundary (0x10001000 on x86-64, 0x10010000 on aarch64), while lld uses the address as-is (0x10000001). So the addresses differ. Should I fix Wild to match lld's behavior (use as-is) or is this acceptable as a design difference?

davidlattimore: I'd say we've got two options (1) use the address that the user specified, but warn that it doesn't meet the alignment requirements - i.e. what the other linkers do or (2) issue an error - i.e. be stricter than the other linkers.

I'd be fine with either of those options, but I don't think we should do what we're doing now which is to issue a warning then use a different address.

deepakshirkem: @davidlattimore  Going with option 2 (error).

wild/tests/sources/elf/image-base/image-base.c
```diff
@@ -0,0 +1,18 @@
+// Verify that --image-base sets the load address of the output binary.
+// Also verifies that an error is emitted when the address is not page-aligned.
+//#AbstractConfig:default
+//#Mode:static
+//#ReferenceLinkers:
```
davidlattimore: Can you enabled bfd and lld as reference linkers? You'll probably need to disable them for the unaligned test. You might also need to pass `--undefined=__executable_start` as a linker argument, otherwise the other linkers won't emit that symbol.

davidlattimore (approved):

PR #2518, opened 2026-09-10
perf: Compute only referenced symbol addresses on later RISC-V relaxation passes

lapla-cogito:
The first RISC-V relaxation pass builds section positions and every canonical symbol address. Later passes already rescan only sections that might still shrink, but they still walked every symbol to refill the address table.

Collect targets still referenced by those sections and compute addresses for just those symbols. Section positions are still recomputed each pass.


davidlattimore (approved):
Another nice speedup :)

Issue #2512, opened 2026-09-09
MachO: emit chained fixups for non-GOT sections as well


marxin:
Right now, we limit us to emit chained fixups only for synthetized GOT entries (part of `__DATA_CONST` segment). However, there are also input sections that need a similar type of fixups:

dyncast.cc:
```c++
struct Base {
    virtual ~Base() = default; // must be polymorphic
};

struct Derived : Base {
    void hello() {
        __builtin_printf("Derived\n");
    }
};

int main() {
    Base* base = new Derived;

    if (auto* derived = dynamic_cast<Derived*>(base)) {
        derived->hello();
    } else {
        __builtin_printf("not a Derived\n");
    }

    delete base;
}
```

```
$ clang++ dyncast.cc
$ llvm-objdump --chained-fixups --macho a.out
...
dyld chained import[5]
  lib_ordinal = 1 (libc++)
  weak_import = 0
  name_offset = 62 (__ZTVN10__cxxabiv117__class_type_infoE)
dyld chained import[6]
  lib_ordinal = 1 (libc++)
  weak_import = 0
  name_offset = 101 (__ZTVN10__cxxabiv120__si_class_type_infoE)
```

We're missing the fixups related to `__DATA,__const` section.

PR #2521, opened 2026-09-11
port(MachO): Emit bind fixups for non-GOT entries

joanna-jo:
Previously, we only emitted chained fixups for GOT entries and assumed that every import had a corresponding GOT entry. We need to emit chained fixups for non-GOT entries as well, including both bind and rebase fixups, although the latter will be covered in a future diff.

For 64-bit absolute relocations that require a strong bind fixup, we store `pending_fixups` in `CommonGroupStateExt` during relocation processing. This state is propagated through to `FinaliseSizesExt` and then consumed by `LayoutExt`, which holds the final state of all sorted fixups (including GOT entry fixups) with their import ordinal and fixup address. This information is ultimately used in the writing phase to encode all binded fixups and their chained fixup table metadata.

Once multiple pages are supported, we'll need to revisit some of the implementation here, which I've documented with TODOs. cc: @marxin 

Part of #2512

### DIFF
--- a/libwild/src/macho_writer.rs
+++ b/libwild/src/macho_writer.rs
@@ -37,7 +37,6 @@ use crate::macho::DylibCommand;
 use crate::macho::DylinkerCommand;
 use crate::macho::EntryPointCommand;
 use crate::macho::FileHeader;
-use crate::macho::GOT_ENTRY_SIZE;
 use crate::macho::MACHO_COMMAND_ALIGNMENT;
 use crate::macho::MACHO_START_MEM_ADDRESS;
 use crate::macho::MAX_SEGMENT_COUNT;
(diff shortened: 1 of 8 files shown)


libwild/src/macho_writer.rs
```diff
@@ -437,16 +404,69 @@ fn write_plt_entries<A: Arch<Platform = MachO>>(
...
+            let page_index = offset_in_segment / MACHO_PAGE_ALIGNMENT.value();
+
+            let next_fixup = fixups
+                .peek()
+                .filter(|next| segment_addresses.contains(&next.fixup_address));
+            let next_offset_in_segment =
+                next_fixup.map(|fixup| fixup.fixup_address - segment.sizes.mem_offset);
+            let next = match next_offset_in_segment {
+                Some(next_offset) if next_offset / MACHO_PAGE_ALIGNMENT.value() == page_index => {
+                    (next_offset - offset_in_segment) / 4
```
davidlattimore: Is it OK if the difference here isn't exactly divisible by 4? If it might matter and we don't want do deal with that now, you could just use `ensure!` to report an error if it isn't divisible by 4.

joanna-jo: Thanks, added that in! Normally this shouldn't happen for layouts since pointers would be 8-byte aligned, but technically it looks like it's still possible for more custom inputs (e.g. manual assembly) or packed structs.

davidlattimore (approved):

PR #2524, opened 2026-09-12
port(wasm): Support single-threaded TLS

### LINKED ISSUE #2474: Wasm: unsupported Wasm relocation type MemoryAddrTlsSleb
main.c:

```c
int foo() { return 0; }

int main() {
  __builtin_printf("hello world\n");
  return foo() + 42;
}
```

```
❯ clang --target=wasm32-wasip1-threads --sysroot=/usr/share/wasi-sysroot m.c -fuse-ld=/home/marxin/Programming/wild/fakes-debug/wasm-ld
wild: error: unsupported Wasm relocation type MemoryAddrTlsSleb
```

```
llvm-readobj -r /usr/share/wasi-sysroot/lib/wasm32-wasip1-threads/libc.a
...
File: /usr/share/wasi-sysroot/lib/wasm32-wasip1-threads/libc.a(fdopendir.c.obj)
Format: WASM
Arch: wasm32
AddressSize: 32bit
Relocations [
  Section (4) CODE {
...
    0x70 R_WASM_MEMORY_ADDR_TLS_SLEB errno 0
  }
```

lapla-cogito:
Lay out TLS segments as a contiguous block after non-TLS data and define immutable `__tls_base` as that address (or `data_start` if there is no TLS).

`R_WASM_MEMORY_ADDR_TLS_SLEB` is resolved as an offset from `__tls_base`. Ordinary `MEMORY_ADDR` relocations against TLS symbols stay absolute, matching wasm-ld without `--shared-memory`. Weak or GC'd TLSREL is 0, including any addend. `--shared-memory` with live TLS is rejected until shared-memory TLS is implemented.

Closes #2474 
Part of #1431

### DIFF
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2273,12 +2273,8 @@ fn data_segment_alignment(input: &WasmObjectLayoutInput<'_>, original_index: u32
 }
 
 fn data_segment_is_tls(input: &WasmObjectLayoutInput<'_>, original_index: u32) -> bool {
-    if input
-        .segment_infos
-        .get(original_index as usize)
-        .is_some_and(|info| info.is_tls())
-    {
-        return true;
+    if let Some(info) = input.segment_infos.get(original_index as usize) {
+        return info.is_tls();
     }
     input.symbols.iter().any(|sym| {
         sym.kind == WasmSymbolKind::Data
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -2317,7 +2317,12 @@ fn tls_reloc_value(abs_addr: u32, tls_base: u32, addend: i64) -> Result<u32> {
     let offset = abs_addr.checked_sub(tls_base).ok_or_else(|| {
         crate::error!("TLS relocation address 0x{abs_addr:x} is before TLS base 0x{tls_base:x}")
     })?;
-    reloc_value_with_addend(offset, addend)
+    let value = i64::from(offset)
+        .checked_add(addend)
+        .ok_or_else(|| crate::error!("Wasm TLS relocation value overflow"))?;
+    let value = i32::try_from(value)
+        .map_err(|_| crate::error!("Wasm TLS relocation value out of range"))?;
+    Ok(value as u32)
 }
 
 fn layout_object_data<'data>(
--- a/libwild/src/wasm.rs
+++ b/libwild/src/wasm.rs
@@ -205,7 +205,7 @@ const UNREACHABLE_FUNCTION_BODY: &[u8] = &[0x00, 0x00, 0x0b];
 /// `i32.const` body for `LINKER_MEMORY_BASE`.
 const LINKER_MEMORY_BASE_INIT_EXPR: &[u8] = &[0x41, 0x80, 0x08];
 
-/// `i32.const 0`. Used for immutable `__tls_base` when no TLS segment is laid out.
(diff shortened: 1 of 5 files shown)


libwild/src/wasm.rs
```diff
@@ -2256,30 +2257,90 @@ fn ensure_stack_size_aligned(stack_size: u32) -> Result {
...
+/// weak-undefined or its segment was GC'd. So do not use the local symbol's `UNDEFINED` flag, which
+/// is also set on cross-object references to a defined TLS symbol.
+fn tls_reloc_value(abs_addr: u32, tls_base: u32, addend: i64) -> Result<u32> {
+    if abs_addr == 0 {
+        return Ok(0);
+    }
+    let offset = abs_addr.checked_sub(tls_base).ok_or_else(|| {
+        crate::error!("TLS relocation address 0x{abs_addr:x} is before TLS base 0x{tls_base:x}")
+    })?;
+    reloc_value_with_addend(offset, addend)
```
davidlattimore: reloc_value_with_addend fails if the value is negative. Does that matter?

libwild/src/wasm.rs
```diff
@@ -2256,30 +2257,90 @@ fn ensure_stack_size_aligned(stack_size: u32) -> Result {
...
+        .any(|i| data_segment_is_tls(input, data_segment_original_index(input, i)))
+}
+
+fn max_tls_alignment(inputs: &[WasmObjectLayoutInput<'_>]) -> Alignment {
+    inputs
+        .iter()
+        .flat_map(|input| {
+            (0..input.data_segments.len()).filter_map(|i| {
+                let original_index = data_segment_original_index(input, i);
+                data_segment_is_tls(input, original_index)
```
davidlattimore, line 2298: It looks like each data segment is going to do a scan over all the object's symbols. Is that likely to be a problem for performance?

libwild/src/wasm.rs
lapla-cogito: Indeed, that's a problem. Fixed, thanks.

libwild/src/wasm.rs
lapla-cogito: In normal cases, I think this doesn't seem like a significant issue, but in any event, I've changed to only scan the symbol table when metadata is absent.

libwild/src/wasm.rs
```diff
@@ -2256,30 +2257,91 @@ fn ensure_stack_size_aligned(stack_size: u32) -> Result {
...
+        })
+        .max()
+        .unwrap_or(crate::alignment::MIN)
+}
+
+/// `R_WASM_MEMORY_ADDR_TLS_*` is an offset from `__tls_base`. `abs_addr == 0` means the symbol is
+/// weak-undefined or its segment was GC'd. So do not use the local symbol's `UNDEFINED` flag, which
+/// is also set on cross-object references to a defined TLS symbol.
+fn tls_reloc_value(abs_addr: u32, tls_base: u32, addend: i64) -> Result<u32> {
+    if abs_addr == 0 {
```
davidlattimore: Could this get a false positive if `--stack-first -z stack-size=0` was set, causing us to skip processing of the addend? Not sure how useful `--stack-first -z stack-size=0` would be though

lapla-cogito: Changed `data_addresses` to be represented as `Vec<Option<u32>>` instead of `Vec<u32>` to distinguish TLS at address 0 and unresolved value.

davidlattimore (approved):

PR #2527, opened 2026-09-12
port(MachO): Set `MH_HAS_TLV_DESCRIPTORS` in header flags

joanna-jo:
If we emit a `S_THREAD_LOCAL_VARIABLES` section, we need to set `MH_HAS_TLV_DESCRIPTORS` in the header flags so that dyld initializes the TLV descriptors. With this and #2521, since we can now emit a bind fixup for the TLV descriptor thunk to `__tlv_bootstrap`, we're now able to link a TLV relocation in an executable for non-weak and non-interposable symbols! cc: @marxin @davidlattimore 

Part of #2071

### DIFF
--- a/wild/tests/sources/macho/trivial-tls/trivial-tls.c
+++ b/wild/tests/sources/macho/trivial-tls/trivial-tls.c
@@ -1,8 +1,5 @@
-// TODO: We don't emit fixups for `__tlv_bootstrap` yet, so we need to disable `RunEnabled` and
-// `DiffEnabled` for now.
-//#RunEnabled:false
-//#DiffEnabled:false
 //#Object:runtime.c
+//#DiffIgnore:section.__unwind_info
 //#ExpectSection:__thread_data
 //#ExpectSection:__thread_bss
 //#ExpectSection:__thread_vars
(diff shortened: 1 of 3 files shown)


wild/tests/sources/macho/trivial-tls/trivial-tls.c
```diff
@@ -1,8 +1,7 @@
-// TODO: We don't emit fixups for `__tlv_bootstrap` yet, so we need to disable `RunEnabled` and
-// `DiffEnabled` for now.
-//#RunEnabled:false
-//#DiffEnabled:false
+//#RunEnabled:true
+//#DiffEnabled:true
```
davidlattimore: These both default to true, so don't need to be explicitly set.

davidlattimore (approved):
Looks good to me. Just one tiny observation

wild/tests/sources/macho/trivial-tls/trivial-tls.c
joanna-jo: Oops thanks for catching that!

davidlattimore (approved):

PR #2529, opened 2026-09-13
feat(args): implement -z force-ibt and -z cet-report flags

deepakshirkem:
Adds CET property reporting flags for x86-64:

- -z force-ibt: warns per input file missing GNU_PROPERTY_X86_FEATURE_1_IBT
- -z cet-report=warning: warns for files missing IBT or SHSTK properties
- -z cet-report=error: errors for files missing IBT or SHSTK properties
- -z cet-report=<invalid>: errors on unknown values

Part of #2380


libwild/src/args/elf.rs
```diff
@@ -199,6 +201,12 @@ pub(crate) enum CompressionKind {
     Zstd,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum CetReport {
```
davidlattimore, line 206: According to https://sourceware.org/binutils/docs/ld/Options.html, "none" is also a valid option

libwild/src/args/elf.rs
```diff
@@ -799,6 +809,26 @@ fn setup_argument_parser() -> ArgumentParser<ElfArgs> {
...
+                args.force_ibt = true;
+                Ok(())
+            },
+        )
+        .sub_option_with_value(
+            "cet-report=",
+            "Report missing CET properties",
+            |args, _, value| {
+                args.cet_report = Some(match value {
+                    "warning" => CetReport::Warning,
```
davidlattimore: Could consider using strum for converting strings to enum values as some of the other enums in this file do

wild/tests/sources/elf/cet-report/cet-report.s
```diff
@@ -0,0 +1,28 @@
...
+# CET (Control-flow Enforcement Technology) is x86-64 specific.
+//#AbstractConfig:default
+//#Arch:x86_64
+//#Mode:static
+//#ReferenceLinkers:
+//#RunEnabled:false
+
+//#Config:force-ibt:default
+//#WildExtraLinkArgs:-z force-ibt
+//#ExpectWarningWild:.*: -z force-ibt: file does not have GNU_PROPERTY_X86_FEATURE_1_IBT property
```
davidlattimore, line 12: What are the expectations around whether this property is set on the output file? Can we have the test check whether it's set?

libwild/src/args/elf.rs
deepakshirkem: Done. It's now the default behavior (no reporting).

libwild/src/args/elf.rs
deepakshirkem: Updated to use strum::EnumString

libwild/src/args/elf.rs
```diff
@@ -125,6 +125,8 @@ pub struct ElfArgs {
     pub(crate) allow_multiple_definitions: bool,
     pub(crate) z_interpose: bool,
     pub(crate) z_isa: Option<NonZeroU32>,
+    pub(crate) force_ibt: bool,
+    pub(crate) cet_report: Option<CetReport>,
```
davidlattimore: What's the difference between `Some(CetReport::None)` and `None`?

libwild/src/elf.rs
```diff
@@ -4384,6 +4385,17 @@ impl LayoutExt {
     ) -> Result<Self> {
         let states = objects_iter(groups).map(|o| &o.format_specific);
         let gnu_property_notes = merge_gnu_property_notes::<C, A>(states.clone(), args.z_isa);
+        if args.force_ibt || args.cet_report.is_some() {
+            let cet_data: Vec<(String, Vec<GnuProperty>)> = objects_iter(groups)
```
davidlattimore: It doesn't look like it should be necessary to clone all this stuff just in order to pass it to `check_cet_properties` then throw it away.

wild/tests/sources/elf/cet-report/cet-report.s
davidlattimore: It doesn't look like this test has changed - we're still not checking that the properties are set correctly

wild/tests/sources/elf/cet-with-ibt/cet-with-ibt.s
```diff
@@ -0,0 +1,24 @@
+# Verify that IBT+SHSTK properties are preserved in the output binary
+# when the input object has them set.
```
davidlattimore, line 2: It looks like the test just checks that there's a `.note.gnu.property` section. It doesn't seem to be checking that the properties in it are as expected.

libwild/src/args/elf.rs
deepakshirkem: Yes, it's redundant. Will fix.

lapla-cogito (commented):
How about moving `z-cet-report.sh` in `mold_skip_tests.toml` to the ignore group because the message format differs?

deepakshirkem:
>How about moving z-cet-report.sh in mold_skip_tests.toml to the ignore group because the message format differs?

Done. Thank You.

wild/tests/sources/elf/cet-report/cet-report.s
davidlattimore: I think that we still need some sort of assertion that the properties emitted when `-z force-ibt` is applied are correct. If as your comment above suggests, there are no property notes emitted, then we should assert that and make sure that this is what other linkers do.

wild/tests/sources/elf/cet-report/cet-report.s
```diff
@@ -0,0 +1,32 @@
+# Test -z force-ibt and -z cet-report flags for CET property reporting.
+# CET (Control-flow Enforcement Technology) is x86-64 specific.
+# Input object has no GNU property notes, so output has no .note.gnu.property.
+//#AbstractConfig:default
+//#Arch:x86_64
+//#Mode:static
+//#ReferenceLinkers:
```
davidlattimore: Why are we disabling all reference linkers? The messages might be different, but the general behaviour should be the same.

davidlattimore (commented):
It looks like anded properties don't behave correctly

```patch
diff --git a/wild/tests/sources/elf/cet-report/cet-report.s b/wild/tests/sources/elf/cet-report/cet-report.s
index 2b1cd991a6..0a406a8b3f 100644
--- a/wild/tests/sources/elf/cet-report/cet-report.s
+++ b/wild/tests/sources/elf/cet-report/cet-report.s
@@ -17,6 +17,9 @@
 //#ExpectWarning:.*: -z force-ibt: file does not have GNU_PROPERTY_X86_FEATURE_1_IBT property
 //#ExpectSectionBytes:.note.gnu.property=0x020000c00400000001000000 16..28
 
+//#Config:force-ibt-mixed:force-ibt
+//#Object:marked.s
+
 //#Config:cet-report-warning:default
 //#ReferenceLinkers:lld
 //#LinkArgs:-z cet-report=warning
diff --git a/wild/tests/sources/elf/cet-report/marked.s b/wild/tests/sources/elf/cet-report/marked.s
new file mode 100644
index 0000000000..98673f9a17
--- /dev/null
+++ b/wild/tests/sources/elf/cet-report/marked.s
@@ -0,0 +1,11 @@
+.section ".note.gnu.property", "a"
+.balign 8
+.long 4
+.long 16
+.long 5
+.asciz "GNU"
+
+.long 0xc0000002 # GNU_PROPERTY_X86_FEATURE_1_AND
+.long 4
+.long 3 # IBT | SHSTK
+.long 0
```

deepakshirkem:
@davidlattimore Thank you very much for your reviews and testing. I’ve addressed the latest comment.

davidlattimore:
It doesn't look like anything has changed. Did something go wrong when you pushed your changes?

davidlattimore (commented):
I should have commented as a review so as to remove this from my to-review list

deepakshirkem:
>It doesn't look like anything has changed. Did something go wrong when you pushed your changes?

Sorry, I force-pushed without staging the changes.

davidlattimore (approved):

PR #2531, opened 2026-09-16
aarch64: implement the Cortex-A53 errata workarounds

### LINKED ISSUE #2260: Support for `--fix-cortex-a53-843419` flag
I just tried switching to Wild linker in CI and it seems to work overall except `--fix-cortex-a53-843419` is being injected by default and causes warnings:
```
warning: linker stderr: wild: --fix-cortex-a53-843419 is not yet supported
```

I didn't find tracking issue for it, so decided to create one.

amaanq:
> [!Note]
> This is my first PR to wild. While working on this, I did use LLMs to write most of the code, though I have taken care to ensure that I've reviewed the code entirely and edited it in a way that made sense to me. I wanted to make sure this was clear up front, so that the maintainers aren't wondering why I made some weird choices in the code, if any. They are likely due to the bot being a bot, and are on me for overlooking them. Thanks!

- Closes #2260

### Problem

Previously, `--fix-cortex-a53-835769` and `--fix-cortex-a53-843419`
were accepted but ignored with a warning. For reference, GNU ld
implements both, and lld implements 843419.

### Solution

Now, the offending instruction is swapped for a branch to a stub that
runs it and branches back. Stubs go in a secondary part appended to
each output section that executable input sections map to, so there
is no extra section header, code addresses stay put, and a linker
script that places `.text` in a memory region keeps the stubs in
reach. Each patch area is rounded up to a whole page, so nothing
after it changes page offset and the scan runs once on provisional
addresses instead of iterating to a fixed point. Additionally,
patching happens in the writer after relocations, so nothing has to
be moved onto the stub. Memory region usage now resets per layout
pass, since the recompute after the scan double counted it.

Stubs sit at the end of their section rather than in the thunk
blocks, so more than 128 MiB of text errors instead of working.
Sorted executable sections, whether from a linker script SORT or
`--sort-section`, have no address at scan time, so those error out
too.


libwild/src/platform.rs
```diff
@@ -5,6 +5,9 @@ use crate::alignment::Alignment;
 use crate::arch::Architecture;
 use crate::bail;
 use crate::env;
+use crate::erratum_aarch64::ErratumFixes;
+use crate::erratum_aarch64::ObjectPatches;
+use crate::erratum_aarch64::PatchSite;
```
davidlattimore: If these types are going to be referenced from the platform module then they shouldn't be platform or arch-specific. One possibility would be to move the type definitions to an erratum module, but leave the instruction decoding, detection etc logic in the erratum_aarch64 module. The Erratum enum would effectively need to be a superset of the erratums supported by all architectures, which for the time being at least would just be aarch64. An alternative would be to make some of these types be generic over Arch and have the Erratum enum be an associated type on Arch. But that could get complicated in other ways, since you're storing some of these types on ObjectLayout which is generic over Platform but not over Arch.

amaanq: Yeah that makes sense. I went ahead with the first option, so `Erratum` and friends now live in a generic erratum module, and `erratum_aarch64` only keeps the decoding and scanning.

libwild/src/erratum.rs
```diff
@@ -0,0 +1,169 @@
+//! Types shared between the generic layout and the per-architecture erratum scanners.
+
+use object::SectionIndex;
+use std::ops::Range;
+
+/// Patches are sized in whole pages so nothing after `.text` changes page offset.
+pub(crate) const PATCH_SIZE_GRANULARITY: u64 = 4096;
```
davidlattimore: This constant seems likely to be arch-specific. Can you think of a way to defer to the arch rather than having it here?

libwild/src/erratum.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// 64-bit register.
+    Mac835769,
+
+    /// An `ADRP` writing `Xn` at page offset `0xff8` or `0xffc`, followed within one or two
+    /// instructions by a load or store using `Xn` as its base register.
+    Adrp843419,
+}
+
+impl Erratum {
+    const ALL: [Erratum; 2] = [Erratum::Mac835769, Erratum::Adrp843419];
```
davidlattimore: I think you can probably derive `strum::EnumIter` instead of doing this

davidlattimore (commented):
It's good to have at least one reference linker on tests wherever possible. I had a quick look at what happens if you enable BFD. It looks like it's applying a different workaround, although possibly if you move the data segment further away it might be forced to apply the same workaround.

Also, there's a couple of likely bugs. Here's a patch that demonstrates them.

```patch
diff --git a/wild/tests/sources/elf/aarch64-erratum-835769/aarch64-erratum-835769.s b/wild/tests/sources/elf/aarch64-erratum-835769/aarch64-erratum-835769.s
index ad09a79d81..ac63c20a65 100644
--- a/wild/tests/sources/elf/aarch64-erratum-835769/aarch64-erratum-835769.s
+++ b/wild/tests/sources/elf/aarch64-erratum-835769/aarch64-erratum-835769.s
@@ -7,6 +7,17 @@
 //#ExpectSectionBytes:.text=0x14 11..12
 //#ExpectSectionBytes:.text=0x1f2003d52204019bfbffff17 24..36
 
+//#Config:regions
+//#RunEnabled:false
+//#Arch:aarch64
+//#Mode:static
+//#LinkerScript:regions.ld
+//#DiffEnabled:false
+//#ExpectSym:value address=0x300000
+//#LinkArgs:--fix-cortex-a53-835769 --no-gc-sections
+//#ReferenceLinkers:bfd
+//#ExpectSectionBytes:.text=0x14 11..12
+
 .text
 .globl _start
 .type _start, @function
diff --git a/wild/tests/sources/elf/aarch64-erratum-835769/regions.ld b/wild/tests/sources/elf/aarch64-erratum-835769/regions.ld
new file mode 100644
index 0000000000..f207056d5b
--- /dev/null
+++ b/wild/tests/sources/elf/aarch64-erratum-835769/regions.ld
@@ -0,0 +1,10 @@
+/* Data is just large enough for content */
+MEMORY {
+    TEXT : ORIGIN = 0x200000, LENGTH = 0x10000
+    DATA : ORIGIN = 0x300000, LENGTH = 8
+}
+
+SECTIONS {
+    .text : { *(.text) } > TEXT
+    .data : { *(.data) } > DATA
+}
diff --git a/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s b/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
index be7c8a7095..9ecba3f350 100644
--- a/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
+++ b/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
@@ -1,18 +1,27 @@
 // -Ttext puts the adrp at page offset 0xff8 so the second ldr gets patched, with
 // byte 11 being the top byte of the branch.
+//#Config:default
 //#Arch:aarch64
 //#Mode:static
 //#LinkArgs:--fix-cortex-a53-843419 --no-gc-sections -Ttext=0x200ff8
-//#ReferenceLinkers:
+//#ReferenceLinkers:bfd
 //#ExpectSectionBytes:.text=0x14 11..12
 //#ExpectSectionBytes:.text=0x010840f9fcffff17 24..32
 
+//#Config:simd:default
+//#CompArgs:-Wa,--defsym,SIMD=1
+
 .text
 .globl _start
 .type _start, @function
 _start:
     adrp x0, value
+.ifdef SIMD
+    // q0 is a separate register to x0, so should not affect outcome.
+    ldr  q0, [x0]
+.else
     ldr  x1, [x0]
+.endif
     ldr  x1, [x0, :lo12:value]
     mov  x8, #93
     mov  x0, x1
```

libwild/src/erratum.rs
amaanq: oh that's a neat one, til

amaanq:
Thanks for the reviews, PTAL :)

davidlattimore (commented):
Here's another couple of tests that fail. Don't feel that you have to make them pass as-is. Reducing the scope say to not support =adr would also be fine.

```patch
diff --git a/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s b/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
index 1c12e96c32..72aff5c43c 100644
--- a/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
+++ b/wild/tests/sources/elf/aarch64-erratum-843419/aarch64-erratum-843419.s
@@ -23,6 +23,24 @@
 //#Config:simd:bfd
 //#CompArgs:-Wa,--defsym,SIMD=1
 
+//#Config:load-address:base
+//#LinkArgs:--fix-cortex-a53-843419 --no-gc-sections
+//#LinkerScript:load-address.ld
+//#ReferenceLinkers:bfd,lld
+//#RunEnabled:false
+//#DiffEnabled:false
+//#ExpectProgramHeader:LOAD flags=RX,vaddr=0x200ff8,paddr=0x800000,sections=[.text]
+
+// ADR-only mode must rewrite the ADRP in place, leaving the load intact.
+//#Config:adr
+//#Arch:aarch64
+//#Mode:static
+//#LinkArgs:--fix-cortex-a53-843419=adr --no-gc-sections -Ttext=0x200ff8 -Tdata=0x210000
+//#ReferenceLinkers:bfd
+//#DiffEnabled:false
+//#ExpectSectionBytes:.text=0x10 3..4
+//#ExpectSectionBytes:.text=0xf9 11..12
+
 .text
 .globl _start
 .type _start, @function
diff --git a/wild/tests/sources/elf/aarch64-erratum-843419/load-address.ld b/wild/tests/sources/elf/aarch64-erratum-843419/load-address.ld
new file mode 100644
index 0000000000..a4371f9669
--- /dev/null
+++ b/wild/tests/sources/elf/aarch64-erratum-843419/load-address.ld
@@ -0,0 +1,4 @@
+SECTIONS {
+    .text 0x200ff8 : AT(0x800000) { *(.text) }
+    .data 0x600000 : { *(.data) }
+}
```

libwild/src/erratum_aarch64.rs
```diff
@@ -0,0 +1,567 @@
...
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct Insn(u32);
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct Reg(u32);
+
+impl Insn {
```
marxin, line 215: I think we should rather make a research and try using an existing crate for the instruction parsing and encoding. Something like https://github.com/capstone-rust/capstone-rs or https://crates.io/crates/yaxpeax-arm might be a good fit.

amaanq: Imo capstone is quite a heavy dependency even in diet mode, though I'm not familiar with yaxpeax-arm, I will look into that. The current implementation for instruction parsing is very tailored for these specific erratum and is quite light, though I can see this being expanded upon in the future for whatever else, in which case using a proper dependency _would_ be nice and warranted....just my two cents

davidlattimore: If the dependencies are heavyweight or slow for our use-case then I'm inclined to agree that we should just do it ourselves. We effectively do that with all the relaxation code too.

PR #2535, opened 2026-09-16
port(MachO): implement `__compact_unwind` section emission

### LINKED ISSUE #2066: Mach-O: add support for `__compact_unwind`
The format is an equivalent to the ELF's `.eh_frame`, as opposed to the format, the relocatable objects on Mach-O contain a per-function entries that needs to be properly merged into a special data structure.
For more information, take a look here: https://github.com/wild-linker/wild/blob/main/libwild/MachO.md#--__compact_unwind-format

marxin:
The PR implements the `__compact_unwind` section emission where we currently limit the support number of encodings to `256`. Once merged, I'm planning to support arbitrary number of encodings, where some of them will be provided in second level page index. A blog post entry I used for the reference implementation: https://faultlore.com/blah/compact-unwinding/.

TODO: Since the #2521, I have a problem with the relocations that point to the personality function:
```
Section __compact_unwind {
...
    0x30 0 3 1 ARM64_RELOC_UNSIGNED 0 ___gxx_personality_v0
...
```

given the fact the `__compact_unwind` is a synthetic section, linking fails due to:
`wild: error: Invalid section address for fixup`

@joanna-jo Do you see a smooth path about how to integrate it into the `pending_fixups` mechanism?

Part of: #2066


marxin:
> @joanna-jo Do you see a smooth path about how to integrate it into the `pending_fixups` mechanism?

I think I find a pretty good way how to special case the unwind info personality relocations.

joanna-jo (commented):
> @joanna-jo Do you see a smooth path about how to integrate it into the `pending_fixups` mechanism?

For the personality function, it looks like lld emits a GOT entry and we need to avoid emitting a regular 8 byte absolute pointer relocation for it. I think the simplest fix would be to have something like `is_emitted` as a check in the logic for pending fixups, as the issue seems to be that `__compact_unwind` is consumed to create `__unwind_info`, and we shouldn't be emitting a bind at that location in `__compact_unwind`.

Longer term, I wonder if we should mirror ELF's `classify_symbol_relocation`, `materialize_relocation_requirements`, and `note_relocation_symbol_reference`, since I saw that you already introduced `ClassifiedSymbolRelocation`? That way, we can go through `process_relocation` for regular relocations but have `process_compact_unwind_section` call `classify_symbol_relocation`, set the GOT flag, then call `note_relocation_symbol_reference`. We can isolate pending fixups logic in `materialize_relocation_requirements`.

joanna-jo:
> I think I find a pretty good way how to special case the unwind info personality relocations.

Oops sorry - I think I missed your newest commit just now while writing my review! LGTM, I don't have a preference between this or the first approach I originally mentioned.

libwild/src/compact_unwind.rs
```diff
@@ -0,0 +1,359 @@
...
+use crate::macho::UnwindInfoWithRelocs;
+use crate::symbol_db::SymbolId;
+use anyhow::Context;
+use hashbrown::HashMap;
+use itertools::Itertools;
+use zerocopy::FromBytes;
+use zerocopy::Immutable;
+use zerocopy::IntoBytes;
+
+// TODO: Copied from `macho-unwind-info` crate and we should upstream it in `object` crate
```
davidlattimore: If stuff is copied from elsewhere, do we need to do anything in terms of license? Perhaps put markers at the start and end of where the different license applies, copy the license file somewhere and refer to it?

libwild/src/macho_writer.rs
```diff
@@ -379,6 +389,66 @@ fn build_exports_trie(layout: &MachOLayout<'_>) -> Result<Vec<u8>> {
...
+            let Some(start_relocation) = entry.start_relocation else {
+                bail!("unwind info missing start relocation");
+            };
+            let start_address = object_layout
+                .section_resolutions
+                .get((start_relocation.r_symbolnum - 1) as usize)
+                .context("cannot resolve start of an unwind info")?
+                .address()
+                .context("missing unwind info relocation")?
+                + entry.entry.start;
```
davidlattimore, line 409: In `symbol_offset_in_section`, it says "On Mach-O the symbol value is the global offset, not a relative to the start of a section". Do we need to take that into account here? i.e. is `entry.entry.start` a global offset? If it is, then we presumably can't just add it directly to the start address of the section.

libwild/src/compact_unwind.rs
```diff
@@ -0,0 +1,359 @@
...
+    let lsda_data_offset = header_size + (compressed_pages + 1) * size_of::<PageEntry>();
+    let second_level_first_offset = lsda_data_offset + lsdas.len() * size_of::<LsdaEntry>();
+
+    let mut lsda_offset = 0;
+    for (i, (start, lsda_count, _)) in second_level_entries.iter().enumerate() {
+        let record = PageEntry {
+            first_address: u32::try_from(*start - text_segment_start)
+                .context("first address does not fit into u32")?,
+            lsda_index_offset: (lsda_data_offset + lsda_offset * size_of::<LsdaEntry>()) as u32,
+            page_offset: u32::try_from(second_level_first_offset + i * size_of::<CompressedPage>())
```
davidlattimore, line 320: What happens if this is larger than the page size?

libwild/src/compact_unwind.rs
marxin: Sure, I feel the best approach would be asking directly @mstange - it's coming from here: https://github.com/mstange/macho-unwind-info/blob/main/src/raw/format.rs. I'm going to write him an email what's the best approach of referencing the code we're going to reuse.

libwild/src/macho_writer.rs
marxin: Correct - addressed that in my latest commit. The reason why it worked is the `__text` section is the very first section and so it's effectively a subtraction of zero.

libwild/src/compact_unwind.rs
marxin: That will mean you'll have 1G functions :) But sure, let's address that properly.

marxin:
> Oops sorry - I think I missed your newest commit just now while writing my review! LGTM, I don't have a preference between this or the first approach I originally mentioned.

No, no, I thank you for the reply. Good we're aligned on the selected approach ;)

> Longer term, I wonder if we should mirror ELF's classify_symbol_relocation, materialize_relocation_requirements, and note_relocation_symbol_reference, since I saw that you already introduced ClassifiedSymbolRelocation? That way, we can go through process_relocation for regular relocations but have process_compact_unwind_section call classify_symbol_relocation, set the GOT flag, then call note_relocation_symbol_reference. We can isolate pending fixups logic in materialize_relocation_requirements.

Definitely, we should come up with a similar function in a future - a good observation we've got an infrastructure on the ELF part of the linker.

libwild/src/compact_unwind.rs
marxin: Got a reply from Markus and he wrote he's fine with whatever makes it most convenient for us. So I explicitly marked what has been copied and provided link to the original blog post.

libwild/src/compact_unwind.rs
```diff
@@ -0,0 +1,362 @@
...
```
davidlattimore: Each page_offset is advanced by `size_of::<CompressedPage>` relative to the previous one, but the loop below that writes them writes `encoding_array` in between the CompressedPages. Don't we need to take the size of the encoding array into account when determining the page_offset?

libwild/src/macho.rs
```diff
@@ -1559,12 +1635,136 @@ impl platform::Platform for MachO {
...
+                .get_mut(address / ENTRY_LEN)
+                .context("missing unwind info entry for a relocation")?;
+
+            match relocation_field_offset {
+                START_FIELD_OFFSET => {
+                    entry.start_relocation = Some(info);
+                    let Some(text_addr) = text_section.map(|s| s.addr.get(LE)) else {
+                        bail!("missing {TEXT_SECTION_NAME} section")
+                    };
+                    entry.entry.start -= text_addr;
```
davidlattimore: Could the relocation be relative to something other than _text? If it can, but we don't want to deal with that now, can we error in that case?

libwild/src/compact_unwind.rs
marxin: Very good observation, addressed with comments explaining we've been currently using full pages of `4KiB` - this part is going to change in my upcoming patch where we'll allow usage of `local_opcodes_offset/local_opcodes_len`.

libwild/src/macho.rs
marxin: Again, you were right, there might be used a different section than `__text`:
```
  Section __compact_unwind {
    0x7A0 0 3 0 ARM64_RELOC_UNSIGNED 0 __StaticInit
...
```

Addressed systematically by taking the section index directly from the relocation.

libwild/src/compact_unwind.rs
marxin: Stress tested the `page_offset` by shrinking `COMPRESSED_PAGE_SIZE` to 16 that effectively means each function lives in its own page!

davidlattimore (approved):

PR #2536, opened 2026-09-16
fix(MachO): use got_address for GOT-related relocations

marxin:
It's a fix for GOT-related relocations.

Part of #757


davidlattimore (approved):
Should we have a test for this? Otherwise, looks good

marxin:
> Should we have a test for this?

Yes, added and will merge the change soon.

PR #2537, opened 2026-09-17
port(MachO): Add support for `__thread_ptrs`

joanna-jo:
For every imported TLV symbol, we need to allocate a `__thread_ptrs` entry, apply the relocation so that it points to that `__thread_ptrs` entry, and emit a bind at the `__thread_ptrs` entry for the TLV descriptor for that symbol.

cc: @marxin @davidlattimore 

Part of #2071


joanna-jo:
Hmmm this test passes locally for me, but it fails in CI under Xcode 26.6. I did some investigating and it looks like the latest version of ld uses a GOT entry for the imported TLV descriptor pointer instead, even though older versions of ld and current lld use a `__thread_ptrs` entry. 

```
Homebrew clang version 23.1.0
Homebrew LLD 23.1.0
Apple ld version: 1267
__thread_ptrs exists: no
__got exists: yes
_value binding:
    __DATA_CONST    __got    0x100004000    bind    foo.lld/_value
```

Do we have any thoughts on what we'd prefer in our linker? Semantically, I don't see any reason not to include this as a GOT entry and benefit from reusing our GOT logic. @marxin @davidlattimore

davidlattimore:
Using a GOT entry sounds good to me

joanna-jo:
> Using a GOT entry sounds good to me

Cool, I'll go ahead and close this then!

PR #2538, opened 2026-09-17
perf: Thin absolute debug relocs and write debug by output section

lapla-cogito:
I've deliberately consolidated these two changes into a single PR because, they share the same goal of accelerating debug info processing and only both modifications together yield measurable performance improvements at least in my environment. Specifically, each commit alone showed about up to only a 2% speedup in Zed builds, but when combined, we achieved approximately a 7% speedup.

commits:

 1. Most debug relocs are absolute 32/64-bit stores. Resolve the value in the same order as the generic path (canonical resolution, then section address), then write it without flags-merge or the relocation-kind table.

2. `.debug_info`, `.debug_ranges`, and `.debug_line` from one input file no longer share a thread. Allocated sections stay on the per-group path.


marxin:
In general, I welcome any performance improvement, however we must be fully aware of the complexity increase and in this particular case, I am not convinced it's worth it:

> Most debug relocs are absolute 32/64-bit stores. Resolve the value in the same order as the generic path (canonical resolution, then section address), then write it without flags-merge or the relocation-kind table.

Given the current logic in `apply_debug_relocation`, can we make a fast-path at the beginning of the function where we'll basically process do similar logic to `write_simple_debug_absolute` (for `RelocationKind::Absolute`)? 

> 2\. `.debug_info`, `.debug_ranges`, and `.debug_line` from one input file no longer share a thread. Allocated sections stay on the per-group path.

Interesting observation! Do you understand the nature of the limitation why it leads to a slower result? Intuitively, with similarly big object files, I would not expect any issue with processing debug info sections from one file one after other. That should lead to a good thread saturation. Or do I miss something?

davidlattimore (commented):
At least on my machine, I don't see quite as large a speedup. For zed-debug I see about 1.6% when --threads=8. With 32 threads, the speedup is less. There is about a 10% reduction in instruction count in both cases though, so I it looks like at least on my machine, the extra instructions without this change were being largely mitigated through parallelism.

For linking wild-debug, the speedup was a bit more. 4.5% at 8 threads and 2.5% at 32.

I also tried benchmarking on my raspberrypi and didn't see a speedup. It reported a 1.1% slowdown, but it's too small to be sure. The instruction count reduction was still present thought.

Here's a couple of tests that pass on main but fail with this change

```patch
diff --git a/wild/tests/sources/elf/debug-abs32-range/debug-abs32-range.s b/wild/tests/sources/elf/debug-abs32-range/debug-abs32-range.s
new file mode 100644
index 0000000000..038b96ccfd
--- /dev/null
+++ b/wild/tests/sources/elf/debug-abs32-range/debug-abs32-range.s
@@ -0,0 +1,25 @@
+/*
+Use lld as reference since GNU ld rejects negative R_AARCH64_ABS32 values.
+//#Arch:aarch64,riscv64
+//#ReferenceLinkers:lld
+//#RunEnabled:false
+//#ExpectSectionBytes:.debug_test=0xffffffff00000080ffffffff
+*/
+
+.text
+.globl _start
+.type _start, %function
+_start:
+    ret
+.size _start, .-_start
+
+.section .debug_base,""
+.Lbase:
+    .byte 0
+
+.section .debug_test,""
+.long .Lbase - 1
+.long .Lbase - 0x80000000
+.long .Lbase + 0xffffffff
+
+.section .note.GNU-stack,""
diff --git a/wild/tests/sources/elf/debug-mixed-merge/debug-mixed-merge.s b/wild/tests/sources/elf/debug-mixed-merge/debug-mixed-merge.s
new file mode 100644
index 0000000000..e431ae43bc
--- /dev/null
+++ b/wild/tests/sources/elf/debug-mixed-merge/debug-mixed-merge.s
@@ -0,0 +1,19 @@
+//#Arch:x86_64
+//#Object:merged.s
+//#ReferenceLinkers:bfd,lld
+//#LinkArgs:--no-gc-sections
+//#DiffIgnore:section.debug_str.entsize
+//#RunEnabled:false
+//#ExpectSection:.debug_str
+//#Contains:unmerged-debug-string
+//#Contains:mergeable-debug-string
+
+.text
+.globl _start
+_start:
+    ret
+
+.section .debug_str,"",@progbits
+.asciz "unmerged-debug-string"
+
+.section .note.GNU-stack,"",@progbits
diff --git a/wild/tests/sources/elf/debug-mixed-merge/merged.s b/wild/tests/sources/elf/debug-mixed-merge/merged.s
new file mode 100644
index 0000000000..c6b178c745
--- /dev/null
+++ b/wild/tests/sources/elf/debug-mixed-merge/merged.s
@@ -0,0 +1,4 @@
+.section .debug_str,"MS",@progbits,1
+.asciz "mergeable-debug-string"
+
+.section .note.GNU-stack,"",@progbits
```

PR #2541, opened 2026-09-18
perf: Release input mappings in parallel on shutdown

lapla-cogito:
Call `madvise(MADV_DONTNEED)` on input mmaps from multiple threads while dropping the layout, so the kernel does not reclaim those pages on a single thread when `Linker` struct is later dropped. This mainly helps `--no-fork`.


lapla-cogito:
With `--no-fork` enabled, I observed a 5% overall speedup on clang-debug, and when focusing specifically on the shutdown phase, the improvement reached approximately 60%.

davidlattimore (approved):
Nice :)

I was wondering if doing this when --fork is active might slow things down, but I can't see any slowdown, so I guess there's perhaps it's fine to always do it.

marxin:
Very cool @lapla-cogito, I was able to measure the following numbers for clang.release:
```
❯ hyperfine -C 'sleep 1' './run-with /tmp/ld --no-fork' './run-with /home/marxin/Programming/wild/ld --no-fork'
Benchmark 1: ./run-with /tmp/ld --no-fork
  Time (mean ± σ):      67.7 ms ±   1.6 ms    [User: 721.5 ms, System: 236.4 ms]
  Range (min … max):    66.3 ms …  72.0 ms    10 runs

Benchmark 2: ./run-with /home/marxin/Programming/wild/ld --no-fork
  Time (mean ± σ):      78.9 ms ±   2.2 ms    [User: 706.3 ms, System: 222.1 ms]
  Range (min … max):    75.7 ms …  81.7 ms    10 runs

Summary
  ./run-with /tmp/ld --no-fork ran
    1.16 ± 0.04 times faster than ./run-with /home/marxin/Programming/wild/ld --no-fork
```

I'm curious if the real reason for the speed up is the `madvise`, or the fact that we are able to do the expensive work (unmapping) from multiple threads. Based on my intuition is, if we dropped `Mmap` (that later calls `libc::munmap`) from multiple threads, we should receive a similar speed-up as the kernel can drop the mapping anyway. Right?

lapla-cogito:
> I'm curious if the real reason for the speed up is the madvise, or the fact that we are able to do the expensive work (unmapping) from multiple threads.

I think it's the latter.

> Based on my intuition is, if we dropped Mmap (that later calls libc::munmap) from multiple threads, we should receive a similar speed-up as the kernel can drop the mapping anyway. Right?

Considering that `Layout` has numerous references to inputs, wouldn't simply unmapping it potentially create dangling pointers?

marxin:
> Considering that `Layout` has numerous references to inputs, wouldn't simply unmapping it potentially create dangling pointers?

Yeah, I meant calling `Drop` on the `InputFile`s, which requires owning the objects. In any case, a nice speed-up!

PR #2544, opened 2026-09-18
chore: a clean up of `-z cet-report`

marxin:
A small code simplification of the previously merged commit from @deepakshirkem.


libwild/src/elf.rs
```diff
@@ -4415,41 +4418,35 @@ impl LayoutExt {
...
-    use object::elf::GNU_PROPERTY_X86_FEATURE_1_IBT;
-    use object::elf::GNU_PROPERTY_X86_FEATURE_1_SHSTK;
-
-    {
-        // Get the feature bits for this file
-        let feature_bits = props
-            .iter()
-            .find(|p| p.ptype == object::elf::GNU_PROPERTY_X86_FEATURE_1_AND)
-            .map_or(0, |p| p.data);
+    // Get the feature bits for this file
```
joanna-jo: Maybe we can even remove this comment as part of cleanup? I think the code is already self-documenting here since the variable is called `feature_bits`.

davidlattimore (approved):

libwild/src/elf.rs
marxin: Yes, good point!

PR #2549, opened 2026-09-19
test: Change CompArgs, CompSoArgs, LinkArgs etc to append

davidlattimore:
Also, because I felt that it would have been confusing, CompArgs no longer applies to shared objects even when CompSoArgs is empty.


davidlattimore:
Part of my motivation here is that I'd like to get rid of the Variant directive - it just seems unnecessary when you can pass `-DWHATEVER=value` via CompArgs - but doing so is a lot cleaner if you don't also need to repeat all the other required CompArgs each time you do so.

PR #2552, opened 2026-09-19
CI: use native Cargo dependency cooldown feature

marxin:
With the increasing number of supply chain attacks ([^1]), Cargo introduced a native cooldown mechanism that will be stabilized in `1.100` (if I'm not mistaken): https://github.com/rust-lang/cargo/pull/17335. So far available on `+nightly` Cargo version.

The check can properly detect if a version from `Cargo.toml` does not fulfill the cooldown period, but there's one drawback connected to `Cargo.lock` file dependencies - these are ignored and only a warning is reported. So until the feature is stabilized, one must use nightly compiler for updating.

Sample output if accidentally updated to latest:
```
❯ cargo +nightly update
    Updating crates.io index
     Locking 6 packages to highest compatible versions as of 3 days ago
 Downgrading cc v1.4.7 -> v1.4.6 (available: v1.4.7, published 19 hours ago)
 Downgrading cfg-if v1.0.5 -> v1.0.4 (available: v1.0.5, published 3 days ago)
 Downgrading find-msvc-tools v0.1.13 -> v0.1.12 (available: v0.1.13, published 19 hours ago)
 Downgrading rustix v1.1.5 -> v1.1.4 (available: v1.1.5, published 2 days ago)
 Downgrading syn v3.0.6 -> v3.0.5 (available: v3.0.6, published 2 days ago)
 Downgrading unicode-ident v1.0.26 -> v1.0.24 (available: v1.0.26, published 2 days ago)
note: pass `--verbose` to see 1 unchanged dependencies behind latest
```

If sticking to a fresh release in `Cargo.toml`:
```
❯ cargo +nightly update
    Updating crates.io index
error: failed to select a version for the requirement `cc = "^1.4.7"`
  version 1.4.7 is too new (published 19 hours ago, minimum age 3 days)
location searched: crates.io index
```

And the linked blog post is also pretty nice reading: https://cooldowns.dev/

[^1]: https://blog.rust-lang.org/2026/08/20/supply-chain-attack-on-arrayref/


.cargo/config.toml.publish-age
```diff
@@ -0,0 +1,6 @@
+[registry]
+# Even one day makes a big difference: https://cooldowns.dev/#does-it-actually-work
+global-min-publish-age = "3 days"
```
davidlattimore, line 3: When building on stable, this and the line below cause warnings to be emitted. I guess one option would be to put this in a config.toml.publish-age and have the CI job copy it.

.github/workflows/ci.yml
```diff
@@ -244,6 +244,19 @@ jobs:
         with:
           command: check
 
+  dependency_cooldown:
+    name: Dependency cooldown
+    runs-on: ubuntu-24.04
+    continue-on-error: true
```
davidlattimore: What's the idea behind setting this? Are we likely to notice a failure if it continues?

davidlattimore (commented):
Nice! I'd been thinking it'd be good if we could take advantage of that feature somehow

.cargo/config.toml.publish-age
marxin: Sure, before it's part of a stack toolchain, let's make it used only the the CI job.

.github/workflows/ci.yml
marxin: It's a copy&paste leftover from a different project - I think we can make it a required job.

davidlattimore (approved):

davidlattimore:
I was just wondering about what dependabot will do and found that it has a 3 day cooldown by default - [introduced relatively recently](https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/). So I guess this CI job will mostly force us to slow down if people do a manual cargo update.

edit: I see this is mentioned in the article you linked, so probably isn't news to you :)

marxin:
Yes, I know it aligns nicely with Dependabot ;)

PR #2553, opened 2026-09-19
feat(x86_64): support SIZE32 and SIZE64 relocations

deepakshirkem:
Adds support for `R_X86_64_SIZE32` and `R_X86_64_SIZE64` relocations. Non-preemptible symbols are resolved using the final symbol size plus addend, while interposable symbols in shared objects emit the corresponding dynamic relocation.

Part of #2380


libwild/src/layout.rs
```diff
@@ -3295,6 +3295,7 @@ pub(crate) fn resolution_flags(rel_kind: RelocationKind) -> ValueFlags {
             ValueFlags::empty()
         }
         RelocationKind::Absolute
+        | RelocationKind::SymbolSize
```
davidlattimore: Mapping SymbolSize as DIRECT might mean that we generate a copy relocation for the symbol when we don't actually need one. Can you verify if that's the case by adding a test for this?

deepakshirkem: I added a regression test with a symbol defined in a shared object and referenced only via foo@SIZE. Thank You.

libwild/src/elf_writer.rs
```diff
@@ -3392,6 +3392,24 @@ fn apply_relocation<
             object_layout,
             layout,
         )?,
+        RelocationKind::SymbolSize => {
```
davidlattimore, line 3419: In think that if section to which the relocation is being applied is read-only, we might still emit the relocation, we should probably fail instead like we do for other relocation kinds. Can you add a test to verify if this is the case?

libwild/src/elf.rs
```diff
@@ -6293,6 +6286,19 @@ fn materialize_relocation_requirements<
...
                 .store(true, atomic::Ordering::Relaxed);
         }
+    } else if rel_kind == RelocationKind::SymbolSize
+        && flags.is_interposable()
+        && (flags.is_dynamic() || symbol_db.output_kind.is_shared_object())
+    {
+        if !section_is_writable {
+            bail!(
+                "Cannot apply relocation {} to read-only section. \
+                Please recompile with -fPIC or link with -no-pie",
```
davidlattimore: I think the message might need some tweaking. We likely can apply that relocation type to a read-only section, just not for a dynamic symbol. I'm also not sure that either of the suggested flags will help in this case. Maybe just "Cannot apply dynamic relocation {} to read-only section." Including the symbol name that was referenced might also be good.

davidlattimore (approved):
Thanks :)

PR #2554, opened 2026-09-19
refactor: Confine host-specific code to a `platforms` module

zackees:
Host (not target) OS selection is currently spread across `fs.rs`, `subprocess.rs`, `perf.rs`, `save_dir.rs`, `linker_plugins.rs` and ~30 `cfg(all(feature = "plugins", unix))` sites, each with its own predicate. Those predicates have drifted: `fs.rs` detects filesystems under `any(target_os = "android", target_os = "linux")` but only defines `kernel_version` for `target_os = "linux"`, so libwild doesn't currently build for Android. This branch fixes that as a side effect.

**Phase 1 (this PR).** `libwild/src/platforms/mod.rs` selects exactly one host tree with a `cfg_select!` that has no fallback arm, so an unlisted host is a compile error there instead of silently inheriting whatever the scattered `cfg(not(...))` branches did. It's the Rust form of an `#if / #elif / #else #error` chain. Facades (`fs`, `host`, `linker_plugin`, `perf`, `process`) re-export the selected tree's items, every tree must provide every item, and a tidy test rejects host `cfg`, `std::os`, `libc` and `nix` outside the trees.

**Phase 2 (next, separate PR).** Replace the textual tidy test with a Dylint rule that works on the pre-expansion AST and ratchets the remaining host selectors out of business logic into the facade. A lint sees what a text scan can't (macro bodies, attribute forms, cfg'd-out code on the linting host, so one Linux job covers every host), and it can cover what phase 1 leaves alone: `wild/tests/integration_tests.rs` (15), `wild/tests/external_tests/lld_tests.rs` (2) and `linker-diff/src/utils.rs` (1). Happy to drop that idea if you'd rather not add a Dylint job.

Behaviour changes, all deliberate:
- The WASI-only test gates become runtime checks, so those tests pass trivially on WASI rather than being reported as ignored.
- libwild builds for Android again.
- With the `plugins` feature, `linker_plugins.rs` now compiles on every host; plugins stay enabled only where the host supports them, via `linker_plugins::ENABLED`. That exposed an `off_t` cast that was wrong on Windows.

Most of the diff is mechanical: 933 of the added lines are the new `platforms/` files, mostly re-export lists, and 73 changed lines are the one `cfg(all(feature = "plugins", unix))` → `cfg(feature = "plugins")` substitution. The parts worth real review are `platforms/mod.rs`, the five facades and `common.rs`. `perf.rs` is a pure rename. The one change I'd most like eyes on is plugin symbol lookup, which now copies the `onload` pointer out of `libloading::Symbol` through the facade.

Checked on every host tree (`aarch64`/`x86_64`/`armv7` Android, Windows MSVC and GNU, macOS, WASI, illumos, FreeBSD, Linux gnu/musl/riscv64) with `cargo clippy -p libwild --lib --tests --target $T --no-default-features --features fork,plugins` and `-D warnings`. Your CI is green on all 24 jobs on my fork: https://github.com/zackees/wild/actions/runs/35473070194

I can split this into smaller PRs if you'd prefer to review it in pieces — the Android fix, the skeleton, `fs.rs`, subprocess, plugins and the tidy test are separable.


davidlattimore (commented):
Thanks for the PR. It is quite large, so if you wouldn't mind splitting some stuff out, that could be good

libwild/src/platforms/mod.rs
davidlattimore, line 1: Do you think calling this module "platforms" is too similar to "platform"? Wondering if we should call this "host" instead

libwild/src/platforms/platform_unix.rs
davidlattimore, line 1: I'm unsure about prefixing the module name here with the name of the parent module

zackees:
i'll split it up

zackees:
Split started: #2555 is the first of three — the `host` module plus fork and perf (17 files, +305/-118). It applies both naming points from the review: `host` instead of `platforms`, and tree modules without the parent-module prefix. Filesystem services (including the Android build fix) and the linker-plugin bits plus the tidy test follow as separate PRs. Happy to close this one now, or leave it open as the overall context until the three have landed — whichever you prefer.

davidlattimore:
Thanks for splitting. I'll close this now, since I have enough trouble keeping track of all the in-flight PRs as it is :)

PR #2555, opened 2026-09-20
refactor: Add a `host` module, starting with fork and perf

zackees:
First of three, splitting #2554 as you asked. This one adds the `host` module and moves two small services into it; the other two follow once this lands.

`libwild/src/host/mod.rs` picks exactly one host tree with a `cfg_select!` that has no fallback arm, so building for an unlisted host is a compile error there rather than silently inheriting whatever the scattered `cfg(not(...))` branches do. It's the Rust form of an `#if / #elif / #else #error` chain. Facades re-export the selected tree's items, and every tree must provide every item.

Moved in here: forking, which drops `subprocess_unsupported.rs`, and performance counters, where `perf.rs` becomes the Linux tree's and `perf_unsupported.rs` becomes a shared stub. No behaviour change.

Naming follows your review on #2554: the module is `host` rather than `platforms`, so it isn't confused with `platform`, and the trees are `unix`, `linux`, `macos`, `illumos`, `other_unix`, `windows` and `wasi`, without the parent-module prefix.

Still to come, as separate PRs:
1. Filesystem services (inputs, preallocate, huge pages, filesystem detection, symlinks). That one also fixes the Android build: `fs.rs` detects filesystems under `any(target_os = "android", target_os = "linux")` but only defines `kernel_version` for `target_os = "linux"`, so libwild doesn't build for Android today.
2. The linker-plugin host bits, plus a tidy test that keeps host `cfg`, `std::os`, `libc` and `nix` out of everything except the trees. After that I'd like to propose a Dylint rule that does the same check on the pre-expansion AST, which catches macro bodies and code that's `cfg`'d out on the linting host, so one Linux job covers every host. Happy to skip that if you'd rather not add a Dylint job.

Checked with `cargo clippy --all-targets --workspace` (default and `--no-default-features`) and `-D warnings`, plus `-p libwild --lib --tests --features fork,plugins` for `aarch64`/`x86_64`/`armv7` Android, Windows MSVC and GNU, macOS, WASI, illumos, FreeBSD, and Linux gnu/musl/riscv64. Android still fails there, from the existing bug that PR 2 fixes.


davidlattimore (approved):
Thanks :) This restructuring looks good to me

PR #2557, opened 2026-09-20
refactor: Move linker-plugin host services into the `host` module, and enforce the boundary

zackees:
Last of three, stacked on #2556 and #2555 — review only the final commit until those land, after which I will rebase onto main.

The linker plugin's host-specific pieces (dlopen, the raw file descriptor, `off_t`, the file-limit bump) move behind a `linker_plugin` facade. That turns the ~30 `cfg(all(feature = "plugins", unix))` sites into plain `feature = "plugins"`, and `linker_plugins.rs` now compiles on every host. Plugins stay enabled only where the host supports them, through `linker_plugins::ENABLED`, so a host that can't load them takes the same path it took when the module was compiled out. Compiling on Windows exposed an `off_t` cast that assumed 64 bits.

With nothing left violating it, `check_host_specific_code` then locks the boundary in: host `cfg` keys, `std::os`, `libc` and `nix` are rejected outside the host trees, as are references to the trees from outside `host/`. It scans sources as text, so code that's cfg'd out on the linting host is still checked and one Linux run covers every host. I verified it catches violations in ordinary code, in the facades, and inside `cfg_select!`; naming a tree from outside `host/` is already a compile error, since `imp` is private.

The piece I'd most like reviewed is plugin symbol lookup, which now copies the `onload` pointer out of `libloading::Symbol` through the facade. The library is kept alive in `LoadedPlugin`, so the pointer stays valid, but it's the one place where the indirection changed the shape of the code rather than moving it.

If you'd like the same check as a lint rather than a text scan, I can follow up with a Dylint rule on the pre-expansion AST, which also sees macro bodies and could cover `wild/tests` and `linker-diff`. Happy to leave it out if a Dylint job isn't something you want to maintain.


libwild/src/linker_plugins.rs
```diff
@@ -432,22 +439,18 @@ impl LoadedPlugin {
...
         // Clear any existing state in case this thread previously made it part way through
         // initialisation.
         CALLBACKS.take();
 
-        let onload_fn: libloading::Symbol<unsafe extern "C" fn(*mut LdPluginTv)> =
-            unsafe { lib.get(b"onload") }
-                .context("Failed to get `onload` function from linker plugin")?;
+        // Safety: `onload` has this type in the linker plugin API, and `lib` outlives our use of
+        // it.
+        let onload_fn: unsafe extern "C" fn(*mut LdPluginTv) = unsafe { lib.symbol(b"onload") }
```
davidlattimore: The problem predates your change, but since your safety comment is now claiming that we match the linker plugin API, I noticed that the API declares that the function returns a status - https://github.com/gcc-mirror/gcc/blob/1d34a925554d036e159abb87418d86d59feadc07/include/plugin-api.h#L616

Although I'd be happy to deal with that in a subsequent PR, since this one is already large.

libwild/src/tidy_tests.rs
```diff
@@ -254,3 +272,114 @@ fn check_elf_specific_code() -> Result {
...
+/// `host/`. See `host/mod.rs`.
+#[test]
+fn check_host_specific_code() -> Result {
+    // Sandboxed hosts (WASI) can't see the source tree or run formatters.
+    if crate::host::os::SANDBOXED {
+        return Ok(());
+    }
+    let src_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("src");
+
+    // `cfg` keys that select the host. `target_arch` and `target_endian` aren't here: outside
```
davidlattimore: This comment doesn't seem right to me. Surely target_arch and target_endian are related to the host we're running on not the host we're linking for?

zackees:
Fixing up.

davidlattimore (approved):

PR #2558, opened 2026-09-20
ci: Enforce the host boundary with a Dylint lint, and ratchet down what's left

zackees:
Follow-up to #2557, stacked on #2555, #2556 and #2557 — review only the last commit until those land, after which I will rebase onto main.

This replaces the text-based `check_host_specific_code` tidy test from #2557 with a Dylint lint that reads the pre-expansion AST. That lets it see `cfg`s inside macro bodies and code that would be `cfg`'d away on the linting host, so one Linux run covers every host. CI lints both feature sets, because a module behind a feature gate only exists when that feature is on.

It denies host `cfg` keys, native OS references (`std::os`, `libc`, `nix`, `windows_sys`) and, outside `libwild/src/host/`, naming a host tree directly. `target_arch` and `target_endian` stay unchecked: they're host facts too, but their uses outside the trees derive a default from the host architecture, as you pointed out on #2557.

The ratchet is the part that matters for the files the host module doesn't cover. `BASELINE` records what still selects the host — 14 findings in `wild/tests/integration_tests.rs`, 1 in `wild/tests/external_tests/lld_tests.rs`, 1 in `linker-diff/src/utils.rs`. Going over an allowance fails, and so does going under one without lowering the number, so those counts can only shrink and new host-dependent code has to go into a tree. I left the existing sites alone rather than growing this PR further.

Cost to you, which is the main reason to say no: a second toolchain (the nightly pinned in `lints/ban_host_cfg/rust-toolchain.toml`, since the lint links against rustc), `cargo-dylint` and `dylint-link` installed via `taiki-e/install-action`, and a job that takes a few minutes. If that isn't a trade you want, #2557's tidy test does most of this job with no new tooling, and I'm happy to drop this one and restore it.

Of the diff, the lint is 773 lines and its `Cargo.lock` about 1,100; the rest is the CI job, two UI fixtures and their expected output.

Verified on my fork: the whole CI suite including the new job is green, https://github.com/zackees/wild/actions/runs/35481420021. The lint has 13 unit tests plus the UI fixtures, and I checked the ratchet in both directions — adding `#[cfg(target_os = "linux")]` to `libwild/src/fs.rs` is rejected, and claiming an allowance of 15 where 14 findings exist reports that the allowance should be lowered.


davidlattimore (commented):
I'll have to think about whether this is worth the cost. One question though, does the use of cfg keys etc outside of wild/libwild actually matter?

zackees:
I should say what I'm working toward, since I never have and it bears on the cost question.

I want wild's host and its output format to be fully independent: ELF from a Windows machine, Mach-O from Linux, any format wild supports from any host it runs on. The far end of that is hosts that aren't a desktop OS at all. As a data point, the whole linker already builds for wasm32-wasip1 on this branch. I haven't run it yet, but it shows how close the host side already is.

That's why I've spent four PRs on the host boundary. If any host can produce any format, "which host am I on" and "what am I producing" have to be separate questions, and a scattered `cfg(target_os)` is exactly where the two get mixed up. On Linux and macOS you mostly get away with that. The more hosts there are, the more each stray cfg is something a person has to find by hand. With `cfg_select!` and no fallback arm, the compiler gives you that list instead.

There's a practical side too. My PRs are AI-assisted, as disclosed, and more of your incoming contributions will be over time. When an AI writes platform-specific code with no structure to put it in, it churns. It adds a cfg wherever the problem showed up, the next change adds another nearby, and the same host fact ends up spelled three ways. Give it a tree to put things in and a check that fails when it doesn't, and the code stays organized, because the error says where the code belongs. A convention in a doc doesn't do that; a lint does.

The Android bug from #2556 is the small version of what the churn produces. Two predicates for "is this Linux-ish" in one file drifted apart, libwild stopped compiling for Android, and nobody noticed because nobody builds it. It has already happened once in code you ship.

So to your question: linker-diff is the weakest of the three and I'm happy to drop it. The integration tests I'd defend, because they bake in the assumption I'm trying to remove. `PlatformKind::host()` maps Linux to ELF, macOS to Mach-O and everything else to nothing, so the platform under test is whatever platform you're running on. However far the linker gets, the suite as written can only ever test host equals target.

This fixes nothing that's broken today. It stops new host assumptions piling up in the one crate that will eventually have to prove cross-linking works. If that isn't worth a second toolchain and the CI time, say so and I'll close this. #2557 covers the code that has to build for another host, and I'd rather have that than nothing.

zackees:
Think if "godbolt" could compile a function into every target completely in browser wasm. That's what this pattern unlocks, and we pay almost nothing for it.

davidlattimore:
Thanks for the explanation. In the case of the tests, cfgs tell us effectively which OS we're running the tests on, which affects what kinds of tests we can reasonably run. My ideal would be if we could run all of the tests on at least some of the platforms. Unfortunately Apple makes it hard to build and run Mac stuff on non-Mac OSes. It also seems pretty hard to test Linux stuff on Mac, although maybe we just haven't tried hard enough. So I'm unsure if the same abstractions that work in libwild will work in the tests - maybe they would, but I'd have to see. Basically until we have a reasonable abstraction that we're happy with in integration_test, does it actually matter if cfgs are added or removed? Once we have an abstraction that we're happy with, we might as well convert all the test code, at which point we don't need the ratchet mechanism.

One concern that occurs to me with using dylint is that while it runs in CI, it doesn't run out of the box for a contributor. Maybe it can if the contributor sets stuff up appropriately, but many wouldn't. That means that a new contributor that puts say ELF-specific code into the layout module isn't going to find out that this is a problem until they send their PR for review.

zackees:
On dylint not running out of the box, I think that can be covered either way you prefer.

First, the lint gives the same answer on every machine. It reads the AST before any `cfg` is evaluated, so a `#[cfg(windows)]` block is flagged on Linux even though Linux would compile it away. A contributor on any OS sees the same findings as CI, and nobody has to lint for a platform they aren't on. The only question is how they get dylint installed, and I can make that a single step in one of two ways.

1. A script in the repo. It checks for `cargo-dylint` and `dylint-link`, installs them if they're missing, and runs the lint with the same arguments CI uses. The pinned nightly comes from the lint's own `rust-toolchain.toml`. It needs nothing but rustup and cargo, and works on Linux, macOS and Windows. I'll add it to this PR along with a line in CONTRIBUTING.
2. [soldr](https://github.com/zackees/soldr), which I maintain. It provisions `cargo-dylint`, `dylint-link` and the driver itself, using prebuilt binaries where they exist, so there's no compile-from-source wait:

   ```
   pip install soldr
   soldr cargo-dylint dylint --path lints/ban_host_cfg --workspace -- --all-targets
   ```

Option 1 keeps wild free of any dependency on my tooling, and I'm happy to go that way. Option 2 is there for anyone who'd rather not wait for `cargo install`.

In both cases a contributor who never runs it locally still finds out from CI, the same way they would for a cross-target build failure. Host code gets touched rarely, and the error says which tree the code belongs in, so that should be one round trip.

Your example of ELF-specific code landing in the layout module is a good one. A text scan would struggle with a format boundary like that, and an AST lint wouldn't. Most of the machinery for it is already in this PR if that's a direction you'd want.

zackees:
As for the Mach-O challenge, I'll follow up with you through privately with some options.

davidlattimore:
> Your example of ELF-specific code landing in the layout module is a good one. A text scan would struggle with a format boundary like that, and an AST lint wouldn't.

I'm not sure. We currently have the test `check_elf_specific_code` and I think it does a reasonable job of preventing people from accidentally adding elf-specific code to various modules. Sure it could be circumvented by using a macro, but it seems pretty unlikely that someone would do that accidentally and I don't think people would be deliberately trying to get around the checks.

Given that the dylint needs a specific nightly installed and the code for the dylint is also much more code than the text-based check it replaces, I'm leaning towards saying that we should hold off on this for now. We could always revisit it later if the text-based checks prove to be a problem.

PR #2559, opened 2026-09-20
feat: Support `--[no-]warn-execstack`, `--warn-execstack-objects`, and `--[no-]error-execstack`

### LINKED ISSUE #2547: Add support for `--warn-execstack` (& associated options)
Thank you very much for your speedy response last time; Now I'm back once more with another compatibility/feature request.

As per `ld.bfd(1)`: <https://sourceware.org/binutils/docs/ld.html#index-_002d_002dwarn_002dexecstack>
There are similar options also present in `ld.bfd` (such as `--error-execstack`), which may also be of interest to implement.

---

Support amongst major linkers [that I'm aware of]:

`ld.bfd`:
Supports `--warn-execstack`;
No stdout;
No stderr;

`ld.lld`:
Supports `--warn-execstack`;
No stdout;
No stderr;

`ld.mold`:
Supports `--warn-execstack`;
No stdout;
No stderr;

`ld.wild`:
Doesn't support `--warn-execstack`;
No stdout;
Stderr:
```text
wild: error: unrecognized option(s): --warn-execstack
```

---

`ld.bfd`, `ld.lld`, and `ld.mold` all seem to accept it fine, although it is possible that some of them may just be silently ignoring the option.

lapla-cogito:
Closes #2547


libwild/src/args/elf.rs
```diff
@@ -491,6 +503,43 @@ impl ElfArgs {
...
+        }
+
+        if self.error_execstack {
+            bail!("creating an executable stack because of -z execstack command line option");
+        }
+        self.warning("enabling an executable stack because of -z execstack command line option");
+        Ok(())
+    }
+
+    fn emit_execstack(&self, message: String) -> Result {
```
marxin: Please put body of the function directly at the single caller -> makes it easier to read and understand what's going on.

libwild/src/args/elf.rs
```diff
@@ -491,6 +503,43 @@ impl ElfArgs {
...
+    ///
+    /// Unlike GNU ld, we never enable an executable stack from input objects. `-z execstack` is
+    /// required. The flags only control whether we error, warn, or stay silent about that request.
+    pub(crate) fn report_object_execstack(&self, object: &impl std::fmt::Display) -> Result {
+        if self.execstack || matches!(self.warn_execstack, WarnExecstack::None) {
+            return Ok(());
+        }
+
+        self.emit_execstack(format!(
+            "{object}: requires executable stack, but -z execstack is not specified"
```
marxin: I think the error message is misleading as the error/warning can be emitted when `WarnExecstack::Always` is on - that should be distinguished, right?

libwild/src/args/elf.rs
```diff
@@ -491,6 +503,43 @@ impl ElfArgs {
...
+            || !matches!(self.warn_execstack, WarnExecstack::Always)
+            || self.should_output_partial_object
+        {
+            return Ok(());
+        }
+
+        if self.error_execstack {
+            bail!("creating an executable stack because of -z execstack command line option");
+        }
+        self.warning("enabling an executable stack because of -z execstack command line option");
```
marxin: I feel the 2 messages should be just unified?

libwild/src/args/elf.rs
```diff
@@ -491,6 +503,43 @@ impl ElfArgs {
...
+        }
+
+        self.emit_execstack(format!(
+            "{object}: requires executable stack, but -z execstack is not specified"
+        ))
+    }
+
+    pub(crate) fn report_z_execstack(&self) -> Result {
+        if !self.execstack
+            || !matches!(self.warn_execstack, WarnExecstack::Always)
```
marxin: Can you please reformat the condition so that it will be not be expressed as conjunction of positive conditions, so e.g. `self.execstack && matches!(self.warn_execstack, WarnExecstack::Always)` - again will make it much easier to get the meaning.

marxin (changes requested):

lapla-cogito:
Thanks. I addressed reviews above.

libwild/src/args/elf.rs
```diff
@@ -491,6 +503,39 @@ impl ElfArgs {
...
+        let message =
+            format!("{object}: requires executable stack, but -z execstack is not specified");
+        if self.error_execstack {
+            bail!("{message}");
+        }
+        self.warning(message);
+        Ok(())
+    }
+
+    pub(crate) fn report_z_execstack(&self) -> Result {
```
marxin: Looking at it again, I think this should go into `parse` (`libwild/src/args/elf.rs:508`) where we make a similar checks related to conflicting options.

libwild/src/args/elf.rs
```diff
@@ -491,6 +503,39 @@ impl ElfArgs {
         self.emulation.architecture()
     }
 
+    /// Report that `object` requested an executable stack via `.note.GNU-stack`.
+    ///
+    /// Unlike GNU ld, we never enable an executable stack from input objects. `-z execstack` is
+    /// required. The flags only control whether we error, warn, or stay silent about that request.
+    pub(crate) fn report_object_execstack(&self, object: &impl std::fmt::Display) -> Result {
+        if self.execstack || matches!(self.warn_execstack, WarnExecstack::None) {
```
marxin, line 514: I would consider adding a comment that at this point `self.execstack && matches!(self.warn_execstack, WarnExecstack::Always)` is already handled and cannot happen.

marxin (changes requested):
Please address 2 minor issues before we merge it.

libwild/src/args/elf.rs
lapla-cogito: It makes sense.

marxin (approved):

PR #2562, opened 2026-09-20
port(MachO): support implicit addend for relocations

marxin:
Unfortunately, I cannot construct a test-case as we're missing support for `rebase` fix-ups as of now (#2563).

An example of the implicit addend:
```
❯ cat unsigned-addend.c
char target[9] = "012345679";
char *pointer = target + 0x5;

int main(void)
{
    if (pointer != target + 5)
        return 1;
    if (*pointer != '5')
        return 2;
    return 0;
}
❯ clang unsigned-addend.c -c
...
  Section __data {
    0x10 0 3 1 ARM64_RELOC_UNSIGNED 0 _target
  }
...
Contents of section __DATA,__data:
 0070 30313233 34353637 39000000 00000000  012345679.......
 0080 05000000 00000000                    ........ <---- HERE the `0x05` is the implicit addend
```

FYI: @joanna-jo

Not sure if we can afford another 1B field in `RelocationKindInfo` struct, @davidlattimore?

Part of #757


davidlattimore (approved):
I ran a couple (ELF) of benchmarks and I can't see any change due to the extra byte. The struct is pretty short-lived, so it's entirely possible that the compiler is optimising it away for platforms that don't use it.

marxin:
Thanks for benchmarking!

PR #2569, opened 2026-09-21
chore: Remove `SkipLinker` / `EnableLinker` test directives

lapla-cogito:
These directives were marked as deprecated and then seem to have been forgotten about.


davidlattimore (approved):
Thanks!

PR #2570, opened 2026-09-21
perf: parallelize full-LTO debug relocation work

zackees:
## Summary

- shard safe large RELA debug-relocation streams and reuse adjacent symbol resolutions;
- populate independent sections of monolithic/full-LTO objects in parallel;
- prepare archive members in parallel while retaining input-order results and deterministic errors;
- preserve distinct `fast`, `md5`, and `sha1` build-ID modes.

The retained-section copy and bounded-ext4 buffering experiments were not included: current-upstream measurements regressed.

## Performance

Captured full-LTO debug link, 8 pinned CPUs, `--build-id=none`, 6 warmups and 40 alternating pairs:

| Baseline | Candidate | Change | Artifact |
|---:|---:|---:|---|
| 215.579 ms median | 124.564 ms median | **42.2% faster** | byte-identical 151.0 MiB ELF (SHA-256 `fa4891032343088af5562e77a86f9d863947bc2e3c90f5024dedc90aafef973ea5`) |

The candidate won all 40 pairs; mean paired saving was 92.013 ms (95% bootstrap CI: 89.816–94.428 ms). This is a host-local captured workload, not a general link-time claim.


davidlattimore:
The PR description is pretty verbose and contains various details that aren't important - e.g. that you ran `cargo fmt --check` - passing CI demonstrates that. Can you edit to just what's actually useful for us to read? Also, please keep in mind that our AI use policy (see `CONTRIBUTING.md`) asks that written communication (e.g. PR descriptions) is from people. If you need to quote something that the AI generated, that's fine, but we want to know when we're talking to a person vs talking to a bot.

davidlattimore:
I'm just running some benchmarks on each of the changes. So far the main speedup seems to come from "perf: shard safe large debug relocation streams". Some of the later changes possibly then regress that, but I need to run more benchmarks. Are there particular kinds of benchmarks I should be looking at for any of these? e.g. debug builds, non-debug builds etc

zackees:
Debug. This is what the full optimizations look like when applied. Not sure if all of the optimizations went into this PR or not.

image

https://github.com/zackees/reld

zackees:
*Response by [clud](https://github.com/zackees/clud) a custom agent made by Zach Vorhies*

The debug-relocation change should be benchmarked primarily with
  large debuginfo-heavy links. Our representative workload used a 249
  MiB full-LTO object with 28,040 sections, including roughly 60 MiB
  of .rela.debug_info. On that workload, relocation sharding reduced
  median end-to-end time by 43.65%.

  For non-debug builds, that optimization should normally take the
  serial path and provide little or no benefit. They are still useful
  as a regression check.

  The parallel XXH3 change should be tested separately with --build-
  id=fast and outputs larger than 4 MiB. It reduced the isolated
  build-ID phase from 8.170 ms to 7.325 ms, or 10.3%, and improved the
  complete link by 3.72%.

  I would suggest testing these configurations independently and
  together:

  - Large debug/full-LTO link: relocation sharding should dominate.
  - Large non-debug link with --build-id=fast: isolates parallel
    hashing.

  - Small debug and non-debug links: verifies that the thresholds and
    serial fallbacks do not regress them.

  - Aggregate large debug link: measures interaction between both
    optimizations.

  We used 20 rotating/alternating runs on a Ryzen 7 3700X with ext4
  and Rust 1.97.1. The aggregate configuration improved the paired
  median by 45.41%.


  All reld-derived optimizations that were both applicable to Wild and
  demonstrated a real improvement made it into PR #2570:

  1. Parallel debug-relocation sharding — 43.65% improvement.
  2. Parallel XXH3 build-ID hashing — 3.72% end-to-end and 10.3% in
     the hashing phase.

  3. Combined — 45.41% improvement.

  Other candidates were tested but intentionally excluded:

  - Retained-output scheduling: 32.1% slower.
  - Buffered ext4 writing: 276.8% slower.
  - Adjacent-symbol cache: 0.6% slower.
  - Monolithic section processing: approximately 0.2% improvement,
    within noise.

  - Archive preparation change: 0.4% slower and primarily a
    determinism change.

  Wild already independently contains output preallocation/fallocate
  support and relocation caching, so those did not need to be added by
  #2570.

  So the precise answer is: yes, every measured reld optimization that
  benefits Wild is included; no, every experimental reld optimization
  was not blindly ported. The final PR contains the two proven
  additions, in parallel form.

zackees:
Correction to the original performance table: the old 1.3% synthetic-link number does not describe the rebuilt PR head and should not be used for this change. During the earlier force-push, `libwild/src/input_data.rs` was left out, along with the follow-up commits for monolithic section population, deterministic archive preparation/error handling, adjacent debug-symbol reuse, and build-ID-mode fixes.

The rebuilt head is `3c09b1aa`. On the captured full-LTO debug link with `--build-id=none`, 8 pinned CPUs, 6 warmups, and 40 alternating pairs, it measured **215.579 ms → 124.564 ms median (42.2% faster)**. The candidate was faster in **40/40** pairs; mean paired saving was **92.013 ms** (95% bootstrap CI **89.816–94.428 ms**). Outputs were byte-identical: 151.0 MiB, SHA-256 `fa4891032343088af5562e77a86f9d863947bc2e3c90f5024dedc90aafef973ea5`.

The previous 1.3% claim was based on the stale, incomplete PR head and a different synthetic workload. It is therefore replaced rather than combined with these numbers.

davidlattimore:
What were you benchmarking that showed this speedup? I just tried a debug build of zed and didn't see much change. Output from a tool like hyperfine or poop would be helpful.

zackees:
I had to downgrade to codex terra as Claude ran out of credits until thursday.

It's now vibe slopping perf benchmark data - made worse as my perf infra is on the reld-wild fork and not the wild-wild fork. There's like 8 other builds running too so I don't trust the numbers here as is.

This PR should be considered a work in progress until I get it sorted out tomorrow.

Thanks for the suggestion on tooling. I'm going to have tighten up the perf benchmarking.

Might be useful to have an automated action for benchmarking triggered by a branch prefix of perf/

davidlattimore:
Automated benchmarking would be nice, but the github infra is I think a little too noisy to get good quality results. Benchmarking on my own hardware, I can make sure that nothing else is running, turn off performance boosting, set the CPUs to a fixed clock rate etc.

mati865:
Remember that you also get different CPUs between runs, in addition to varying CPU load.

There are tools like codspeed that run on their own machines but I haven't used them.

zackees:
Tell me if this is the wrong strategy:

A/B benchmarking on a pinned runner.

I've found the github runners to be surprisingly stable. This is good for performance differences > 5%.

Otherwise they have 3% noise in their performance. Good for significant performance improvements and catching regressions.

davidlattimore:
I think that's a good strategy, but 5% is too high for what we'd want. If I see a change that regresses performance by 1.5% I'd be wanting to look into what could be done about it.

PR #2574, opened 2026-09-21
chore: fix build after recently merged commits

marxin:


davidlattimore (approved):
Any idea why CI didn't pick it up?

marxin:
Well, I merged 2 PRs one right after another and the CI runs were run independently. It's caused by us not requiring a PR be based on the recent `main` branch tip (which is a good compromise to be honest) :)

lapla-cogito:
The merge queue feature can resolve this type of problem.

marxin:
> The merge queue feature can resolve this type of problem.

Yeah, good point. Given the situation happens very rarely, I feel we don't have to invest in more complex workflow.

mati865:
There should be an option to require ci re-run when base branch changes. I'll look at it.

mati865:
Under branch protection rules there is 

Enabled it to take the screenshot, but without saving the changes, so it's still disabled. We might want to set it up.

davidlattimore:
That would be an easy option, but I feel it might have some substantial downsides. If two people send PRs overnight, then I review them when I get up and merge one, unless I'm mistaken, I couldn't then merge the second one without getting the author to rebase and repush their change. Or maybe I'm missing something and there's a way to make this behave more like lapla-cogito's suggestion of a merge queue where the rebasing happens automatically?

mati865:
I'm not entirely sure how it works in regard to auto rebase with GitHub but at least there are buttons to update the PR: 
obraz

EDIT: Oh, that's a part of Github Refined: https://github.com/refined-github/refined-github#editing-pull-requests
I thought it was incorporated by now.

davidlattimore:
Huh, neat! I just turned on that option and now we have a button to update the branch. I say we try it for a while and see how it goes

mati865:
At the first sight it looks like it might be annoying with how GitHub implemented it. We can try the merge queues too if you want, I haven't used it myself, so no feedback from my side.

davidlattimore:
I just merged a PR and the process seemed reasonable. I clicked the button to rebase it, that then triggered CI to rerun. With CI running, I clicked the merge-when-ready button (I forget the exact wording on the button). That then waited until the tests had completed and then automatically merged.

One somewhat annoying thing is that in the branch protection rule, you have list all the required checks that need to pass. We have quite a few and you have add them all, otherwise PRs could be merged with those checks failing. It'd definitely be nicer if you could say all checks must pass then maybe opt some out if you want them to be optional.

I guess the current scheme might not work well if we're trying to merge two PRs at once. Presumably one will win then the other one will fail to merge because it needs rebasing again. If that becomes a problem, then I'd definitely be up for giving merge queues a try.

davidlattimore:
Although one mitigating factor is that the checks pane on any PR shows which checks that were run are and aren't required, so seeing checks here that don't have the "required" tag next to them can serve as a reminder that we need to add them.

PR #2575, opened 2026-09-21
fix(aarch64): resolve undefined weak branch relocations

deepakshirkem:
Handles undefined weak AArch64 branch relocations according to the ABI by resolving the branch target to the next instruction.

Adds support for `R_AARCH64_CALL26`, `R_AARCH64_JUMP26`, `R_AARCH64_CONDBR19`, and `R_AARCH64_TSTBR14`, with regression coverage compared against lld.

Part of #2380


davidlattimore (commented):
I'm not sure that we always want to do that. The following patch adds a test that passes at head, but fails with this change. Note that the added test leans heavily on linker-diff to actually do the checking.

```patch
diff --git a/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s b/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
index 1c40d61937..7ff8105b04 100644
--- a/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
+++ b/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
@@ -1,12 +1,22 @@
 // Check that AArch64 branches to an undefined weak symbol resolve
-// to the next instruction.
+// to the next instruction in static output and to the PLT in shared output.
 //
+//#AbstractConfig:default
 //#Arch:aarch64
 //#ReferenceLinkers:lld
 //#RunEnabled:false
 //#LinkArgs:--image-base=0x10000000 --no-gc-sections
+
+//#Config:static:default
 //#ExpectSectionBytes:.text=0x01000014010000942000005421000036c0035fd6
 
+//#Config:shared:default
+//#Mode:dynamic
+//#LinkArgs:-shared
+//#DiffIgnore:.dynamic.DT_RELA
+//#DiffIgnore:.dynamic.DT_RELAENT
+//#DiffIgnore:section.got.plt.entsize
+
 .weak target
 
 .text
```

libwild/src/elf_writer.rs
```diff
@@ -3396,6 +3396,15 @@ fn apply_relocation<
...
 
+    let definition = layout.symbol_db.definition(local_symbol_id);
+
+    let undefined_weak_target =
+        if layout.symbol_db.is_undefined(definition) && layout.symbol_db.is_weak(local_symbol_id) {
+            A::undefined_weak_target(r_type, place)
+        } else {
+            None
+        };
+
```
davidlattimore: I've benchmarked your change (linking wild for aarch64 on x86_64). It shows about a 1% slowdown. I think it's due to doing this stuff here on the hot path.

deepakshirkem: I’ve moved the undefined/weak detection out of the hot path now. Please rerun the benchmark and check whether there’s still any measurable slowdown?

davidlattimore (commented):
Thanks. Yep, that seems to have fixed the performance regression.

But here's another patch that shows a failing case:

```patch
diff --git a/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s b/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
index d71dab642a..0044f6cb82 100644
--- a/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
+++ b/wild/tests/sources/elf/aarch64-undefined-weak-branch/aarch64-undefined-weak-branch.s
@@ -16,16 +16,21 @@
 //#DiffIgnore:.dynamic.DT_RELAENT
 //#DiffIgnore:section.got.plt.entsize
 
+//#Config:shared-conditional-only:shared
+//#CompArgs:-Wa,--defsym,CONDITIONAL_ONLY=1
+
 .weak target
 
 .text
 .global _start
 _start:
+.ifndef CONDITIONAL_ONLY
     // R_AARCH64_JUMP26
     b target
 
     // R_AARCH64_CALL26
     bl target
+.endif
 
     // R_AARCH64_CONDBR19
     b.eq target
```

davidlattimore (approved):
Thanks :)

PR #2581, opened 2026-09-22
fix: Align TLS segment p_vaddr to p_align

### LINKED ISSUE #2578: TLS segment p_vaddr not aligned to p_align with varied alignment of thread-local variables
The following C program:
```c
#include <stdio.h>
__thread long a = 0x1122334455667788;
__thread _Alignas(256) char b[8];
int main(void) { printf("a=%lx b=%d\n", a, b[0]); return 0; }
```

When built as follows:
```
clang -static -fuse-ld=wild repro.c -o out.wild
```

We'd expect it to print:
```
a=1122334455667788 b=0
```

However, on my machine (Ubuntu 24.04.3), this segfaults. ld, lld, and mold all give the expected result.

`readelf -lW` on the wild-linked program shows:
```
Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
...
  TLS            0x0aa6c8 0x00000000004ac6c8 0x00000000004ac6c8 0x000020 0x000174 R   0x100
```
which as I understand it is the issue: VirtAddr isn't aligned to Align as expected

I reproduced this on the latest trunk (34b9b72eec9d2acdded9350b611c1dbf40d6c1bf)

For disclosure: this was found by a fuzzer I wrote w/ LLM assistance to differentially test linker behavior. The reproducer is artificial rather than from real-world code

lapla-cogito:
Some loaders, including glibc, require PT_TLS p_vaddr to be a multiple of p_align. With mixed TLS alignments, p_align is the maximum of all TLS sections, but p_vaddr was the start of .tdata, which can have a weaker alignment. Align the first TLS section to the segment alignment so that this holds.

Fixes #2578


wild/tests/sources/elf/tls-segment-align/tls-segment-align.c
```diff
@@ -0,0 +1,31 @@
...
+__thread char tls_var_b[8] __attribute__((aligned(256)));
+
+int main(void) {
+  if (tls_var_a != 0x1122334455667788UL) {
+    return 1;
+  }
+  if (tls_var_b[0] != 0) {
+    return 2;
+  }
+  if (((unsigned long)tls_var_b) & 255UL) {
```
davidlattimore: `objdump -dr wild/tests/build/elf/x86_64/tls-segment-align/static/tls-segment-align.c.o` shows that GCC optimises this if statement away - the dumped assembly is identical to if I delete the if statement entirely. `ptr_black_box.h` would likely help here

libwild/src/layout.rs
davidlattimore: If we moved this line down in the function, would it allow us to skip setting it in the if-statement you added below?

davidlattimore (commented):
Thanks for the fix :)

wild/tests/sources/elf/tls-segment-align/tls-segment-align.c
lapla-cogito: Ah, I didn't notice that. Thanks.

libwild/src/layout.rs
lapla-cogito: That makes sense.

davidlattimore (approved):

